PR #5, opened 2019-06-13
MasterUserRecord and NSTemplateTier CRDs

alexeykazakov:


alexeykazakov:
> Looks good, but we will need to update it when we fix: codeready-toolchain/api#3 (comment) but can be as separated PR

Yeah. I will take care of it.

PR #8, opened 2019-06-16
Add freshly generated MasterUserRecordController

sbryzak:
Empty controller, reconcile loop performs no business logic in this PR.

Fixes https://jira.coreos.com/browse/CRT-132


pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -0,0 +1,153 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Define a new Pod object
+	pod := newPodForCR(instance)
```
alexeykazakov: Let's remove the pod logic here since it's not relevant. It's OK if the controller doesn't do much yet but let's not make it do something we don't want.

sbryzak: All pod-generation sample code removed.

PR #10, opened 2019-06-18
Rename CRD files to align with api generator

alexeykazakov:
The CRDs are regenerated with the new `make generate` target from `api` repo. There are a few differences:
- `_crd` prefix is removed.
- the generator uses a different version of **k8s controller-tools** (the last stable one, which is 0.1.10) comparing to operator-sdk. Some changes look suspicious but let's fix them later if they cause any problems. We should fix them by using the proper version of the generator tools rather than manually changing the CRDs.

### DIFF
--- a/deploy/crds/toolchain_v1alpha1_masteruserrecord.yaml
+++ b/deploy/crds/toolchain_v1alpha1_masteruserrecord.yaml
@@ -1,14 +1,15 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
   name: masteruserrecords.toolchain.dev.openshift.com
 spec:
   group: toolchain.dev.openshift.com
   names:
     kind: MasterUserRecord
-    listKind: MasterUserRecordList
     plural: masteruserrecords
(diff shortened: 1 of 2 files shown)


xcoulon:
> The CRDs are regenerated with the new `make generate` target from `api` repo. There are a few differences:
> 
> * `_crd` prefix is removed.
> * the generator uses a different version of **k8s controller-tools** (the last stable one, which is 0.1.10) comparing to operator-sdk. Some changes look suspicious but let's fix them later if they cause any problems. We should fix them by using the proper version of the generator tools rather than manually changing the CRDs.

FYI, I had to use the `v0.10.0` of the k8 controller tools to be able to run the `go` command. With a later version, I got some compilation errors. I can revisit this if we need a more recent version of the tools to generate the `.go` and CRD files.

alexeykazakov:
This is what is used by `operator-sdk`: https://github.com/operator-framework/operator-sdk/blob/master/Gopkg.toml#L41

I will try to add some CRD tests today and see if it works. But we need to align it with the api generator anyway (so I would merge this PR for now).

nurali-techie:
@alexeykazakov any particular reason for removing `_crd` prefix ??  It was good indication/filter to find diff CRDs.

alexeykazakov:
The reason is in the description. To align with the crd generator in api repo.

PR #14, opened 2019-06-21
feat(CRT-150): add caching controller for KubeFedCluster

MatousJobanek:
Changes introduced in this PR:
* slightly changed startup of the KubeFedCluster controllers in main.go so it makes more sense
* starts KubeFedCluster caching controller taken from https://github.com/codeready-toolchain/toolchain/pull/4

These are the same changes that go also to the host repo: https://github.com/codeready-toolchain/member-operator/pull/22


pkg/cluster/service.go
```diff
@@ -0,0 +1,107 @@
...
+		Client:            cl,
+		ClusterStatus:     &fedCluster.Status,
+		Type:              Type(fedCluster.Labels[labelType]),
+		OperatorNamespace: fedCluster.Labels[labelNamespace],
+	}
+	if cluster.OperatorNamespace == "" {
+		if cluster.Type == Host {
+			cluster.OperatorNamespace = defaultHostOperatorNamespace
+		} else {
+			cluster.OperatorNamespace = defaultMemberOperatorNamespace
```
alexeykazakov: So, if no labels set then Type will be an empty string which implies member? Wouldn't it be safer to explicitly set Type = "member" too then?

pkg/controller/kubefedcluster.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+	}
+	clusterHealthCheckConfig := &util.ClusterHealthCheckConfig{
+		PeriodSeconds:    util.DefaultClusterHealthCheckPeriod,
+		TimeoutSeconds:   util.DefaultClusterHealthCheckTimeout,
+		FailureThreshold: util.DefaultClusterHealthCheckFailureThreshold,
+		SuccessThreshold: util.DefaultClusterHealthCheckSuccessThreshold,
+	}
+	klog.InitFlags(nil)
+	return kubefedcluster.StartClusterController(controllerConfig, clusterHealthCheckConfig, stopChan)
```
alexeykazakov: Does the health check controller uses all the kubefed clusters from the given namespace to health check them?
What happens if some health check fail?

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +82,16 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+	for _, account := range userRecord.Spec.UserAccounts {
+		syncStatus := syncWithUserAccount(account, userRecord)
+		if syncStatus.Requeue {
+			return reconcile.Result{
+				Requeue:      syncStatus.Requeue,
+				RequeueAfter: syncStatus.RequeueAfter},
+				syncStatus.Err
+		}
+		continue
+	}
```
alexeykazakov: We will also need to update the status of the MasterUserRecord but I guess this is in the TODO list :)
Just don't forget that we will need to update the status if sync failed too.

MatousJobanek:
Thanks a lot for your review @alexeykazakov 

> This PR already contains pretty big chunk of functionality. Let's try to follow the rule of "multiple small PRs instead of one big PR".
...
It will improve the quality of our PR reviews :)

That was my initial intention. The only purpose of this "WIP PR" was to share the idea with you. It wasn't a request for the proper review at all - I just wanted to show you what I'm working on and what it could look like. That was also the reason why I added totally incomplete changes in `masteruserrecord_controller.go ` file in this PR - to show the usage and that I'll cover this communication between host and member.
I also mentioned that I wanted to spread it over multiple PRs and repos in the description of this PR, so I think that there was no need to point it out again ;-)
I created the PR very quickly, I knew that the code had a lot of deficiencies, formal and informal mistakes, todo things and so on. The only purpose was to give you a chance to see the idea, my work and rough code, nothing else.  ;-)

pkg/cluster/cache.go
```diff
@@ -0,0 +1,57 @@
...
+
+	Name              string
+	Type              Type
+	OperatorNamespace string
+
+	// clusterStatus is the cluster status as of last sampling.
+	ClusterStatus *v1beta1.KubeFedClusterStatus
+}
+
+func (c kubeFedClusterClients) addFedCluster(cluster *FedCluster) {
```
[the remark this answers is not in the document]
MatousJobanek: That was on my todo list as well :-) - I have already some local changes, including this and some more so I won't commit your suggestion as it would introduce conflicts on my side

pkg/cluster/cache.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (c kubeFedClusterClients) deleteFedCluster(name string) {
+	c.Lock()
+	defer c.Unlock()
+	delete(c.clusters, name)
+}
+
+func (c kubeFedClusterClients) getFedCluster(name string) (*FedCluster, bool) {
+	c.Lock()
```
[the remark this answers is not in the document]
MatousJobanek: thanks for this proposal - on Friday evening I realized that my approach wasn't too effective an I wanted to check it, but you already gave me the answer :-D

pkg/cluster/service.go
```diff
@@ -0,0 +1,107 @@
...
+	defaultHostOperatorNamespace   = "toolchain-host-operator"
+	defaultMemberOperatorNamespace = "toolchain-member-operator"
+)
+
+type KubeFedClusterService struct {
+	LocalConfig *rest.Config
+	Log         logr.Logger
+}
+
+func (s KubeFedClusterService) AddKubeFedCluster(obj interface{}) {
```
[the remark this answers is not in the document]
MatousJobanek: I know :-). These pointers is something that I check at once as part of the final cleanup of the code :-)...

pkg/cluster/service.go
MatousJobanek: yup, for sure

pkg/controller/kubefedcluster.go
MatousJobanek: > Does the health check controller uses all the kubefed clusters from the given namespace to health check them?

yes, why it wouldn't?

> What happens if some health check fail?

Then it sets the condition `Ready` to `false` with the corresponding reason (For more info see this: https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/controller/kubefedcluster/clusterclient.go#L77-L108) and updates the CR with this new status

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: yes, there are much more things that are missing in the `masteruserrecord_controller.go` code

pkg/controller/kubefedcluster.go
alexeykazakov: > yes, why it wouldn't?

Just wanted to confirm ;)

pkg/controller/kubefedcluster.go
```diff
@@ -1,8 +1,8 @@
 package controller
 
-
 import (
 	"fmt"
+	"github.com/codeready-toolchain/toolchain/pkg/controller"
```
alexeykazakov:
```suggestion
	"github.com/codeready-toolchain/toolchain-common/pkg/controller"
```

PR #24, opened 2019-07-03
Add conditions to MUR status

alexeykazakov:
See https://github.com/codeready-toolchain/api/pull/15

### DIFF
--- a/deploy/crds/toolchain_v1alpha1_masteruserrecord.yaml
+++ b/deploy/crds/toolchain_v1alpha1_masteruserrecord.yaml
@@ -30,9 +30,15 @@ spec:
           type: object
         spec:
           properties:
-            state:
-              description: 'Desired state of the user record: approved|banned|deactivated'
-              type: string
+            deprovisioned:
+              description: If set to true then the corresponding UserAccount should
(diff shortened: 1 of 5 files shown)


deploy/crds/toolchain_v1alpha1_masteruserrecord.yaml
```diff
@@ -30,9 +30,15 @@ spec:
...
-              type: string
+            deprovisioned:
+              description: If set to true then the corresponding UserAccount should
+                be deleted "false" is assumed by default
+              type: boolean
+            disabled:
+              description: If set to true then the corresponding user should not be
+                able to login (but the underlying UserAccounts still exists) "false"
+                is assumed by default
+              type: boolean
```
nurali-techie: Though it may be fine to add all this stuff related to deprovisioning / disabled .. but ideally this only should be introduce when we plan to add these features .. like currently we are only working on user_provisioning feature and nothing related to user_deprovisioning and user_disable.

The problem is, we are adding more stuff than required as of now.  Also, we really don't know if these attributes are enough when we actually implement user_deprovisioning and user_disabled and there is chance that we need to refactor it.

PR #28, opened 2019-07-10
Add e2e tests make target

dipak-pawar:


dipak-pawar:
`images` and `e2e` stanzas are failing because of wrong output path. I have opened PR for it https://github.com/openshift/release/pull/4329. 

Please note that we need to merge at least one of them with failed check

make/test.mk
```diff
@@ -1,20 +1,28 @@
...
 ## runs the tests without coverage and excluding E2E tests
 test:
 	@echo "running the tests without coverage and excluding E2E tests..."
 	$(Q)go test ${V_FLAG} -race $(shell go list ./... | grep -v /test/e2e) -failfast
+
 	
-.PHONY: test-ci
-# runs the tests and uploads the coverage report on codecov.io
-# DO NOT USE LOCALLY: must only be called by OpenShift CI when processing new PR and when a PR is merged! 
-test-ci: test-with-coverage upload-codecov-report
```
alexeykazakov: Why are you removing `test-ci`?

alexeykazakov: I thought it's still being used? But apparently it's not?
I'm confused... How do we upload codecov reports then if test-ci is not used?

Btw, regarding codecov reports. Ideally we should collect coverage for both unit and e2e tests. Not sure how we could do it practically... (cc: @xcoulon )

dipak-pawar: It was used before. Now it's removed.

alexeykazakov: But where/when we report to codecov now?

alexeykazakov: Also I see we still have it in member repo. Can we align our repos?

dipak-pawar: @alexeykazakov: we are using two make targets instead of it, so I think we can make it. Look at https://github.com/openshift/release/pull/4266

dipak-pawar: > Also I see we still have it in member repo. Can we align our repos?

I will remove it

alexeykazakov: > we are using two make targets instead of it
Ah.. right. Missed that.
So, looks good!

xcoulon: @alexeykazakov yes, we discussed with @dipak-pawar and realized that the `test-ci` goal had little or no benefit, and we could just set the command to execute to `make test-with-coverage upload-codecov-report` in the job, which gives the same result.

PR #38, opened 2019-07-31
Add operator deploymentconfig and imagestream for continuous deployment

dipak-pawar:
*Changes Proposed in this PR*
- Adds ImageStream which points to image from openshift CI cluster.
- Add DeploymentConfig which uses above ImageStream
- Fix minor make targets

To Do
- [ ] Add Documentation
- [ ] Testing on actual clusters

### DIFF
--- a/deploy/operator_dev_cluster.yaml
+++ b/deploy/operator_dev_cluster.yaml
@@ -1,0 +1,86 @@
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: host-operator-v0.1
+  namespace: toolchain-host-operator
+spec:
+  lookupPolicy:
+    local: false
+  tags:
+  - from:
+      kind: DockerImage
+      name: registry.svc.ci.openshift.org/codeready-toolchain/host-operator-v0.1:host-operator
+    importPolicy:
+      scheduled: true
+    name: host-operator
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    app: host-operator
+  name: host-operator
+  namespace: toolchain-host-operator
+spec:
+  replicas: 1
+  revisionHistoryLimit: 10
+  selector:
+    app: host-operator
+    deploymentconfig: host-operator
+  strategy:
+    activeDeadlineSeconds: 21600
+    resources: {}
+    rollingParams:
+      intervalSeconds: 1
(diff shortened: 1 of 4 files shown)


nurali-techie:
@dipak-pawar the name `operator_dev_cluster.yaml` suggest this is specifically for `dev_cluster` .. but I guess it's not specific to any cluster and can be applied to any cluster !!

deploy/operator_dev_cluster.yaml
```diff
@@ -0,0 +1,93 @@
+apiVersion: image.openshift.io/v1
+kind: ImageStream
```
xcoulon: aren't we supposed to have a `DeploymentConfig` that uses this `ImageStream`?

xcoulon:
> @dipak-pawar the name `operator_dev_cluster.yaml` suggest this is specifically for `dev_cluster` .. but I guess it's not specific to any cluster and can be applied to any cluster !!

for now the target is the `dev` clusters since we want to continuously deploy on this environment, as soon as a new Docker Image was pushed on the container registry. For the staging and production environments, the deployment policy may be slightly different

dipak-pawar:
> the name operator_dev_cluster.yaml suggest this is specifically for dev_cluster .. but I guess it's not specific to any cluster and can be applied to any cluster !!

@nurali-techie  yes, you can apply it to any cluster, but not to any production cluster. At this point, we are just using it with dev cluster, this is why I have given this name. I don't think I'll use this imagestream setup for my local developement where I can directly use container image

deploy/operator_dev_cluster.yaml
dipak-pawar: we are using it here https://github.com/codeready-toolchain/host-operator/pull/38/files#diff-caad408dc9a5d42c733a0e159f579e64R18

nurali-techie:
@dipak-pawar @xcoulon fine :+1:  thanks for clarification :-)

deploy/operator_dev_cluster.yaml
xcoulon: ah sorry, I missed that  🤦‍♂

dipak-pawar:
thank you @xcoulon  for the review. merging this as wanted to see actual rollout of deployment  in https://console-openshift-console.apps.host.crt-stage.com once we have new image in openshift ci image registry.

PR #39, opened 2019-07-31
update usersingup/status permission

dipak-pawar:


nurali-techie:
@dipak-pawar have you tested or know that this will work.  like `usersignups/status` .. because I remembered that I tried something for `status` sub-resource permission and that was not working.

dipak-pawar:
> have you tested or know that this will work. like usersignups/status .. because I remembered that I tried something for status sub-resource permission and that was not working.

@nurali-techie I have tried it, it's working for me. Btw, not sure which triggers this status update action as I don't have tests related to it. Maybe @sbryzak  can tell us if he is still having the issue. attaching screenshot for reference.
Screenshot from 2019-07-31 18-56-24

nurali-techie:
@dipak-pawar as far as you have some reference for this change .. I am fine :+1:   It just for double cross-check :-)

PR #41, opened 2019-08-08
feat(CRT-136): create UserAccount in member cluster 

MatousJobanek:
This is the first step of the whole host -> member synchronization of MUR and UserAccount.
This PR is derived from https://github.com/codeready-toolchain/host-operator/pull/23

This PR brings the code that only creates UserAccount in the member when a new MasterUserRecord is created.
Additional changes of the status and synchronization will be done in following PRs, but everything is available in #23 

I haven't added e2e tests as it requires both operators running and we still don't have the support for it - there is Jira issue created for it: https://jira.coreos.com/browse/CRT-202


dipak-pawar:
@MatousJobanek I have approved this PR. but It'll be good, if we have e2e tests as a part of this PR as we have required make target available in this repository to run both the operators on single clusters

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -2,27 +2,36 @@ package masteruserrecord
...
 var log = logf.Log.WithName("controller_masteruserrecord")
 
-/**
-* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
-* business logic.  Delete these comments after modifying this file.*
- */
+const (
+	// Status condition reasons
+	unableToGetUserAccountReason = "UnableToGetUserAccount"
+	clusterNotReady              = "ClusterNotReady"
```
alexeykazakov: Maybe "TargetClusterNotReady" would be more clear?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+		// then
+		require.Error(t, err)
+		msg := "the fedCluster member-cluster not found in the registry"
+		assert.Contains(t, err.Error(), msg)
+		assertUaNotFound(t, memberClient)
+		assertMurCondition(t, hostClient, toBeNotReady(clusterNotReady, msg))
+	})
+
+	t.Run("when member cluster does not exist", func(t *testing.T) {
```
alexeykazakov:
```suggestion
	t.Run("when member cluster is not ready", func(t *testing.T) {
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+const (
+	hostClusterName   = "host-cluster"
+	memberOperatorNs  = "toolchain-member-operator"
+	memberClusterName = "member-cluster"
+	hostOperatorNs    = "toolchain-host-operator"
+)
+
+type getMemberCluster func(cl client.Client) func(name string) (*cluster.FedCluster, bool)
+
+func TestCreateUserAccountSuccessful(t *testing.T) {
```
alexeykazakov: What do you think about adding a test for the case when the UA already exists? Basically in this case nothing should fail or change (yet).

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+	test.AssertConditionsMatch(t, record.Status.Conditions, expected)
+}
+
+func assertUaNotFound(t *testing.T, memberClient client.Client) {
+	ua := &toolchainv1alpha1.UserAccount{}
+	err := memberClient.Get(context.TODO(), namespacedName(memberOperatorNs, "john"), ua)
+	require.Error(t, err)
+	assert.IsType(t, metav1.StatusReasonNotFound, errors.ReasonForError(err))
+}
+
```
alexeykazakov: Let's also add a test(s) for the case when the status update failed (something like https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/useraccount/useraccount_controller_test.go#L414)

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+func newMasterUserRecord(userName string) *toolchainv1alpha1.MasterUserRecord {
+	mur := &toolchainv1alpha1.MasterUserRecord{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      userName,
+			Namespace: hostOperatorNs,
+		},
+		Spec: toolchainv1alpha1.MasterUserRecordSpec{
+			UserID: "12345abcdef",
+			UserAccounts: []toolchainv1alpha1.UserAccountEmbedded{{
```
alexeykazakov: Ideally we would want to test multiple UserAccounts in one MUR but it would be pretty tricky to do in the current e2e setup (one host and one member)...

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: you are right, renamed in f793b66

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: When the UA already exists then it should synchronize and this is already covered in the following PR #23. This is why I skipped this case for this PR (to limit the conflicts and my work :-))

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: Good point. I started working with multiple UserAccounts in the following PR, but it makes sense to do it here. 
As for e2e setup, yes it would be very tricky, but I'll cover it in the unit test

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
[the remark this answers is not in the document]
MatousJobanek: yeah, the result of my "copy-paste driven development". changed daa7869

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: done c0939f4

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: added the test here: 13cea3e see https://github.com/codeready-toolchain/host-operator/pull/41/commits/13cea3e32a7e039d95c6ac284eada3b6429066dd#diff-93c8b7a9cc0875ddd6d3002d3d0b4253R51-R75
to make the code (and my life) easier, I also introduced some test doubles and assertions (that were originally introduced in the PR #23) as part of this commit. Please take a look at it, thanks and sorry for bringing new additional changes, but the code would look horrible without doing it :-)

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
alexeykazakov: Fair enough.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -2,27 +2,36 @@ package masteruserrecord
...
 var log = logf.Log.WithName("controller_masteruserrecord")
 
-/**
-* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
-* business logic.  Delete these comments after modifying this file.*
- */
+const (
+	// Status condition reasons
+	unableToGetUserAccountReason = "UnableToGetUserAccount"
+	targetClusterNotReady        = "TargetClusterNotReady"
```
nurali-techie: `Reason` suffix missing for variable name.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+		if err != nil {
+			reqLogger.Error(err, "unable to synchronize with member UserAccount")
+			return reconcile.Result{}, err
+		}
+	}
+
 	return reconcile.Result{}, nil
 }
+
+func (r *ReconcileMasterUserRecord) ensureUserAccount(log logr.Logger, recAccount toolchainv1alpha1.UserAccountEmbedded, record *toolchainv1alpha1.MasterUserRecord) error {
```
nurali-techie: it seems that `record` param is not used.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+		}
+		return r.wrapErrorWithStatusUpdate(log, record, r.setStatusFailed(unableToGetUserAccountReason), err,
+			"failed to get userAccount '%s' from cluster '%s'", record.Name, recAccount.TargetCluster)
+	}
+
+	// TODO do the synchronization
+	return nil
+}
+
+func (r *ReconcileMasterUserRecord) getMemberCluster(log logr.Logger, recAccount toolchainv1alpha1.UserAccountEmbedded, record *toolchainv1alpha1.MasterUserRecord) (*cluster.FedCluster, error) {
```
nurali-techie: it seems that `record` param is not used.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+			return reconcile.Result{}, err
+		}
+	}
+
 	return reconcile.Result{}, nil
 }
+
+func (r *ReconcileMasterUserRecord) ensureUserAccount(log logr.Logger, recAccount toolchainv1alpha1.UserAccountEmbedded, record *toolchainv1alpha1.MasterUserRecord) error {
+	// get & check fed cluster
+	fedCluster, err := r.getMemberCluster(log, recAccount, record)
```
nurali-techie: `fedCluster` and `getMemberCluster`; wrt naming use either `fed` / `member` for consistency :-)

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+
+	// TODO do the synchronization
+	return nil
+}
+
+func (r *ReconcileMasterUserRecord) getMemberCluster(log logr.Logger, recAccount toolchainv1alpha1.UserAccountEmbedded, record *toolchainv1alpha1.MasterUserRecord) (*cluster.FedCluster, error) {
+	// get & check fed cluster
+	fedCluster, ok := r.retrieveMemberCluster(recAccount.TargetCluster)
+	if !ok {
+		return nil, fmt.Errorf("the fedCluster %s not found in the registry", recAccount.TargetCluster)
```
nurali-techie: Here as it's error message, `Member cluster` goes better than `fed cluster`.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+				return err
+			}
+			return nil
+		}
+		return r.wrapErrorWithStatusUpdate(log, record, r.setStatusFailed(unableToGetUserAccountReason), err,
+			"failed to get userAccount '%s' from cluster '%s'", record.Name, recAccount.TargetCluster)
+	}
+
+	// TODO do the synchronization
+	return nil
```
nurali-techie: shall we set status=`Ready+True` here ??

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+	fedCluster, err := r.getMemberCluster(log, recAccount, record)
+	if err != nil {
+		return r.wrapErrorWithStatusUpdate(log, record, r.setStatusFailed(targetClusterNotReady), err,
+			"failed to get the member cluster '%s'", recAccount.TargetCluster)
+	}
+
+	// get UserAccount from member
+	nsdName := namespacedName(fedCluster.OperatorNamespace, record.Name)
+	userAccount := &toolchainv1alpha1.UserAccount{}
+	err = fedCluster.Client.Get(context.TODO(), nsdName, userAccount)
```
nurali-techie: I like the `fedCluster.Client.Get()`, it feels similar to getting object from normal client.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
-	// Do nothing for now
+	for _, account := range userRecord.Spec.UserAccounts {
+		err = r.ensureUserAccount(reqLogger, account, userRecord)
+		if err != nil {
+			reqLogger.Error(err, "unable to synchronize with member UserAccount")
+			return reconcile.Result{}, err
+		}
+	}
+
 	return reconcile.Result{}, nil
```
nurali-techie: should we set status=`Ready+True` here ??

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -95,6 +101,111 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, err
 	}
 
-	// Do nothing for now
+	for _, account := range userRecord.Spec.UserAccounts {
+		err = r.ensureUserAccount(reqLogger, account, userRecord)
+		if err != nil {
+			reqLogger.Error(err, "unable to synchronize with member UserAccount")
+			return reconcile.Result{}, err
```
nurali-techie: This comment is about brainstorming with team:

Should we add delay for next `Reconcile()` call with `RequeueAfter`??  Basically this `err` could be originated from member_cluster (remote call) and we may need to treat it differently ??  The probable error could be n/w issue and re-trying in short delay would not make sense ??

MatousJobanek: This is a very big topic. In one of my versions of synchronization, I had different strategies (with different delays) for different types of error. In KubeFed they use something similar so I took the idea from them. The reasons why I haven't introduced it in our code yet are:
* I wanted to start with simple code and make it more complex later - so for now just use errors and default delays
* I think that it would be better to firstly know which types of errors we can face and based on this knowledge we can bring the strategies.

I'll create an Jira issue for this

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: not yet - this is part of the synchronization pr

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: I'm blind - thanks: 8d25c79

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: good point, no idea why my IDE didn't scream: 7d65f1e

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: renamed: 53e9ca6

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: changed: 53e9ca6

MatousJobanek:
@dipak-pawar I added the e2e tests here 5c47f3f  (I forgot to remove the replace clause from the go.mod file but I fixed it here: 21b9d41)

pkg/controller/masteruserrecord/masteruserrecord_controller.go
alexeykazakov: I agree. Let's keep it simple for now and improve later (if needed).
It seems that OpenShift (k8s?) already have some mechanism which increases delays between failed reconciles between attempts.
I noticed that if there is a constant failure then the reconcile is invoked with increasing delay (something like NextDelay = PreviousDelay * 2) and I think it's good enough for all our errors for now.

PR #44, opened 2019-08-12
fix: use scheme with registered types from api repo

MatousJobanek:
Use scheme with registered types from api repo
see codeready-toolchain/api#25


pkg/apis/apis.go
```diff
@@ -1,13 +1,12 @@
...
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// AddToSchemes may be used to add all resources defined in the project to a Scheme
-var AddToSchemes runtime.SchemeBuilder
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme) error {
-	return AddToSchemes.AddToScheme(s)
+	return apis.AddToSchemes.AddToScheme(s)
```
nurali-techie: What's the motivation for this change ??  why we want to reuse such stuff ?? .. because I see that in future we may have different set of scheme required for api/host/member and in such case we need to revert this changes.

MatousJobanek: why we would need to revert the change? see my PR for the member repo https://github.com/codeready-toolchain/member-operator/pull/43/files#diff-2384f451d80dcad3f603d0c962d38f6a there is used another scheme, so you can still have different sets of schemes in the operators.
As for the motivation - those schemes registered in API are the same for both operators, so it looks logical to register them at one place and reuse it either here or in e2e tests.

MatousJobanek: if you don't want to reuse it, then there is no point in having the `apis` package in api repo at all https://github.com/codeready-toolchain/api/tree/master/pkg/apis (I'm talking about `apis` package, not `v1alpha1` that contains the types

nurali-techie: Ok with this change https://github.com/codeready-toolchain/member-operator/pull/43/files#diff-2384f451d80dcad3f603d0c962d38f6a it allows operator specific type registration.  This is what make sense :+1:

PR #48, opened 2019-08-14
fix: change role to have full permissions for toolchain.dev.openshift.com group

MatousJobanek:
With the original configuration, it was failing for e2e tests with `no RBAC policy matched` error

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -60,14 +60,6 @@ rules:
 - apiGroups:
   - toolchain.dev.openshift.com
   resources:
-  - masteruserrecords
-  - nstemplatetiers
-  - usersignups
+  - '*'
   verbs:
   - '*'
-- apiGroups:
-  - toolchain.dev.openshift.com
-  resources:
-  - usersignups/status
-  verbs:
-  - update


MatousJobanek:
Currently, there are no e2e tests in host-operator repository - that's why they are green in CI :-).
I'm working on e2e tests for the creation of UserAccount and I wasn't able to understand why it was always failing (doing nothing) and after some time I figured out that it was the problem with RBAC. 
Itl screamed that it didn't have permissions to modify MasterUserRecord (which was on the list), but when I changed it to `*` it started working - don't ask me why :-/ 
But apart from that, the original file contained unnecessary duplication and I think having `*` is fine and will cover all future renaming of the CRDs

alexeykazakov:
Hm.. ok :)
And I agree, '*' for our own group is fine.

nurali-techie:
@MatousJobanek may be for sub-resource `status` it need to specify separately (in case RBAC is coming for status).  But fine with `*` anyway for our resources .. but I would like to see such changes comes along with e2e test PR rather separate PR just to make such changes.  May be for future changes :+1:

MatousJobanek:
> but I would like to see such changes comes along with e2e test PR rather separate PR just to make such changes.

And I thought that it would be better to push it separately in case anyone else has the same problem so (s)he doesn't have to wait for implementing/approving the whole e2e tests. 
In addition, this is not related only to e2e tests but to the whole functionality.

PR #50, opened 2019-08-19
 feat(CRT-176): add garbage collection for MasterUserRecord that deletes UserAccounts

MatousJobanek:
This PR:
* adds a finalizer to `MasterUserRecord` for the first reconcile (after the creation of `MUR` and before creating `UserAccouts`)
* deletes `UserAccounts` when the `MasterUserRecord` is going to be deleted
* removes the finalizer from `MasterUserRecord` after successful deletion of all `UserAccounts`
* adds unit tests as well as e2e test

### DIFF
--- a/test/e2e/masteruserrecord_test.go
+++ b/test/e2e/masteruserrecord_test.go
@@ -75,6 +75,36 @@ func TestMasterUserRecord(t *testing.T) {
 		// TODO: verify expected condition when the member operator has a logic that updates NsTemplateSet and its status
 		verifyResources(awaitility, extraMur, nil, expectingUaConditions(toBeProvisioned()))
 	})
+
+	t.Run("delete MasterUserRecord and expect UserAccount to be deleted", func(t *testing.T) {
+		// given
+		currentMur := NewHostAwaitility(awaitility).GetMasterUserRecord(mur.Name)
+
+		// when
+		err := awaitility.Client.Delete(context.TODO(), currentMur)
+
+		// then
+		require.NoError(t, err)
+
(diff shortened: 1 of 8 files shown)


test/e2e/masteruserrecord_test.go
```diff
@@ -73,6 +73,36 @@ func TestMasterUserRecord(t *testing.T) {
 		// TODO: verify expected condition when the member operator has a logic that updates NsTemplateSet and its status
 		verifyResources(awaitility, extraMur)
 	})
+
+	t.Run("delete MasterUserRecord and expect UserAccount to be deleted", func(t *testing.T) {
+		// given
+		currentMur := NewHostAwaitility(awaitility).GetMasterUserRecord(mur.Name)
```
alexeykazakov: 1. Wait for mur & user accounts to be created first.
2. Then delete and wait for all the resources to be deleted.

You already have 2 but missed 1. If you don't do 1 then 2 doesn't guarantee that you user accounts were deleted (maybe they controller hasn't created them yet).

MatousJobanek: I would say that 1. is part of the test "setup" at the beginning of the whole test: 
https://github.com/codeready-toolchain/host-operator/pull/50/files/564e8c7285b0150c7aef2714e5258ef582104585..e50c92515c2d042e29529eed3df2fbcccd50d060#diff-20c771505f30e89a2a3e8efdee8c003fR29-R36
notice that it's not a subtest, in other words, it will be executed every time even if you run only this subtest

alexeykazakov: Oh... right. I overlooked that. Never mind then :)

PR #58, opened 2019-08-28
chore(CRT-222): move e2e tests to toolchain-e2e

MatousJobanek:
* This PR moves all e2e tests to the toolchain-e2e repo - see this PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/1
* changes test makefile to execute e2e tests from toolchain-e2e repo
* adds a description to README describing the process and logic


MatousJobanek:
> How is the coverage report now generated with the e2e tests in a separate codebase?

I know that we've already discussed it in a meeting, but for transparency, I'm putting the explanation also here:
There is no code-coverage taken for the e2e tests currently and we don't plan any investigation in this area for now.
You can also see that in the codecov report - there is no change in the coverage after the test deletion https://github.com/codeready-toolchain/host-operator/pull/58#issuecomment-527431556

alexeykazakov:
> There is no code-coverage taken for the e2e tests currently and we don't plan any investigation in this area for now.

We do need to start generate code coverage reports for e2e test too (not just for unit tests as for now). But I agree that it should be a separate task not related to this e2e test restructure.

make/test.mk
```diff
@@ -73,82 +73,51 @@ HOST_NS := host-operator-$(shell date +'%s')
...
+		# get branch ref of the fork the PR was created from
+		$(eval BRANCH_REF := $(shell curl ${AUTHOR_LINK}/host-operator.git/info/refs?service=git-upload-pack --output - /dev/null 2>&1 | grep -a ${PULL_SHA} | awk '{print $$2}'))
+		@echo "detected branch ref ${BRANCH_REF}"
+		# check if a branch with the same ref exists in the user's fork of toolchain-e2e repo
+		$(eval REMOTE_E2E_BRANCH := $(shell curl ${AUTHOR_LINK}/toolchain-e2e.git/info/refs?service=git-upload-pack --output - /dev/null 2>&1 | grep -a ${BRANCH_REF} | awk '{print $$2}'))
+		@echo "branch ref of the user's fork: \"${REMOTE_E2E_BRANCH}\" - if empty then not found"
+		# check if the branch with the same name exists, if so then merge it with master and use the merge branch, if not then use master
+		if [[ -n "${REMOTE_E2E_BRANCH}" ]]; then \
+			# retrieve the branch name; \
+			BRANCH_NAME=`echo ${BRANCH_REF} | awk -F'/' '{print $$3}'`; \
```
alexeykazakov: You want to remove `\` from all the comments in this block, don't you? :) Unless you want to keep entire `then` block as a big comment.

MatousJobanek: no, the slash is needed there, otherwise, the next line is evaluated as a new Makefile line - not as part of the "one-line" bash command.

alexeykazakov: Ah right. But it means you can’t use `#` as comments here at all. Unless I’m missing something.

MatousJobanek: it's possible - the back-slack just escapes a new line. It works, in the same way as you would write a normal bash script - you can put comments there as well ;-) 
just paste this into your terminal and you will see that it's working:
```bash
if [[ true ]]; then \
echo "before comment"; \
# comment \
echo "after comment"; \
fi;
```
in bash, it actually works in the same way also without the back-slack behind the comment:
```bash
if [[ true ]]; then \
echo "before comment"; \
# comment 
echo "after comment"; \
fi;
```
but to be properly parsed by Makefile, I need to have it there in my case. 
The only thing I can remove is the semicolon at the end of the comments, the back-slash has to stay there

MatousJobanek: but if the comments are confusing, I can remove them completely - I just thought that it would be good to have the steps commented :-)

alexeykazakov: No, no, keep the comments. They are useful.
I just missed `;` in the end of the comments which separates the comment from the following code in the same line. Bash was never my strong side :)

So, all looks good here.

alexeykazakov: Oh, `;` are actually not needed. Bash seems to be smart enough to convert
```
# some comment \
some code;
```
to `# some comment; some code;`

Anyway, the code looks good.

README.adoc
```diff
@@ -43,6 +43,25 @@ There are a few more targets that you can find useful:
...
+The logic that executes tests in openshift-ci automatically tries to pair PR opened for this (host-operator) repository with a branch that potentially exists in the developer's fork of the https://github.com/codeready-toolchain/toolchain-e2e[toolchain-e2e] repo. This pairing is based on a branch name.
+
+For example, if a developer with GH account `cooljohn` opens a PR (for host-operator repo) from a branch `fix-reconcile`, then the logic checks if there is a branch `fix-reconcile` also in the `cooljohn/toolchain-e2e` fork.
+If there is, then the logic:
+
+1. clones latest changes from https://github.com/codeready-toolchain/toolchain-e2e[codeready-toolchain/toolchain-e2e]
+2. fetches the `fix-reconcile` branch from `cooljohn/toolchain-e2e` fork
+3. merges `master` branch with the changes from `fix-reconcile` branch
+4. runs e2e tests for both operators from this merged branch. As deployment for `host-operator` it uses the code from the PR.
```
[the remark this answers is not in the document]
MatousJobanek: added: e51ec5e

README.adoc
```diff
@@ -43,6 +43,25 @@ There are a few more targets that you can find useful:
 
 === End-to-End tests
 
+==== Background & pairing
+
+E2E tests are not located in this repository - all e2e tests are in https://github.com/codeready-toolchain/toolchain-e2e[toolchain-e2e] repo, however, it's still possible to run them locally from this repo - see <<Running End-to-End Tests>>.
```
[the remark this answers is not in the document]
MatousJobanek: Yes, there are duplications indeed. But each readme is looking on the logic from a different point of view. 
But as you said, when we have the flow finished and pushed to all repos, I'll try to consolidate/merge the READMEs to minimize the duplications.

PR #64, opened 2019-09-18
add sample templates to be created at operator startup 

xcoulon:
First step to the resolution of https://jira.coreos.com/browse/CRT-231: use `go-bindata` to wrap 2 Template Tier resources (yaml files).

Fixes CRT-257


xcoulon:
I'm proposing this first PR so we have a couple of NSTemplateTier resources which will be created/updated during the operator startup in a subsequent PR. Cutting the work in smaller bits to ease and reduce the the review cycles ;)

xcoulon:
> I probably don't understand the purpose of the NSTemplateTiers provided in this PR. I thought that we would embed only the templates and the corresponding NSTemplateTiers will be created during the operator startup. see the detailed doc part: https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.reonf823irn
> In other words, instead of storing the whole NSTemplateTiers in one file we would store the templates in multiple files.
> Just to be clear, I'm not saying that this approach is wrong and that one is better, so if you see any benefit of having everything in one file, then I'm fine ;-)

ah right, that's a very good point! 🤦‍♂ 
Let me change these yaml file to pure template manifests, and in a subsequent PR, I'll generate the `NSTemplateTier` manifest to be applied at startup!

make/go.mk
```diff
@@ -19,4 +19,9 @@ $(OUT_DIR)/operator:
 
 .PHONY: vendor
 vendor:
-	$(Q)go mod vendor
\ No newline at end of file
+	$(Q)go mod vendor
+
+.PHONY: generate
```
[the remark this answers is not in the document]
xcoulon: yes, if we exclude the generated file, it would make sense to include the `generate` as a dependency to the `build` goal

openshift-ci/Dockerfile.tools
```diff
@@ -52,6 +51,10 @@ RUN curl -L -s https://github.com/operator-framework/operator-sdk/releases/downl
 
 RUN mkdir -p ${GOPATH}/src/${GO_PACKAGE_PATH}/
 
+RUN go get -u github.com/go-bindata/go-bindata/...
```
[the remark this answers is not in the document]
xcoulon: right, let me add the dependency in `tools.go` as we have for other generators, instead

go.mod
```diff
@@ -53,3 +54,5 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20180711000925-0cf8f7e6ed1d
 	k8s.io/kubernetes => k8s.io/kubernetes v1.14.6
 )
+
+go 1.13
```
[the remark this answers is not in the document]
xcoulon: no, that's a left over that happened locally when re-installing `operator-sdk` via the `brew` pkg manager on macOS. I'll revert that change

pkg/templates/template_bindata.go
```diff
@@ -0,0 +1,610 @@
+// Package templates Code generated by go-bindata. (@generated) DO NOT EDIT.
```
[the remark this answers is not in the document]
xcoulon: ok ;)

pkg/templates/basic.yml
```diff
@@ -0,0 +1,93 @@
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: NSTemplateTier
+metadata:
+  name: basic
+spec:
+  namespaces:
+    - type: ide
+      revision: abcdef
```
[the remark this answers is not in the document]
xcoulon: as @MatousJobanek pointed in https://github.com/codeready-toolchain/host-operator/pull/64#pullrequestreview-290379034, I'm going to refactor these manifests

pkg/templates/basic.yml
```diff
@@ -0,0 +1,93 @@
...
+              provider: codeready-toolchain
+              version: ${COMMIT}
+            name: ${PROJECT_NAME}
+        parameters:
+        - name: PROJECT_NAME
+          required: true
+        - name: USER_NAME
+          required: true
+        - name: COMMIT
+          required: true
```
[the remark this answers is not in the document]
xcoulon: ok, let me replace `${PROJECT_NAME}` with `${USERNAME}` but let's address the `revision`/`commit` param in a subsequent since I'm going to refactor the manifests anyways

pkg/templates/basic.yml
```diff
@@ -0,0 +1,93 @@
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: NSTemplateTier
+metadata:
+  name: basic
+spec:
+  namespaces:
+    - type: ide
```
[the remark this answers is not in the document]
xcoulon: ok, sure

pkg/templates/basic.yml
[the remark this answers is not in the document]
xcoulon: done in fafe224, while extracting the templates from the NSTemplateTier manifests

pkg/templates/basic.yml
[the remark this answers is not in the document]
xcoulon: outdated, since in fafe224, the templates were extracted from the NSTemplateTier manifests. The `revision` value will be set during an subsequent PR

pkg/templates/template_bindata.go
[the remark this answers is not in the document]
xcoulon: done in fafe224

go.mod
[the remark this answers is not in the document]
xcoulon: done in f911523

openshift-ci/Dockerfile.tools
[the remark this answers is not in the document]
xcoulon: fixed in fafe224

make/go.mk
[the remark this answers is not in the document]
xcoulon: done in fafe224

xcoulon:
> > I probably don't understand the purpose of the NSTemplateTiers provided in this PR. I thought that we would embed only the templates and the corresponding NSTemplateTiers will be created during the operator startup. see the detailed doc part: https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.reonf823irn
> > In other words, instead of storing the whole NSTemplateTiers in one file we would store the templates in multiple files.
> > Just to be clear, I'm not saying that this approach is wrong and that one is better, so if you see any benefit of having everything in one file, then I'm fine ;-)
> 
> ah right, that's a very good point! 🤦‍♂
> Let me change these yaml file to pure template manifests, and in a subsequent PR, I'll generate the `NSTemplateTier` manifest to be applied at startup!

@MatousJobanek I extracted the Template manifests in fafe224, and I'll take care of generating the `NSTemplateTier` ones after this PR

templates/advanced-code.yml
```diff
@@ -0,0 +1,24 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-code
+      openshift.io/display-name: ${USERNAME}-code
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+      version: ${REVISION}
```
MatousJobanek: I'm not sure if we should add the version here, because in the NsTemplateSet controller, there is also added the `revision` label. based on that label we will know if the namespace should be provisioned/updated or not. I know that it's different label - `version` vs `revision` but it can be very confusing... WDYT?

go.mod
```diff
@@ -52,4 +53,4 @@ replace (
 	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.0.0-20190228175259-3e0149950b0e
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20180711000925-0cf8f7e6ed1d
 	k8s.io/kubernetes => k8s.io/kubernetes v1.14.6
-)
+)
```
MatousJobanek: looks like weird formatting

xcoulon: I believe this was fixed in 1ab41c6

templates/advanced-code.yml
xcoulon: ok, so we can remove it from the templates here, and add them at the `NSTemplateTier` level:
```
kind: NSTemplateTier
      metadata:
          name: basic
      spec:
    namespaces:
    - type: code
    	template: <yaml-with-ns-template>
    	revision: y8f907f6
    - type: dev
    	template: <yaml-with-ns-template>
    	revision: f8q907f4
    - type: stage
    	template: <yaml-with-ns-template>
    	revision: 907fy8f6
    - type: default
    	template: <yaml-with-ns-template>
    	revision: 23dfrt8
```
Later, the `NSTemplateSetController` will use this `revision` field from the `NSTemplateTier` to compare with the label set on the corresponding Namespace to determine if an update is needed. Is it ok for you?

MatousJobanek: yup :+1:

xcoulon: ok, so let me remove those fields here, then ;)

xcoulon: done in e8ea902

templates/basic-code.yml
```diff
@@ -0,0 +1,20 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-code
+      openshift.io/display-name: ${USERNAME}-code
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+          name: ${USERNAME}
```
MatousJobanek: Is this formatting correct?
```suggestion
    name: ${USERNAME}
```

templates/basic-stage.yml
```diff
@@ -0,0 +1,21 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-stage
+      openshift.io/display-name: ${USERNAME}-stage
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+          name: ${USERNAME}
```
MatousJobanek:
```suggestion
    name: ${USERNAME}
```

templates/basic-dev.yml
```diff
@@ -0,0 +1,20 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-dev
+      openshift.io/display-name: ${USERNAME}-dev
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+          name: ${USERNAME}
```
MatousJobanek:
```suggestion
    name: ${USERNAME}
```

templates/advanced-code.yml
```diff
@@ -0,0 +1,20 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-code
+      openshift.io/display-name: ${USERNAME}-code
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+    name: ${USERNAME}
```
alexeykazakov:
```suggestion
    name: ${USERNAME}-code
```

templates/advanced-dev.yml
```diff
@@ -0,0 +1,20 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-dev
+      openshift.io/display-name: ${USERNAME}-dev
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+    name: ${USERNAME}
```
alexeykazakov:
```suggestion
    name: ${USERNAME}-dev
```

templates/advanced-stage.yml
```diff
@@ -0,0 +1,21 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}-stage
+      openshift.io/display-name: ${USERNAME}-stage
+      openshift.io/requester: ${USERNAME}
+    labels:
+      provider: codeready-toolchain
+    name: ${USERNAME}
```
alexeykazakov:
```suggestion
    name: ${USERNAME}-stage
```

templates/basic-code.yml
alexeykazakov:
```suggestion
    name: ${USERNAME}-code
```

templates/basic-dev.yml
alexeykazakov:
```suggestion
    name: ${USERNAME}-dev
```

templates/basic-stage.yml
alexeykazakov:
```suggestion
    name: ${USERNAME}-stage
```

alexeykazakov: Yep. Need to fix formatting but also the namespace name. See my suggestion below.

templates/basic-dev.yml
alexeykazakov: Same here. See my suggestion below.

templates/basic-code.yml
alexeykazakov: See my suggestion below.

xcoulon:
thanks for the review @alexeykazakov and @MatousJobanek, and in particulat, thanks for catching the formatting and naming problems. I've applied all of @alexeykazakov's suggestions. Could you give it a look again?

xcoulon:
thanks for the reviews, @MatousJobanek and @alexeykazakov 🙌

PR #67, opened 2019-09-20
generate metadata with git short commits

xcoulon:
use a new makefile goal to generate an extra
yaml file in the `deploy/nstemplatetiers` directory,
which contains an entry for each template manifest,
along with the latest commit (short) hash.

This file is then included in the generated
`pkg/templates/templates_content.go` file
so the metadata can be retrieved later

this generated `deploy/nstemplatetiers/metadata.yaml`
is excluded from the git repository

fixes CRT-262


make/go.mk
```diff
@@ -21,8 +21,21 @@ $(OUT_DIR)/operator:
...
 	@echo "generating templates bindata..."
 	@go install github.com/go-bindata/go-bindata/...
-	@$(GOPATH)/bin/go-bindata -pkg templates -o ./pkg/templates/template_contents.go -nocompress -prefix deploy/nstemplatetiers deploy/nstemplatetiers
\ No newline at end of file
+	@$(GOPATH)/bin/go-bindata -pkg templates -o ./pkg/templates/templates.go -nocompress -prefix $(NSTEMPLATES_DIR) $(NSTEMPLATES_DIR)
+
+.PHONY: generate-metadata
+generate-metadata:
+	@echo "generating namespace templates metadata in $(YAML_TEMPLATES)" 
+	@-rm $(NSTEMPLATES_DIR)/metadata.yaml
```
dipak-pawar: can we silently ignore when `metadata.yaml` is  absent, I can see
```
rm: cannot remove 'deploy/nstemplatetiers/metadata.yaml': No such file or directory
make: [make/go.mk:36: generate-metadata] Error 1 (ignored)
```

xcoulon: yes: if the command starts with a `-`, the error is ignored. (see https://www2.ipp.mpg.de/~dpc/gmake/make_47.html)

MatousJobanek: I think that @dipak-pawar meant ignoring also the error message: 
```
@-rm $(NSTEMPLATES_DIR)/metadata.yaml 2> /dev/null
```
but that's a detail

xcoulon: ah sorry, I misunderstood @dipak-pawar 's comment. That's a good idea, actually! I'll add that in a subsequent PR ;)

PR #72, opened 2019-09-27
chore(CRT-229): add targets that help with deploying/installing operator

MatousJobanek:
Adds make targets for using resources manipulating with CSV and CRDs
builds on top of: https://github.com/codeready-toolchain/host-operator/pull/73
Paired e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/9

- [x] update README


make/docker.mk
```diff
@@ -1,4 +1,11 @@
+IMAGE := docker.io/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}:${GIT_COMMIT_ID_SHORT}
```
alexeykazakov: Let's switch to quay.io.
Also, wouldn't we need to create a `codeready-toolchain` repository in quay.io/docker.io for `docker push` and then grand permissions our developers to push to it?

Also, since it's purely for development env I would rather create a dev repo for these images. And maybe moved it to dev.mk... (but I'm not sure about dev.mk though).

Alternatively we could push to the cluster registry instead of quay.io but not sure if it's easy to do... So, the flow would be:
1. Setup a dev cluster (`export KUBECONFIG=...`, etc)
2. `make deploy-csv` which will login/push to the dev cluster (from step 1) image registry then deploy the csv.

make/dev.mk
```diff
@@ -68,4 +68,14 @@ add-member-to-host:
...
-	@${ADD_CLUSTER_SCRIPT_PATH} host host-cluster
\ No newline at end of file
+	@${ADD_CLUSTER_SCRIPT_PATH} host host-cluster
+
+.PHONY: deploy-csv
+## Create the test namespace
+deploy-csv: docker-push
+	sed -e 's|REPLACE_IMAGE|${IMAGE}|g' hack/deploy_csv.yaml | oc apply -f -
+
+.PHONY: install-operator
```
alexeykazakov: It probably could depend on `deploy-csv`, right? So, `make install-operator` is enough to deploy the CSV and install it.

make/dev.mk
```diff
@@ -68,4 +68,14 @@ add-member-to-host:
 .PHONY: add-host-to-member
 ## Run script add-cluster.sh host host-cluster
 add-host-to-member:
-	@${ADD_CLUSTER_SCRIPT_PATH} host host-cluster
\ No newline at end of file
+	@${ADD_CLUSTER_SCRIPT_PATH} host host-cluster
+
+.PHONY: deploy-csv
+## Create the test namespace
```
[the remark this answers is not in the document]
MatousJobanek: fixed 0f85a44

make/dev.mk
```diff
@@ -68,4 +68,14 @@ add-member-to-host:
...
+deploy-csv: docker-push
+	sed -e 's|REPLACE_IMAGE|${IMAGE}|g' hack/deploy_csv.yaml | oc apply -f -
+
+.PHONY: install-operator
+## Create the test namespace
```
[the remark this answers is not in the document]
MatousJobanek: fixed 0f85a44

make/dev.mk
MatousJobanek: done 0f85a44

make/docker.mk
MatousJobanek: Yes, this is something I wanted to discuss with the whole team during ou meeting. For now, we use minishift for local dev/testing, but it's getting to be complicated as we more and more rely on OS 4.2 functionality.

> Also, wouldn't we need to create a codeready-toolchain repository in quay.io/docker.io for docker push and then grand permissions our developers to push to it?

that was one of my first ideas, but then I realized that pushing to the same repository (quay namespace) is not a good idea. Consider case when two devs are both pushing at the same time - we would end up with rewriting latest image of the other dev. Sure, we could solve it by tags or by modified name, but it doesn't look so nice to me.
I would rather think of exporting a variable eg. `QUAY_NAMESPACE` in all dev's PCs where the variable would contain quay username of the developer. Then, push the image to quay.io/${QUAY_NAMESPACE}/host-operator:dev That means that the devs would push the images to their own namespaces, which would ensure:
* isolation 
* the fact that it's purely dev thingy
* easy permission management

As for the cluster registry - yes, I was trying to do that, but I failed with configuring my docker daemon to ignore the unknown certificates. There is probably some error between my chair and keyboard, but I wasn't able to push anything to cluster registry no matter what I set to any of the configuration files of docker daemon.

alexeykazakov: Yeah.. `QUAY_NAMESPACE` env var could be a good solution. Isolation is a good point.
+1 to that.

> As for the cluster registry - yes, I was trying to do that, but I failed with configuring my docker daemon to ignore the unknown certificates.

@dipak-pawar maybe you know some quick-fix for that problem? :)

MatousJobanek: > @dipak-pawar maybe you know some quick-fix for that problem? :)

just letting you know that I tried really everything:
* adding it as `insecure-registries` to `/etc/docker/daemon.json`
* adding certificates to `/etc/docker/certs.d/*devcluster.openshift.com/` dirs
* adding it as a `--insecure-registry` option to `/etc/sysconfig/docker`
* setting `--tlsverify` to `false` when running `docker login`
* and maybe else more that I don't remember...

dipak-pawar: @MatousJobanek, I already tried using docker/podman, it worked for me.
Steps Followed:

1.  Make sure that you have publicly exposed registry route created using 
`oc get routes -n openshift-image-registry`. 
2. If you don't see publicly exposed route, in above command output, create it using following command:
` oc patch configs.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge`
3. Login to cluster with your user e.g. `oc login -u your_user -p your_password`
3. Now you can try using podman like following:
```
HOST=$(oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}')
podman login -u $(oc whoami) -p $(oc whoami -t) --tls-verify=false $HOST
podman tag docker.io/library/nginx:latest default-route-openshift-image-registry.apps.dpawar-1570073282.devcluster.openshift.com/default/nginx:latest
podman push default-route-openshift-image-registry.apps.dpawar-1570073282.devcluster.openshift.com/default/nginx:latest --tls-verify=false
```
OR
4. we can try using docker  by updating ca trust store using following

```
to_registry=$(oc get images.config.openshift.io/cluster  -o jsonpath={.status.externalRegistryHostnames[0]})

oc get secret router-certs-default -n openshift-ingress -o json |jq -r '.data["tls.crt"]' | base64 -d >ca.crt
sudo cp ca.crt /etc/pki/ca-trust/source/anchors/${to_registry}.crt
sudo update-ca-trust enable
sudo systemctl daemon-reload
sudo systemctl restart docker
docker login -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps.dpawar-1570073282.devcluster.openshift.com 
docker push default-route-openshift-image-registry.apps.dpawar-1570073282.devcluster.openshift.com/openshift/hello-openshift-1
```

or update docker configuration to use insecure registry as our cluster registry(I haven't tried this)

MatousJobanek: awesome, it works - I did the same steps but didn't add the certs to pki ca-trust and tried only to `/etc/docker/certs.d/` however, it's a good point and it's much better approach. No idea why I haven't tried it :-/...
Anyway, thanks a lot @dipak-pawar

MatousJobanek: @alexeykazakov please check last changes: b486b90

README.adoc
```diff
@@ -40,6 +40,44 @@ There are a few more targets that you can find useful:
...
+
+```bash
+$ export QUAY_NAMESPACE=<quay-username>
+```
+
+Then, to install the operator run:
+```bash
+$ make install-operator
+```
+
```
alexeykazakov: There was one more step required for me. `make install-operator` created `host-operator` repository in my quay.io but after that I had to go to quay.io configuration and made that repo public (it was created as private). Until then OS could not pull the image.
Let's mention this step here too.

PR #74, opened 2019-10-02
Fix make deploy-crd

alexeykazakov:


xcoulon:
> Looks good, but wouldn't be better to run `oc apply -f deploy/crds/` (for the whole directory) so we avoid such problems when we rename the CRD again?
> I forgot that there is a target calling apply for every file separately

right, I didn't know we can run the `oc apply -f` on a directory. That would make this Makefile goal more robust, indeed

alexeykazakov:
> Looks good, but wouldn't be better to run oc apply -f deploy/crds/ (for the whole directory) so we avoid such problems when we rename the CRD again?

I was thinking about that too but we might want to keep something in this directory which we don't want to apply. But it's probably better to apply to entire dir now and switch back to individual files when/if we need to exclude something in the future.

alexeykazakov:
> Looks good, but wouldn't be better to run oc apply -f deploy/crds/ (for the whole directory) so we avoid such problems when we rename the CRD again?

Done in e778a8b503e7b3c8ffb4fb0769dc8da7b099aa10

MatousJobanek:
yup, we do it in e2e tests https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L157 so I think when we have something else in the dir, we will hit this problem in e2e tests first :-) 
Anyway, when we fully move to installing CRDs only via CSV, we won't need this `apply` command any more

MatousJobanek:
As for the build error - it looks like openshift CI started using go 1.13 - I see the same error when I'm building the project with `1.13` but not with `1.12`. 
Let me fix that ;-)

PR #78, opened 2019-10-04
create/update NSTemplateTiers at controller startup

xcoulon:
call the `CreateOrUpdateResources` func during the
controller startup, but only after the client cache
has been initialized, otherwise, the `client.Get()`
func will return an error, even if the resource exists
on the cluster.

Fixes CRT-263


pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -0,0 +1,235 @@
...
+				if obj, ok := obj.(*toolchainv1alpha1.NSTemplateTier); ok {
+					if obj.ObjectMeta.ResourceVersion != "foo-"+obj.ObjectMeta.Name {
+						return errors.Errorf("a matching 'resourceVersion' field must be returned by the client during object update")
+					}
+					obj.ObjectMeta.ResourceVersion = "bar-" + obj.ObjectMeta.Name
+				}
+				return client.Client.Update(ctx, obj)
+			}
+			assets, revisions := generateRevisions("123456")
+			err := nstemplatetiers.CreateOrUpdateResources(s, client, namespace, assets)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

xcoulon: fixed in 78c1b4d

cmd/manager/main.go
```diff
@@ -154,6 +153,14 @@ func main() {
 
 	log.Info("Starting the Cmd.")
 
+	go func() {
+		mgr.GetCache().WaitForCacheSync(stopChannel)
```
alexeykazakov:
```suggestion
		if !mgr.GetCache().WaitForCacheSync(stopChannel) {
			log.Error(errors.New("timed out waiting for caches to sync"), "")
			os.Exit(1)
		}
```

cmd/manager/main.go
```diff
@@ -154,6 +153,14 @@ func main() {
...
+		// create or update all NSTemplateTiers on the cluster at startup
+		if err := nstemplatetiers.CreateOrUpdateResources(mgr.GetScheme(), mgr.GetClient(), namespace, nstemplatetiers.Asset); err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
+	}()
```
alexeykazakov: I'm a bit concerned that we run this in a separate go routine because of two reasons:
1. We create tiers asynchronously and I guess theoretically tiers might not be available when we need them. But probably our reconcile loops will handle it properly...
2. If something goes wrong then we exit here (but we don't know when it can happen). I'm not sure how all the controller update flow works... does it have any health check thing, etc so we can control rolling updates, etc.

But this is just something to keep in mind for now. I don't see how we could better.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -10,15 +11,57 @@ import (
...
+			if err != nil {
+				return errors.Wrapf(err, "unable to get the NSTemplateTiers '%s' in namespace '%s'", tier.Name, tier.Namespace)
+			}
+			// retrieve the current 'resourceVersion' before updating the resource
+			tier.ObjectMeta.ResourceVersion = existing.ObjectMeta.ResourceVersion
+			if err := client.Update(context.TODO(), tier); err != nil {
+				return errors.Wrapf(err, "unable to update the NSTemplateTiers '%s' in namespace '%s'", tier.Name, tier.Namespace)
+			}
+			log.Info("NSTemplateTier resource updated", "namespace", tier.Namespace, "name", tier.Name)
+		}
```
alexeykazakov: Add `} else {`. Otherwise you log two messages in case of updating:
```
log.Info("NSTemplateTier resource updated", "namespace", tier.Namespace, "name", tier.Name)
...
log.Info("NSTemplateTier resource created", "namespace", tier.Namespace, "name", tier.Name)
```
Which is confusing.

xcoulon: ah right! 🤦‍♂

xcoulon: fixed in beb7b99

cmd/manager/main.go
xcoulon: @alexeykazakov I understand your concerns. 

For 1., yes, our reconcile loop on the member-cluster operator should requeue the request if the NSTemplateTier resources are not available
For 2., the controller also waits for the cache to be in sync during its `Start` (see https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L140-L147), so I doubt we may face the situation where the go routine above kills the controller while it is already processing a request.

cmd/manager/main.go
```diff
@@ -7,28 +7,28 @@ import (
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
-
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/rest"
+	"github.com/pkg/errors"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

xcoulon: fixed in ee5d0c6

cmd/manager/main.go
alexeykazakov: > would it be too naive to wait for the cache to have it in sync before creating the tiers and starting the controllers?

@MatousJobanek what do you mean?

xcoulon: @MatousJobanek I thought that the controllers started in the last block of this `main()` func, ie:
```
if err := mgr.Start(stopChannel); err != nil {
	log.Error(err, "Manager exited non-zero")
	os.Exit(1)
}
```
so I'm not sure what you mean here. Also, the controllers don't start before the cache is sync'd, as we can see in https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L140-L158. IMO, the controllers will start and at the same time, the NSTemplateTier resources will be created. So yes, there might be a very tiny time window when the controllers are running but there is no NSTemplateTier resources available. But then that would be the responsibility of the member-operator's controllers to requeue the requests until the NSTemplateTier resources exist. Or am I missing something?

MatousJobanek: > what do you mean?

I meant to not start a separated goroutine and do it in the main one - which would mean to wait until the cache is synced directly in the main function - then create the tiers & start the controllers

> I thought that the controllers started in the last block of this main()

you are right, sorry for the confusion. I mixed it with the `StartKubeFedClusterControllers` function, but you don't have to care about this one.

>  So yes, there might be a very tiny time window

yes, I was referring to this tiny window


Anyway, you are right that we don't need to deal with it. I'm just trying to optimize it when it's not needed. 
So, you can ignore my comments above ;-).

alexeykazakov: > I meant to not start a separated goroutine and do it in the main one - which would mean to wait until the cache is synced directly in the main function - then create the tiers & start the controllers

Can't do that because otherwise we would create a dead lock :(

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	s := scheme.Scheme
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+
+	t.Run("ok", func(t *testing.T) {
+
+		t.Run("create only", func(t *testing.T) {
+			// given
+			namespace := "host-operator" + uuid.NewV4().String()[:7]
```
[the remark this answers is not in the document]
xcoulon: just for the sake of readability in the logs ;)

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -0,0 +1,236 @@
...
+					obj.ObjectMeta.ResourceVersion = "foo-" + obj.ObjectMeta.Name
+				}
+				return client.Client.Create(ctx, obj)
+			}
+			// verify that no "advanced" nor "basic" NSTemplateTier resources exist prior to creation
+			for _, tierName := range []string{"advanced", "basic"} {
+				tier := toolchainv1alpha1.NSTemplateTier{}
+				err = client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: tierName}, &tier)
+				require.Error(t, err)
+				assert.Equal(t, fmt.Sprintf("nstemplatetiers.toolchain.dev.openshift.com \"%s\" not found", tierName), err.Error())
```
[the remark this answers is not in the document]
xcoulon: ok, let me do that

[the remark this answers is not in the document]
xcoulon: done in 7cedaff

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -0,0 +1,237 @@
...
+	"fmt"
+	"testing"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/host-operator/pkg/apis"
+	"github.com/codeready-toolchain/host-operator/pkg/templates/nstemplatetiers"
+	testnstemplatetiers "github.com/codeready-toolchain/host-operator/test/templates/nstemplatetiers"
+	testsupport "github.com/codeready-toolchain/toolchain-common/pkg/test"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -0,0 +1,236 @@
...
+			// given
+			namespace := "host-operator" + uuid.NewV4().String()[:7]
+			client := testsupport.NewFakeClient(t)
+			// set the 'resourceVersion' when creating the object
+			client.MockCreate = func(ctx context.Context, obj runtime.Object) error {
+				if obj, ok := obj.(*toolchainv1alpha1.NSTemplateTier); ok {
+					if obj.ObjectMeta.ResourceVersion != "" {
+						return errors.Errorf("'resourceVersion' field must be specified by the server during object creation")
+					}
+					obj.ObjectMeta.ResourceVersion = "foo-" + obj.ObjectMeta.Name
```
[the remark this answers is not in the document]
xcoulon: I've changed the test to increment the `Generation` instead of changing the `Revision`, while trying to reproduce what the "real" server does.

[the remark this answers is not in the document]
xcoulon: > BTW, why you didn't have to do it in case of creating the inner objects of namespaces in the member operator? https://github.com/codeready-toolchain/member-operator/blob/master/pkg/template/processor.go#L60-L80

I'm not sure I understand what you mean?

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
xcoulon: fixed in https://github.com/codeready-toolchain/host-operator/pull/78/commits/70b6bff32d770efb333c6af9f16ebfd7c41fde5e

xcoulon:
> codeready-toolchain/toolchain-e2e#14

thanks @alexeykazakov!

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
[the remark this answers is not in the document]
xcoulon: @MatousJobanek is it better with https://github.com/codeready-toolchain/host-operator/pull/78/commits/e88f525cca19474f5da64795f80f59951808a27e ? and yes, at least on Minishift, I found that it's necessary to set the ResourceVersion from the object returned by `client.Get()` into the object to pass to `client.Update()` otherwise the server would respond with an error, as I commented in the code. I'll see if we need to do the same in https://github.com/codeready-toolchain/member-operator/blob/master/pkg/template/processor.go#L60-L80

PR #79, opened 2019-10-05
Add 'tier' as short name for NSTemplateTier CRD

alexeykazakov:
This PR adds a short name 'tier' for NSTemplateTier CRD. So we can do `oc get tier`.

Auto-generated by https://github.com/codeready-toolchain/api/pull/41


deploy/crds/toolchain_v1alpha1_nstemplatetier_crd.yaml
```diff
@@ -10,6 +10,8 @@ spec:
   names:
     kind: NSTemplateTier
     plural: nstemplatetiers
+    shortNames:
```
xcoulon: should we also have `singular: nstemplatetier` to be consistent with the plural name?

MatousJobanek: isn't it set by default - that you can use lowercase `nstemplatetier` without specifying it in `shortNames`?

alexeykazakov: Yes. `nstemplatetier` works OOTB. No need to add it to `shortNames` explicitly.

xcoulon: ok, sorry for the dumb question 🤦‍♂

PR #80, opened 2019-10-08
add commit and buildtime vars

xcoulon:
Fixes CRT-282

### DIFF
--- a/cmd/manager/main_test.go
+++ b/cmd/manager/main_test.go
@@ -1,0 +1,16 @@
+package main_test
+
+import (
+	"testing"
+
+	main "github.com/codeready-toolchain/host-operator/cmd/manager"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVars(t *testing.T) {
+	// simply verify that the vars exist.
+	// They will be populated by the `-ldflags="-X ..."` at build time
+	assert.Equal(t, "unknown", main.Commit)
+	assert.Equal(t, "unknown", main.BuildTime)
+}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -33,9 +33,11 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost               = "0.0.0.0"
-	metricsPort         int32 = 8383
-	operatorMetricsPort int32 = 8686
+	metricsHost                = "0.0.0.0"
+	metricsPort         int32  = 8383
+	operatorMetricsPort int32  = 8686
+	Commit              string = "unknown"
+	BuildTime           string = "unknown"
 )
 
 var log = logf.Log.WithName("cmd")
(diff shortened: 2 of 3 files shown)


xcoulon:
> Same here as in the member PR. Let's add a unit test.

done in https://github.com/codeready-toolchain/host-operator/commit/44d73ddc9ac1cca8c6dac28239fba3a90ce06e39

cmd/manager/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+	main "github.com/codeready-toolchain/host-operator/cmd/manager"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVars(t *testing.T) {
+	// simply verify that the vars exist.
+	// They will be populated by the `-ldflags="-X ..."` at build time
+	assert.Equal(t, "unknown", main.Commit)
+	assert.Equal(t, "unknown", main.BuildTime)
```
nurali-techie: Not getting point of this test.  What's the objective and what it test.  What I see here is, is just asserting the variable are initialized with particular value and nothing else ??

xcoulon: @nurali-techie yes, this ensures that the variables exist. We can't do much at this level (unit testing), but in the future, we could have e2e tests to check the actual values (matching a commit hash and a datetime). Does this clarify?

xcoulon:
> # [Codecov](https://codecov.io/gh/codeready-toolchain/host-operator/pull/80?src=pr&el=h1) Report
> > Merging [#80](https://codecov.io/gh/codeready-toolchain/host-operator/pull/80?src=pr&el=desc) into [master](https://codecov.io/gh/codeready-toolchain/host-operator/commit/61d7c514b958a1f30f19e746d71dacba5f2af001?src=pr&el=desc) will **decrease** coverage by `11.21%`.
> > The diff coverage is `0%`.
> 
> ```diff
> @@             Coverage Diff             @@
> ##           master      #80       +/-   ##
> ===========================================
> - Coverage   82.06%   70.85%   -11.22%     
> ===========================================
>   Files           4        5        +1     
>   Lines         474      549       +75     
> ===========================================
>   Hits          389      389               
> - Misses         72      147       +75     
>   Partials       13       13
> ```
> 
> [Impacted Files](https://codecov.io/gh/codeready-toolchain/host-operator/pull/80?src=pr&el=tree)	Coverage Δ	
> [cmd/manager/main.go](https://codecov.io/gh/codeready-toolchain/host-operator/pull/80/diff?src=pr&el=tree#diff-Y21kL21hbmFnZXIvbWFpbi5nbw==)	`0% <0%> (ø)`

I don't understand how the coverage can drop by 11% with this commit!

cmd/manager/main_test.go
nurali-techie: Test for variable exists, doesn't make much sense for statically defined variable.  If these variable are not exists then there test will failed with compile error.

xcoulon: Sure, but this test is a tiny extra harness. 
We're not going to debate on whether we should keep or remove this test, are we?

alexeykazakov: As Xavier said the point is to make sure we have main.Commit and main.BuildTime available. We didn't before (and it was a bug). So, the test could be just as simple as:
```
var _ = main.Commit
var _ = main.BuildTime
```
But `assert.Equal(t, "unknown", main.Commit)` would work fine too.

PR #83, opened 2019-10-10
Exclude cmd/manager (main) from coverage report

alexeykazakov:
We can't really test cmd/manager properly in unit tests (it's coverd in e2e tests) so excluding it from the codecov report. See https://github.com/codeready-toolchain/host-operator/pull/80

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -26,7 +26,7 @@ test-with-coverage: generate
 	@echo "running the tests with coverage..."
 	@-mkdir -p $(COV_DIR)
 	@-rm $(COV_DIR)/coverage.txt
-	$(Q)go test -vet off ${V_FLAG} $(shell go list ./... | grep -v /test/e2e) -coverprofile=$(COV_DIR)/coverage.txt -covermode=atomic ./...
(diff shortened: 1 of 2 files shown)


make/test.mk
```diff
@@ -26,7 +26,7 @@ test-with-coverage: generate
 	@echo "running the tests with coverage..."
 	@-mkdir -p $(COV_DIR)
 	@-rm $(COV_DIR)/coverage.txt
-	$(Q)go test -vet off ${V_FLAG} $(shell go list ./... | grep -v /test/e2e) -coverprofile=$(COV_DIR)/coverage.txt -covermode=atomic ./...
+	$(Q)go test -vet off ${V_FLAG} $(shell go list ./... | grep -v /cmd/manager) -coverprofile=$(COV_DIR)/coverage.txt -covermode=atomic ./...
```
MatousJobanek: doesn't that mean that it won't run the tests inside `cmd/manager` directory? We have one small test there already: https://github.com/codeready-toolchain/host-operator/blob/master/cmd/manager/main_test.go and I'm not sure if skipping it is a good idea

alexeykazakov: No, no. It doesn't affect the tests itself. Only the coverage reporting. I've tested before creating a PR :)

MatousJobanek: Ok :+1:

xcoulon: good! I had the same concern as @MatousJobanek ;)

PR #86, opened 2019-10-14
support commit hashes as numbers

xcoulon:
in some rare cases, the short commit hash of a
namespace template file in deploy/templates/nstemplatetiers
may be a number.

Fixes CRT-293


pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -96,7 +96,7 @@ func TestParseAllRevisions(t *testing.T) {
 		assert.Equal(t, "123456c", revisions["advanced"]["stage"])
 		assert.Equal(t, "123456d", revisions["basic"]["code"])
 		assert.Equal(t, "123456e", revisions["basic"]["dev"])
-		assert.Equal(t, "123456f", revisions["basic"]["stage"])
+		assert.Equal(t, "1234567", revisions["basic"]["stage"])
```
nurali-techie: As I understood, commit_hash is string, it can be `alpha-numeric` or `numeric-only`.  Why do we need to handle commit_hash `numeric-only` in some special way ??

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -109,14 +110,19 @@ func parseAllRevisions(metadata []byte) (map[string]map[string]string, error) {
 		tierKind := data[0]
 		nsKind := data[1]
 		// create a new entry if needed
-		r, ok := revision.(string)
-		if !ok {
-			return nil, errors.Errorf("invalid namespace template filename revision for '%[1]s'. Expected a string, got a %[2]T ('%[2]v')", filename, revision)
+		var rev string
+		switch r := revision.(type) {
```
alexeykazakov: Can we just fix our metadata.yaml generation so it uses SHAs as strings?
```
advanced-code: "1234567"
```
Then we can unmarshal it to `map[string]string`

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
xcoulon: because the YAML parser will convert the value into an integer if it can, resulting in an error since the we expect a `string` in https://github.com/codeready-toolchain/host-operator/pull/86/files/4d2b183c98d113de5ba58161cd4c864c2e89c446#diff-6d635ca031a9d241865459f7ef30f167L112

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
xcoulon: I guess we could do that, but when we need to ensure that the generated file would have this format, and this will only fail when a commit hash is a number. So I'd rather ensure that we support both cases (alphanum or nums) at this level.

xcoulon:
@alexeykazakov @nurali-techie I refactored the code in https://github.com/codeready-toolchain/host-operator/pull/86/commits/c9e951dce54ba3d8451ec0d0f642b12c18f0fa22 to make sure that `metadata.yml` contains string values

PR #87, opened 2019-10-14
refactor Commit and BuildTime variables

xcoulon:
moving to the `version` pkg since these vars were
not correctly set by the `-ldflags="-X ..."`, probably
because of the `cmd/manager` pkg called `main`

also, use a `.dockerfile` to reduce the size of the uploaded
 context when building the Docker image

Fixes CRT-282


xcoulon:
I will apply the same changes in `member-operator` once this PR is approved ;)

PR #89, opened 2019-10-15
add missing permissions on host-operator service account

xcoulon:
fixes CRT-301


xcoulon:
I'm not sure how we can test these changes (besides checking the logs once the operator is deployed), so any suggestion is welcome

xcoulon:
@dipak-pawar @alexeykazakov I re-generated the CSV in https://github.com/codeready-toolchain/host-operator/pull/89/commits/c8c384d88e37522f95f2fbe7d947d03b49e8f10f

deploy/role.yaml
```diff
@@ -15,6 +15,27 @@ rules:
...
+  - processedtemplates
+  - replicationcontrollers
+  - limitranges
+  verbs:
+  - get
+  - list
+- apiGroups: 
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
```
dipak-pawar: @xcoulon, I don't think we need to put it in role, as it's clusterscoped resource, and it's already added in `cluster_role.yaml

deploy/role.yaml
```diff
@@ -15,6 +15,27 @@ rules:
   - secrets
   verbs:
   - "*"
+- apiGroups:
+  - ""
+  resources:
+  - podtemplates
+  - namespaces
+  - componentstatuses
+  - nodes
```
dipak-pawar: As per my understanding, `nodes/namespaces` can be put in `cluster_role.yaml` only, we don't need it to be available in the role

deploy/role.yaml
xcoulon: good catch, indeed. I missed the `cluster_role.yaml` file. So I guess the problem we've seen is due to something else, then

xcoulon: so let me revert the last 2 commits

deploy/role.yaml
alexeykazakov: Also wondering why would we need nodes even for reading/listing in cluster role... But maybe there are some indirect dependencies?

MatousJobanek: We should need only list permission for nodes - it's needed by KubeFed controller that checks the health of the clusters.
This is already part of the `cluster_role.yaml` file, so no need for additional changes

PR #91, opened 2019-10-24
Updated host operator to remove UserSignup.Spec.UserID

sbryzak:
This PR updates the host operator to remove usage of the `UserSignup.Spec.UserID` field.  Instead of using the `UserID` field to populate the MasterUserRecord name, the UserSignup.Name value is now used instead.

Fixes https://jira.coreos.com/browse/CRT-315

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -82,6 +82,8 @@ github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de h1:bExhq0r
 github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de/go.mod h1:/Is7p+crldTqYG0f8LQL7OqmdQRnMJ7u//N82tQdf28=
 github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285 h1:cEyMmDOG7f0+DDt+Kepj9UV0IRr+XLMfEGdL+0Sbhfg=
 github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285/go.mod h1:9uSs8YlapNzaxIubqp2P9TA4L4iO4gCuqW1rrhCwsV4=
+github.com/codeready-toolchain/api v0.0.0-20191023231919-300ac3dd8c36 h1:OZuIMbGR1jo2E0VIV1elaBaByZBqIElSBut9M88yOis=
+github.com/codeready-toolchain/api v0.0.0-20191023231919-300ac3dd8c36/go.mod h1:9uSs8YlapNzaxIubqp2P9TA4L4iO4gCuqW1rrhCwsV4=
 github.com/codeready-toolchain/toolchain-common v0.0.0-20191010043304-822e291d04cb h1:2Al5qMR1lSVpIbD10GIDyDg7OaX1KM4kggivVlg7fbc=
 github.com/codeready-toolchain/toolchain-common v0.0.0-20191010043304-822e291d04cb/go.mod h1:dyqWm9EXBYMXzdfaHqipJSujNiK5BvZ3YjukErym4BE=
 github.com/coreos/bbolt v1.3.0/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
--- a/go.sum
+++ b/go.sum
@@ -80,8 +80,6 @@ github.com/chai2010/gettext-go v0.0.0-20170215093142-bf70f2a70fb1/go.mod h1:/iP1
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de h1:bExhq0ri3+O/3rHHBtJIfJkEev9a9yDQgjD7XhMDln0=
 github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de/go.mod h1:/Is7p+crldTqYG0f8LQL7OqmdQRnMJ7u//N82tQdf28=
-github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285 h1:cEyMmDOG7f0+DDt+Kepj9UV0IRr+XLMfEGdL+0Sbhfg=
-github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285/go.mod h1:9uSs8YlapNzaxIubqp2P9TA4L4iO4gCuqW1rrhCwsV4=
 github.com/codeready-toolchain/api v0.0.0-20191023231919-300ac3dd8c36 h1:OZuIMbGR1jo2E0VIV1elaBaByZBqIElSBut9M88yOis=
 github.com/codeready-toolchain/api v0.0.0-20191023231919-300ac3dd8c36/go.mod h1:9uSs8YlapNzaxIubqp2P9TA4L4iO4gCuqW1rrhCwsV4=
 github.com/codeready-toolchain/toolchain-common v0.0.0-20191010043304-822e291d04cb h1:2Al5qMR1lSVpIbD10GIDyDg7OaX1KM4kggivVlg7fbc=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -82,6 +82,8 @@ github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de h1:bExhq0r
 github.com/codeready-toolchain/api v0.0.0-20191009081803-8f4e395353de/go.mod h1:/Is7p+crldTqYG0f8LQL7OqmdQRnMJ7u//N82tQdf28=
 github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285 h1:cEyMmDOG7f0+DDt+Kepj9UV0IRr+XLMfEGdL+0Sbhfg=
 github.com/codeready-toolchain/api v0.0.0-20191013054335-17f0782f9285/go.mod h1:9uSs8YlapNzaxIubqp2P9TA4L4iO4gCuqW1rrhCwsV4=
```
alexeykazakov: `go mod tidy`

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -113,7 +113,7 @@ func TestUserSignupWithManualApprovalApproved(t *testing.T) {
 			UID:       types.UID(uuid.NewV4().String()),
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			UserID:            "foo",
+			Username:          "foo",
 			CompliantUsername: "foo",
```
alexeykazakov: Let's use different ComplaintUserName (so it doesn't equal to Username).

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -178,7 +178,7 @@ func TestUserSignupWithNoApprovalPolicyTreatedAsManualApproved(t *testing.T) {
...
```
alexeykazakov: Same here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -242,7 +242,7 @@ func TestUserSignupWithManualApprovalNotApproved(t *testing.T) {
...
```
alexeykazakov: Same here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -288,7 +288,7 @@ func TestUserSignupWithAutoApprovalClusterSet(t *testing.T) {
...
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -353,7 +353,7 @@ func TestUserSignupWithMissingApprovalPolicyTreatedAsManual(t *testing.T) {
 			UID:       types.UID(uuid.NewV4().String()),
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			UserID:            "bar",
+			Username:          "bar",
 			CompliantUsername: "bar",
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -391,7 +391,7 @@ func TestUserSignupMURCreateFails(t *testing.T) {
 			UID:       types.UID(uuid.NewV4().String()),
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			UserID:            "foo",
+			Username:          "foo",
 			CompliantUsername: "foo",
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -425,7 +425,7 @@ func TestUserSignupMURCreateAlreadyExists(t *testing.T) {
...
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -474,7 +474,7 @@ func TestUserSignupMURReadFails(t *testing.T) {
 			UID:       types.UID(uuid.NewV4().String()),
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			UserID:            "foo",
+			Username:          "foo",
```
alexeykazakov: Here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -508,7 +508,7 @@ func TestUserSignupSetStatusApprovedByAdminFails(t *testing.T) {
...
 			CompliantUsername: "foo",
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -542,7 +542,7 @@ func TestUserSignupSetStatusApprovedAutomaticallyFails(t *testing.T) {
...
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -575,7 +575,7 @@ func TestUserSignupSetStatusNoClustersAvailableFails(t *testing.T) {
...
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -609,7 +609,7 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
 			Approved:          false,
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -662,7 +662,7 @@ func TestUserSignupNoMembersAvailableFails(t *testing.T) {
 			UID:       types.UID(uuid.NewV4().String()),
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			UserID:            "foo",
+			Username:          "foo",
 			CompliantUsername: "foo",
```
alexeykazakov: And here.

go.sum
sbryzak: Done

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed all instances of duplicate values.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -213,11 +213,11 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
 
 	mur := &toolchainv1alpha1.MasterUserRecord{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      userSignup.Name,
+			Name:      userSignup.Spec.CompliantUsername,
```
MatousJobanek: There are two things to be fixed:
1. when checking if the MUR already exists [here](https://github.com/codeready-toolchain/host-operator/pull/91/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R130) you should use `userSignup.Spec.CompliantUsername` in the namespaced name
2. because we are now using `compliantUsername` created on our own that is not unique, it can easily happen that such a MUR already exists so we would end-up here https://github.com/codeready-toolchain/host-operator/pull/91/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R133-R136. In that case, we would need to modify the name a bit (eg. ad suffix `-2`) to get the unique one

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -240,7 +240,7 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
 			"Error creating MasterUserRecord")
 	}
 
-	logger.Info("Created MasterUserRecord", "Name", userSignup.Name, "TargetCluster", targetCluster)
+	logger.Info("#### Created MasterUserRecord", "Name", mur.Name, "TargetCluster", targetCluster)
```
MatousJobanek: @sbryzak Do you know that you forgot to remove this diagnosis prefix?

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: @sbryzak I see that you fixed the first point but the second point is still missing. Could you please shed some light on why you don't do such a check or modification of MUR name? 
If the task is postponed to be done in a separated PR, do we have a Jira issue created for that?

PR #93, opened 2019-10-29
Removed CompliantUsername from UserSignup.Spec

sbryzak:
This PR contains changes made as a result of running `make generate` on the api module, after removing the `CompliantUsername` field from UserSignup.Spec.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/29

See this issue for more details:

https://jira.coreos.com/browse/CRT-330


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -240,8 +288,8 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
 			"Error creating MasterUserRecord")
 	}
 
-	logger.Info("#### Created MasterUserRecord", "Name", mur.Name, "TargetCluster", targetCluster)
-	return nil
+	logger.Info("Created MasterUserRecord", "Name", mur.Name, "TargetCluster", targetCluster)
+	return r.updateStatusCompliantUsername(userSignup)
```
alexeykazakov: Oh... if we fail here (during status update) then we are screwed. It means the MUR already exists but we don’t know about that and we will create another one with regenerated compliant username in the next reconcile cycle. And in case of some constant error (due to some bug for example) we would end up with a 100/1000 MURs created for the single UserSignup.

So we have to use userID label in MUR as I mentioned in my comment above. UserID is always available to us and it doesn’t change.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -378,3 +426,7 @@ func (r *ReconcileUserSignup) updateStatusConditions(userSignup *toolchainv1alph
 	}
 	return r.client.Status().Update(context.TODO(), userSignup)
 }
+
+func (r *ReconcileUserSignup) updateStatusCompliantUsername(userSignup *toolchainv1alpha1.UserSignup) error {
+	return r.client.Status().Update(context.TODO(), userSignup)
```
alexeykazakov: Reuse updateStatus() function here. It will help you with testing.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: That won't happen.. check the logic in `generateCompliantUsername()` again - if a MUR is found with the same UserID as the UserSignup, then that one will be used.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -193,6 +210,37 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+	replaced := strings.ReplaceAll(strings.ReplaceAll(instance.Spec.Username, "@", "-at-"), ".", "-")
+
+	errs := validation.IsQualifiedName(replaced)
+	if len(errs) > 0 {
+		return "", NewSignupError(fmt.Sprintf("transformed username [%s] is invalid", replaced))
+	}
+
+	transformed := replaced
+
+	for i := 1; i < 1001; i++ { // No more than 1000 attempts to find a vacant name
```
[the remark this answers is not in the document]
sbryzak: i'll reduce it to 100, good idea.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,10 +126,23 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
 
+	// Flag that we set to true if a compliantUsername is set
+	compliantUsername := instance.Status.CompliantUsername
+
+	// If Status.CompliantUsername is not set, then generate a compliant Username
+	if compliantUsername == "" {
+		compliantUsername, err = r.generateCompliantUsername(instance)
+		if err != nil {
+			// Some error occurred - requeue the request
+			reqLogger.Error(err, "Error generating compliant username", instance.Spec.Username)
```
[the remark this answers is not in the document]
sbryzak: I forgot to set the key name, fixed.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: Yeah I missed that but unfortunately it won't solve the problem.
We may end up with multiple usersignups with the same compliant username. And as soon the first usersignup is approved and creates the corresponding MUR that MUR will be recognized ad the MUR for all these signups - https://github.com/codeready-toolchain/host-operator/pull/93/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R158
We could check user ID there too and trigger username regeneration but IMO the logic will be clearer and less error-prone if we rely on the userID label in MUR and generate the username only when the signup is approved. Right before creating a MUR.

sbryzak: I checked my business logic again and I'm pretty sure I got it correct.  I don't see how it is possible for multiple usersignups to have the same compliant username, the code shouldn't allow it.  Also, the code I have [here](https://github.com/codeready-toolchain/host-operator/pull/93/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R134) is (I'm pretty sure) correct too.  The `generateCompliantUsername()` call will ensure that if the MUR for that UserID already exists then the `compliantUsername` will be set to that value.  This takes care of the situation where the MUR exists already but the `UserSignup.Status.CompliantUsername` isn't set yet (for whatever strange reason).

sbryzak: This is actually really complex business logic, and it took me a few hours to work out the exact process to handle all possible scenarios.  I think it might be worth having a bluejeans call to go over the details here.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -213,7 +261,7 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
 
 	mur := &toolchainv1alpha1.MasterUserRecord{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      userSignup.Spec.CompliantUsername,
+			Name:      userSignup.Status.CompliantUsername,
```
alexeykazakov: Wouldn't `userSignup.Status.CompliantUsername` be always empty here? It's set only if MUR already exists: https://github.com/codeready-toolchain/host-operator/pull/93/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R157

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: Yeah right. I missed that status set only if the MUR with such compliant username **and** userID found. And not when the generated name available. But unless I'm missing something again it causes another issue - https://github.com/codeready-toolchain/host-operator/pull/93#discussion_r341835246
It can be fixed by passing the generated name to MUR provisioning method instead of using the status (which is not set at that point) but I still have a feeling that using UserID label in MUR would simplify the logic.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: You're right, this is one spot that I missed. :)  Fixed now.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Good catch here, I did miss this one.  It should now handle all scenarios - I'll work on determining them all and get them covered in the e2e tests.

deploy/olm-catalog/host-operator/host-operator.package.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
```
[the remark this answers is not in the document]
MatousJobanek: This is not supposed to be here - it was removed in https://github.com/codeready-toolchain/host-operator/commit/fef8d36c8bfe7ba0801d97dea23c545b01856cd6 so I guess that this file is a result of a wrong merge/rebase

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,10 +126,23 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Flag that we set to true if a compliantUsername is set
+	compliantUsername := instance.Status.CompliantUsername
+
+	// If Status.CompliantUsername is not set, then generate a compliant Username
+	if compliantUsername == "" {
+		compliantUsername, err = r.generateCompliantUsername(instance)
```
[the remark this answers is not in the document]
MatousJobanek: Agree with Alexey that the generation should be done after the approval just before creating the MUR

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -240,8 +291,8 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
 			"Error creating MasterUserRecord")
```
MatousJobanek: don't forget to change the lines [285-289](https://github.com/codeready-toolchain/host-operator/pull/93/files#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R285-R289) - if the MUR already exists, then I guess that we should change (increment) the suffix of the compliant username

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,21 +128,27 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
-		// We generally EXPECT the MasterUserRecord to not be found here, so we only deal with other error types
-		if !errors.IsNotFound(err) {
-			// Error reading the MasterUserRecord, requeue the request
-			return reconcile.Result{}, err
+	// List all MasterUserRecord resources that have a UserID label equal to the UserSignup.Name
+	labels := map[string]string{toolchainv1alpha1.MasterUserRecordUserIDLabelKey: instance.Name}
+	opts := client.MatchingLabels(labels)
+	murList := &toolchainv1alpha1.MasterUserRecordList{}
+	if err = r.client.List(context.TODO(), murList, opts); err != nil {
+		return reconcile.Result{}, err
```
```suggestion
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, <somereason>, err, "")
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,21 +128,27 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		mur := murs[0]
+		if mur.Spec.UserID != instance.Name {
+			return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusInvalidMURUserID)
 		}
-	} else {
 		// If we successfully found an existing MasterUserRecord then our work here is done, set the status
 		// to Complete and return
 		reqLogger.Info("MasterUserRecord exists, setting status to Complete")
+		instance.Status.CompliantUsername = mur.Name
 		return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusComplete)
```
alexeykazakov: We also need to set Complete condition to true here.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -378,3 +446,7 @@ func (r *ReconcileUserSignup) updateStatusConditions(userSignup *toolchainv1alph
 	}
 	return r.client.Status().Update(context.TODO(), userSignup)
 }
+
+func (r *ReconcileUserSignup) updateStatusCompliantUsername(userSignup *toolchainv1alpha1.UserSignup) error {
```
alexeykazakov: This function name is misleading. It's just updating the status. There is nothing compliant username specific. Not sure why do we need this function at all?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -211,37 +253,41 @@ func (r *ReconcileUserSignup) provisionMasterUserRecord(userSignup *toolchainv1a
...
+
 	mur := &toolchainv1alpha1.MasterUserRecord{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      userSignup.Spec.CompliantUsername,
+			Name:      compliantUsername,
 			Namespace: userSignup.Namespace,
+			Labels:    labels,
 		},
 		Spec: toolchainv1alpha1.MasterUserRecordSpec{
 			UserID:       userSignup.Name,
```
alexeykazakov: We should stop using UserID field, shouldn't we? UserID is already set in the label. This filed is a duplicate.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -193,6 +204,37 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		// Check if a MasterUserRecord exists with the same transformed name
+		namespacedMurName := types.NamespacedName{Namespace: instance.Namespace, Name: transformed}
+		err := r.client.Get(context.TODO(), namespacedMurName, mur)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return "", err
+			}
+			// If there was a NotFound error looking up the mur, it means we found an available name
+			return transformed, nil
+		} else if mur.Spec.UserID == instance.Name { // If the found MUR has the same UserID as the UserSignup, then *it* is the correct MUR
```
alexeykazakov: 1. We use label now to store UserID.
2. If for whatever reason there is already MUR with the same UserID label then we should return an error here. Theoretically it can happen if there are multiple reconciles executed in the same time for the same UserSignup resource. But in this case returning an error (which will re-queue) will work. The next reconcile will be successful.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I've refactored this logic now.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -193,6 +201,38 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		err := r.client.Get(context.TODO(), namespacedMurName, mur)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return "", err
+			}
+			// If there was a NotFound error looking up the mur, it means we found an available name
+			return transformed, nil
+		} else if mur.Labels[toolchainv1alpha1.MasterUserRecordUserIDLabelKey] == instance.Name {
+			// If the found MUR has the same UserID as the UserSignup, then *it* is the correct MUR
+			return transformed, nil
```
alexeykazakov: As I mentioned in https://github.com/codeready-toolchain/host-operator/pull/93#discussion_r343944609
If for whatever reason there is already MUR with the same UserID label then we should return an error here. Theoretically it can happen if there are multiple reconciles executed in the same time for the same UserSignup resource. But in this case returning an error (which will re-queue) will work. The next reconcile will be successful.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,21 +127,25 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+	murList := &toolchainv1alpha1.MasterUserRecordList{}
+	if err = r.client.List(context.TODO(), murList, opts); err != nil {
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidMURState, err, "Failed to list MUR")
+	}
+
+	murs := murList.Items
+	// If we found more than one MasterUserRecord, then die
+	if len(murs) > 1 {
+		err = NewSignupError("multiple matching MUR resources found")
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidMURState, err, "Multiple MUR found")
```
alexeykazakov: `err` is nil here

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Fixed.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -64,16 +66,19 @@ func TestUserSignupWithAutoApproval(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, reconcile.Result{}, res)
 
+	// Lookup the user signup again
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+	require.NoError(t, err)
+
 	mur := &v1alpha1.MasterUserRecord{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Spec.CompliantUsername, Namespace: req.Namespace}, mur)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Status.CompliantUsername, Namespace: req.Namespace}, mur)
```
MatousJobanek: at this point, you shouldn't use `userSignup.Status.CompliantUsername` as it is still empty - it will be present after the second reconcile.
So you have a few options:
* list all MURs and verify that there is only one
* list MUR with the label selector
* use the expected name of the MUR: `"foo-at-redhat-com"`

MatousJobanek: btw, it would be good to have the two reconcile calls in two separated tests - it would be easier to read, better to verify (as it is isolated) and check when anything is broken

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -128,19 +133,18 @@ func TestUserSignupWithManualApprovalApproved(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, reconcile.Result{}, res)
 
+	// Lookup the userSignup again
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+	require.NoError(t, err)
+
 	mur := &v1alpha1.MasterUserRecord{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Spec.CompliantUsername, Namespace: req.Namespace}, mur)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Status.CompliantUsername, Namespace: req.Namespace}, mur)
```
MatousJobanek: same here

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -193,19 +197,18 @@ func TestUserSignupWithNoApprovalPolicyTreatedAsManualApproved(t *testing.T) {
...
```
MatousJobanek: same here

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -257,16 +260,16 @@ func TestUserSignupWithManualApprovalNotApproved(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, reconcile.Result{}, res)
 
+	// Lookup the userSignup again
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+	require.NoError(t, err)
+
 	mur := &v1alpha1.MasterUserRecord{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, mur)
```
MatousJobanek: I don't think that the MUR should be found using `Name: userSignup.Name`

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -304,19 +306,18 @@ func TestUserSignupWithAutoApprovalClusterSet(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, reconcile.Result{}, res)
 
+	// Lookup the userSignup again
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+	require.NoError(t, err)
+
 	mur := &v1alpha1.MasterUserRecord{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Spec.CompliantUsername, Namespace: req.Namespace}, mur)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Status.CompliantUsername, Namespace: req.Namespace}, mur)
```
MatousJobanek: same here

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -654,6 +598,172 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
 	require.Equal(t, v1.ConditionTrue, cond.Status)
 }
```
MatousJobanek: Don't forget to verify that the `Status.CompliantUsername` is properly set
```suggestion
	assert.Equal(t, mur.Name, instance.Status.CompliantUsername)
}
```

deploy/olm-catalog/host-operator/host-operator.package.yaml
```diff
@@ -0,0 +1,5 @@
...
+- currentCSV: host-operator.v0.0.1
+  name: alpha
+defaultChannel: alpha
+packageName: host-operator
```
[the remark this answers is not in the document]
sbryzak: Directory removed.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,21 +127,25 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
-			// Error reading the MasterUserRecord, requeue the request
-			return reconcile.Result{}, err
-		}
-	} else {
+	// List all MasterUserRecord resources that have a UserID label equal to the UserSignup.Name
+	labels := map[string]string{toolchainv1alpha1.MasterUserRecordUserIDLabelKey: instance.Name}
+	opts := client.MatchingLabels(labels)
+	murList := &toolchainv1alpha1.MasterUserRecordList{}
+	if err = r.client.List(context.TODO(), murList, opts); err != nil {
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidMURState, err, "Failed to list MUR")
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -123,21 +127,25 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+	}
+
+	murs := murList.Items
+	// If we found more than one MasterUserRecord, then die
+	if len(murs) > 1 {
+		err = NewSignupError("multiple matching MUR resources found")
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -193,6 +201,39 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return "", err
+			}
+			// If there was a NotFound error looking up the mur, it means we found an available name
+			return transformed, nil
+		} else if mur.Labels[toolchainv1alpha1.MasterUserRecordUserIDLabelKey] == instance.Name {
+			// If the found MUR has the same UserID as the UserSignup, then *it* is the correct MUR -
+			// Return an error here and allow the reconcile() function to pick it up on the next loop
+			return "", NewSignupError(fmt.Sprintf("could not generate compliant username as MUR [%s] already exists", mur.Name))
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Good catch here.. i've modified this test (and the others you mentioned) to return a list of MURs instead.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Good catch, fixed.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Done.

sbryzak:
@MatousJobanek thanks for the great review.  I'm pretty sure I've addressed all of your comments - you can find a new test `TestUserSignupWithExistingMURDifferentUserIDOK()` which tests reconciliation with an existing MUR with different UserID to cover the case where `generateCompliantUsername()` needs to iterate past the first transformed username.  There is also the test `TestDeathBy100Signups()` which exercises this code also.  There's no direct tests for `generateCompliantUsername()` as it is a private function, however I am confident we now have sufficient indirect testing of this function.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -594,6 +604,77 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
 		}
 	}
 
+	require.Equal(t, mur.Name, instance.Status.CompliantUsername)
+	require.NotNil(t, cond)
+	require.Equal(t, v1.ConditionTrue, cond.Status)
```
alexeykazakov: You can use https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/condition.go#L26 instead of manually look up the condition.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -594,6 +604,77 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+			Namespace: operatorNamespace,
+			UID:       types.UID(uuid.NewV4().String()),
+		},
+		Spec: v1alpha1.UserSignupSpec{
+			Username: "foo@redhat.com",
+			Approved: true,
+		},
+	}
+
+	// Create a MUR with the same UserID
```
alexeykazakov: This MUR has different UserID (not the same as in UserSignup above).

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
 	err = r.client.Get(context.TODO(), key, instance)
 	require.NoError(t, err)
 
+	require.NotEmpty(t, instance.Status.CompliantUsername)
```
alexeykazakov:
```suggestion
	require.Equal(t, "foo-at-redhat-com-1", instance.Status.CompliantUsername)
```

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+		},
+	}
+
+	r, req, _ := prepareReconcile(t, userSignup.Name, userSignup, configMap(config.UserApprovalPolicyAutomatic))
+
+	createMemberCluster(r.client)
+	defer clearMemberClusters(r.client)
+
+	_, err := r.Reconcile(req)
+	require.Error(t, err)
```
alexeykazakov: Check the error message.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+
+	var cond *v1alpha1.Condition
+	for _, condition := range instance.Status.Conditions {
+		if condition.Type == v1alpha1.UserSignupComplete {
+			cond = &condition
+		}
+	}
+
+	require.NotNil(t, cond)
+	require.Equal(t, v1.ConditionFalse, cond.Status)
```
alexeykazakov: Use the same check as in https://github.com/codeready-toolchain/host-operator/pull/93/files#diff-d7d6486f50f2c1cd16101f429e4ef843R601

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+		})
+	}
+
+	r, req, _ := prepareReconcile(t, userSignup.Name, args...)
+
+	createMemberCluster(r.client)
+	defer clearMemberClusters(r.client)
+
+	res, err := r.Reconcile(req)
+	require.Error(t, err)
```
alexeykazakov: Check the error message.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+		},
+	}
+
+	r, req, _ := prepareReconcile(t, userSignup.Name, userSignup, mur, mur2, configMap(config.UserApprovalPolicyAutomatic))
+
+	createMemberCluster(r.client)
+	defer clearMemberClusters(r.client)
+
+	_, err := r.Reconcile(req)
+	require.Error(t, err)
```
alexeykazakov: Check the error message.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -643,17 +654,193 @@ func TestUserSignupWithExistingMUROK(t *testing.T) {
...
+
+	var cond *v1alpha1.Condition
+	for _, condition := range instance.Status.Conditions {
+		if condition.Type == v1alpha1.UserSignupComplete {
+			cond = &condition
+		}
+	}
+
+	require.NotNil(t, cond)
+	require.Equal(t, v1.ConditionFalse, cond.Status)
```
alexeykazakov: Same here. No need to manually iterate the conditions. Use our custom assert as above.

alexeykazakov:
@sbryzak tests are failing now because you changed the error message when there are more than 1 MUR with the specific user id - https://github.com/codeready-toolchain/host-operator/pull/93/commits/8058d6a2d6aded4fdd46b590fe748f692883e925#diff-5b4b18a0fc2e0d708e4e6334cdb6bc69R141

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.0.0+incompatible
+	github.com/codeready-toolchain/api => github.com/sbryzak/api v0.0.0-20191101214508-cf5d4d46f1cd
```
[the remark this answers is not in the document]
sbryzak: Done.

alexeykazakov:
There are merge conflicts in e2e repo.

PR #103, opened 2019-11-14
Add target cluster details to UserAccount Status for MUR

alexeykazakov:
Fixes https://jira.coreos.com/browse/CRT-348

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/51


alexeykazakov:
Sorry, I updated the wrong PR. It's still WIP...

pkg/controller/masteruserrecord/sync.go
```diff
@@ -61,6 +87,55 @@ func (s *Synchronizer) synchronizeStatus() error {
 	return nil
 }
 
+// withClusterDetails returns the given user account status with additional information about
+// the target cluster such as API endpoint and Console URL if not set yet
+func (s *Synchronizer) withClusterDetails(status toolchainv1alpha1.UserAccountStatusEmbedded) (toolchainv1alpha1.UserAccountStatusEmbedded, error) {
+	if status.Cluster.Name != "" && (status.Cluster.APIEndpoint == "" || status.Cluster.ConsoleURL == "") {
```
MatousJobanek: I'm just curious: when the `status.Cluster.Name` could be empty?

pkg/controller/masteruserrecord/sync.go
```diff
@@ -61,6 +87,55 @@ func (s *Synchronizer) synchronizeStatus() error {
...
+	if err != nil {
+		return url, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return url, errors.Errorf("status is not 200 OK: %s", resp.Status)
+	}
+	defer func() {
+		_, _ = ioutil.ReadAll(resp.Body)
+		_ = resp.Body.Close()
+	}()
```
MatousJobanek: Should we put this right after the `Get` call? I think that we should read&close it also when status is not OK

pkg/controller/masteruserrecord/sync_test.go
```diff
@@ -236,5 +360,37 @@ func testSyncMurStatusWithUserAccountStatus(t *testing.T, userAccount *toolchain
...
 		AllUserAccountsHaveCondition(condition)
 }
+
+func newMemberCluster(cl client.Client) *cluster.FedCluster {
+	return &cluster.FedCluster{
+		Name:              test.MemberClusterName,
+		APIEndpoint:       fmt.Sprintf("https://api.%s:6433", test.MemberClusterName),
+		Client:            cl,
+		Type:              cluster.Member,
+		OperatorNamespace: test.HostOperatorNs,
```
MatousJobanek: OperatorNamespace is a name of a namespace (in the cluster) the operator is running in - in this case it's representing member operator that is running in `test.MemberOperatorNs`
```suggestion
		OperatorNamespace: test.MemberOperatorNs,
```

pkg/controller/masteruserrecord/sync.go
alexeykazakov: When there is no status (yet). In this case the status struct is an empty struct with empty strings by default.

pkg/controller/masteruserrecord/sync.go
alexeykazakov: Oh.. good catch. We definitely should.

alexeykazakov: Fixed in https://github.com/codeready-toolchain/host-operator/pull/103/commits/0bb6353313de16924fa58bc8666a5d191936c073

pkg/controller/masteruserrecord/sync_test.go
alexeykazakov: Right. But the member cluster resource exists in the Host cluster. It's used by the Host operator (which runs in the host NS) to access Member cluster.
So, the operator namespace is Host Operator NS, or I'm missing something?

The API docs could be clearer though ;)

MatousJobanek: The host controllers need to now the member namespace - to GET/POST/UPDATE resources inside of the member cluster. To identify the member namespace we use the label `namespace` and store it as `OperatorNamespace` inside of the `FedCluster` type. The idea is that the `FedCluster` represents some cluster (either host or member) and the `OperatorNamespace` is saying which namespace the toochain operator is running in. So here the `FedCluster` represents a member cluster so it should contain  `test.MemberOperatorNs`. There is no point in storing the `test.HostOperatorNs` as the host controllers know the namespace they are running in.

I'll try to improve the docs ;-)

alexeykazakov: That makes sense :)
Maybe targetNamespace would be a better name... but we don’t have to change the field name now. But let’s improve the docs at least.
Also I wonder why tests pass...

alexeykazakov: @MatousJobanek could you please make sure it’s covered by tests if it’s not too tricky?

PR #106, opened 2019-11-16
Add edit role to ns templates

alexeykazakov:
So, the user has edit role in -code, -dev, and -stage namespaces.

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/49

TODO:

- [x] Add tests

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced-code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced-code.yaml
@@ -15,6 +15,18 @@ objects:
     labels:
       provider: codeready-toolchain
     name: ${USERNAME}-code
+- apiVersion: v1
+  kind: RoleBinding
+  metadata:
+    labels:
+      provider: codeready-toolchain
+    name: code-edit
+    namespace: ${USERNAME}-code
+  roleRef:
+    name: edit
+  subjects:
+  - kind: User
+    name: ${USERNAME}
 parameters:
 - name: USERNAME
   required: true
--- a/deploy/templates/nstemplatetiers/advanced-code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced-code.yaml
@@ -20,10 +20,12 @@ objects:
   metadata:
     labels:
       provider: codeready-toolchain
-    name: codeready-toolchain-code
+    name: user-edit
     namespace: ${USERNAME}-code
   roleRef:
-    name: admin
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: edit
   subjects:
   - kind: User
     name: ${USERNAME}
--- a/deploy/templates/nstemplatetiers/advanced-code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced-code.yaml
@@ -20,7 +20,7 @@ objects:
   metadata:
     labels:
       provider: codeready-toolchain
-    name: toolchain-code
+    name: codeready-toolchain-code
     namespace: ${USERNAME}-code
   roleRef:
     name: admin
--- a/deploy/templates/nstemplatetiers/advanced-code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced-code.yaml
@@ -20,10 +20,10 @@ objects:
   metadata:
     labels:
       provider: codeready-toolchain
-    name: code-edit
+    name: toolchain-code
     namespace: ${USERNAME}-code
   roleRef:
-    name: edit
+    name: admin
   subjects:
   - kind: User
     name: ${USERNAME}
--- a/pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
+++ b/pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
@@ -3,16 +3,17 @@ package nstemplatetiers
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"testing"
 	texttemplate "text/template"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	testnstemplatetiers "github.com/codeready-toolchain/host-operator/test/templates/nstemplatetiers"
-	"github.com/pkg/errors"
 
 	"github.com/codeready-toolchain/host-operator/pkg/apis"
 	templatev1 "github.com/openshift/api/template/v1"
-	uuid "github.com/satori/go.uuid"
+	"github.com/pkg/errors"
+	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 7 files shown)


pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -160,6 +160,10 @@ func TestNewNSTemplateTier(t *testing.T) {
 							_, _, err = decoder.Decode(asset, nil, tmplObj)
 							require.NoError(t, err)
 							assert.Equal(t, *tmplObj, ns.Template)
+
+							// Assert expected objects in the template
+							// Each template should have one Namespace and one RoleBinding object
+							require.Len(t, ns.Template.Objects, 2)
```
MatousJobanek: if we know that each template should have at least one `Namespace` and one `RoleBinding` object, why we cannot verify it directly, instead of just checking the length...?

deploy/templates/nstemplatetiers/advanced-code.yaml
```diff
@@ -15,6 +15,18 @@ objects:
     labels:
       provider: codeready-toolchain
     name: ${USERNAME}-code
+- apiVersion: v1
+  kind: RoleBinding
+  metadata:
+    labels:
+      provider: codeready-toolchain
+    name: codeready-toolchain-code
```
MatousJobanek: From my point of view, `codeready-toolchain-code` name doesn't bring much information - we know that it's inside of the `code` namespace and using the label we also know about the provider `codeready-toolchain`. 
I would rather name it `user-admin`, `user-edit` or `edit-code` to highlight the purpose of the resource (`RoleBinding`)

deploy/templates/nstemplatetiers/advanced-code.yaml
```diff
@@ -15,6 +15,18 @@ objects:
...
+    namespace: ${USERNAME}-code
+  roleRef:
+    name: admin
```
MatousJobanek: May I know why `admin` and not `edit`? By giving the admin rights we also give the ability to delete the namespace, don't we?

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
alexeykazakov: Checking the objects inside the template is not that easy. We would need to process the template. Basically would need to bring the template processor from member operator repo here. So we would need to move it common and reuse everywhere. It’s not a trivial change.
A simple decoding doesn’t seem to be enough as well. Maybe I missed something though since I didn’t spent much time on it.
But we do proper testing of the roles and namespaces in our e2e tests anyway. So I think it’s not worth significantly complicate our unit tests for the cases which already covered by e2e tests.

deploy/templates/nstemplatetiers/advanced-code.yaml
alexeykazakov: Good point

deploy/templates/nstemplatetiers/advanced-code.yaml
alexeykazakov: I think edit doesn’t allow creating/updating rbac. And it would be nice to allow advanced users to share the namespace. But allowing deleting namespaces is dangerous...
Let me double check the permissions.

deploy/templates/nstemplatetiers/advanced-code.yaml
alexeykazakov: Done in https://github.com/codeready-toolchain/host-operator/pull/106/commits/d17c13c1e5c0d47270b843b45f7e33389089bfce

deploy/templates/nstemplatetiers/advanced-code.yaml
alexeykazakov: OK I checked it. "admin" role gives access to deleting namespaces and manipulating with rbac.
"edit" doesn't allow any of that.
I changed admin to edit for now. We can address the missing rbac thing for advanced tier later.
See https://github.com/codeready-toolchain/host-operator/pull/106/commits/d17c13c1e5c0d47270b843b45f7e33389089bfce

alexeykazakov: BTW deleting a namespace should not be a big deal. The controller will recreate it. But anyway let's deal with that in a separate PR. For now we will use "edit" role for all tiers.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
MatousJobanek: I see, but we could do a bit simpler check - either this:
```go
assert.Contains(t, string(ns.Template.Objects[1].Raw), `"kind":"RoleBinding","metadata":{"labels":{"provider":"codeready-toolchain"},"name":"user-edit"`)
```
or more safe:
```go
rbFound := false
for _, object := range ns.Template.Objects {
	if strings.Contains(string(object.Raw), `"kind":"RoleBinding","metadata":{"labels":{"provider":"codeready-toolchain"},"name":"user-edit"`) {
		rbFound = true
		break
	}
}
assert.True(t,rbFound, "the user-edit RoleBinding wasn't found in the namespace of the type", "ns-type", nsType)
```
take it just as an suggestion - it's not a must ;-)

deploy/templates/nstemplatetiers/advanced-code.yaml
```diff
@@ -15,6 +15,20 @@ objects:
...
+- apiVersion: v1
+  kind: RoleBinding
+  metadata:
+    labels:
+      provider: codeready-toolchain
+    name: user-edit
+    namespace: ${USERNAME}-code
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
```
MatousJobanek: I guess that we need to define the group and kind to differentiate between `rbac.authorization.k8s.io` and `authorization.openshift.io`, correct? Or is there anything else behind it?

alexeykazakov: Well.. both options with 1) apiGroup and kind and 2) without them work fine. But it seems that some admission webhook modifies roleRef without apiGroup/kind because it always there after creation. There must be some reason for such admission webhook. So I decided to add it there explicitly.
Referring to the cluster role IMO makes sense regardless.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
alexeykazakov: Done in https://github.com/codeready-toolchain/host-operator/pull/106/commits/8202e03f2c85262e9359ea749643eda44c4fdde0

PR #117, opened 2019-12-06
feat(CRT-354): create RegistrationService CR during the operator startup

MatousJobanek:
This is the last step in moving the deployment of registration service into the host-operator.

The changes in the PR use the loaded configuration and creates/updates RegistrationService during the operator startup 

paired e2e PR https://github.com/codeready-toolchain/toolchain-e2e/pull/56


deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,10 @@ spec:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
                   value: host-operator
+                - name: REGISTRATION_SERVICE_IMAGE
+                  value: REPLACE_REGISTRATION_SERVICE_IMAGE
+                - name: REGISTRATION_SERVICE_ENVIRONMENT
+                  value: REPLACE_REGISTRATION_SERVICE_ENVIRONMENT
```
alexeykazakov: These two params are not enough.
We also need to pass:
REGISTRATION_AUTH_CLIENT_LIBRARY_URL
REGISTRATION_AUTH_CLIENT_CONFIG_RAW
REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL

because they will point to different sso in different env.

alexeykazakov: It would be nice if we have a set of configurations for each env presented by some simple files.
Something like:
deploy/env/prod.config
deploy/env/dev.config
deploy/env/e2e-test.config
deploy/env/stage.config
etc.
The format of these files can be anything we want (a text file with <key>=<value>, a yaml, viper config file, json, whatever we want). The important part is that is should be easy to edit by human and contains the values for configuration we need to pass to Host-operator/Reg-service deployments.

It wouldn't change your approach here. It's rather an addition to it.

Our CD (and toolchain-infra scripts) will use `deploy/env/prod.config` to populate `REGISTRATION_SERVICE_ENVIRONMENT`, `REGISTRATION_AUTH_CLIENT_LIBRARY_URL`, etc.

e2e-test make files will use 'deploy/env/e2e-test.config', etc.

We could even create a separate repo(s?) for these configurations and keep both host and reg-service (and member later if/when we need it):
https://github.com/codeready-toolchain/toolchain-config
with:
- prod
- stage (don't need it for now)
- dev
- e2e-tests
folders.
So we have one place were we can control configuration of all relevant operators.

But let's do it in a separate PR(s). For this PR let's just make sure we don't miss important params in deploy/operator.yaml.

alexeykazakov: I'm basically talking about GitOps here.

MatousJobanek: > These two params are not enough.

The reason why I introduced only two of them was that I wanted to limit the number of params to the only necessary ones. I mean necessary for starting the service - the rest could be introduced by modifying the created CR itself. But to be honest, I wasn't sure about that approach, so let add the additional ones

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,133 @@
...
+// file path. If the provided config file path is empty, a default configuration
+// will be created.
+func New(configFilePath string) (*Registry, error) {
+	c := CreateEmptyRegistry()
+	if configFilePath != "" {
+		for _, reg := range []*viper.Viper{c.host, c.regService} {
+			reg.SetConfigType("yaml")
+			reg.SetConfigFile(configFilePath)
+			err := reg.ReadInConfig() // Find and read the config file
+			if err != nil { // Handle errors reading the config file.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			if err != nil {           // Handle errors reading the config file.
```

golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			if err != nil {           // Handle errors reading the config file.
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
alexeykazakov: I see the added auth params but we do not set them anywhere.

For now we have three environments:
- `dev`. It's used by `make dev-deploy-*` targets in e2e repo and in the corresponding reg-service make target:
**ENVIRONMENT** should be set to "dev".
**AUTH** params should be set to the values which currently is used by reg-service by default - https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go#L71-L92
- "e2e-test". When running `make test-e2e-*`
**ENVIRONMENT** should be set to "e2e-tests".
**AUTH** params do not matter because they will be ignored by reg-service if running in "e2e-tests" env.
- "prod". Everything else.
**ENVIRONMENT** should be set to "prod".
**REGISTRATION_AUTH_CLIENT_LIBRARY_URL** param should be set to:
`https://sso.redhat.com/auth/js/keycloak.js`
**REGISTRATION_AUTH_CLIENT_CONFIG_RAW** param should be set to:
`'{"realm":"redhat-external","auth-server-url":"https://sso.redhat.com/auth","ssl-required":"none","resource":"crtoolchain-public","clientId":"crtoolchain-public","public-client":true}'`
**REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL** param should be set to:
`'https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs'`

And we have the following places were we can set the values:
- Default values in reg-service viper configuration: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go
- Default values in host operator configuration: https://github.com/codeready-toolchain/host-operator/pull/117/files#diff-ec24f891497e9e435f8f9956e23cb379R107
- Make targets (to use *_REPLACE strings).

What we have now:
- For **e2e** env **we are good** because env param is set in e2e make targets. Other params do not matter.
- For **dev** env we have an **issue**. Env is set properly in make: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/dev.mk#L4 but AUTH params will be set to *_REPLACE strings since we don't really replace them by anything (and since they are not an empty strings then reg-service viper won't use the default values we set in https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go#L71-L92
To fix that we can set these AUTH_REPLACE vars to empty strings in https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/dev.mk#L4 so the default values will be used (or we could set them explicitly to the same default values from https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go#L71-L92 - both options will work). And we will need to pass these params properly in https://github.com/codeready-toolchain/toolchain-e2e/pull/56 PR
- For **prod** env we have **issues** with both env var and auth vars.
env var will be set to ..._REPLACE string so the default value (prod) won't be used by viper.
AUTH vars will have ..._REPLACE strings as well.
We could solve it either in host configuration by checking if the value is ..._REPLACE and then return the default (prod env + auth values for prod I provided above) values or by explicitly setting replace strings to empty stings when building an image and then let viper to use the default values in host operator configuration.

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                - name: REGISTRATION_AUTH_CLIENT_LIBRARY_URL
```
alexeykazakov:
```suggestion
                - name: REGISTRATION_SERVICE_AUTH_CLIENT_LIBRARY_URL
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                  value: REPLACE_REGISTRATION_AUTH_CLIENT_LIBRARY_URL
```
alexeykazakov:
```suggestion
                  value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_LIBRARY_URL
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                - name: REGISTRATION_AUTH_CLIENT_CONFIG_RAW
```
alexeykazakov:
```suggestion
                - name: REGISTRATION_SERVICE_AUTH_CLIENT_CONFIG_RAW
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                  value: REPLACE_REGISTRATION_AUTH_CLIENT_CONFIG_RAW
```
alexeykazakov:
```suggestion
                  value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_CONFIG_RAW
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                - name: REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL
```
alexeykazakov:
```suggestion
                - name: REGISTRATION_SERVICE_AUTH_CLIENT_PUBLIC_KEYS_URL
```

deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+                  value: REPLACE_REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL
```
alexeykazakov:
```suggestion
                  value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_PUBLIC_KEYS_URL
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
               fieldPath: metadata.name
         - name: OPERATOR_NAME
           value: "host-operator"
+        - name: REGISTRATION_SERVICE_IMAGE
+          value: REPLACE_REGISTRATION_SERVICE_IMAGE
+        - name: REGISTRATION_SERVICE_ENVIRONMENT
+          value: REPLACE_REGISTRATION_SERVICE_ENVIRONMENT
+        - name: REGISTRATION_AUTH_CLIENT_LIBRARY_URL
```
alexeykazakov:
```suggestion
        - name: REGISTRATION_SERVICE_AUTH_CLIENT_LIBRARY_URL
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
...
+          value: REPLACE_REGISTRATION_AUTH_CLIENT_LIBRARY_URL
```
alexeykazakov:
```suggestion
          value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_LIBRARY_URL
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
...
+        - name: REGISTRATION_AUTH_CLIENT_CONFIG_RAW
```
alexeykazakov:
```suggestion
        - name: REGISTRATION_SERVICE_AUTH_CLIENT_CONFIG_RAW
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
...
+          value: REPLACE_REGISTRATION_AUTH_CLIENT_CONFIG_RAW
```
alexeykazakov:
```suggestion
          value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_CONFIG_RAW
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
...
+        - name: REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL
```
alexeykazakov:
```suggestion
        - name: REGISTRATION_SERVICE_AUTH_CLIENT_PUBLIC_KEYS_URL
```

deploy/operator.yaml
```diff
@@ -31,3 +31,13 @@ spec:
...
+          value: REPLACE_REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL
```
alexeykazakov:
```suggestion
          value: REPLACE_REGISTRATION_SERVICE_AUTH_CLIENT_PUBLIC_KEYS_URL
```

PR #125, opened 2019-12-15
Reduce user permissions

alexeykazakov:
Requires https://github.com/codeready-toolchain/member-operator/pull/109 to be merged first to pass e2e tests.

After some investigation it looks like che requires edit permissions for roles and rolebindings in the namespace to create workspaces. This is on top of the edit role (to create other resources like deployments, routes, services, etc).

This PR reduces permissions to edit role for all namespaces and adds the missing edit permissions for roles and rolebindings for -code namespace.

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/63


xcoulon:
thanks for the PR. I like the idea of not giving full `admin` privileges to users in their namespaces.

PR #126, opened 2019-12-17
Support user deactivation in UserSignup controller

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-387

As per the design doc:

A user may be deactivated due to an extended period of activity or for some other reason.  Deactivating a user causes all of their data to be lost, however they may be reactivated via the registration service by signing up again.  When reactivating again, the process is essentially the same as a fresh user signup from the user's point of view.

When deactivating a user, their UserSignup resource is retained but all other resources are removed.  To deactivate a user, an Admin (or automated process) sets the Deactivated property of the user's UserSignup resource to true.  Once the Deactivated property is set, the UserSignupController will reconcile the UserSignup resource, and delete the user's MasterUserRecord.  From here, garbage collection will take care of the deleting of the user's other resources in the member cluster.


go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/codeready-toolchain/host-operator
 require (
 	cloud.google.com/go v0.46.3 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.6.0 // indirect
-	github.com/codeready-toolchain/api v0.0.0-20200106162618-9b2c84309f9d
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20191206100349-1d3b3ba54c59
+	github.com/codeready-toolchain/api v0.0.0-20191212002517-882a587d15ce
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20191206153324-4205c5ebe624
```
alexeykazakov: Don't downgrade.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -142,14 +144,33 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
 		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidMURState, err, "Multiple MasterUserRecords found")
 	} else if len(murs) == 1 {
+		mur := murs[0]
+
+		// If the user has been deactivated, then we need to delete the MUR
+		if instance.Spec.Deactivated {
+			err = r.client.Delete(context.TODO(), &mur)
+			if err != nil {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusFailedToDeleteMUR, err,
+					"Error creating MasterUserRecord")
```
```suggestion
					"Error deleting MasterUserRecord")
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -430,6 +462,17 @@ func (r *ReconcileUserSignup) setStatusComplete(userSignup *toolchainv1alpha1.Us
 		})
 }
 
+func (r *ReconcileUserSignup) setStatusDeactivated(userSignup *toolchainv1alpha1.UserSignup, message string) error {
+	return r.updateStatusConditions(
+		userSignup,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.UserSignupComplete,
+			Status:  corev1.ConditionTrue,
```
alexeykazakov: Do you think we should set the Complete condition to `true` for deactivated signups?
I think it depends on the logic for the re-signup in the reg service we want to implement...

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +813,144 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+		v1alpha1.Condition{
+			Type:   v1alpha1.UserSignupApproved,
+			Status: v1.ConditionTrue,
+			Reason: "ApprovedAutomatically",
+		},
+		v1alpha1.Condition{
+			Type:   v1alpha1.UserSignupComplete,
+			Status: v1.ConditionTrue,
+		})
+
```
alexeykazakov: We don't really have to add all these steps in the test and could just add all the expected resources when preparing reconcile, do we?
I mean if we just add the expected initial Signup and MUR resources to the fake client then we can leave lines 922-951 in this test and significantly simplify the test.

The whole e2e flow (creating a user signup from scratch, approving it, then deactivating) will be tested by the e2e tests. In our unit test we just need to add the Signup and MUR in desired stated to the fake client before deactivating it.

alexeykazakov:
Also, we need to cover errors during MUR deletion by tests.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -142,14 +144,33 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		// If the user has been deactivated, then we need to delete the MUR
+		if instance.Spec.Deactivated {
+			err = r.client.Delete(context.TODO(), &mur)
+			if err != nil {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusFailedToDeleteMUR, err,
+					"Error creating MasterUserRecord")
+			}
+
+			reqLogger.Info("Deleted MasterUserRecord", "Name", mur.Name)
+			return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusDeactivated)
```
alexeykazakov: The MUR has not been deleted yet actually. So, instead of setting status to deactivated right away after we send the delete request I think our status management should be a bit more complex.
If MUR exists then:
1. Delete MUR
2. Update status to Deactivating, exit reconcile

When the MUR is actually deleted then another reconcile will be triggered and since there is no MUR anymore the status will be updated to Deactivated.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I was wondering about this, however it seems to me that the `Complete` condition should be set to `true` once the resources reach the desired state based on their specification.  In the case of a new user signup, we set complete to true once the MUR is created, for a deactivation, it seems correct that we would set it to true once the MUR is deleted, as there is no longer any work remaining.  Can you think of an alternative way to approach this?

pkg/controller/usersignup/usersignup_controller.go
sbryzak: That's reasonable, I've simply deleted the status update from here since we will catch it in the subsequent reconcile further down in the code.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: Yeah I see your point. It makes sense. We could just check the reason in reg-service when checking the signup/mur statuses. If Completed is true but the reason is Deactivated then we know that user has been deactivated.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: By why not setting `Deactivating` reason?
Just thinking about the scenario when something goes wrong during deactivation:
T0. User provisioned. Signup CR status set to Complete with ApprovedManually/Automaticaly reason (here is a bug in our code! we set "" as the reason when complete==true)
T1. Set spec.Deactivated = true
T2. Cntrl reconcile requests MUR to be deleted and currently it doesn't update the status so it's still Complete=true and the original reason (which should be Approved when we fix the bug in T0)
T3. Something goes wrong and MUR is not deleted. Then we are stuck with wrong Status from T0.

So, I would do the following:
- Fix Approved reason/status for T0.
- Then in T3: Set Complete to False with reason: Deactiavating.

sbryzak: I've made the changes in T3, I think they are a good idea so now Complete is set to false with a reason of "Deactivating".  I don't agree with T0 though, remember that UserSignup has two status conditions, one for `Approved` and one for `Complete`.  The `Approved` status will already contain a reason indicating whether the signup was approved automatically or manually, and it would be redundant to duplicate this information for the `Complete` reason also.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: I've simplified the test, and also taken the opportunity to confirm the new `Deactivating` status is set.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: OK I forgot that we keep Approved status in a separate condition. Just looks a bit weird that we have an empty reason when Complete is false. But I think that’s ok if we have only one state when Complete == true (it’s when MUR is created).

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +813,141 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+	}
+
+	mur := &v1alpha1.MasterUserRecord{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "john-doe-at-redhat-com",
+			Namespace: operatorNamespace,
+			UID:       types.UID(uuid.NewV4().String()),
+			Labels:    map[string]string{toolchainv1alpha1.MasterUserRecordUserIDLabelKey: userSignup.Name},
+		},
+	}
```
MatousJobanek: There are some functions in toolchain-common repo that will help you with creating some of the CRs:
```
mur := murtest.NewMasterUserRecord("john-doe-at-redhat-com", murtest.MetaNamespace(operatorNamespace))
mur.Labels = map[string]string{toolchainv1alpha1.MasterUserRecordUserIDLabelKey: userSignup.Name}
```
see https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/test/masteruserrecord
it would be good if we had something similar for UserSignup too - the tests would be much shorter

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +813,141 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+
+	r, req, _ := prepareReconcile(t, userSignup.Name, userSignup, mur, configMap(config.UserApprovalPolicyAutomatic), basicNSTemplateTier)
+
+	// Lookup the UserSignup
+	key := types.NamespacedName{
+		Namespace: operatorNamespace,
+		Name:      userSignup.Name,
+	}
```
MatousJobanek: same for this - there is `test.NamespacedName()` function
```
key := test.NamespacedName(operatorNamespace, userSignup.Name)
```

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +813,141 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+			Status: v1.ConditionFalse,
+			Reason: "Deactivating",
+		})
+
+	murs := &v1alpha1.MasterUserRecordList{}
+
+	// The MUR should have now been deleted
+	err = r.client.List(context.TODO(), murs)
+	require.NoError(t, err)
+	require.Len(t, murs.Items, 0)
```
MatousJobanek: could you please cut the test here and verify the next reconcile in a new test? It would be much better if we had one unit test for one operation - it's much easier to read as well as debug in case of failures

MatousJobanek: you should also verify a case when:
`UserSignup.Reason = Deactivating` but the MUR still exists - I guess that in that case, the client will return an error for another attempt of the MUR deletion - you should verify that use-case too.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Thanks Matous, these functions make it much cleaner.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: I'd prefer not to cut the test at this point.  Moving the remainder of the lines to a different test will still require all of the setup that comes before.  In fact, everything before the line `	userSignup.Spec.Deactivated = true` is essentially setup, and the remainder of the test is simply simulating the controller running in a live environment and validating that the resource states are correct.

MatousJobanek: let me try to clarify it here - in your test you verify two things:
* for the first reconcile, that the MUR is deleted and the UserSignup status condition is set to `Deactivating`
* for the second reconcile, you verify that when there is no MUR present then the  UserSignup status condition is set to `Deactivated`

for me, these are two cases that should be tested separately and not together. In addition, setting `userSignup.Spec.Deactivated = true` doesn't have to be done via client call, but during the UserSignup initialization - there is no difference and it will make the test shorter and easier to read.

If you are afraid of doing the test setup again, you can use subtests and reuse the test setup.

MatousJobanek: and please, don't forget to verify also the cases I mentioned in https://github.com/codeready-toolchain/host-operator/pull/126#discussion_r366989267

sbryzak: @MatousJobanek I've made your suggested changes, thanks.  It seems a bit redundant to me though to use subtests for this.. the second subtest cannot exist in isolation without the previous subtest executing first.  The value of having the subtests at all then seems to be merely as a syntactic visual delineation between the sections, or am I missing something here?

MatousJobanek: no, they are not dependent - both subtests can be executed either on their own or both of them together. There is nothing in the second subtest that would require anything from the first one. 
Try it out ;-)

sbryzak: I've also added a test for when the condition status is set to `Deactivating` but the MUR still exists, in this case the controller simply deletes the MUR, as the workflow isn't based on the status value.

sbryzak: Ah you're right, my apologies... I didn't take into account that `prepareReconcile` sets up the complete state of the resources for each test.

MatousJobanek: > I've also added a test for when the condition status is set to Deactivating but the MUR still exists

thanks for adding this, but I was talking about covering the case when the `client.Delete` call returns an error.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +814,204 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
 	)
 }
 
+func TestUserSignupDeactivatedWithNoMUR(t *testing.T) {
+	userSignup := &v1alpha1.UserSignup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      uuid.NewV4().String(),
+			Namespace: operatorNamespace,
+			UID:       types.UID(uuid.NewV4().String()),
```
xcoulon: I believe you don't need to set the `UID`

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +814,204 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+	)
+}
+
+func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
+	// given
+	userSignup := &v1alpha1.UserSignup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      uuid.NewV4().String(),
+			Namespace: operatorNamespace,
+			UID:       types.UID(uuid.NewV4().String()),
```
xcoulon: same here, no need to specify the `UID`

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +814,204 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+		murs := &v1alpha1.MasterUserRecordList{}
+
+		// The MUR should have now been deleted
+		err = r.client.List(context.TODO(), murs)
+		require.NoError(t, err)
+		require.Len(t, murs.Items, 0)
+
+	})
+
+	t.Run("when MUR doesn't exist, then the condition should be set to Deactivated", func(t *testing.T) {
```
xcoulon: isn't it the same test as `TestUserSignupDeactivatedWithNoMUR` ?

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -813,6 +814,204 @@ func TestUserSignupWithInvalidNameNotOK(t *testing.T) {
...
+	})
+}
+
+func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
+	// given
+	userSignup := &v1alpha1.UserSignup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      uuid.NewV4().String(),
+			Namespace: operatorNamespace,
+			UID:       types.UID(uuid.NewV4().String()),
```
xcoulon: same here, no need to declare the `UID`

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: This was an artifact of copying the previous tests, I've removed it from this PR.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: You're right, good catch.  I've removed the standalone test.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Fixed.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: > > I've also added a test for when the condition status is set to Deactivating but the MUR still exists
> 
> thanks for adding this, but I was talking about covering the case when the `client.Delete` call returns an error.

Ah, that wasn't clear but no problem, I will add a test to validate the behaviour when the MUR delete fails.  In this case the reason won't be set to `Deactivating` though, instead we should expect a reason value of `UnableToDeleteMUR`.

sbryzak: I've added this test now.

MatousJobanek: Thanks a lot and sorry for not being clear :-)

alexeykazakov:
The PR looks good! But please do not merge until the e2e test PR is ready to be merged too - https://github.com/codeready-toolchain/toolchain-e2e/pull/68

PR #132, opened 2020-01-06
fix: common api sheme contain only CRs

MatousJobanek:
The scheme should contain only CRs apis - this will get rid of the last `reflector.go:95: Failed to list` errors from our logs - we should create metrics for the extensions
see: https://github.com/codeready-toolchain/api/pull/83

### DIFF
--- a/pkg/apis/apis.go
+++ b/pkg/apis/apis.go
@@ -4,13 +4,13 @@ import (
 	"github.com/codeready-toolchain/api/pkg/apis"
 
 	routev1 "github.com/openshift/api/route/v1"
-	templatev1 "github.com/openshift/api/template/v1"
+	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme) error {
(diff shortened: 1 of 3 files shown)


pkg/apis/apis.go
```diff
@@ -4,13 +4,13 @@ import (
...
 
 	routev1 "github.com/openshift/api/route/v1"
-	templatev1 "github.com/openshift/api/template/v1"
+	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme) error {
-	addToSchemes := append(apis.AddToSchemes, templatev1.Install)
```
alexeykazakov: Don't we really need the template schema anymore?

MatousJobanek: look at the next line that was there: 
`addToSchemes = append(apis.AddToSchemes, routev1.Install)`
the template schema wasn't used there before, there was a small bug - when appending the `routev1.Install` there was used `apis.AddToSchemes` instead of `addToSchemes` so it overrode the original `addToSchemes` that contained the template schema. 
Actually, when I tried to fix it and add the template schema then it started failing. I'm not sure why, but as it's working with this setup I decided to leave it as it is and investigate the failure later.

alexeykazakov: hm.. ok

PR #133, opened 2020-01-07
chore: remove unneeded permissons

MatousJobanek:
Since the RBAC issue with `reflector.go:95: Failed to list` errors should be fixed I decided to remove unneeded permissions from our roles. I didn't spend much time investigating which particular permissions are used and which are not, I just removed those that were apparently added just as a workaround for making our logs clean and happy.

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -7,14 +7,8 @@ rules:
   - ""
   resources:
   - nodes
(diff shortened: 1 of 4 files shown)


MatousJobanek:
Good point @sbryzak
I unified the format so all cases use the double quotes `"`  e1537c0

PR #134, opened 2020-01-08
Regenerated from api - added BannedUser type

sbryzak:

### DIFF
--- a/deploy/olm-catalog/host-operator/host-operator.package.yaml
+++ b/deploy/olm-catalog/host-operator/host-operator.package.yaml
@@ -1,0 +1,5 @@
+channels:
+- currentCSV: host-operator.v0.0.1
+  name: alpha
+defaultChannel: alpha
+packageName: host-operator
--- a/deploy/olm-catalog/host-operator/host-operator.package.yaml
+++ b/deploy/olm-catalog/host-operator/host-operator.package.yaml
@@ -1,5 +1,0 @@
-channels:
-- currentCSV: host-operator.v0.0.1
-  name: alpha
-defaultChannel: alpha
-packageName: host-operator
--- a/deploy/crds/toolchain_v1alpha1_banneduser_crd.yaml
+++ b/deploy/crds/toolchain_v1alpha1_banneduser_crd.yaml
@@ -1,0 +1,54 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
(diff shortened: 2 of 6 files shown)


deploy/olm-catalog/host-operator/host-operator.package.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
```
alexeykazakov: There should not be any package yamls

sbryzak: I've removed this yaml file, but I'd like to know why it is getting generated in the first place.  I've removed my vendor dir and run `make clean` in the api project, so I'm not sure where this is coming from.

MatousJobanek: That's weird. Are you sure that you use the latest versions of both projects? What target do you use for generating and propagating the CRD changes in api project? You should use `make generate` - that should do everything necessary for you including creating/updating the correct CSV and operator package

alexeykazakov:
> Please, don't forget to add also the BannedUser to the list of owned CRDs in the base template CSV: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml#L115
you should also add an example of the CR here: https://github.com/codeready-toolchain/host-operator/tree/master/examples
When it's done, then you need to propagate the change also to the hack files - to do that you need to either run make generate-csv or directly by calling the script: ../api/scripts/olm-catalog-generate.sh -pr ../host-operator

I think we need to document it somewhere. Maybe in api makefile (generate target)?

alexeykazakov:
> I think we need to document it somewhere. Maybe in api makefile (generate target)?

Or/and in Readme. There is a few places which need to be changed in case of new CRD introduction.

MatousJobanek:
@sbryzak you didn't do all the steps that I wrote in the comment: https://github.com/codeready-toolchain/host-operator/pull/134#pullrequestreview-339892906 - talking about this one:

>When it's done, then you need to propagate the change also to the hack files - to do that you need to either run make generate-csv or directly by calling the script: ../api/scripts/olm-catalog-generate.sh -pr ../host-operator

without having it propagated in the hack files, you won't be able to verify the changes locally.

MatousJobanek:
I created PR with the missing bits https://github.com/codeready-toolchain/host-operator/pull/135

PR #138, opened 2020-01-15
feat: store reg-service parameters in a map

MatousJobanek:
Instead of storing reg-service parameters as standalone variables we will use a map to have a more generic behavior

see https://github.com/codeready-toolchain/api/pull/89


go.mod
```diff
@@ -46,6 +46,8 @@ require (
 	sigs.k8s.io/testing_frameworks v0.1.2 // indirect
 )
 
+replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200115111617-f3c1e80a5abe
```
MatousJobanek: a note for myself to not forget to remove it once the api PR is merged

MatousJobanek: updated in 3014136

PR #141, opened 2020-01-29
Fix MUR status synchronization in case of target availability error

alexeykazakov:
Currently ff MUR reconcile failed with some error (target member cluster is temporally not available for example) then the error is stuck in MUR.Status.Ready condition if there is no changes in the MUR.Status.UserAccount[] conditions.

Steps to reproduce:
- Create a MUR and get it provisioned. UserAccount is provisioned and MUR is Ready.
- Break the member kubefedcluster, so member is not available from host.
- Trigger MUR reconcile
- MUR.Status.Ready condition is false now with the error about unavailable member
- Fix member cluster
- MUR reconcile is called but there is nothing to update in MUR.Status.UserAccount[] since the corresponding UserAccount statuses didn't change in member. So, MUR do not update Ready condition either and we stuck with the error.

This PR fixes the problem and force Ready condition update even if there is no UserAccount status changes.

TODO:
- [x] Add Unit tests

Writing e2e tests for that fix is pretty tricky... We can't make an existing kubefed cluster unhealthy by updating the spec of the existing kubefedcluster resource. The update is ignored if it breaks the cluster.

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
@@ -410,7 +410,7 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
 			Cluster: toolchainv1alpha1.Cluster{
 				Name: "member3-cluster",
 			},
-			UserAccountStatus: userAccount.Status,
+			UserAccountStatus: userAccount3.Status,
 		})
 
 		memberClient := test.NewFakeClient(t, userAccount, consoleRoute())
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
@@ -393,27 +393,27 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
 
 		userAccount := uatest.NewUserAccountFromMur(mur,
 			uatest.StatusCondition(toBeNotReady("Provisioning", "")), uatest.ResourceVersion("123abc"))
-		mur.Status.UserAccounts = []toolchainv1alpha1.UserAccountStatusEmbedded{}
-
 		userAccount2 := uatest.NewUserAccountFromMur(mur,
 			uatest.StatusCondition(toBeNotReady("Provisioning", "")), uatest.ResourceVersion("123abc"))
-		mur.Status.UserAccounts = append(mur.Status.UserAccounts, toolchainv1alpha1.UserAccountStatusEmbedded{
-			SyncIndex: "111aaa",
-			Cluster: toolchainv1alpha1.Cluster{
-				Name: "member2-cluster",
-			},
-			UserAccountStatus: userAccount2.Status,
-		})
-
 		userAccount3 := uatest.NewUserAccountFromMur(mur,
 			uatest.StatusCondition(toBeNotReady("Provisioning", "")), uatest.ResourceVersion("123abc"))
-		mur.Status.UserAccounts = append(mur.Status.UserAccounts, toolchainv1alpha1.UserAccountStatusEmbedded{
-			SyncIndex: "123abc",
-			Cluster: toolchainv1alpha1.Cluster{
-				Name: "member3-cluster",
+
+		mur.Status.UserAccounts = []toolchainv1alpha1.UserAccountStatusEmbedded{
+			{
+				SyncIndex: "111aaa",
+				Cluster: toolchainv1alpha1.Cluster{
+					Name: "member2-cluster",
+				},
+				UserAccountStatus: userAccount2.Status,
 			},
-			UserAccountStatus: userAccount3.Status,
-		})
+			{
+				SyncIndex: "123abc",
+				Cluster: toolchainv1alpha1.Cluster{
+					Name: "member3-cluster",
+				},
+				UserAccountStatus: userAccount3.Status,
+			},
+		}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -380,71 +380,116 @@ func TestModifyUserAccounts(t *testing.T) {
...
+		mur.Status.UserAccounts = []toolchainv1alpha1.UserAccountStatusEmbedded{}
+
+		userAccount2 := uatest.NewUserAccountFromMur(mur,
+			uatest.StatusCondition(toBeNotReady("Provisioning", "")), uatest.ResourceVersion("123abc"))
+		mur.Status.UserAccounts = append(mur.Status.UserAccounts, toolchainv1alpha1.UserAccountStatusEmbedded{
+			SyncIndex: "111aaa",
+			Cluster: toolchainv1alpha1.Cluster{
+				Name: "member2-cluster",
+			},
+			UserAccountStatus: userAccount.Status,
```
xcoulon: why do you use `userAccount.Status` and not `userAccount2.Status`?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -380,71 +380,116 @@ func TestModifyUserAccounts(t *testing.T) {
...
+		})
+
+		userAccount3 := uatest.NewUserAccountFromMur(mur,
+			uatest.StatusCondition(toBeNotReady("Provisioning", "")), uatest.ResourceVersion("123abc"))
+		mur.Status.UserAccounts = append(mur.Status.UserAccounts, toolchainv1alpha1.UserAccountStatusEmbedded{
+			SyncIndex: "123abc",
+			Cluster: toolchainv1alpha1.Cluster{
+				Name: "member3-cluster",
+			},
+			UserAccountStatus: userAccount.Status,
```
xcoulon: same here: why don't you use `userAccount3.Status`?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -380,71 +380,116 @@ func TestModifyUserAccounts(t *testing.T) {
...
 func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
-	// given
 	logf.SetLogger(logf.ZapLogger(true))
 	s := apiScheme(t)
 
-	mur := murtest.NewMasterUserRecord("john",
-		murtest.StatusCondition(toBeNotReady(provisioningReason, "")),
-		murtest.AdditionalAccounts("member2-cluster", "member3-cluster"))
+	t.Run("mur status synced with updated user account statuses", func(t *testing.T) {
+		// given
```
xcoulon: could you add some comments to explain the setup you're doing, please?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -380,71 +380,116 @@ func TestModifyUserAccounts(t *testing.T) {
...
+		mur := murtest.NewMasterUserRecord("john",
+			murtest.StatusCondition(toBeNotReady(provisioningReason, "")),
+			murtest.AdditionalAccounts("member2-cluster", "member3-cluster"))
+
+		userAccount := uatest.NewUserAccountFromMur(mur,
```
xcoulon: could you refactor this code to something like https://github.com/codeready-toolchain/host-operator/pull/141/files#diff-93c8b7a9cc0875ddd6d3002d3d0b4253R460 which is more readable?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: because @alexeykazakov didn't write this code - he just moved it to a subtest :-D It was me who wrote it :-) 

technically, it doesn't matter as all the UserAccounts have the same status condition, but I can imagine that it can be a bit confusing

MatousJobanek:
```suggestion
			UserAccountStatus: userAccount2.Status,
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: same as the previous one - that was me who wrote it
```suggestion
			UserAccountStatus: userAccount3.Status,
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: my code too, so let me do it:
```suggestion
		// given
		// setup MUR that wil contain UserAccountStatusEmbedded fields for UserAccounts from "member2-cluster" and "member3-cluster" but will miss from test.MemberClusterName
		// then the reconcile should add the misssing UserAccountStatusEmbedded for the missing test.MemberClusterName cluster without updating anything else
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
alexeykazakov: It's the same old test I didn't change in the PR but sure... here we go: https://github.com/codeready-toolchain/host-operator/pull/141/commits/e81e9ec1667f125813afce6afa651df25f5bc768

alexeykazakov:
Thanks @MatousJobanek for addressing the PR comments!

PR #142, opened 2020-01-30
Implement ban user logic in UserSignupController

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-396

This PR adds support for banning users via the BannedUser resource.

E2e tests are currently under development and will be included in this PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/68

IMPORTANT NOTE: This PR introduces new validations on the `UserSignup` resource, which now require that both the `user-email` and `email-hash` label values (see [here](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L16)) are set at time of creation.  If these label values aren't set, the resource will be left in an invalid state.


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -81,6 +115,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.BannedUser{}}, &handler.EnqueueRequestsFromMapFunc{
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -46,6 +46,40 @@ const (
...
+func (b BannedUserToUserSignupMapper) Map(obj handler.MapObject) []reconcile.Request {
+	if bu, ok := obj.Object.(*toolchainv1alpha1.BannedUser); ok {
+		// look-up any associated UserSignup using the BannedUser's "toolchain.dev.openshift.com/emailHash" label
+		if emailHashLbl, exists := bu.Labels[toolchainv1alpha1.BannedUserEmailHashLabelKey]; exists {
+
+			labels := map[string]string{toolchainv1alpha1.BannedUserEmailHashLabelKey: emailHashLbl}
+			opts := client.MatchingLabels(labels)
+			userSignupList := &toolchainv1alpha1.UserSignupList{}
+			if err := b.client.List(context.TODO(), userSignupList, opts); err != nil {
+				// What should we do here??
```
xcoulon: I believe we can't do much except logging the error...

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +176,50 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+				banned = true
+				break
+			}
+		}
+	} else {
+		// If there isn't an emailHash label, then the state is invalid
+		return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusMissingEmailHash)
+	}
+
+	if banned {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I've added a log statement here.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -517,3 +652,11 @@ func (r *ReconcileUserSignup) updateStatusConditions(userSignup *toolchainv1alph
 	}
 	return r.client.Status().Update(context.TODO(), userSignup)
 }
+
+// validateEmailHash calculates an md5 hash value for the provided userEmail string, and compares it to the provided
+// userEmailHash.  If the values are the same the function returns true, otherwise it will return false
+func validateEmailHash(userEmail, userEmailHash string) bool {
+	md5hash := md5.New()
+	md5hash.Write([]byte(userEmail))
```
golangcibot: Error return value of `md5hash.Write` is not checked (from `errcheck`)

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +180,45 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Is the user banned? To determine this we query the BannedUser resource for any matching entries.  The query
+	// is based on the user's emailHash value - if there is a match, and the e-mail addresses are equal, then the
+	// user is banned.
+	banned := false
+
+	// Lookup the user email label
+	if emailLbl, exists := instance.Labels[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
```
```suggestion
	if emailLbl, exists := instance.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -485,6 +614,50 @@ func (r *ReconcileUserSignup) setStatusDeactivated(userSignup *toolchainv1alpha1
...
+		userSignup,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.UserSignupComplete,
+			Status:  corev1.ConditionFalse,
+			Reason:  failedToReadBannedUsersReason,
+			Message: message,
+		})
+}
+
+func (r *ReconcileUserSignup) setStatusInvalidMissingUserEmailLabel(userSignup *toolchainv1alpha1.UserSignup, message string) error {
```
alexeykazakov: It's an annotation. Not a label.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -517,3 +690,11 @@ func (r *ReconcileUserSignup) updateStatusConditions(userSignup *toolchainv1alph
 	}
 	return r.client.Status().Update(context.TODO(), userSignup)
 }
+
+// validateEmailHash calculates an md5 hash value for the provided userEmail string, and compares it to the provided
+// userEmailHash.  If the values are the same the function returns true, otherwise it will return false
+func validateEmailHash(userEmail, userEmailHash string) bool {
+	md5hash := md5.New()
+	_, _ = md5hash.Write([]byte(userEmail))
```
alexeykazakov: Don't hide the error. Return it.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: Let's maybe move the entire banned check to a separate method?
So we can do something like:

```
banned, err := isBanned(...)
if err != nil {
}

File: pkg/controller/usersignup/usersignup_controller.go [hunk 8]
```diff
@@ -147,6 +235,18 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		if banned {
+			err = r.client.Delete(context.TODO(), &mur)
+			if err != nil {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusFailedToDeleteMUR, err,
+					"Error deleting MasterUserRecord for banned UserSignup")
+			}
+
+			reqLogger.Info("Deleted MasterUserRecord", "Name", mur.Name)
+			return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusBanning)
+		}
```
alexeykazakov: It's almost the same as the following (`if instance.Spec.Deactivated`) block.
WDYT about unifing them?
Something like this:
```
if banned {
    return r.deleteMur(mur, r.setStatusBanning, r.setStatusFailedToBan)
}
if deactivated {
    return r.deleteMur(mur, r.setStatusBanning, r.setStatusFailedToBan)
}
...
func (r *ReconcileUserSignup) deleteMur(mur, inProgressStatusUpdater, failedStatusUpdater func(...)) (reconcile.Result, error) {
	if err := r.updateStatus(reqLogger, instance, inProgressStatusUpdater); err != nil {
		logError...
	}
	if err := r.client.Delete(context.TODO(), &mur); err != nil {
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.failedStatusUpdater, err,
					"Error deleting MasterUserRecord")
	}

	reqLogger.Info("Deleted MasterUserRecord", "Name", mur.Name)
}
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -81,6 +123,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.BannedUser{}}, &handler.EnqueueRequestsFromMapFunc{
```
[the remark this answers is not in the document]
alexeykazakov: Actually no, this use case a bit different. The label doesn't contain the name of the resource. It contains the value of another label. So, never mind :)

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -36,16 +38,56 @@ const (
...
+			req := []reconcile.Request{}
+
+			for _, userSignup := range userSignupList.Items {
+				req = append(req, reconcile.Request{
+					NamespacedName: types.NamespacedName{Namespace: "toolchain-host-operator", Name: userSignup.Name},
+				})
+			}
+
+			return req
+		}
```
alexeykazakov: I wonder if we it's actually OK if we just add the email-hash label here if it's not set... Maybe in a separate go routine..

go.sum
```diff
@@ -75,6 +75,8 @@ github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDk
 github.com/codeready-toolchain/api v0.0.0-20191203182149-f994640853b0/go.mod h1:K2iSNSOwpQf+Wo3LdmrGaPlha2BLwuywwWrnbuiy8Pg=
 github.com/codeready-toolchain/api v0.0.0-20200120091303-e66748df07aa h1:P5F3iJphPPf+WEYQBYJV+MNHXcXGnJ5MBoB269HrZjs=
 github.com/codeready-toolchain/api v0.0.0-20200120091303-e66748df07aa/go.mod h1:zPbdWiEiv9SpZMvyRkY6E6n7Eynm5Y0uhJhBiGXjkg0=
+github.com/codeready-toolchain/api v0.0.0-20200204002410-9b0bcf96feca h1:u30N0mGaeR3aJrehwh+/PxH49LVZZ3EI71Z9f55jMFo=
```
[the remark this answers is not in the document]
sbryzak: Done

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -36,16 +38,56 @@ const (
...
 
 var log = logf.Log.WithName("controller_usersignup")
 
+type BannedUserToUserSignupMapper struct {
+	client client.Client
+}
+
+func (b BannedUserToUserSignupMapper) Map(obj handler.MapObject) []reconcile.Request {
+	if bu, ok := obj.Object.(*toolchainv1alpha1.BannedUser); ok {
+		// look-up any associated UserSignup using the BannedUser's "toolchain.dev.openshift.com/emailHash" label
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -36,16 +38,56 @@ const (
...
+		if emailHashLbl, exists := bu.Labels[toolchainv1alpha1.BannedUserEmailHashLabelKey]; exists {
+
+			labels := map[string]string{toolchainv1alpha1.UserSignupUserEmailHashAnnotationKey: emailHashLbl}
```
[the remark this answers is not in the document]
sbryzak: Correct.. this is my fault, I will need to fix this in the API

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -36,16 +38,56 @@ const (
...
+			if err := b.client.List(context.TODO(), userSignupList, opts); err != nil {
+				log.Error(err, "Could not list UserSignup resources with label value", toolchainv1alpha1.UserSignupUserEmailHashAnnotationKey, emailHashLbl)
+				return nil
+			}
+
+			req := []reconcile.Request{}
+
+			for _, userSignup := range userSignupList.Items {
+				req = append(req, reconcile.Request{
+					NamespacedName: types.NamespacedName{Namespace: "toolchain-host-operator", Name: userSignup.Name},
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Done.. I've added a new status updater to reflect an error here also.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I tried this, however there are too many places within the ban check where the function returns.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Good idea, I've created a unified function as suggested.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I actually do like this idea, and I'm sure it would help to reduce the difficulty in manual creation of UserSignup resources.  I'll spend a little time exploring this option to see if we break any rules by doing so.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: But all these returns are supposed to return an error, right? Or I'm missing something?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +191,48 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+			for _, bannedUser := range bannedUserList.Items {
+				if bannedUser.Spec.Email == emailLbl {
+					banned = true
+					break
+				}
+			}
+
+			hashIsValid, err := validateEmailHash(emailLbl, emailHashLbl)
+			if err != nil {
+				return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusFailedToValidateEmailHash)
```
alexeykazakov: we should return the original error here:
```suggestion
				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusFailedToValidateEmailHash, err, "Failed to validate email hash")
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +191,48 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+			} else if !hashIsValid {
+				return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusInvalidEmailHash)
```
alexeykazakov: Same here:
```suggestion
				err := errors.New("invalid hash") 
				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidEmailHash, err, "")
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +191,48 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+			}
+		} else {
+			// If there isn't an email-hash label, then the state is invalid
+			return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusMissingEmailHash)
```
alexeykazakov: Same here. Use wrapErrorWithStatus

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -131,6 +191,48 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		}
+	} else {
+		return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusInvalidMissingUserEmailAnnotation)
```
alexeykazakov: And here.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: https://github.com/codeready-toolchain/host-operator/pull/142#pullrequestreview-353435994

pkg/controller/usersignup/usersignup_controller.go
sbryzak: If we return an error here, won't the controller then attempt to reconcile again ad infinitum?  Looking at the code, it seems that it is impossible for this error to actually ever manifest, and the only reason that the md5 digest's Write() function has an error in the return value is because it implements the `Writer` interface - the actual implementation of the method doesn't allow for an error to be returned.  Check it out for yourself - `err` is never assigned:

```
func (d *digest) Write(p []byte) (nn int, err error) {
	// Note that we currently call block or blockGeneric
	// directly (guarded using haveAsm) because this allows
	// escape analysis to see that p and d don't escape.
	nn = len(p)
	d.len += uint64(nn)
	if d.nx > 0 {
		n := copy(d.x[d.nx:], p)
		d.nx += n
		if d.nx == BlockSize {
			if haveAsm {
				block(d, d.x[:])
			} else {
				blockGeneric(d, d.x[:])
			}
			d.nx = 0
		}
		p = p[n:]
	}
	if len(p) >= BlockSize {
		n := len(p) &^ (BlockSize - 1)
		if haveAsm {
			block(d, p[:n])
		} else {
			blockGeneric(d, p[:n])
		}
		p = p[n:]
	}
	if len(p) > 0 {
		d.nx = copy(d.x[:], p)
	}
	return
}
```

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Similar to above - if we return an error here, the controller continually re-attempt to reconcile, with the identical result every time.  If the e-mail hash is invalid, then the current design requires user intervention to correct the hash value, and this intervention will re-trigger the reconciliation anyway.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Same as above, a missing email hash will require user intervention.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Same again here, if the user email is missing then it will require a user to update the resource.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: They won't all return an error, many of them will simply update the `UserSignup` status (even if it's an invalid status) and then return.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: ok I see...
then lets just ignore  the error (with a comment!) in `validateEmailHash()` as you did previously:
```
func validateEmailHash(...) bool {
   ...
    _, _ := d.write(...)
   ...
}
```

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: They should return errors (see my other comment) but even if they didn't then returning no error means returning nil :) So it still should work in both cases.

pkg/controller/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Fixed

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -356,6 +500,12 @@ func TestUserSignupWithNoApprovalPolicyTreatedAsManualApproved(t *testing.T) {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "foo",
 			Namespace: operatorNamespace,
+			Annotations: map[string]string{
```
alexeykazakov: This object meta is duplicated in many places. Could we replace it by a method?
```
ObjectMeta: newObjectMeta("foo"),
```
or
```
ObjectMeta: newObjectMeta(""),
```
```
func newObjectMeta(name string) metav1.ObjectMeta {
    if name == "" {
        name = uuid.NewV4().String()
    }
    meta := metav1.ObjectMeta {
        Name:      name,
			Namespace: operatorNamespace,
			Annotations: map[string]string{
				toolchainv1alpha1.UserSignupUserEmailAnnotationKey: "foo@redhat.com",
			},
			Labels: map[string]string{
				toolchainv1alpha1.UserSignupUserEmailHashLabelKey: "fd2addbd8d82f0d2dc088fa122377eaa",
			},
	}
}
```

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1036,12 +1264,164 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+	// when
+	_, err := r.Reconcile(req)
+
+	// then
+	require.NoError(t, err)
+	err = r.client.Get(context.TODO(), test.NamespacedName(operatorNamespace, userSignup.Name), userSignup)
+	require.NoError(t, err)
+
+	// Confirm the status is set to Banned
+	test.AssertConditionsMatch(t, userSignup.Status.Conditions,
```
alexeykazakov: Also check there is no MUR created.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1036,12 +1264,164 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+
+	// Reconcile again
+	_, err = r.Reconcile(req)
+	require.NoError(t, err)
+
+	err = r.client.Get(context.TODO(), key, userSignup)
+	require.NoError(t, err)
+
+	// Confirm the status is now set to Banned
+	test.AssertConditionsMatch(t, userSignup.Status.Conditions,
```
alexeykazakov: Also check there is still no MUR.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Check added

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Check added

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: I've implemented this where it makes sense.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: I've decided against doing this for now, I wasn't able to find any examples of this being done in any of my research, plus if we go with the CLI admin command for banning users it will be handled there.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -24,13 +27,104 @@ import (
...
 )
 
+const (
+	// Status condition reasons
+	userBanningReason                = "Banning"
+	userBannedReason                 = "Banned"
+	failedToReadBannedUsersReason    = "FailedToReadBannedUsers"
+	missingUserEmailAnnotationReason = "MissingUserEmailAnnotation"
+	missingEmailHashLabelReason      = "MissingEmailHashLabel"
+	invalidEmailHashLabelReason      = "InvalidEmailHashLabel"
```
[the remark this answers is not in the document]
sbryzak: Reasons have now been moved.

go.mod
```diff
@@ -40,6 +39,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/alexeykazakov/toolchain-common v0.0.0-20200317013642-7b0b4734e580
```
alexeykazakov: Just a reminder to remove it.

PR #149, opened 2020-02-25
Disabling useraccount

tinakurian:
When the disabled property is set on MasterUserRecord, the controller should propagate this value to the UserAccount resource in the member cluster.

- Jira task
  - https://issues.redhat.com/browse/CRT-383
- Related PRs
  - common: https://github.com/codeready-toolchain/toolchain-common/pull/82
  - e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/75
  - api: https://github.com/codeready-toolchain/api/pull/108


pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +550,54 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+	logf.SetLogger(logf.ZapLogger(true))
+	s := apiScheme(t)
+	mur := murtest.NewMasterUserRecord("john", murtest.DisabledMur(true))
+	memberClient := test.NewFakeClient(t)
+	hostClient := test.NewFakeClient(t, mur)
+	cntrl := newController(hostClient, s, newGetMemberCluster(true, v1.ConditionTrue),
+		clusterClient(test.MemberClusterName, memberClient))
+
+	// when
+	_, err := cntrl.Reconcile(newMurRequest(mur))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +550,54 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+
+	memberClient.MockGet = func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+		console := types.NamespacedName{Namespace: consoleNamespace, Name: "console"}
+		che := types.NamespacedName{Namespace: cheNamespace, Name: "che"}
+		nsdName := namespacedName("toolchain-member-operator", mur.Name)
+		if _, ok := obj.(*routev1.Route); ok && (key == console || key == che ){
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		if _, ok := obj.(*routev1.Route); ok && (key == console || key == che) {
```

go.mod
```diff
@@ -53,6 +53,9 @@ require (
 
 // Pinned to kubernetes-1.14.1
 replace (
+	//TODO: replace this
```
tinakurian: Note to myself to remove this before merging

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,14 +207,33 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
...
+func (s *Synchronizer) alignDisabled() {
+	for _, uaStatus := range s.record.Status.UserAccounts {
+		if !isDisabled(uaStatus.Conditions) {
+			return
+		}
+	}
+	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeDisabled())
+}
+
+func isDisabled(conditions []toolchainv1alpha1.Condition) bool {
```
alexeykazakov: This would be useful to have in common: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go
Can we introduce a new function there. Something like: 
```
// HasConditionReason returns true if the first Condition with given conditionType from the given slice has the specified reason
func HasConditionReason(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType, reason string) bool {
	con, found := FindConditionByType(conditions, conditionType)
	return found && con.Reason == reason
}
```

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,14 +207,33 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
...
 		if !isReady(uaStatus.Conditions) {
-			return
+			return false
 		}
 	}
 	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisioned())
+	return true
 }
 
 func isReady(conditions []toolchainv1alpha1.Condition) bool {
```
alexeykazakov: I know you are not changing this code but since you are editing the file anyway.. let's replace this `isReady()` function by https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go#L39

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,14 +207,33 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
...
+func isDisabled(conditions []toolchainv1alpha1.Condition) bool {
+	for _, con := range conditions {
+		if con.Type == toolchainv1alpha1.ConditionReady {
+			return con.Reason == toolchainv1alpha1.UserAccountDisabledReason
+		}
+	}
+	return false
+}
+
 // alignReadiness checks if all embedded SAs are ready
```
alexeykazakov:
```suggestion
// alignReadiness updates the status to Provisioned and returns true if all the embedded UserAccounts are ready
```

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,14 +207,33 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
 	return url, nil
 }
 
+func (s *Synchronizer) alignDisabled() {
```
alexeykazakov:
```suggestion
// alignDisabled updates the status to Disabled if all all the embedded UserAccounts have Disabled status
func (s *Synchronizer) alignDisabled() {
```

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,14 +207,33 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
...
+	for _, uaStatus := range s.record.Status.UserAccounts {
+		if !isDisabled(uaStatus.Conditions) {
+			return
+		}
+	}
+	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeDisabled())
```
alexeykazakov: Just to be on the safe side in case there is no embedded UserAccounts.
```suggestion
	if len(s.record.Status.Conditions) > 0 {
		s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeDisabled())
	}
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +550,55 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+			userAccount := obj.(*toolchainv1alpha1.UserAccount)
+			userAccount.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(userAccount.Status.Conditions, toBeDisabled())
+			return nil
+		}
+		return memberClient.Client.Get(ctx, key, obj)
+	}
+
+	memberClient.MockUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
+		return nil
+	}
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov maybe it's not needed? when trying to run the test i ran into a problem where this stack trace of funcs is called: `ensureUserAccount()` -> `synchronizeStatus()`-> `withClusterDetails()`. Inside `withClusterDetails()` we attempt to call both `withConsoleURL()` and `withCheDashboardURL()`. So in order to get the test to complete, i mocked these calls. Is there another way of doing this?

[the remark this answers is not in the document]
alexeykazakov: Just add a console route when you creating a fake client. Thine it will be available in the client which will imitate the actual cluster client: https://github.com/codeready-toolchain/host-operator/blob/d7d62146578a94d447c20228eb702471865c79d3/pkg/controller/masteruserrecord/sync_test.go#L40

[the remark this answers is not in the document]
tinakurian: ohhh! facepalm* i did use this in other tests in member operator, i forgot. thanks :-)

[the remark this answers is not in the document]
alexeykazakov: OK. But why do you also mock update and update the status in the get?

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +548,33 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+	hostClient := test.NewFakeClient(t, mur)
+	cntrl := newController(hostClient, s, newGetMemberCluster(true, v1.ConditionTrue),
+		clusterClient(test.MemberClusterName, memberClient))
+
+	// when
+	res, err := cntrl.Reconcile(newMurRequest(mur))
+	require.NoError(t, err)
+	assert.Equal(t, reconcile.Result{}, res)
+
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +548,33 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+	// then call again to set the disabled field on user account
+	res, err = cntrl.Reconcile(newMurRequest(mur))
+	require.NoError(t, err)
+	assert.Equal(t, reconcile.Result{}, res)
+
+	userAcc := &toolchainv1alpha1.UserAccount{}
+	err = memberClient.Get(context.TODO(), types.NamespacedName{Name: mur.Name, Namespace: "toolchain-member-operator"}, userAcc)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

go.sum
```diff
@@ -117,8 +117,12 @@ github.com/codeready-toolchain/api v0.0.0-20200224173120-61614082a545 h1:sy6fTvt
 github.com/codeready-toolchain/api v0.0.0-20200224173120-61614082a545/go.mod h1:zPbdWiEiv9SpZMvyRkY6E6n7Eynm5Y0uhJhBiGXjkg0=
 github.com/codeready-toolchain/api v0.0.0-20200226022556-bb2d9dd17828 h1:lWjwDBdMDJqXWhntvoqdaEio7jT9B/lRnVssDiMP4Xo=
```
alexeykazakov: go mod tidy? :)

tinakurian: yes! fixed :-)

pkg/controller/masteruserrecord/sync.go
```diff
@@ -205,23 +206,28 @@ func (s *Synchronizer) openShift3XConsoleURL(apiEndpoint string) (string, error)
...
+func (s *Synchronizer) alignDisabled() {
 	for _, uaStatus := range s.record.Status.UserAccounts {
-		if !isReady(uaStatus.Conditions) {
+		if !condition.HasConditionReason(uaStatus.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.UserAccountDisabledReason) {
 			return
 		}
 	}
-	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisioned())
+
+	if len(s.record.Status.Conditions) > 0 {
```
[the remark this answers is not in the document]
tinakurian: @MatousJobanek this was a PR suggestion which i thought made sense to be on the safe side: https://github.com/codeready-toolchain/host-operator/pull/149#discussion_r384135230

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +548,33 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+	mur := murtest.NewMasterUserRecord("john", murtest.DisabledMur(true))
+	memberClient := test.NewFakeClient(t, consoleRoute(), cheRoute(false))
+	hostClient := test.NewFakeClient(t, mur)
+	cntrl := newController(hostClient, s, newGetMemberCluster(true, v1.ConditionTrue),
+		clusterClient(test.MemberClusterName, memberClient))
+
+	// when
+	res, err := cntrl.Reconcile(newMurRequest(mur))
+	require.NoError(t, err)
+	assert.Equal(t, reconcile.Result{}, res)
```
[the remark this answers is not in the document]
tinakurian: @MatousJobanek it's when the useraccount is created. in the next reconcile we sync the spec and status. i guess it could be considered part of the setup

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
MatousJobanek: ok, could you please explain to me why @tinakurian? I don't really see any purpose or reason for having this check there. Why should we check the length of the conditions in MUR.status - it doesn't have any impact on the result. In addition, if there is no condition, then we should add a new one containing the disabled reason, which is exactly what will do the next line `condition.AddOrUpdateStatusConditions`

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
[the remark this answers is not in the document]
MatousJobanek: so why don't you create the UserAccount as part of the test setup and add it to the fake client while initializing it - it should be pretty easy

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
alexeykazakov: It was me who asked to add this check. But I made a type in the proposed snippet.
The app idea was to check `s.record.Status.UserAccounts` (not `s.record.Status.Conditions`). So we wouldn’t set the mur status to disabled if there is no embedded user accounts/user account statues (== status is unknown).
Such a check would make the alignDusabled function more robust (set mur to disabled only if all (if any) the embedded ua statues are disabled). But in the context of the whole sync process it would not make any difference because by the time this function is called there is always at least one embedded status in the mur.
So I’m ok with both options:
1. Remove this check
2. Fix the check, which may help in the future if we refactor the code which calls this function.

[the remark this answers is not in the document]
MatousJobanek: > The idea was to check s.record.Status.UserAccounts (not s.record.Status.Conditions). So we wouldn’t set the mur status to disabled if there is no embedded user accounts/user account statues (== status is unknown).

yes, that makes sense

> Fix the check, which may help in the future if we refactor the code which calls this function.

+1

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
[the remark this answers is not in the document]
tinakurian: @MatousJobanek yeah, good idea! i've changed the test to do what you suggested!

tinakurian:
> > When the disabled property is set on MasterUserRecord, the controller should propagate this value to the UserAccount resource in the member cluster.
> 
> correct me if I'm wrong - the changes in the PR:
> 
> * replaces MUR reasons by those from the api repo
> * sets MUR.status to disabled when statuses in all UserAccounts are set to disabled
> 
> nothing else

@MatousJobanek that's right

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +548,29 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
 		DoesNotHaveFinalizer()
 }
 
+func TestDisablingMasterUserRecord(t *testing.T) {
+	// given
+	logf.SetLogger(logf.ZapLogger(true))
+	s := apiScheme(t)
+	mur := murtest.NewMasterUserRecord("john", murtest.DisabledMur(true))
+	userAccount := uatest.NewUserAccountFromMur(mur)
```
MatousJobanek: be aware that the `disabled` field is copied from MUR, which means that you initialize the test with `UserAccount.disabled == true`, which means that you don't test anything here.

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -546,6 +548,29 @@ func TestDeleteMultipleUserAccountsViaMasterUserRecordBeingDeleted(t *testing.T)
...
+	memberClient := test.NewFakeClient(t, userAccount, consoleRoute(), cheRoute(false))
+	hostClient := test.NewFakeClient(t, mur)
+	cntrl := newController(hostClient, s, newGetMemberCluster(true, v1.ConditionTrue),
+		clusterClient(test.MemberClusterName, memberClient))
+
+	// then call again to set the disabled field on user account
```
MatousJobanek:
```suggestion
	// when
```

PR #151, opened 2020-02-28
Add get/list bannedusers to reg-service SA

alexeykazakov:


MatousJobanek:
This should be changed in registration-service first and then copied to host-operator - the source of the template is stored there. 
https://github.com/codeready-toolchain/registration-service/pull/94

alexeykazakov:
@MatousJobanek we have already discussed it with @sbryzak. Shane is going to make the same changes in reg-service in https://github.com/codeready-toolchain/registration-service/pull/94
There are multiple other PRs which depend on this change so we decided to make this change in host-operator first.

MatousJobanek:
Ok. It just have a small side effect while running `make generate` in `api` repo - the target copies the file from reg-service to host-operator so I need to revert that change every time.

alexeykazakov:
> Ok. It just have a small side effect while running make generate in api repo - the target copies the file from reg-service to host-operator so I need to revert that change every time.

Oh.. I missed that. So, yeah, then updating the reg-service together with that change makes sense indeed.

PR #152, opened 2020-03-02
chore(CRT-475): add team tier

MatousJobanek:
Adds a new tear called `team` - this tier will be used for namespaces that will be shared between team members. The templates contain additional Role that enables RBAC modifications on that namespaces.

As soon as these two PRs are merged:
https://github.com/codeready-toolchain/toolchain-e2e/pull/76
https://github.com/codeready-toolchain/member-operator/pull/129
then I'll be able to provide e2e tests for this tier


deploy/templates/nstemplatetiers/team-dev.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  labels:
+    provider: codeready-toolchain
+  name: basic-dev
```
alexeykazakov:
```suggestion
  name: team-dev
```
But maybe we should call this tier something like "pipelines" or "cicd" or "team-pipelines" or "team-cicd"? Just in case we want to introduce other types of "team" tiers later. WDYT?

deploy/templates/nstemplatetiers/team-stage.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  labels:
+    provider: codeready-toolchain
+  name: basic-stage
```
alexeykazakov:
```suggestion
  name: team-stage
```

deploy/templates/nstemplatetiers/team-dev.yaml
MatousJobanek: copy-paste :-p thanks for the fix.

As for the name of the tier - the expected format of the name is: `<tier>-<nsType>`
so by changing the suffix, it will cause that we will provision namespaces `<username>-cicd` which is totally fine, but I would rather start using it when we introduce the real support of teams/organizations. For now, it would provision `mjobanek-cicd` which doesn't bring much clarity IMHO.  
I used the name `team-dev` because that is exactly the purpose of the tier - to share the namespace between other team members - there is nothing related to Tekton in the templates

MatousJobanek:
> Also, let's add some basic e2e tests. We could at least check if the tier is available (created). 

good point @alexeykazakov 
https://github.com/codeready-toolchain/toolchain-e2e/pull/77

deploy/templates/nstemplatetiers/team-dev.yaml
alexeykazakov: Sure, but I was thinking about renaming the tier only. Not the namespaces.
So, the template names would be something like `teamcicd-dev`, `teamcicd-stage` which would create the following namespaces: `<username>-dev` and `<username>-stage`.
My concern that maybe we would want to create other "team"/shared tiers later...
But we could probably stick with just "team" for now and see how it goes.

PR #160, opened 2020-03-09
feat(CRT-478): modify ChangeTierRequest_controller to change tier baseded on request

MatousJobanek:
The controller will change the tier in MUR based on `ChangeTierRequest`. If the tier is successfully changed, then it will set the status to complete and delete the `ChangeTierRequest` with a 24hours delay.
To achieve the deletion the controller will:
1. set the condition `complete` to `status = true`
2. set the reconcile requeue after 24h
3. when the reconcile is triggered again then it checks if it contains condition `complete` where `status == true`
4. if there is, then it checks the `LastTransitionTime` of the condition 
  a. if it was before 24h then it deletes the resource
  b. if it wasn't before 24h then it requeues with the duration that is the diff of `LastTransitionTime` and 24h

paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/79


go.mod
```diff
@@ -28,6 +28,8 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
+replace github.com/codeready-toolchain/toolchain-common => ../toolchain-common
```
MatousJobanek: a note for myself to remove it when it's merged

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +81,122 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+
+	err = r.changeTier(reqLogger, changeTierRequest, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, r.changeIsComplete(reqLogger, changeTierRequest)
+}
+
+func (r *ReconcileChangeTierRequest) changeIsComplete(log logr.Logger, changeTierRequest *toolchainv1alpha1.ChangeTierRequest) error {
```
golangcibot: `(*ReconcileChangeTierRequest).changeIsComplete` - `log` is unused (from `unparam`)

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,333 @@
...
+		// when
+		err := controller.wrapErrorWithStatusUpdate(log, changeTierRequest, statusUpdater, apierrors.NewBadRequest("oopsy woopsy"), "failed to create namespace")
+
+		// then
+		require.Error(t, err)
+		assert.Equal(t, "failed to create namespace: oopsy woopsy", err.Error())
+	})
+}
+
+func AssertThatChangeTierRequestIsDeleted(t *testing.T, cl client.Client, name string, condition v1alpha1.Condition) {
```
golangcibot: `AssertThatChangeTierRequestIsDeleted` - `condition` is unused (from `unparam`)

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +81,122 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, nil
+	// if is complete, then we can delete it
+	if condition.IsTrue(changeTierRequest.Status.Conditions, toolchainv1alpha1.ChangeTierRequestComplete) {
```
alexeykazakov: We will never get to that, will we? Our predicate will trigger reconcile only for NotComplete resources.

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +81,122 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+
+	err = r.changeTier(reqLogger, changeTierRequest, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, r.changeIsComplete(changeTierRequest)
+}
+
+func (r *ReconcileChangeTierRequest) changeIsComplete(changeTierRequest *toolchainv1alpha1.ChangeTierRequest) error {
```
alexeykazakov: Or maybe just complete (so it sounds more like an action)?
```suggestion
func (r *ReconcileChangeTierRequest) complete(changeTierRequest *toolchainv1alpha1.ChangeTierRequest) error {
```

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +81,122 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+	}
+
+	if err := r.client.Delete(context.TODO(), changeTierRequest, deleteIn(int64(86400))); err != nil {
+		return errs.Wrapf(err, "unable to delete ChangeTierRequest object '%s'", changeTierRequest.Name)
+	}
+
+	return nil
+}
+
+func deleteIn(seconds int64) *client.DeleteOptions {
```
alexeykazakov: WDYT about the following instead?
```
deleteIn(24 * time.Hour)
...
func deleteIn(duration time.Duration) *client.Deletoptions {
    seconds := int64(d.Seconds())
    return &client.DeleteOptions{GracePeriodSeconds: &seconds}
}
```
But feel free to keep it as is. It's very minor.

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: > Our predicate will trigger reconcile only for NotComplete resources.

I'm not 100% sure how it will behave if:
1. the tier has been changed
2. the status is set to complete
3. the deletion of the ChangeTierRequest fails
4. we return the error, but keeping the status = complete
5. will be the reconcile triggered again based on the error or filtered out based on the predicate?

This is the reason for this check - just to be on the safe side :-)

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: sure, why not :-) 88e9fad

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: your version looks much better 8e4757f

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,355 @@
...
+			Name:      tierName,
+		},
+		Spec: v1alpha1.NSTemplateTierSpec{
+			Namespaces: namespaces,
+		},
+	}
+}
+
+var (
+	notComplete = v1alpha1.ChangeTierRequestStatus{
```
golangcibot: `notComplete` is unused (from `deadcode`)

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,355 @@
...
+		Conditions: []v1alpha1.Condition{toBeNotComplete("")},
+	}
+
+	noStatus = v1alpha1.ChangeTierRequestStatus{}
+
+	complete = v1alpha1.ChangeTierRequestStatus{
```
golangcibot: `complete` is unused (from `deadcode`)

pkg/configuration/configuration.go
```diff
@@ -76,7 +79,7 @@ func New(configFilePath string) (*Registry, error) {
 			reg.SetConfigType("yaml")
 			reg.SetConfigFile(configFilePath)
 			err := reg.ReadInConfig() // Find and read the config file
-			if err != nil {           // Handle errors reading the config file.
+			if err != nil { // Handle errors reading the config file.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			if err != nil {           // Handle errors reading the config file.
```

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,355 @@
...
+func toBeNotComplete(msg string) v1alpha1.Condition {
+	return v1alpha1.Condition{
+		Type:    v1alpha1.ChangeTierRequestComplete,
+		Status:  apiv1.ConditionFalse,
+		Reason:  v1alpha1.ChangeTierRequestChangeFiledReason,
+		Message: msg,
+	}
+}
+
+func newChangeTierRequest(murName, tierName, targetCluster string, status v1alpha1.ChangeTierRequestStatus) *v1alpha1.ChangeTierRequest {
```
golangcibot: `newChangeTierRequest` - `murName` always receives `"johny"` (from `unparam`)

cmd/manager/main.go
```diff
@@ -232,6 +232,11 @@ func printConfig(cfg *configuration.Registry) {
 		logWithValues = logWithValues.WithValues("key", key, "value", value)
 	}
 	logWithValues.Info("Registration Service configuration variables:")
+
+	for key, value := range cfg.GetAllHostOperatorParameters() {
+		logWithValues = logWithValues.WithValues("key", key, "value", value)
+	}
+	logWithValues.Info("Host Operator configuration variables:")
```
xcoulon: we may not want to do that if in the future we have secrets (credentials, etc.) in the config.

pkg/configuration/configuration_test.go
```diff
@@ -125,3 +126,65 @@ func TestGetDynamicKeys(t *testing.T) {
...
+		config := getFileConfiguration(t, `ANYTHING: "SOMETHING"`)
+		assert.Empty(t, config.GetAllHostOperatorParameters())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		u, err := uuid.NewV4()
+		require.NoError(t, err)
+		newVal := u.String()
+		newVal2 := "foo=bar=baz"
+		//todo fix this
```
xcoulon: is this comment left on purpose?

pkg/configuration/configuration.go
```diff
@@ -86,24 +89,40 @@ func New(configFilePath string) (*Registry, error) {
...
+}
+
 // GetRegServiceImage returns the registration service image.
 func (c *Registry) GetRegServiceImage() string {
 	return c.regService.GetString(varImage)
 }
 
 // GetAllRegistrationServiceParameters returns the map with key-values pairs of parameters that have REGISTRATION_SERVICE prefix
 func (c *Registry) GetAllRegistrationServiceParameters() map[string]string {
+	return c.getAllParameters(RegServiceEnvPrefix)
```
xcoulon: what about having a config specific to the registration service, backed by `viper` and set with a specific prefix, instead

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, nil
+	// if is complete, then check when status was changed delete it if the requested duration has passed
+	completeCond, found := condition.FindConditionByType(changeTierRequest.Status.Conditions, toolchainv1alpha1.ChangeTierRequestComplete)
+	if found && completeCond.Status == corev1.ConditionTrue {
```
xcoulon: you could probably combine the 2 statements above into:
```
if completeCond, found := condition.FindConditionByType(changeTierRequest.Status.Conditions, toolchainv1alpha1.ChangeTierRequestComplete);  found && completeCond.Status == corev1.ConditionTrue {
...
}
```

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+		deleted, requeueAfter, err := r.checkTransitionTimeAndDelete(reqLogger, changeTierRequest, completeCond)
+		if deleted || err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{
```
xcoulon: so I assume this will work even if the operator is updated/redeployed in the interval? At startup, all `ChangeTierRequest` objects will be reconciled and a request will be requeued for the remaining time (or the time left), so that 24h after it was complete, it will be deleted. Correct?

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+
+	reqLogger.Info("Change of the tier is completed")
+	return reconcile.Result{
+		Requeue:      true,
+		RequeueAfter: r.config.GetDurationBeforeChangeRequestDeletion(),
+	}, r.setStatusChangeComplete(changeTierRequest)
+}
+
+func (r *ReconcileChangeTierRequest) checkTransitionTimeAndDelete(log logr.Logger, changeTierRequest *toolchainv1alpha1.ChangeTierRequest, completeCond toolchainv1alpha1.Condition) (bool, time.Duration, error) {
+	log.Info("the ChangeTierRequest is completed so we can ignore it")
```
xcoulon:
```suggestion
log.Info("the ChangeTierRequest is completed so we can deal with its deletion")
```

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+	timeFromLastChange := time.Since(completeCond.LastTransitionTime.Time)
```
xcoulon: I would rename this variable to `timeSinceCompletion` instead, but it's a minor suggestion.

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+
+	if timeFromLastChange >= r.config.GetDurationBeforeChangeRequestDeletion() {
+		log.Info("the ChangeTierRequest is completed for longer time than '%s' so it's ready to be deleted", r.config.GetDurationBeforeChangeRequestDeletion().String())
```
xcoulon:
```suggestion
log.Info("the ChangeTierRequest has been completed for longer time than '%s' so it's ready to be deleted", r.config.GetDurationBeforeChangeRequestDeletion().String())
```

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+		if err := r.client.Delete(context.TODO(), changeTierRequest, &client.DeleteOptions{}); err != nil {
+			return false, 0, errs.Wrapf(err, "unable to delete ChangeTierRequest object '%s'", changeTierRequest.Name)
+		}
+		return true, 0, nil
+	}
+	return false, r.config.GetDurationBeforeChangeRequestDeletion() - timeFromLastChange, nil
```
xcoulon: maybe add a `log.Info(...)` here to print the value of `r.config.GetDurationBeforeChangeRequestDeletion() - timeFromLastChange` ?

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -74,5 +85,132 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
...
+		}
+	}
+	newNsTemplateSet := toolchainv1alpha1.NSTemplateSetSpec{
+		TierName:   changeTierRequest.Spec.TierName,
+		Namespaces: namespaces,
+	}
+	changed := false
+	for i, ua := range mur.Spec.UserAccounts {
+		if changeTierRequest.Spec.TargetCluster != "" {
+			if ua.TargetCluster == changeTierRequest.Spec.TargetCluster {
```
xcoulon: these 2 `if` statements could be combined

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,345 @@
...
+		controller, request, cl := newController(t, changeTierRequest, mur, teamTier)
+
+		// when
+		_, err := controller.Reconcile(request)
+
+		// then
+		require.NoError(t, err)
+		murtest.AssertThatMasterUserRecord(t, "johny", cl).
+			AllUserAccountsHaveTier(*teamTier)
+		AssertThatChangeTierRequestHasCondition(t, cl, changeTierRequest.Name, toBeComplete())
```
xcoulon: what about having the same fluent API that we use for other CRDs ? eg, something like: 
```
AssertThatChangeTierRequest(t, cl, changeTierRequest.Name).HasCondition(Complete())
```

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -0,0 +1,345 @@
...
+		murtest.AssertThatMasterUserRecord(t, "johny", cl).
+			UserAccountHasTier("another-cluster", *teamTier).
+			UserAccountHasTier(test.MemberClusterName, *basicTier)
+		AssertThatChangeTierRequestHasCondition(t, cl, changeTierRequest.Name, toBeComplete())
+	})
+
+	t.Run("will not do anything and return requeue with shorter duration that 10s", func(t *testing.T) {
+		// given
+		mur := murtest.NewMasterUserRecord("johny")
+		changeTierRequest := newChangeTierRequest("johny", "team", "", noStatus)
```
xcoulon: what about refactoring the `newChangeTierRequest` func to a variadic func, where the last param would be the conditions (and if none is given, set to `noStatus`?)

pkg/controller/changetierrequest/changetierrequest_controller.go
```diff
@@ -83,34 +85,43 @@ func (r *ReconcileChangeTierRequest) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	// if is complete, then we can delete it
-	if condition.IsTrue(changeTierRequest.Status.Conditions, toolchainv1alpha1.ChangeTierRequestComplete) {
-		return reconcile.Result{}, r.complete(changeTierRequest)
+	// if is complete, then check when status was changed delete it if the requested duration has passed
```
alexeykazakov:
```suggestion
	// if is complete, then check when status was changed and delete it if the requested duration has passed
```

pkg/controller/changetierrequest/changetierrequest_controller.go
alexeykazakov: This is how it should work, yes.

pkg/configuration/configuration.go
MatousJobanek: I don't understand what you mean by that?

xcoulon: My question was: what about having 2 config objects, one for the host operator and one for the registration service, instead of collecting all env vars and filtering by prefix?

pkg/controller/changetierrequest/changetierrequest_controller.go
xcoulon: ok, thanks for confirming ;)

pkg/configuration/configuration.go
MatousJobanek: > My question was: what about having 2 config objects, one for the host operator and one for the registration service

well, we already have something like that - we use two Viper configs with different prefixes:
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/configuration/configuration.go#L54-L55

> instead of collecting all env vars and filtering by prefix?

we need to use different prefixes anyway. I guess that you just don't like the `getAllParameters` function. 
* For the registration-service it is necessary because we don't know which keys are going to be used there. So we need to get all env-vars and filter them by the prefix. After all, IIRC it was your idea :-) 
* for the host-operator, sure I don't really need this function because I know the keys so I can call directly the functions that return the specific values. It is just way easier to get all of the variables together with the keys.

cmd/manager/main.go
MatousJobanek: Good point. So we have two ways to avoid that:
1. filter out the variables that contain the secrets here - where we print the vars in the logs
2. call directly the functions and collect the list of the keys and values

The proper way is the second solution, but we cannot do it for the registration-service :-/ 
Technically, the secrets for the registration-service should not have been stored within the host-operator variable, so we should be safe in that case.
Let me try to rewrite it so we gather all vars and keys of the host-operator directly by calling the functions.

pkg/configuration/configuration_test.go
MatousJobanek: whoops, zombie :ghost: cf54e04

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: I could, but I don't like such long if statements - it's hard to read it. I would rather keep it as it is if you don't mind.

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: yes exactly :-)

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: done 1b63eac

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: done 7ca89d6

pkg/controller/changetierrequest/changetierrequest_controller.go
MatousJobanek: not really. I need to cover the `else` part appropriately. There are these cases possible
1. targetCluster is specified && is equal to the one in userAccount spec - then:
   * we need to change the tier inside of the userAccount and exit the loop 
2. targetCluster is specified && is NOT equal to the one in userAccount spec
   * we need to skip this UserAccount and wait until the targetCluster is matched
3. targetCluster is NOT specified
   * we need to change tier in all UserAccounts

If I combined it, then I wouldn't be able to cover all the cases

pkg/controller/changetierrequest/changetierrequest_controller_test.go
MatousJobanek: It would make sense if we had more assertions to cover with this fluent API - as we have in case of MUR or other resources. But this is not the case of `ChangeTierRequest` - I guess that it's not really worth it creating the struct and all methods to achieve the fluent API.

pkg/controller/changetierrequest/changetierrequest_controller_test.go
MatousJobanek: actually, we don't need to pass the status at all - we don't use it anywhere. It's probably some zombie from previous versions :ghost: fixed in 8280048

deploy/env/dev.yaml
```diff
@@ -11,4 +11,6 @@ registration-service:
                   "resource": "crt",
                   "clientId": "crt",
                   "public-client": true
-                }'
\ No newline at end of file
+                }'
+host-operator:
+  duration-before-change-request-deletion: '1m'
```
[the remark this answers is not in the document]
MatousJobanek: I have two reasons:
1. I needed the short duration for the demo on Wednesday :-) 
2. I was thinking if the long duration makes any sense for the dev environment because:
    * we deploy the dev environment on 10h clusters
    * when we use the dev, we usually want to try it out if it works as expected. In case I want to try the deletion of the ChangeTierRequest, then I don't want to wait for 24h for that :-)

pkg/configuration/configuration.go
```diff
@@ -86,24 +89,40 @@ func New(configFilePath string) (*Registry, error) {
 
 func (c *Registry) setConfigDefaults() {
 	c.host.SetTypeByDefaultValue(true)
+	c.host.SetDefault(varDurationBeforeChangeRequestDeletion, "24h")
+
 	c.regService.SetTypeByDefaultValue(true)
 }
 
+// GetRegServiceImage returns the registration service image.
```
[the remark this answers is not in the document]
MatousJobanek: copy-paste driven development :stuck_out_tongue:

cmd/manager/main.go
MatousJobanek: done in 26f377e

deploy/env/dev.yaml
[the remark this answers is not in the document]
alexeykazakov: These are valid reasons but... When we have some UI (admin console or something) which uses these ChangeTierRequiests then one minute timeout could be too short. Also, having dev env as close to prod as possible is almost always a good idea.
We have tests for short timeouts so we don't necessarily need to keep different timeouts in dev and prod for the testing purposes.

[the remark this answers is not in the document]
MatousJobanek: fair enough - removed in 732d4ed

pkg/controller/changetierrequest/changetierrequest_controller.go
xcoulon: ok, fair enough. I prefer the combined style of statements,  but it's a matter of taste ;)

pkg/controller/changetierrequest/changetierrequest_controller.go
xcoulon: ok, thanks

pkg/controller/changetierrequest/changetierrequest_controller.go
xcoulon: ok, got it.

pkg/controller/changetierrequest/changetierrequest_controller_test.go
xcoulon: Sure, but it's also a matter of consistency in our assertions

pkg/controller/changetierrequest/changetierrequest_controller_test.go
xcoulon: oh, ok. No need for such a change indeed, then.

cmd/manager/main.go
xcoulon: but why do we need to print the key/values, especially if there's a risk of leaking credentials?

xcoulon: I mean, if we can't avoid such a leak, shouldn't we just avoid dumping all key/value pairs in the logs?

pkg/configuration/configuration.go
xcoulon: ok, let's address this in a separate PR, by refactoring the existing code.

cmd/manager/main.go
MatousJobanek: The reason why I started printing out the variables was that I found out very difficult to figure out what was going on when the deployment of the registration service failed. 
I thought that it would be good to know which parameters are used for the reg-service deployment - mainly when we have different configuration values/parameters for different environments. I need to say that this printing was very helpful a few times and it saved me a lot of time.

pkg/configuration/configuration.go
```diff
@@ -62,7 +65,7 @@ func New(configFilePath string) (*Registry, error) {
 		c.host.SetConfigType("yaml")
 		c.host.SetConfigFile(configFilePath)
 		err := c.host.ReadInConfig() // Find and read the config file
-		if err != nil {              // Handle errors reading the config file.
+		if err != nil { // Handle errors reading the config file.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		if err != nil {              // Handle errors reading the config file.
```

cmd/manager/main.go
xcoulon: ok, let's address this issue in a separate PR, then

PR #161, opened 2020-03-10
add cluster-scoped resources in NSTemplateTiers

xcoulon:
refactored the NSTemplateTier generator using
pure functions.

also, moving templates in subdirs (one per tier),
with a strict naming scheme:
- namespace templates MUST be named `ns_<kind>.yaml`
- optional cluster resources template MUST be named `cluster.yaml`

any other file will be ignored.

Fixes https://issues.redhat.com/browse/CRT-459


xcoulon:
this is an alternative PR to https://github.com/codeready-toolchain/host-operator/pull/157 (which may be closed in favor of this one). Note that for now, the "codeready-toolchain/api" module version I'm using here does not include https://github.com/codeready-toolchain/api/pull/118

alexeykazakov:
> Note that for now, the "codeready-toolchain/api" module version I'm using here does not include codeready-toolchain/api#118

Why not?

deploy/templates/nstemplatetiers/basic/cluster.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: template.openshift.io/v1
```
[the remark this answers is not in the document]
xcoulon: ok, I'll add that one. And yes, I'll keep the "team" tier without any cluster-scoped resource manifest so we can make sure we support this kind of deployment

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -164,43 +186,47 @@ func (g nstemplatetierGenerator) newNSTemplateTiers(namespace string) (map[strin
...
-		_, _, err = g.decoder.Decode(tmpl, nil, tmplObj)
+		_, _, err := decoder.Decode(tmpls.clusterTemplate.content, nil, tmplObj)
 		if err != nil {
 			return nil, errors.Wrapf(err, "unable to generate '%s' NSTemplateTier manifest", tier)
 		}
-		// add it to the NSTemplateTier obj
-		obj.Spec.Namespaces = append(obj.Spec.Namespaces, toolchainv1alpha1.NSTemplateTierNamespace{
-			Type:     nsType,
-			Revision: revisions[nsType],
+		result.Spec.ClusterResources = toolchainv1alpha1.NSTemplateTierClusterResources{
```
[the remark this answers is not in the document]
xcoulon: To be honest, I still don't understand why we have to use a pointer (I know it works), whereas in 
```
// NSTemplateSetNamespace the namespace definition in an NSTemplateSet resource
type NSTemplateSetNamespace struct {
	// The type of the namespace. For example: ide|cicd|stage|default
	Type string `json:"type"`

	// The revision of the corresponding template
	Revision string `json:"revision"`

	// Optional field. Used to specify a custom template
	Template string `json:"template,omitempty"`
}
```
the `Template` field is correctly omitted when empty and it works. In other words, I would like to make sure that the pointer is really the way we have to proceed here. We don't use pointers anywhere else in our API types, so I'm a bit surprised we need that now.

xcoulon:
> The overall approach looks good! I like the new template structure. 

thanks for your feedback @alexeykazakov ;)

> But is there any reason for ignoring my fixes in #157? :)

please see my response here: https://github.com/codeready-toolchain/host-operator/pull/161#discussion_r390813299

(An alternative would be to use an array for the cluster-scoped resource, as this can be `nil`)

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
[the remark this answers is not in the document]
xcoulon: ok, so as mentioned in https://github.com/kubernetes/community/blob/6a41010589b0ba425fbf8b84872538e586619d45/contributors/devel/sig-architecture/api-conventions.md:
> Using a pointer allows distinguishing unset from the zero value for that type. There are some cases where, in principle, a pointer is not needed for an optional field since the zero value is forbidden, and thus implies unset. There are examples of this in the codebase. However:
> - it can be difficult for implementors to anticipate all cases where an empty value might need to be distinguished from a zero value
> - structs are not omitted from encoder output even where omitempty is specified, which is messy;
> - having a pointer consistently imply optional is clearer for users of the Go language client, and any other clients that use corresponding types
>
> Therefore, we ask that pointers always be used with optional fields that do not have a built-in nil value.

so I'll incorporate your change in https://github.com/codeready-toolchain/host-operator/commit/f93bbe11a15087b1783692393d5588d45f7f38e4 ;)

xcoulon:
> > Note that for now, the "codeready-toolchain/api" module version I'm using here does not include [codeready-toolchain/api#118](https://github.com/codeready-toolchain/api/pull/118)
> 
> Why not?

added in fef2870

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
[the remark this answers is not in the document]
xcoulon: done in fef2870

xcoulon:
> > The overall approach looks good! I like the new template structure.
> 
> thanks for your feedback @alexeykazakov ;)
> 
> > But is there any reason for ignoring my fixes in #157? :)
> 
> please see my response here: [#161 (comment)](https://github.com/codeready-toolchain/host-operator/pull/161#discussion_r390813299)
> 
> (An alternative would be to use an array for the cluster-scoped resource, as this can be `nil`)

done in fef2870

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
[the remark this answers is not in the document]
alexeykazakov: To addition to that quote the problem here is that `omitempty` works only with specific types:

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

See https://golang.org/pkg/encoding/json/
So it doesn't work with empty arbitrary structs like `Template`. And it's not just theoretical. If you switch to an empty template vs. a pointer and remove ClusterResources yaml from the team template then our e2e tests will fail with the following error:
```
{"level":"error","ts":1583774144.763569,"logger":"cmd","msg":"","error":"unable to create the NSTemplateTiers 'basic' in namespace 'codeready-toolchain-host-09165437': NSTemplateTier.toolchain.dev.openshift.com \"basic\" is invalid: spec.clusterResources.template.objects: Invalid value: \"null\": spec.clusterResources.template.objects in body must be of type array: \"null\"","errorVerbose":"NSTemplateTier.toolchain.dev.openshift.com \"basic\" is invalid: spec.clusterResources.template.objects: Invalid value: \"null\": spec.clusterResources.template.objects in body must be of type array: \"null\"\nunable to create the NSTemplateTiers 'basic' in namespace 'codeready-toolchain-host-09165437'\ngithub.com/codeready-toolchain/host-operator/pkg/templates/nstemplatetiers.CreateOrUpdateResources\n\t/tmp/codeready-toolchain/host-operator/pkg/templates/nstemplatetiers/nstemplatetier_generator.go:40\nmain.main.func1\n\t/tmp/codeready-toolchain/host-operator/cmd/manager/main.go:193\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1357","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nmain.main.func1\n\t/tmp/codeready-toolchain/host-operator/cmd/manager/main.go:194"} 
``` 
It's because the ClusterResources fields is not omitted but it contains an empty Template with empty objects array which is forbidden, so the tier doesn't load and the operator doesn't start.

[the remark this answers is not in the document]
xcoulon: yes, that's the error I got (before using a pointer), indeed.

xcoulon:
> Thanks for addressing the comments @xcoulon!
> But since you merged with force it's not easy to review the changes from the previous review only. Overall looks good though, so I'm approving ;)
> 

Yes, sorry about that. I will avoid `force` pushes in the future.

> But before merging please create a new e2e PR. The current [codeready-toolchain/toolchain-e2e#78](https://github.com/codeready-toolchain/toolchain-e2e/pull/78) is using the old host-operator branch from the closed PR.

Ah right, good point. Let me open a new PR with a "pairing" branch name!

xcoulon:
> > Thanks for addressing the comments @xcoulon!
> > But since you merged with force it's not easy to review the changes from the previous review only. Overall looks good though, so I'm approving ;)
> 
> Yes, sorry about that. I will avoid `force` pushes in the future.
> 
> > But before merging please create a new e2e PR. The current [codeready-toolchain/toolchain-e2e#78](https://github.com/codeready-toolchain/toolchain-e2e/pull/78) is using the old host-operator branch from the closed PR.
> 
> Ah right, good point. Let me open a new PR with a "pairing" branch name!

see https://github.com/codeready-toolchain/toolchain-e2e/pull/81

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -45,25 +45,31 @@ func TestCreateOrUpdateResources(t *testing.T) {
...
+			for _, tierName := range []string{"advanced", "basic", "team"} {
 				tier := toolchainv1alpha1.NSTemplateTier{}
 				err = clt.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: tierName}, &tier)
 				require.NoError(t, err)
 				assert.Equal(t, int64(1), tier.ObjectMeta.Generation)
+				if tier.Name == "team" {
+					assert.Nil(t, tier.Spec.ClusterResources) // "team" tier should not have cluster resources set
+				} else {
+					assert.NotNil(t, tier.Spec.ClusterResources)
+				}
```
MatousJobanek: it looks a bit weird to me that we have team tier in prod with the ClusterResources set, but here we are missing it. At first glance, it was very confusing. Wouldn't it be better to use a different name for this tier? eg  `nocluster`? 
It's just my point of view - nothing significant.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -8,191 +8,218 @@ import (
...
+			assert.NotEmpty(t, tmpls["basic"].clusterTemplate.revision)
+			assert.NotEmpty(t, tmpls["basic"].clusterTemplate.content)
+			// "team" tier
+			assert.NotEmpty(t, tmpls["team"].namespaceTemplates["dev"].revision)
+			assert.NotEmpty(t, tmpls["team"].namespaceTemplates["dev"].content)
+			assert.NotEmpty(t, tmpls["team"].namespaceTemplates["stage"].revision)
+			assert.NotEmpty(t, tmpls["team"].namespaceTemplates["stage"].content)
+			require.NotNil(t, tmpls["team"].clusterTemplate)
+			assert.NotEmpty(t, tmpls["team"].clusterTemplate.revision)
+			assert.NotEmpty(t, tmpls["team"].clusterTemplate.content)
```
MatousJobanek: couldn't we this whole part replace by:
```go
for _, tier := range []string{"advanced", "basic", "team"} {
	for _, nsType := range []string{"code", "dev", "stage"} {
		if tier != "team" || nsType != "code" {
			assert.NotEmpty(t, tmpls[tier].namespaceTemplates[nsType].revision)
			assert.NotEmpty(t, tmpls[tier].namespaceTemplates[nsType].content)
		}
	}
	require.NotNil(t, tmpls[tier].clusterTemplate)
	assert.NotEmpty(t, tmpls[tier].clusterTemplate.revision)
	assert.NotEmpty(t, tmpls[tier].clusterTemplate.content)
}
```

test/templates/nstemplatetiers/metadata.yaml
```diff
@@ -1,6 +1,10 @@
-advanced-code: "123456a"
-advanced-dev: "123456b"
-advanced-stage: "123456c"
-basic-code: "123456d"
-basic-dev: "123456e"
-basic-stage: "1234567" # here, a number string
```
MatousJobanek: didn't we forget to test the number string?

deploy/templates/nstemplatetiers/advanced/cluster.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: Template
+metadata:
+  labels:
+    toolchain.dev.openshift.com/provider: codeready-toolchain
+  name: advanced-cluster-resources
+objects:
+- apiVersion: quota.openshift.io/v1
+  kind: ClusterResourceQuota
+  metadata:
+    name: for-${USERNAME}
```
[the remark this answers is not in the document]
xcoulon: yes, that sounds like a good idea, indeed.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -24,115 +23,138 @@ var log = logf.Log.WithName("templates")
...
+		tiers = append(tiers, tier)
+	}
+	sort.Strings(tiers)
 	for _, tier := range tiers {
-		log.Info("creating or updating NSTemplateTier", "namespace", tier.Namespace, "name", tier.Name)
-		if err := client.Create(context.TODO(), tier); err != nil {
+		tierObj := tierObjs[tier]
+		log.Info("creating or updating NSTemplateTier", "namespace", tierObj.Namespace, "name", tierObj.Name)
+		if err := client.Create(context.TODO(), tierObj); err != nil {
 			if !apierrors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
xcoulon: because I missed it ;) let me update that

deploy/templates/nstemplatetiers/advanced/cluster.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  labels:
+    toolchain.dev.openshift.com/provider: codeready-toolchain
```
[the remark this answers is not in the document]
xcoulon: ok, let's address this in a separate PR, for all templates at once (as you suggested on Slack)

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
xcoulon: good point. I will add a `cluster.yaml` file in the `team` tier in the `test/templates/nstemplatetiers` pkg, and an extra one called `nocluster` to avoid this confusion

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
xcoulon: yes, I was lazy 😬 but let me do that

xcoulon: done in 780068f

test/templates/nstemplatetiers/metadata.yaml
xcoulon: done in 780068f

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
xcoulon: done in 780068f

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -246,60 +276,73 @@ func TestNewNSTemplateTiers(t *testing.T) {
...
+				require.NotNil(t, tier.Spec.ClusterResources)
+				assert.NotEmpty(t, tier.Spec.ClusterResources.Revision)
+				assert.NotEmpty(t, tier.Spec.ClusterResources.Template)
+			}
+		}
 	})
+}
 
+var ExpectedRevisions = map[string]map[string]string{
+	"advanced": map[string]string{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	"advanced": {
```

deploy/templates/nstemplatetiers/advanced/cluster.yaml
```diff
@@ -0,0 +1,25 @@
...
+    quota:
+      hard:
+        limits.cpu: 1750m
+        limits.memory: 7Gi
+        requests.storage: 5Gi
+        persistentvolumeclaims: "2"
+    selector:
+      annotations:
+        openshift.io/requester: ${USERNAME}
+      labels: null
```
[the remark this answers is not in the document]
alexeykazakov: These annotation is set in our templates. So, I think it's OK to relay on something which is defined in our templates from another related template.
Why using these annotations in the first place? This is what is used in the OpenShift docs (because all namespaces created via `ProjectRequest` have that annotation) :)
But I'm OK with switching to owner label too.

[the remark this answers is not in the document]
alexeykazakov: What is important here is to eventually improve our e2e tests which will not only check that the resources created. We should actually check that these resources gives us what we want:
- The limits are actually enforced (it fails if we try to create pods and other resources if they exceed the limits).
- Users can't do stuff outside of expected roles
- etc.
IMO it's very important but will take awhile to implement. So, we can create a separate JIRA for that.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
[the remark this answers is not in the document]
xcoulon: done in 2cb547a

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -21,13 +20,16 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
golangcibot: SA1019: Package sigs.k8s.io/controller-runtime/pkg/runtime/log is deprecated: use pkg/log  (from `staticcheck`)

xcoulon: fixed in https://github.com/codeready-toolchain/host-operator/pull/161/commits/9d8ebcd2cb78a1dd3692763480822e9891895f70

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -163,44 +185,52 @@ func (g nstemplatetierGenerator) newNSTemplateTiers(namespace string) (map[strin
...
-		return nil, errors.Errorf("tier '%s' does not exist", tier)
+func newNSTemplateTier(decoder runtime.Decoder, namespace, tier string, tmpls templates) (*toolchainv1alpha1.NSTemplateTier, error) {
+	// retrieve the namespace types and order them, so we can compare
+	// with the expected templates during the tests
+	namespaceKinds := make([]string, 0, len(tmpls.namespaceTemplates))
+	for kind := range tmpls.namespaceTemplates {
+		namespaceKinds = append(namespaceKinds, kind)
 	}
-	obj := &toolchainv1alpha1.NSTemplateTier{
+	sort.Strings(namespaceKinds)
```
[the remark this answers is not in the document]
xcoulon: here we sort the namespace kinds (`code`, `dev`, etc.) whereas in the `CreateOrUpdateResources` func we sort by tier name (`advanced`, `basic`, etc.). I needed to add both to make sure that some tests failed with the expected error.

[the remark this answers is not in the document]
MatousJobanek: ah, sure :man_facepalming: thanks for clarifying

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -24,115 +23,138 @@ var log = logf.Log.WithName("templates")
...
+			content:  content,
+		}
+		switch {
+		case strings.HasPrefix(filename, "ns_"):
+			kind := strings.TrimSuffix(strings.TrimPrefix(filename, "ns_"), ".yaml")
+			results[tier].namespaceTemplates[kind] = tmpl
+		case filename == "cluster.yaml":
+			results[tier].clusterTemplate = tmpl
+		default:
+			log.Info("skipping asset: unknown scope for file", "name", name)
```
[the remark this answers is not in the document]
xcoulon: done

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -24,115 +23,138 @@ var log = logf.Log.WithName("templates")
...
+		if name == "metadata.yaml" {
+			continue
 		}
-		tierKind := data[0]
-		nsKind := data[1]
-		if _, ok := revisions[tierKind]; !ok {
-			revisions[tierKind] = make(map[string]string, 3) // expect 3 entries: 'code', 'dev' and 'stage'
+		// split the name using the `/` separator
+		parts := strings.Split(name, "/")
+		// skip any name that does not have 3 parts
```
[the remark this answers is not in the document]
xcoulon: done

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -24,115 +21,122 @@ var log = logf.Log.WithName("templates")
...
+		cl := commonclient.NewApplyClient(client, s)
+		created, err := cl.CreateOrUpdateObject(tierObj, true, nil)
+		if err != nil {
+			return errors.Wrapf(err, "unable to create or update the '%s' NSTemplateTiers in namespace '%s'", tierObj.Name, tierObj.Namespace)
+		}
+		if created {
+			log.Info("NSTemplateTier resource created", "namespace", tierObj.Namespace, "name", tierObj.Name)
 		} else {
-			log.Info("NSTemplateTier resource created", "namespace", tier.Namespace, "name", tier.Name)
+			log.Info("NSTemplateTier resource updated", "namespace", tierObj.Namespace, "name", tierObj.Name, "ResourceVersion", tierObj.ResourceVersion)
```
MatousJobanek: the first returned value of `cl.CreateOrUpdateObject` tells if the object was either created or updated - not only created.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -8,191 +8,235 @@ import (
...
+					for _, kind := range []string{"code", "dev", "stage"} {
+						t.Run(kind, func(t *testing.T) {
+							if tier == "team" && kind == "code" {
+								// not applicable
+								return
+							}
+							assert.NotEmpty(t, tmpls[tier].namespaceTemplates[kind].revision)
+							assert.NotEmpty(t, tmpls[tier].namespaceTemplates[kind].content)
+						})
+						t.Run("cluster", func(t *testing.T) {
```
MatousJobanek: give the context about the kind the cluster is tested for:
```suggestion
						t.Run("cluster " + kind, func(t *testing.T) {
```

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
xcoulon: right, I forgot to update this message after our discussion on BJN

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
xcoulon: actually, the `t.Run("cluster", ...)` should not be in the loop that ranges on the namespace kinds 🤦‍♂

xcoulon: refactored in 98dca2f

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
xcoulon: fixed in 98dca2f

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
MatousJobanek: ah right :+1:

PR #166, opened 2020-03-26
chore: rename toolchain-*-edit role to rbac-edit

MatousJobanek:
let's use some name that is unified across all templates and expresses the purpose of the role. 
We can then easily reuse the role while adding permissions to collaborators

paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/84

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
@@ -20,7 +20,7 @@ objects:
   metadata:
     labels:
       toolchain.dev.openshift.com/provider: codeready-toolchain
-    name: toolchain-che-edit
+    name: rbac-edit
     namespace: ${USERNAME}-code
   rules:
     - apiGroups:
@@ -50,12 +50,12 @@ objects:
   metadata:
     labels:
       toolchain.dev.openshift.com/provider: codeready-toolchain
-    name: user-toolchain-che-edit
+    name: user-rbac-edit
     namespace: ${USERNAME}-code
   roleRef:
     apiGroup: rbac.authorization.k8s.io
     kind: Role
-    name: toolchain-che-edit
+    name: rbac-edit
   subjects:
     - kind: User
       name: ${USERNAME}
--- a/deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/toolchain-host-operator/0.0.1/toolchain-host-operator.v0.0.1.clusterserviceversion.yaml
@@ -111,6 +111,12 @@ metadata:
           "apiVersion": "toolchain.dev.openshift.com/v1alpha1",
           "kind": "UserSignup",
           "metadata": {
+            "annotations": {
+              "toolchain.dev.openshift.com/user-email": "johnsmith@redhat.com"
+            },
+            "labels": {
+              "toolchain.dev.openshift.com/email-hash": "57e46a4968bab87c552924607e46be82"
+            },
             "name": "1a03ecac-7c0b-44fc-b66d-12dd7fb21c40"
           },
           "spec": {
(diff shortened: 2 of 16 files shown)


deploy/templates/nstemplatetiers/advanced/ns_code.yaml
```diff
@@ -50,12 +50,12 @@ objects:
   metadata:
     labels:
       toolchain.dev.openshift.com/provider: codeready-toolchain
-    name: user-toolchain-che-edit
+    name: user-rbac-edit
```
xcoulon: in https://github.com/codeready-toolchain/host-operator/blob/master/deploy/templates/nstemplatetiers/advanced/ns_dev.yaml#L18-L23 the name is `user-edit`.

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
```diff
@@ -20,7 +20,7 @@ objects:
   metadata:
     labels:
       toolchain.dev.openshift.com/provider: codeready-toolchain
-    name: toolchain-che-edit
+    name: rbac-edit
```
xcoulon: what about `user-edit`? This would match with the rolebinding as suggested in https://github.com/codeready-toolchain/host-operator/pull/166/files#diff-0d5d074f131b625ee50490c7d955290aR53

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
MatousJobanek: It's different `RoleBinding`. In that `RoleBinding` we set the `ClusterRole` `edit` to the user, here we set the `rbac-edit` `Role` to the user

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
MatousJobanek: as I wrote, we already have RoleBinding `user-edit` that is used for different `ClusterRole`.

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
xcoulon: ok, my bad. 
I understand your intent and I agree with it, but how can we easily differentiate between `user-edit` and `user-rbac-edit` ? IMO `rbac` in the `user-rbac-edit` role name does not help/clarify much :(

MatousJobanek: my understanding of the format is:
`<who-gets-the-permission>-<name-of-role>`

in the same way, we will able to set the `RoleBindings` for collaborators. So in a case I wanted to invite you to my team namespace (as an "administrator" - giving you the "full access"), I would create `RoleBindings`:
`xcoulon-at-redhat-com-edit` and `xcoulon-at-redhat-com-rbac-edit`
This is the template I would use for it: https://gist.github.com/MatousJobanek/81f552e889f5c8b6756c251b2a98074f 
If I wanted to give you only view permissions, then I would create `xcoulon-at-redhat-com-view` RB

so we could change the names to: `${USERNAME}-edit` `${USERNAME}-rbac-edit`, but I wanted to have a way to differentiate between collaborators and the owner of the namespace. So collaborators'  RBs would have names:
`<name-of-collaborator>-<name-of-role>`
but the owner of the namespace has always:
`user-<name-of-role>`

> IMO rbac in the user-rbac-edit role name does not help/clarify much :(

`user-rbac-edit` is a `RoleBinding` not a `Role`
The `Role` has the name `rbac-edit` expressing that the role is giving you edit permissions (all CRUD verbs) for the RBACs resources.

But if you have some better names, I'm open for a discussion :-)

xcoulon: ok, fine.

xcoulon: thanks for the clarification.

PR #167, opened 2020-03-30
chore: increase default cpu and memory limits

MatousJobanek:
after applying the LimitRanges (tier templates) to my namespaces, our Tekton pipelines started failing because of OOM error. We could change it directly in the steps of the Tasks, but we should take it as a sign that the default limits were set probably to low. I've increased them so:
* our pipelines should run without any problems
* it should hopefully cover most of the cases

After all, they are only defaults - everyone can change it.

paired PR https://github.com/codeready-toolchain/toolchain-e2e/pull/85


deploy/templates/nstemplatetiers/basic/ns_dev.yaml
```diff
@@ -40,8 +40,8 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 150m
-        memory: 512Mi
+        cpu: 500m
+        memory: 2Gi
```
alexeykazakov: Well... in Cluster Quota we have 
```
limits.cpu: 1750m
limits.memory: 7Gi
```
It means that users won't able to run more than 3 pods without specifying the limits. Do you think we could set these default numbers to something lower? Maybe 300m CPU and 1500Mi memory? Would it be enough for many tekton task runs? WDYT?

alexeykazakov: We can keep 500m/2Gi limits for the team tier though.

MatousJobanek: The CPU limit is probably not a problem - this can be lower, but the memory is the tricky one. Firstly I tried 1400m of memory, but our toolchain-operator pipeline failed with OOM so I raised it to 2G - to have a bigger cushion there. 
To be honest, I had the same feeling that 2G is probably too high. So let me make it lower - 1400 for memory and 300m for cpu - and set the higher limit directly in our Task

MatousJobanek: done ac773f6 
let's have all tiers with the same limits and increase the team tier when we increase also the ClusterQuosta there.
and thanks for confirming that my concern was valid :-)

PR #172, opened 2020-04-02
Delete unused image stream manifest

alexeykazakov:
This ImageStream was previously used to update the operator in our crt-stage cluster. Now we use OLM for that.


xcoulon:
> This ImageStream was previously used to update the operator in our crt-stage cluster. Now we use OLM for that.

This is an acknowledgement that we rely on OpenShift 4.x (unless someone wants to install OLM on OS3), so should we also remove code like https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/test.mk#L227-L233 too? 

@MatousJobanek WDYT?

MatousJobanek:
@xcoulon please, do no remove anything related to OS3 (Minishift). For now, it's the only way I can run e2e tests in a reasonable time.

xcoulon:
> @xcoulon please, do no remove anything related to OS3 (Minishift). For now, it's the only way I can run e2e tests in a reasonable time.

Ok, I understand. I thought you had some limitations with Minishift and it did not work for you.
I wouldn't remove anything without your review and approval anyways ;)

PR #176, opened 2020-04-03
fix: improve transformation of compliant username

MatousJobanek:
improvement of the transformation of the compliant username:
* drop `*@domain.com` suffix from the username. The reason is that a label in a domain name may contain 63 chars max. We need to keep in mind that the username is used also for the names of the namespaces and the namespace name is part of routes created in it. We should avoid creating unnecessarily long usernames otherwise someone could run out of chars per label while creating routes.
* if there is any special character, then we replace it by a hyphen `-`
* if there are more hyphens in a sequence, then we replace it by the single one
* if there is a name that starts with a hyphen, then we add a prefix `crt-`
* if there is a name that ends with a hyphen, then we add a suffix `-crt`
* if there is a name that contains only numbers, then we add a prefix `crt-`

paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/94


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -390,6 +395,24 @@ func (r *ReconcileUserSignup) generateCompliantUsername(instance *toolchainv1alp
 	return "", NewSignupError(fmt.Sprintf("unable to transform username [%s] even after 100 attempts", instance.Spec.Username))
 }
 
+func transformUsername(username string) string {
+	userName := specialCharRegexp.ReplaceAllString(strings.Split(username, "@")[0], "-")
```
alexeykazakov, line 399: It's actually possible to create a user with username starting with `@`. I've just checked it. In this case the username will be an empty string if I'm not missing anything. Let's cover this case too.
For example we could transform `@spiderman` to `crt-spiderman` or to `at-spiderman`. I like `at-spiderman` better but `crt-spiderman` would work too.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1866,3 +1846,33 @@ func newObjectMeta(name, email string) metav1.ObjectMeta {
 		},
 	}
 }
+
+func TestTransformUsername(t *testing.T) {
+	assertName(t, "some", "some@email.com")
+	assertName(t, "so-me", "so-me@email.com")
+	assertName(t, "some", "some")
+	assertName(t, "so-me", "so-me")
```
alexeykazakov, line 1854:
```suggestion
	assertName(t, "so-me", "so-me")
	assertName(t, "so-me", "so-----me")
```

xcoulon:
> improvement of the transformation of the compliant username:
> 
> * drop `*@domain.com` suffix from the username. The reason is that a label in a domain name may contain 63 chars max. We need to keep in mind that the username is used also for the names of the namespaces and the namespace name is part of routes created in it. We should avoid creating unnecessarily long usernames otherwise someone could run out of chars per label while creating routes.

I may be missing something, but is it possible that 2 users have similar usernames, one as the form of an email address (eg: `johndoe@domain.com`) and one as just a regular name (eg: `johndoe`). In that case, will be have a collision, or will we end with something like `johndoe` and `johndoe-2`?

alexeykazakov:
We will end up with johndoe-2 (if it’s not taken) in this case. Or johndoe-3
if *-2 is taken and so on.

xcoulon:
> We will end up with johndoe-2 (if it’s not taken) in this case. Or johndoe-3 if *-2 is taken and so on.
> […](#)
> 

Ok, so that logic of avoiding the collisions is still in place and is not impacted by this change. That's what I wanted to confirm. Thanks!

pkg/controller/usersignup/usersignup_controller.go
xcoulon: WWIW, I prefer the `at-` prefix (or `-at` suffix)

pkg/controller/usersignup/usersignup_controller_test.go
MatousJobanek: this is already indirectly covered by `assertName(t, "so-me", "so:#$%!$%^&me")` but it's good to add also the direct test. Thanks :+1:

MatousJobanek:
> Ok, so that logic of avoiding the collisions is still in place and is not impacted by this change. That's what I wanted to confirm. Thanks!

yes, that logic is not impacted and it should be also covered by this test: https://github.com/codeready-toolchain/host-operator/blob/d142a45476f4712dd22b003df1b03f6beb8db841/pkg/controller/usersignup/usersignup_controller_test.go#L901-L964
and here you can see that the logic will try 100 numbers and then it will fail: https://github.com/codeready-toolchain/host-operator/blob/d142a45476f4712dd22b003df1b03f6beb8db841/pkg/controller/usersignup/usersignup_controller_test.go#L1368-L1430

MatousJobanek:
> We will end up with johndoe-2 (if it’s not taken) in this case. Or johndoe-3
if *-2 is taken and so on.

it actually starts appending the numbers with `-1`, which is probably not intuitive/obvious - I'll change it to `-2`

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: Thanks for pointing this out - I missed this case.
done: 5b4422c

MatousJobanek:
> it actually starts appending the numbers with -1, which is probably not intuitive/obvious - I'll change it to -2

changed in 032ad1d and modified e2e tests in: https://github.com/codeready-toolchain/toolchain-e2e/pull/94/commits/ccb941d87833d378f3a60ff75c29677f971f5959

PR #179, opened 2020-04-07
Prevent privilege escalation in user namespaces

alexeykazakov:
Instead of allowing `"*"` verbs for `role/rolebinding` resources in user namespaces we should explicitly allow all verbs but `"escalate" `.
Since `"*"` also includes `"escalate"` it allows users to escalate privileges and create roles/rolebindings  with permissions the user doesn't already have. User would be able to modify limitranges for example.
With this change users won't be able to escalate permissions.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/98

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
@@ -30,7 +30,13 @@ objects:
         - roles
         - rolebindings
       verbs:
-        - '*'
+        - get
+        - list
+        - watch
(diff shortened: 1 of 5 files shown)


alexeykazakov:
> wow, nice @alexeykazakov I wasn't aware of the verb escalate

Yeah... There is also `bind` verb for `rolebindings` which is related to privilege escalation too. But this PR takes care of it too (we don't include `bind` in our `rbac-edit` role anymore).

PR #189, opened 2020-04-21
chore: balance values in LimitRanges/Quotas to fit in memory/cpu limits

MatousJobanek:
My explanation for the values:
We need to take into account that if we aggregate all values taken from the running pods/containers, then the resulting values cannot exceed the values set in `ClusterResourceQuotas`:
* basic/advanced:
   * cpu: 1750m
   * memory: 7Gi
* team:
   * cpu: 2000m
   * memory: 15Gi (increased from 10Gi to have some cushion - we will be able to run tasks that eat more memory in parallel) 

In addition, when there is an `EventListener` running in the namespace, it immediately takes the default values (we run 2 EventListeners in our codeready-toolchain-dev namespace). That lowers the memory/millicpu that has left for the actual Pipelines/Tasks. 

Then, the actual `TaskRun` contains (apart from the steps) also some additional init/internal containers (eg. for cloning repo, creating a directory or exporting image digest) - these containers also take the default values (since there is nothing set by default).

Counting this together, we can expect that there will be always at least 3 or 4 containers running plus the containers processing the steps. It can be even more in the case of our pipelines running in codeready-toolchain-dev namespace.

Thus, I had to change a few values:
* I decreased the `cpu` limit in all tiers to `150m`, otherwise, we could easily hit the quota limit so no pod would be initialized - that was exactly the root cause of the failure in our pipelines - there were too many pods with too many of the milicores set in their limits
* I decreased the `memory` limit for dev namespace in basic/advanced tiers to 750Mi so there is a possibility to run two TaskRuns in parallel.
* After checking the Prometheus/Grafana metrics I decreased the `memory` limit in team tier to 1Gi - it should be enough in a combination with this PR: https://github.com/codeready-toolchain/toolchain-cicd/pull/10/ - In the PR I merged steps into a single step (where it was possible) and used increased limits for the steps where it is needed - for the rest the default 1Gi should be enough. I more or less tested it and it was working fine.
* As I mentioned above, I increased the memory in ClusterResourceQuota of the team tier to 15Gi to feel a bit more relaxed and to have some cushion.

Paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/103

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_dev.yaml
@@ -64,8 +64,8 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 250m
-        memory: 1Gi
+        cpu: 150m
+        memory: 750Mi
       defaultRequest:
         cpu: 100m
         memory: 64Mi
--- a/deploy/templates/nstemplatetiers/basic/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/basic/ns_dev.yaml
@@ -32,8 +32,8 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 250m
-        memory: 1Gi
+        cpu: 150m
+        memory: 750Mi
       defaultRequest:
         cpu: 100m
         memory: 64Mi
--- a/deploy/templates/nstemplatetiers/team/cluster.yaml
+++ b/deploy/templates/nstemplatetiers/team/cluster.yaml
@@ -11,10 +11,10 @@ objects:
     quota:
       hard:
         limits.cpu: 2000m
-        limits.memory: 10Gi
+        limits.memory: 15Gi
         limits.ephemeral-storage: 5Gi
         requests.cpu: 2000m
-        requests.memory: 10Gi
+        requests.memory: 15Gi
         requests.storage: 5Gi
         requests.ephemeral-storage: 5Gi
         persistentvolumeclaims: "2"
--- a/deploy/templates/nstemplatetiers/team/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/team/ns_dev.yaml
@@ -64,8 +64,8 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 500m
-        memory: 2Gi
+        cpu: 150m
+        memory: 1Gi
       defaultRequest:
         cpu: 100m
         memory: 64Mi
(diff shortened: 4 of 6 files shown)


MatousJobanek:
> Since this is the ratio of available compute resources in the nodes. 

ok, I didn't know that

To be honest, I wanted to raise a question (or open a discussion) regarding the cpu limits at the end of the description, but I forgot to do that - It was probably too late for my brain :sleeping: 

So, as for the cpu limits - I'd like to find a better agreement between all of us (@alexeykazakov @xcoulon @michaelkleinhenz) on the values we will use. These are more thoughts:

### Decrease CPU limits

Personally, I would decrease the default CPU values even more. Eg: `requests: 20m` `limits: 50m` - these values should be more than enough for running the internal containers, event listeners or some simple steps.

These low values are fine for the namespace where we run the pipelines (`*-dev`), but what about the rest of the namespaces? I guess that for `*-code` we will need a bit more (not sure if Che workspaces are initiated with some values or ont) - in tenant service we use 1 cpu for `*-che` namespaces, so maybe something similar? If yes, then we would need to definitely raise the quotas (also for basic/advanced tiers)

In `*-stage` namespaces, we could keep it small (the same as for `*-dev`) - it's after all dedicated for running stage deployment so the deployment should contain their own limits.

### CPU quota
Technically, 2 cores per user could be enough just for running pipelines plus some small deployment (considering that we lower the default limits to what I suggested above). 
However, as was mentioned by Alexey, we should keep the ratio so it would be worth raising it too. 
I didn't raise it as it didn't seem to be important to me - the insufficient CPU won't cause a failure, it just makes the build slower.

### Memory limits/quota
As soon as the Tekton team fixes this issue https://issues.redhat.com/browse/SRVKP-676 then we will be able to decrease also the limits for the memory (and thus technically also decrease the overall quota).

### Costs
I'm not sure about the costs - what is more expensive cpu or memory? What we should focus on? And does it really matter? 
For now, we only know the values we need for building our stuff, as soon as we start onboarding other teams we can be surprised how much memory/cpu they need.

alexeykazakov:
@MatousJobanek yes, let’s discuss it. But I think a new jira issue would be a better place than a closed github issue. And now ignoring my own comment let me add couple of points here :)
We don’t pay for cpu or memory. We pay for nodes. Currently we use m.5 xlarge nodes. Each provides 4 cores and 16 Gi of memory. So the ratio is 1:4.

Regarding lowering the limits... there are some features which do not set the limits but consume a lot. For example build config.
If we set the default limits too low it would break the getting started experience in dev console. Basically “add new application” won’t work for most of the imports. And this is what users see first when logging to the dev console.
Btw we need to talk to Shoubhik to make sure the default limit thing is addressed in Build v2 in OpenShift which is currently under development.

And we don’t have to worry about -code namespaces. Che explicitly set all requests and limits for all the containers (ata boys).

PR #195, opened 2020-04-28
create template for user provisioned notification

tinakurian:

PR #196, opened 2020-04-30
managing notification templates

tinakurian:
This story entailed taking the generated assets and from this PR: https://github.com/codeready-toolchain/host-operator/pull/195, loading them and returning it to the caller. 

Jira:  https://issues.redhat.com/browse/CRT-517


pkg/templates/notificationtemplates/notification_binassets.go
```diff
@@ -0,0 +1,25 @@
+package notificationtemplates
```
xcoulon, line 1: We could probably move this file in `pkg/templates/assets` to avoid the code duplication.

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,84 @@
...
+	}
+}
+
+func WithEmptyCache() Option {
+	return func(_ *Assets) {
+		templates = nil
+	}
+}
+
+// GetNotificationTemplates returns a notification subject and body or an error
```
xcoulon, line 35: could you also mention the meaning of the returned `bool` value in the description?

pkg/templates/notificationtemplates/notification_binassets.go
tinakurian: fixed here: e936505542d70ecc34defdf3f036aec30798d88d, i also added comments to WithAssets and WithEmptyCache here: 5dadc99ff8ad9fab80cdf8b1a213dac5e725f533

pkg/templates/notificationtemplates/notification_generator.go
tinakurian: done here: 41d79047577d9c630ddcfa119727316fa607bf84

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,90 @@
...
+// WithEmptyCache clears the cache of templates
+func WithEmptyCache() Option {
+	return func(_ *assets.Assets) {
+		templates = nil
+	}
+}
+
+// GetNotificationTemplates returns a notification subject, body and a boolean
+// indicating whether or not a template was found. Otherwise, an error will be returned
+func GetNotificationTemplates(name string, opts ...Option) (*NotificationTemplate, bool, error) {
```
sbryzak, line 42: I suggest renaming this to `GetNotificationTemplate`, as it's returning a singular template.

pkg/templates/notificationtemplates/notification_generator_test.go
```diff
@@ -0,0 +1,97 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func TestGetNotificationTemplates(t *testing.T) {
+	logf.SetLogger(zap.Logger(true))
+	t.Run("ok", func(t *testing.T) {
+		t.Run("get notification templates", func(t *testing.T) {
+			// when
+			template, found, err := notificationtemplates.GetNotificationTemplates("userprovisioned", notificationtemplates.WithEmptyCache())
```
sbryzak, line 21: Does the first call to this function require additional parameters beyond the template name?  If this is so, can it be initialized during controller startup?

tinakurian: No, just the name.The options are to clear cache (if needed) or used for testing..

pkg/templates/notificationtemplates/notification_generator.go
tinakurian: done here: 775cd43fa80a0e0d0238aa4dd9c5b9bd44a51745

pkg/templates/notificationtemplates/notification_generator_test.go
```diff
@@ -0,0 +1,97 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func TestGetNotificationTemplate(t *testing.T) {
+	logf.SetLogger(zap.Logger(true))
+	t.Run("ok", func(t *testing.T) {
+		t.Run("get notification templates", func(t *testing.T) {
+			// when
+			template, found, err := notificationtemplates.GetNotificationTemplate("userprovisioned", notificationtemplates.WithEmptyCache())
```
MatousJobanek, line 21: instead of using the `WithEmptyCache` option, you could use a function that sets the global var to nil and call it using `defer` - see this test for an inspiration https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/cluster/cache_whitebox_test.go#L20

this will ensure that every test cleans the global variable up

tinakurian: sure, i thought the options was a nicer way of doing this. since it's just an option, it doesn't need to be passed in and hence doesn't need to be used in production. it also allows me to keep the tests in a tests package, rather than in the same package as the code. however, both ways achieve the same outcome. i'll replace it with a reset func

go.sum
```diff
@@ -119,6 +119,10 @@ github.com/codeready-toolchain/api v0.0.0-20200402215020-c7a5434db5fb h1:jKoTRIP
 github.com/codeready-toolchain/api v0.0.0-20200402215020-c7a5434db5fb/go.mod h1:w07CIyd2cCp1i38hdPrG4aZPbwISzMkxOmtCWE4EtdY=
 github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49 h1:A1QJx5ZrTFnQVbBYuiS6bqnFcPy7xS3fIpzrmEvKbdQ=
 github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49/go.mod h1:w07CIyd2cCp1i38hdPrG4aZPbwISzMkxOmtCWE4EtdY=
+github.com/codeready-toolchain/api v0.0.0-20200417005724-848dc61f6f9d h1:KvVlYn+JLfEltxgtk3pL7+WSGXT8FRtH58LNrTKrxmw=
+github.com/codeready-toolchain/api v0.0.0-20200417005724-848dc61f6f9d/go.mod h1:w07CIyd2cCp1i38hdPrG4aZPbwISzMkxOmtCWE4EtdY=
+github.com/codeready-toolchain/toolchain-common v0.0.0-20200417111659-353104c908c7 h1:rYAwwh1O5tGXijOfF8YN/Ox1QhyDqKBPVdH0rGHm/wM=
+github.com/codeready-toolchain/toolchain-common v0.0.0-20200417111659-353104c908c7/go.mod h1:OhMfZoXLHTu4b6pXxWd6MAuCoT0OcKYiDZdEn3oCPGI=
```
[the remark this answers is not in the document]
tinakurian: done as part of: 50b9d8a36f5bc8d799ed56bf5f07cce1b5224571

pkg/templates/assets/assets.go
```diff
@@ -1,25 +1,25 @@
...
+func (a Assets) GetNames() []string {
+	return a.Names()
 }
 
 // Asset returns the content of the given file, or an error
-func (a Assets) Asset(name string) ([]byte, error) {
-	return a.asset(name)
+func (a Assets) GetAsset(name string) ([]byte, error) {
+	return a.Asset(name)
 }
```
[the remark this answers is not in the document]
tinakurian: done as part of: 50b9d8a36f5bc8d799ed56bf5f07cce1b5224571

pkg/templates/notificationtemplates/notification_generator_test.go
```diff
@@ -0,0 +1,97 @@
...
+			// then
+			require.NoError(t, err)
+			require.NotNil(t, template)
+			assert.True(t, found)
+			assert.NotEmpty(t, template.Content)
+			assert.NotEmpty(t, template.Subject)
+			assert.Equal(t, template.Subject, "Notice: Your Red Hat CodeReady Toolchain account has been provisioned")
+			assert.Contains(t, template.Content, "You are receiving this email because you have an online <a href={{.RegistrationURL}}>Red Hat CodeReady Toolchain</a>")
```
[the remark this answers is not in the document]
tinakurian: done as part of: 50b9d8a36f5bc8d799ed56bf5f07cce1b5224571

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,90 @@
...
+	"strings"
+
+	"github.com/codeready-toolchain/host-operator/pkg/templates/assets"
+
+	"github.com/pkg/errors"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("notification-templates")
+var templates map[string]NotificationTemplate
```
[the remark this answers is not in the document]
tinakurian: done as part of: 50b9d8a36f5bc8d799ed56bf5f07cce1b5224571

pkg/templates/notificationtemplates/notification_generator_test.go
tinakurian: removed opts here: 2afd35318551e963d97e30d90eec4ac4e08c3b26

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,90 @@
...
+		assets.Asset = a.Asset
+	}
+}
+
+// WithEmptyCache clears the cache of templates
+func WithEmptyCache() Option {
+	return func(_ *assets.Assets) {
+		templates = nil
+	}
+}
```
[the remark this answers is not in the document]
tinakurian: it's removed: 2afd35318551e963d97e30d90eec4ac4e08c3b26

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,90 @@
...
+
+// GetNotificationTemplate returns a notification subject, body and a boolean
+// indicating whether or not a template was found. Otherwise, an error will be returned
+func GetNotificationTemplate(name string, opts ...Option) (*NotificationTemplate, bool, error) {
```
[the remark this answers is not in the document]
tinakurian: removed here: 2afd35318551e963d97e30d90eec4ac4e08c3b26

pkg/templates/assets/assets.go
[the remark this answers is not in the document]
xcoulon: > Having both the fields exported and also the methods in place doesn't make any sense.

yes it does.
In 
```
assets := `NewAssets(testnstemplatetiers.AssetNames, testnstemplatetiers.Asset)
```
we use the "regular" test assets, whereas in 
```
fakeAssets := nstemplatetiers.NewAssets(testnstemplatetiers.AssetNames, func(name string) ([]byte, error) {
	if name == "metadata.yaml" {
		return []byte("advanced-code: abcdef"), nil
	}
	// error occurs when fetching the content of the 'advanced-code.yaml' template
	return nil, errors.Errorf("an error")
})
```
we can replace the `Asset` with a mock func which returns a different result, so we can test for errors.

pkg/templates/notificationtemplates/notification_generator.go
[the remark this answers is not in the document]
xcoulon: it's a variadic func. Why is it an issue?

pkg/templates/notificationtemplates/notification_generator_test.go
xcoulon: the templates map used to cache the results should not be public/exposed, should it? So this is why I suggested with `WithEmptyCache` option at first.

pkg/templates/notificationtemplates/notification_generator.go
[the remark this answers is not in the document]
MatousJobanek: Because the Options are used only and only for testing purposes. In prod, it just brings confusion. Also, see https://github.com/codeready-toolchain/host-operator/pull/196#discussion_r418437916

pkg/templates/assets/assets.go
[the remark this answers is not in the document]
MatousJobanek: @xcoulon I understand the purpose of this struct - I was referring to the newly exported fields. When the fields are exported, then we don't need the methods to call the functions and vice versa.

pkg/templates/notificationtemplates/notification_generator_test.go
MatousJobanek: > the templates map used to cache the results should not be public/exposed, should it? 

not at all, but that can be solved by a "whitebox test" :-)

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1 +1,67 @@
...
+import (
+	"strings"
+
+	"github.com/codeready-toolchain/host-operator/pkg/templates/assets"
+
+	"github.com/pkg/errors"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("notification-templates")
```
MatousJobanek, line 12: this is not used anywhere 
```suggestion
```

tinakurian: good catch - removed here: 25161ddc4f96a7a916643f77d2c8fdcc908c49e6

PR #197, opened 2020-05-01
Tune default resources & qouta/limits

alexeykazakov:
- Overcommit cpu limits. Since this PR only changes the limit and not the requests it should not cause additional resource usage (thanks to cpu throttling). But it will speed up some tools like Che which still don't set cpu limits properly. This PR also increases the default CPU limit accordingly.
- Decrease the default CPU request to 10m.
- Increase storage limits and PVCs number from 5Gi/2 to 7Gi/5. Che workspace creates one PVC and Tekton needs another one. So, having more than 2 PVC limit would allow users to create their own PVCs too.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/106

### DIFF
--- a/deploy/crds/toolchain_v1alpha1_masteruserrecord_crd.yaml
+++ b/deploy/crds/toolchain_v1alpha1_masteruserrecord_crd.yaml
@@ -16,6 +16,9 @@ spec:
   - JSONPath: .spec.userAccounts[].targetCluster
     name: Cluster
     type: string
+  - JSONPath: .spec.userAccounts[].spec.nsTemplateSet.tierName
+    name: Tier
+    type: string
   - JSONPath: .spec.banned
     name: Banned
     priority: 1
--- a/deploy/crds/toolchain_v1alpha1_tiertemplate_crd.yaml
+++ b/deploy/crds/toolchain_v1alpha1_tiertemplate_crd.yaml
@@ -137,6 +137,10 @@ spec:
               required:
               - objects
               type: object
+            tierName:
+              description: 'The tier of the template. For example: "basic", "advanced",
+                or "team"'
+              type: string
             type:
               description: 'The type of the template. For example: "code", "dev",
                 "stage" or "cluster"'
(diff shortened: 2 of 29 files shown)


MatousJobanek:
> Since this PR only changes the limit and not the requests it should not cause additional resource usage

not the usage, but I guess that it influences the node scheduler and theoretically also the autoscaler in the sense that it could ask for another node to be spined up. IIRC the scheduler doesn't count only the sum of the requests, but also the sum of the limits.

> But it will speed up some tools like Che which still don't set cpu limits properly. This PR also increases the default CPU limit accordingly.

:disappointed: :man_facepalming: 

> Decrease the default CPU request to 10m.

great :+1: 

As for the limits - I thought that you wanted to keep the memory/cpu ratio at 4:1 since the same ratio is used for available resources in the nodes... https://github.com/codeready-toolchain/host-operator/pull/189#pullrequestreview-397744501 So shouldn't we raise also the memory limits?

alexeykazakov:
@MatousJobanek as far as I know the scheduler takes into account requests only. So, limits do not affect it directly. The only thing which could affect auto-scaler if some user takes all 4 cores and there is no CPU left. Then the available CPU will be fewer and another request from another user would spin up a new node.
But overall, overcommitting CPU is different comparing to memory situation. When there is no enough available CPU (after scheduling) then cpu is throttled. But if there is no enough memory then the container is killed by OOM.

So, I’m not sure about 4000m as the limit... but I wouldn’t expect too many concurrent Che workspaces for now. So I think it should not affect the autoscaler significantly. But it would improve Che UX on the other hand.
What do you think of try that and adjust it later if needed?

MatousJobanek:
> as far as I know the scheduler takes into account requests only. 

ok, then I got confused with the quotas... This is still kind of hidden magic for me - all the scheduling/scaling conditions etc.. 

>  What do you think of try that and adjust it later if needed?

sure sure, I didn't want to block it - I just added my thoughts ;-) 

And how about the ratio?

PR #199, opened 2020-05-05
Update UserSignup status when MUR is updated

rajivnathan:
### Related Issue(s)
https://issues.redhat.com/browse/CRT-527

### What does this PR do?

Fixes the problem where users would not be able to log in if the master user record name changes. The code was previously only handling first time creation of the MUR but not updates. The fix is to update `UserSignup.CompliantUsername` when the master user record name is changed.

### Acceptance Criteria
- [x] Allow users to continue using the toolchain after the MUR has been updated
- [x] Add unit tests
- [ ] ~Add e2e test~ (See comments below)
Thanks @alexeykazakov for the help with this one


rajivnathan:
@alexeykazakov I have a working unit test but I'm not sure how to get the e2e test to work.

I'm not sure it's possible to create a UserSignup with a “wrong” compliant username in the status for the e2e tests since status can only be changed via the kube system and controllers. If anyone has ideas please let me know. But I think the unit tests cover the scenario well. :)

alexeykazakov:
@rajivnathan yeah... updating Status sub-resource from the e2e client can be tricky...
What we could try to do:
- Create a UserSignup. It will automatically create a MUR. 
- Create another MUR with the same UserSignup as the owner:
```
metadata:
  labels:
    toolchain.dev.openshift.com/user-id: "<userSignupResourceName>"
  name: <NewCompliantName>
  ownerReferences:
  - apiVersion: toolchain.dev.openshift.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: UserSignup
    name: "<userSignupResourceName>"
```
So you now have two MURs with different (complaint) name for the same UserSignup. UserSignup reconcile will start throwing errors now since it detects multiple MURs but that's fine.
- Delete the first MUR. Now you have a UserSignup with the child MUR but the compliant usernames do not match. The UserSignup reconcile is supposed to update the compliant username now and use the second MUR name instead of the first/deleted MUR.

But if it doesn't work for some reason then I think we can just stick with unit tests.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1878,3 +1878,103 @@ func assertName(t *testing.T, expected, username string) {
...
+	require.Equal(t, userSignup.Name, mur.Labels[v1alpha1.MasterUserRecordUserIDLabelKey])
+	require.Equal(t, mur.Name, "foo-old")
+	require.Equal(t, userSignup.Status.CompliantUsername, "foo-old")
+
+	// delete the old MUR to trigger creation of a new MUR using the new username
+	murToDelete := &v1alpha1.MasterUserRecord{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: mur.Name, Namespace: req.Namespace}, murToDelete)
+	require.NoError(t, err)
+	err = r.client.Delete(context.TODO(), murToDelete)
+	require.NoError(t, err)
```
MatousJobanek, line 1942: you don't need to get the resource since you already have the MUR object
```suggestion
	err = r.client.Delete(context.TODO(), oldMur)
	require.NoError(t, err)
```

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1878,3 +1878,103 @@ func assertName(t *testing.T, expected, username string) {
...
+	require.Equal(t, reconcile.Result{}, res)
+
+	// after the 1st reconcile verify that the MUR still exists and its name still matches the initial UserSignup CompliantUsername
+	murs := &v1alpha1.MasterUserRecordList{}
+	err = r.client.List(context.TODO(), murs)
+	require.NoError(t, err)
+	require.Len(t, murs.Items, 1)
+	mur := murs.Items[0]
+	require.Equal(t, operatorNamespace, mur.Namespace)
```
MatousJobanek, line 1932: you can list MURs in a specific namespace, then you wouldn't need to verify the namespace in the retrieved resource

```suggestion
	err = r.client.List(context.TODO(), murs, client.InNamespace(operatorNamespace))
	require.NoError(t, err)
	require.Len(t, murs.Items, 1)
	mur := murs.Items[0]
```

rajivnathan: Good point, thanks!

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: Also a good point 🙂

rajivnathan:
@MatousJobanek 
>I like the solution with the completeStatusUpdater and still using updateStatus method

That was Alexey's idea 🙂

PR #200, opened 2020-05-05
add notification condition to mur

tinakurian:
In order to track the status of the user provisioned notification, we need to add a new Status condition to MasterUserRecord.

This PR adds the conditions for both provisioned notification created and provisioned notification failed. 

Jira: https://issues.redhat.com/browse/CRT-521

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/codeready-toolchain/host-operator
 require (
 	cloud.google.com/go v0.46.3 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.6.0 // indirect
-	github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49
+	github.com/codeready-toolchain/api v0.0.0-20200506130228-a821d9c227f1
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20200427080755-4d31a88fd82c
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v0.1.0
@@ -31,8 +31,6 @@ require (
 
 // Pinned to kubernetes-1.16.2
 replace (
-    // TODO: tina remove this
-    github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49 => github.com/tinakurian/api v0.0.0-20200505154630-c9659c388982
 	k8s.io/api => k8s.io/api v0.0.0-20191016110408-35e52d86657a
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20191016113550-5357c4baaf65
 	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20191004115801-a2eda9f80ab8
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -31,6 +31,8 @@ require (
 
 // Pinned to kubernetes-1.16.2
 replace (
+    // TODO: tina remove this
+    github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49 => ../api
```
tinakurian, line 35: note to myself to remove this

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -310,6 +310,22 @@ func toBeDisabled() toolchainv1alpha1.Condition {
 	}
 }
 
+func toBeProvisionedNotificationCreated() toolchainv1alpha1.Condition {
```
tinakurian, line 313: @alexeykazakov not sure if this name is good enough. not a huge fan but i couldn't think of anything better any suggestions?

go.mod
```diff
@@ -31,6 +31,8 @@ require (
 
 // Pinned to kubernetes-1.16.2
 replace (
+    // TODO: tina remove this
+    github.com/codeready-toolchain/api v0.0.0-20200403111848-efd1f59b2e49 => github.com/tinakurian/api v0.0.0-20200505154630-c9659c388982
```
tinakurian, line 35: note to myself to remove this

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -310,6 +310,22 @@ func toBeDisabled() toolchainv1alpha1.Condition {
...
+		Status: corev1.ConditionTrue,
+		Reason: toolchainv1alpha1.MasterUserRecordNotificationCRCreatedReason,
+	}
+}
+
+func toBeProvisionedNotificationFailed() toolchainv1alpha1.Condition {
+	return toolchainv1alpha1.Condition{
+		Type:   toolchainv1alpha1.MasterUserRecordUserProvisionedNotificationCreated,
+		Status: corev1.ConditionFalse,
+		Reason: toolchainv1alpha1.MasterUserRecordNotificationCRCreationFailedReason,
```
alexeykazakov, line 325: Add a message too. When there is an error then we set the error message as the condition message.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
alexeykazakov: Yeah... That "toBe" prefix IMO is a bit confusing. But let's keep it for the consistency sake.

tinakurian: Yeah agreed, sounds good.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
tinakurian: Good idea, added it here: 84fadd766bb49d3a0303a5d271948dbe8e461a5c

PR #202, opened 2020-05-11
chore: Remove config file and command flags configuration

rajivnathan:
Similar to the recent changes on the member operator as part of https://github.com/codeready-toolchain/member-operator/pull/150 we don't need configuration based on files or command flags. We can use config maps and environment variables for configuration.

For background, see comments https://github.com/codeready-toolchain/member-operator/pull/150#discussion_r417802075 and https://github.com/codeready-toolchain/member-operator/pull/150#discussion_r417804412


pkg/controller/registrationservice/creation_test.go
```diff
@@ -3,9 +3,10 @@ package registrationservice
 import (
 	"context"
 	"fmt"
-	commonclient "github.com/codeready-toolchain/toolchain-common/pkg/client"
 	"testing"
 
+	commonclient "github.com/codeready-toolchain/toolchain-common/pkg/client"
+
```
xcoulon, line 9: no need for a blank line here

rajivnathan: Fixed

PR #203, opened 2020-05-11
creating notification

tinakurian:
This PR adds the creation of the notification resource when a useraccount has been provisioned. 

Jira: https://issues.redhat.com/browse/CRT-509


pkg/controller/masteruserrecord/sync.go
```diff
@@ -220,6 +224,27 @@ func (s *Synchronizer) alignReadiness() bool {
 			return false
 		}
 	}
+
+	if condition.IsTrue(s.record.Status.Conditions, toolchainv1alpha1.ConditionReady) && !notificationCreated {
```
alexeykazakov, line 228: Instead of that you should do the following after the line 248:
```
	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisioned())
	if condition.IsNotTrue(s.record.Status.Conditions, toolchainv1alpha1.MasterUserRecordUserProvisionedNotificationCreated) {
		// TODO Create Notification in host. Not in member!
		...
		s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, NotificationCreated())
	}

	return true.

...
// IsNotTrue returns true if the condition with the given condition type has Unknown or False status
func IsNotTrue(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType) bool {
	condition, found := FindConditionByType(conditions, conditionType)
	return !found || condition.Status != apiv1.ConditionTrue
}
```

`IsNotTrue()` function worth moving to common. Next to `IsTrue()`.

tinakurian: @alexeykazakov ohhh! good idea! okay, i'll make that change

tinakurian: @alexeykazakov just wondering - the resource here is being created in host right? i see the TODO comment and got a bit confused. i think i've addressed this comment here: 912606199bdd2edef96701b3e4b7ba48d55a6e60

MatousJobanek: His comment is related to this line:
https://github.com/codeready-toolchain/host-operator/pull/203/files/fdea53c1e7e61a5a730a3d53caf1a354770af8a0#diff-5d9536e84d383cec9fa71e746d09dc7cR240
you should not create the Notification CR in member - you use memberCluster there.

tinakurian: ohhh okay, i think it's fixed here: 6a278695be3aecb4ed2e8ba27cab0e92ef420ac4

pkg/controller/masteruserrecord/sync.go
```diff
@@ -220,6 +222,27 @@ func (s *Synchronizer) alignReadiness() bool {
...
+				Namespace: s.record.Namespace,
+			},
+			Spec: toolchainv1alpha1.NotificationSpec{
+				UserID:   s.record.Spec.UserID,
+				Template: "userprovisioned",
+			},
+		}
+
+		if err := s.hostClient.Create(context.TODO(), notification); err != nil {
+			s.logger.Error(err, "failed to create user provisioned notification")
```
alexeykazakov, line 240: logging the error is not enough. I think instead of logging we should return an error here (wrap the original error) and wrap it with the status in the controller.
So it will be logged by the controller, error status will be added, the reconcile will be re-queued.

pkg/controller/masteruserrecord/sync.go
```diff
@@ -220,6 +222,27 @@ func (s *Synchronizer) alignReadiness() bool {
...
+	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisioned())
+	if condition.IsNotTrue(s.record.Status.Conditions, toolchainv1alpha1.MasterUserRecordUserProvisionedNotificationCreated) {
+		notification := &toolchainv1alpha1.Notification{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      s.record.Name + "-provisioned",
+				Namespace: s.record.Namespace,
+			},
+			Spec: toolchainv1alpha1.NotificationSpec{
+				UserID:   s.record.Spec.UserID,
+				Template: "userprovisioned",
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov sounds good, template loading gets this name from the directory name. i think keeping it as a const is a good idea - where should we store this const? it would need to be used by the controller as well in order to call `GetNotificationTemplate` in `notification_generator.go`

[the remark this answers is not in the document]
alexeykazakov: It's not only about keeping that template name as a const. It's also about making sure we are referring to an existing template here. So I would keep the "userprovisioned" template object available in notification_generator.go (ideally imitable). So you can access it via `notificationtemplates.UserProvisioned` in the notification controller. And as `notificationtemplates.UserProvisioned.Name` here, in the MUR sync.

Or maybe `notificationtemplates.UserProvisioned()` and `notificationtemplates.UserProvisioned().Name` if we want to get access to a deep copy of the "const" template to make sure we don't accidentally modify the template in the controllers.

@sbryzak WDYT?

pkg/controller/masteruserrecord/sync.go
```diff
@@ -220,6 +222,27 @@ func (s *Synchronizer) alignReadiness() bool {
...
+		}
+
+		if err := s.hostClient.Create(context.TODO(), notification); err != nil {
+			s.logger.Error(err, "failed to create user provisioned notification")
+			return false
+		}
+		s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisionedNotificationCreated())
+	}
+
 	s.record.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(s.record.Status.Conditions, toBeProvisioned())
```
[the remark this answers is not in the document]
tinakurian: fixed here: b6cf442251deb5042ee3030e3442f3f56d804a39

pkg/controller/masteruserrecord/sync.go
tinakurian: sounds good, it looks like the controller already wraps the error, is this what you were looking for: 00a78eca82abc149ee7203742b9fe16ac7016476?

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
sbryzak: +1 to Alexey's idea of making `UserProvisioned` a var, something like:

`var UserProvisioned, _, _ = GetNotificationTemplate("userprovisioned")`

On a side note, looking at the `GetNotificationTemplate()` function - I recommend we move this line:

`assets := assets.NewAssets(AssetNames, Asset)`

into the `loadTemplates()` function, below the cache check. So that you have is more like this:

```
func loadTemplates() (map[string]NotificationTemplate, error) {
	if notificationTemplates != nil {
		return notificationTemplates, nil
	}

	assets := assets.NewAssets(AssetNames, Asset)
```

This will avoid the `NewAssets()` call every time we call `GetNotificationTemplate()`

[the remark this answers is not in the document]
tinakurian: @sbryzak, the reason i didn't do this was because it will disallow me to have 100% code coverage. By having loadTemplates(...assets...) i'm able to pass in custom assets into the func forcing errors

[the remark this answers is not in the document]
tinakurian: @sbryzak @alexeykazakov is this what you had in mind then: d673eb3729c09fd32c2963da641ecb11a1b42a42

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -9,17 +9,18 @@ import (
 )
 
 var notificationTemplates map[string]NotificationTemplate
+var UserProvisioned, _, _ = GetNotificationTemplate("userprovisioned")
```
alexeykazakov, line 12: Make sure we have a test to check that UserProvisioned is not an empty struct and actually equals the expected template (with "userprovisioned" name).

tinakurian: done here: c523c03158f3fd42722c96bd2446973d15a3a2bd

pkg/controller/masteruserrecord/sync.go
alexeykazakov: Yes. But we don't need to log this error here. It will be logged by the controller.

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
alexeykazakov: You can do something like this to keep the code testable:
```
func loadTemplates() (map[string]NotificationTemplate, error) {
	if notificationTemplates != nil {
		return notificationTemplates, nil
	}
	return templatesForAssets(assets.NewAssets(AssetNames, Asset)
}

func templatesForAssets(assets assets.Assets) (map[string]NotificationTemplate, error {
	paths := assets.Names()
	notificationTemplates = make(map[string]NotificationTemplate)
	for _, path := range paths {
	    ...
```

Now you can test `templatesForAssets()` with whatever assets you want.

pkg/templates/notificationtemplates/notification_generator_test.go
```diff
@@ -36,9 +36,18 @@ func TestGetNotificationTemplate(t *testing.T) {
...
 		})
+		t.Run("ensure cache is used", func(t *testing.T) {
+			// when
+			defer resetNotificationTemplateCache()
+
+			require.NotEmpty(t, UserProvisioned)
+			assert.Equal(t, "userprovisioned", UserProvisioned.Name)
+			assert.Equal(t, "Notice: Your Red Hat CodeReady Toolchain account has been provisioned", UserProvisioned.Subject)
+			assert.Contains(t, UserProvisioned.Content, "You are receiving this email because you have an online <a href={{.RegistrationURL}}>Red Hat CodeReady Toolchain</a>")
+		})
```
alexeykazakov, line 50: Or just simply:
```suggestion
			assert.Equal(t, template["userprovisioned"], UserProvisioned)
		})
```

pkg/templates/notificationtemplates/notification_generator_test.go
```diff
@@ -36,9 +36,18 @@ func TestGetNotificationTemplate(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, template)
 			require.NotNil(t, template["userprovisioned"])
```
alexeykazakov, line 38: It can never be nil, can it?
```suggestion
			require.NotEmpty(t, template["userprovisioned"])
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -427,6 +427,63 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
...
+			AllUserAccountsHaveStatusSyncIndex("123abc").
+			AllUserAccountsHaveCondition(userAccount.Status.Conditions[0])
+
+		// Get the notification resource
+		notification := toolchainv1alpha1.Notification{}
+		hostClient.Get(context.TODO(), namespacedName(mur.Namespace, userAccount.Name+"-provisioned"), &notification)
+		assert.Equal(t, notification.Spec.UserID, userAccount.Spec.UserID)
+		assert.Equal(t, notification.Spec.Template, "userprovisioned")
+		assert.Equal(t, notification.Name, userAccount.Name+"-provisioned")
+		assert.Equal(t, notification.Namespace, mur.Namespace)
```
alexeykazakov, line 484: Can we just add these status and resource assertions to the existing test instead of copy-pasting it?

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
tinakurian: sounds good, this is unrelated to this PR though - will create a defect maybe and fix it under that.

pkg/controller/masteruserrecord/sync.go
tinakurian: fixed as part of: 43b2953aa3f7922169954600dd7d4cabfbe0a2ac

pkg/templates/notificationtemplates/notification_generator_test.go
tinakurian: fixed as part of: 43b2953aa3f7922169954600dd7d4cabfbe0a2ac

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
tinakurian: fixed as part of: 43b2953aa3f7922169954600dd7d4cabfbe0a2ac

pkg/templates/notificationtemplates/notification_generator_test.go
tinakurian: fixed as part of: 43b2953aa3f7922169954600dd7d4cabfbe0a2ac

pkg/controller/masteruserrecord/sync.go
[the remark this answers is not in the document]
alexeykazakov: It's a tiny change which could be easily addressed here IMO instead of going trough creating JIRA/planning/PRs, etc.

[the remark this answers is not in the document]
tinakurian: addressed here: cd81b3da5005ec2e509e4e40d74ddc4615312a8b

PR #206, opened 2020-05-18
chore: update git package to git224-all

MatousJobanek:


openshift-ci/Dockerfile.tools
```diff
@@ -16,7 +16,7 @@ RUN yum install epel-release -y \
     && yum install  https://centos7.iuscommunity.org/ius-release.rpm -y \
     && yum install --enablerepo=centosplus install -y --quiet \
```
xcoulon: should we also remove the `install` pkg here? the command looks like `yum install install` and there is no such `install` pkg

PR #220, opened 2020-06-09
chore: add icon to CSV

MatousJobanek:
add icon to CSV so the formatting in OLM page in OpenShift console is better


MatousJobanek:
the icon:
image

PR #221, opened 2020-06-09
feat: NSTemplateTierController creates TemplateUpdateRequests

xcoulon:
The new NSTemplateTier controller watches:
- NSTemplateTier resources as its primary resources
- TemplateUpdateRequests as its secondary resources

When the current number of active (ie, non-deleted) TemplateUpdateRequest
resources associated with the NStemplateTier that was modified
is lower than `maxPoolSize` and there are MasterUserRecords to update,
then the controller creates an extra TemplateUpdateRequest.

MasterUserRecords to update are those labeled with:
- the tier name found in the NSTemplateTier
- but with a 'hash' label whose value does not match with the NSTemplateTier hash

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/124

Fixes https://issues.redhat.com/browse/CRT-683


go.mod
```diff
@@ -33,4 +33,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20200609152033-ef48cc411b71
```
alexeykazakov, line 36: Just a reminder.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,236 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("nstemplatetier-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource NSTemplateTier
+	err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.NSTemplateTier{}}, &handler.EnqueueRequestForObject{})
```
alexeykazakov, line 54: Do you really need to reconcile when the status changed too?
If not (we usually don't need to watch status changes of primary resources) then use `predicate.GenerationChangedPredicate{}`
```suggestion
	err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.NSTemplateTier{}}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})
```

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	log.Info("done with creating TemplateUpdateRequest resources after update of NSTemplateTier", "tier", instance.Name)
+	return reconcile.Result{}, nil
+}
+
+func newLabelSelector(instance toolchainv1alpha1.NSTemplateTier, member cluster.FedCluster) (client.MatchingLabelsSelector, error) {
+	// compute the hash of the `.spec.namespaces[].templateRef` + `.spec.clusteResource.TemplateRef`
+	hash, err := computeTemplateRefsHash(instance)
+	if err != nil {
+		log.Error(err, "unable to generate the hash for the NSTemplateTier")
```
alexeykazakov, line 167: Here and everywhere else in the controller, let's use `logger` from `Reconcile()` instead of `log`.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,236 @@
...
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	logger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	logger.Info("Reconciling NSTemplateTier")
+
+	// Fetch the NSTemplateTier instance
+	instance := toolchainv1alpha1.NSTemplateTier{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, &instance)
+	if err != nil {
+		log.Error(err, "unable to get the current NSTemplateTier instance")
```
alexeykazakov, line 96:
```suggestion
		logger.Error(err, "unable to get the current NSTemplateTier instance")
```
like that (should be using logger everywhere else too)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,236 @@
...
+		if errors.IsNotFound(err) {
```
alexeykazakov, line 97: Move this check to line 96 (before logging the error).

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		log.Error(err, "unable to get active TemplateUpdateRequests")
+		return reconcile.Result{}, err
+	}
+
+	members := r.retrieveMemberClusters()
+loop:
+	for _, member := range members {
+		// create a TemplateUpdateRequest if active count < max pol size
+		if activeTemplateUpdateRequests < maxPoolSize {
```
alexeykazakov, line 114: Let's check `activeTemplateUpdateRequests < maxPoolSize` before retrieving the members. You don't need them if you have reached the max pool size.

xcoulon:
> There is no e2e tests. Do you want to add them when introducing the TemplateUpdateRequest controller?

I will start adding E2E tests with this PR, which will be completed when the TemplateUpdateRequest is added (ie, when working on https://issues.redhat.com/browse/CRT-685)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ah right, of course 🤦‍♂️

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: yes, good point! there is no need to reconcile after changes in the status (or at least not yet)

go.mod
xcoulon: thanks ;)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ah right, of course 🤦‍♂️

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: so what happens if there's an error which is not a `NotFound` one? Shouldn't we log the `err` as well?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: sure, it makes more sense the other way around (checking `activeTemplateUpdateRequests < maxPoolSize` first)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
alexeykazakov: Yes. I would log all errors but NotFound.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: done in d0458e8

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: done in d0458e8

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ok, thanks for the clarification

xcoulon: done in 31aa93a

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: done in 628d278

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,238 @@
...
+					}
+				}
+			}
+		}
+	}
+	logger.Info("done with creating TemplateUpdateRequest resources after update of NSTemplateTier", "tier", instance.Name)
+	return reconcile.Result{}, nil
+}
+
+func newLabelSelector(instance toolchainv1alpha1.NSTemplateTier, member cluster.FedCluster) (client.MatchingLabelsSelector, error) {
```
alexeykazakov, line 168: Can we add a comment explaining what this function does and why?

xcoulon: sure! done in 135416c

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_nstemplatetier")
+
+const (
+	// MaxPoolSize the maximum number of TemplateUpdateRequest resources that can exist at the same time
+	MaxPoolSize = 5
```
alexeykazakov, line 32: Alternatively you could move it to configuration: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/configuration/configuration.go
But I'm fine with hardcoding it for now.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,251 @@
...
+			// find a MasterUserRecord which is not already up-to-date
+			// and for which there is no TemplateUpdateRequest yet
+			// fetch by subsets of "MaxPoolSize + 1" size until a MasterUserRecord candidate is found
+			murs := toolchainv1alpha1.MasterUserRecordList{}
+			matchingLabels, err := newLabelSelector(instance, *member)
+			if err != nil {
+				logger.Error(err, "unable to get MasterUserRecords to update")
+				return reconcile.Result{}, err
+			}
+			r.client.List(context.Background(), &murs,
```
alexeykazakov, line 129: golangci-lint would caught it... you missed the error handling.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,251 @@
...
+					accountTmpls := account.Spec.NSTemplateSet
+					logger.Info("checking mur account", "name", mur.Name, "tier", accountTmpls.TierName)
+					if accountTmpls.TierName != instance.Name {
+						// skip if the user account is not based on the current Tier
+						continue
+					}
+					// compare all TemplateRefs in the current MUR.UserAccount vs NSTemplateTier instance
+					if !reflect.DeepEqual(templateRefs(accountTmpls), templateRefs(instance.Spec)) {
+						logger.Info("creating a TemplateUpdateRequest to update the MasterUserRecord", "name", mur.Name, "tier", instance.Name)
+						r.client.Create(context.TODO(), &toolchainv1alpha1.TemplateUpdateRequest{
```
alexeykazakov, line 146: Error handling!

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	logger.Info("done with creating TemplateUpdateRequest resources after update of NSTemplateTier", "tier", instance.Name)
+	return reconcile.Result{}, nil
+}
+
+// newLabelSelector creates a label selector to find MasterUserRecords which are not up-to-date with
+// the templateRefs of the given NSTemplateTier.
+//
+// (longer explanation)
+// newLabelSelector creates a selector to find MasterUserRecords with the following labels:
+// - toolchain.dev.openshift.com/<member>-templates-tier: <tier>
```
alexeykazakov, line 174: Why do we care about `<member>` here? Shouldn't we just use `toolchain.dev.openshift.com/templates-tier` and `toolchain.dev.openshift.com/templates-tier-hash` labels instead?
I'm a bit concerned about binding MURs to target clusters. Our UserAccounts are bind to target/member clusters. Not MURs. And MURs can have multiple UserAccounts with different target clusters. We don't use it now but we will when supporting namespace sharing between clusters and/or migrating between clusters.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,251 @@
...
+
+// compute the hash of the `.spec.namespaces[].templateRef` + `.spec.clusteResource.TemplateRef`
+func computeTemplateRefsHash(instance toolchainv1alpha1.NSTemplateTier) (string, error) {
+	spec, err := json.Marshal(instance.Spec)
+	if err != nil {
+		return "", err
+	}
+	md5hash := md5.New()
+	// Ignore the error, as this implementation cannot return one
+	md5hash.Write([]byte(spec))
```
alexeykazakov, line 224: Let's ignore the errors explicitly. I'm still hopping we will use some linter to fetch error handling errors in the future. And that would make the linter happy ;)
```suggestion
	_, _ = md5hash.Write([]byte(spec))
```

xcoulon:
> > There is no e2e tests. Do you want to add them when introducing the TemplateUpdateRequest controller?
> 
> I will start adding E2E tests with this PR, which will be completed when the TemplateUpdateRequest is added (ie, when working on https://issues.redhat.com/browse/CRT-685)

@alexeykazakov here are the e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/124

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: hmm, I thought a user could have different tier, depending on the cluster she uses. At least the design of the MasterUserRecord allows it, AFAICS.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: yes, I've seen it and it's fixed now ;)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: fixed in 776b35b

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ok, let me apply this suggestion.

xcoulon: done in 31ef427

pkg/controller/nstemplatetier/nstemplatetier_controller.go
alexeykazakov: I see the reasoning.. But there are some issues with this approach:
- You have to iterate over all member clusters to fetch all MUR to update. So, if there are 10 member clusters you will have to list murs 10 times for each tier.
- The name of the label can't be longer than 63 characters. If we have too long cluster name then we may exceed that limit. As the matter of fact it won't work with our current crt-stage clusters. For our member cluster the label would be `toolchain.dev.openshift.com/member-member-8bs24-templates-tier-hash` which is 68 characters.

What if we use `toolchain.dev.openshift.com/<tier-name>-tier-version` label instead of these two labels?
So, we can have multiple labels (one per tier) within a MUR:
- `toolchain.dev.openshift.com/advanced-tier-version: xxx`
- `toolchain.dev.openshift.com/basic-collaborator-tier-version: yyy`

And we can look up all MURs with the outdated advanced tier by using `toolchain.dev.openshift.com/advanced-tier-version != zzz` selector.

Also I was thinking what is better to use here... `-version` or `-hash` as the label name. We could use `-version-hash` but it's too long... WDYT? I'm OK with both options.

xcoulon: > I see the reasoning.. But there are some issues with this approach:
> 
> * You have to iterate over all member clusters to fetch all MUR to update. So, if there are 10 member clusters you will have to list murs 10 times for each tier.

That's right :/ 

> * The name of the label can't be longer than 63 characters. If we have too long cluster name then we may exceed that limit. As the matter of fact it won't work with our current crt-stage clusters. For our member cluster the label would be `toolchain.dev.openshift.com/member-member-8bs24-templates-tier-hash` which is 68 characters.

Damn'it! (as a side note, why did we name the cluster `member-member-8bs24`?)

> 
> What if we use `toolchain.dev.openshift.com/<tier-name>-tier-version` label instead of these two labels?
> So, we can have multiple labels (one per tier) within a MUR:
> 
> * `toolchain.dev.openshift.com/advanced-tier-version: xxx`
> * `toolchain.dev.openshift.com/basic-collaborator-tier-version: yyy`
> 
> And we can look up all MURs with the outdated advanced tier by using `toolchain.dev.openshift.com/advanced-tier-version != zzz` selector.
>

ok, I can do that! BTW, I realize that the current solution in the PR would not work because we use the TemplateUpdateRequests have the same name as their associated MUR, so we would have an error if a user had accounts with the same tier to update on several clusters! 🤦‍♂️ 
 
> Also I was thinking what is better to use here... `-version` or `-hash` as the label name. We could use `-version-hash` but it's too long... WDYT? I'm OK with both options.

I'd rather stay with `-hash` to keep the label slightly shorter ;)

xcoulon: actually, the selector should check that the MUR has a label with key `toolchain.dev.openshift.com/<tiername>-tier-hash` but with value `!= <hash>`, because we don't want to select MUR associated with another tier (I will have to test that too, BTW!)

xcoulon: also, we need to have this new `toolchain.dev.openshift.com/<tiername>-tier-hash` label set on all MasterUserRecords prior to be able to update the templates. Otherwise, none of the MasterUserRecords will be selected!

xcoulon: done in 6a6ff2a

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: yes, I'd rather use the configuration in a separate PR.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: > > actually, the selector should check that the MUR has a label with key toolchain.dev.openshift.com/-tier-hash but with value `!= <hash>`
> 
> This is exactly what I said, no? :)

well, I wanted to emphasize that with a single selector such as:
```
toolchain.dev.openshift.com/advanced-tier-version != zzz
```
we would select MURs using an outdated "advanced" tier, but we may also select all MURs NOT using this tier. Hence, we need to verify that the `toolchain.dev.openshift.com/advanced-tier-version` label exists, whatever its value.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -142,6 +148,30 @@ func (r *ReconcileMasterUserRecord) addFinalizer(logger logr.Logger, mur *toolch
 	return nil
 }
 
+// check all tiers in use, and make sure that there's a label with a hash of the templates for each of them.
```
alexeykazakov, line 151: +1!
This is an important migration step.

```suggestion
// ensureTierTemplatesLabel checks all tiers in use, and make sure that there's a label with a hash of the templates for each of them.
```

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -124,6 +125,11 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 			return reconcile.Result{}, err
 		}
 	}
+	if err = r.ensureTierTemplatesLabel(logger, mur); err != nil {
+		logger.Error(err, "unable to check labels on MasterUserRecord")
```
alexeykazakov, line 129: Let's use `r.wrapErrorWithStatusUpdate()` instead of just logging. So it will log the error and set the status.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -114,72 +115,69 @@ func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcil
...
-			if err != nil {
-				logger.Error(err, "unable to get MasterUserRecords to update")
+		// create a TemplateUpdateRequest if active count < max pol size
+		// find a MasterUserRecord which is not already up-to-date
+		// and for which there is no TemplateUpdateRequest yet
+		// fetch by subsets of "MaxPoolSize + 1" size until a MasterUserRecord candidate is found
+		murs := toolchainv1alpha1.MasterUserRecordList{}
+		matchingLabels, err := newLabelSelector(instance)
+		if err != nil {
+			logger.Error(err, "unable to get MasterUserRecords to update")
```
alexeykazakov, line 124: Use `wrapErrorWithStatusUpdate()` instead.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -114,72 +115,69 @@ func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcil
...
+			return reconcile.Result{}, err
+		}
+		err = r.client.List(context.Background(), &murs,
+			client.InNamespace(instance.Namespace),
+			client.Limit(MaxPoolSize+1),
+			matchingLabels,
+		)
+		if err != nil {
+			logger.Error(err, "unable to get MasterUserRecords to update")
```
alexeykazakov, line 133: Use `wrapErrorWithStatusUpdate()` instead.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -114,72 +115,69 @@ func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcil
...
+			// check if there's already a TemplateUpdateRequest for this MasterUserRecord
+			templateUpdateRequest := toolchainv1alpha1.TemplateUpdateRequest{}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{
+				Namespace: instance.Namespace,
+				Name:      mur.Name,
+			}, &templateUpdateRequest); err == nil {
+				logger.Info("MasterUserRecord already has an associated TemplateUpdateRequest", "name", mur.Name)
+				continue
+			} else if !errors.IsNotFound(err) {
+				logger.Error(err, "unable to get TemplateUpdateRequest for MasterUserRecord", "name", mur.Name)
```
alexeykazakov, line 147: Use `wrapErrorWithStatusUpdate()` instead.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -114,72 +115,69 @@ func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcil
...
-						// so the request has to be requeued, in order to create other TemplateUpdateRequests
-						return reconcile.Result{Requeue: true}, nil
+						},
+						Spec: toolchainv1alpha1.TemplateUpdateRequestSpec{
+							TierName:         instance.Name,
+							Namespaces:       instance.Spec.Namespaces,
+							ClusterResources: instance.Spec.ClusterResources,
+						},
+					}); err != nil {
+						return reconcile.Result{}, err
```
alexeykazakov, line 174: And here too.
But you know what? Maybe it's worth moving all this code to a method called `ensureTemplateUpdateRequests()` and if it returns an error then wrap it in the status.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -192,38 +190,32 @@ func (r *ReconcileNSTemplateTier) Reconcile(request reconcile.Request) (reconcil
...
+	hash, err := ComputeHashForNSTemplateTier(*tier)
 	if err != nil {
 		return client.MatchingLabelsSelector{}, err
 	}
 	selector := labels.NewSelector()
-	tierLabel, err := labels.NewRequirement(TemplateTierNameLabel(member.Name), selection.Equals, []string{instance.Name})
+	tierLabel, err := labels.NewRequirement(TemplateTierHashLabelKey(tier.Name), selection.Exists, []string{})
 	selector = selector.Add(*tierLabel)
-	templateHashLabel, err := labels.NewRequirement(TemplateTierHashLabel(member.Name), selection.NotEquals, []string{hash})
+	templateHashLabel, err := labels.NewRequirement(TemplateTierHashLabelKey(tier.Name), selection.NotEquals, []string{hash})
```
alexeykazakov, line 208: Hm.. I thought that this selector would be enough and you don't need to use `selection.Exists` but maybe you are right and without `Exists` it will return resources without that label too.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
xcoulon: done in 145c0d3 (along with tests)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: I'm not sure what you have in mind with regards to the NSTemplateTier status when such an error occurred. Could you provide me with an example?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current NSTemplateTier tier")
+		return reconcile.Result{}, err
+	}
+	if requeue, err := r.ensureTemplateUpdateRequest(logger, tier); err != nil {
+		log.Error(err, "unable to ensure TemplateRequestUpdate resource after NSTemplateTier changed.")
```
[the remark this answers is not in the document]
xcoulon: I'm a bit concerned with the error message in the status. What if multiple (many!) errors happen during the process? Are we going to fill the `NSTemplateTier` status with each error? And how do we get back to a normal state (`ready=true`) afterwards?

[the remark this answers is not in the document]
alexeykazakov: How it works:
When there is an error occurred you don't just return it and end the reconcile. You also put it to the status and exit. If there was an error in the status before it will update it with the current one.
If everything is fine then in the end of the reconcile loop you set Ready=true and erase the message. We do it in all controllers which have statuses.
It looks like:
```
reconcile() {
    err:= ensureSomething()
    if err != nil {
        wrapWithStatus(err)
        return err
    }
    err= ensureSomethingElse()
    if err != nil {
        wrapWithStatus(err)
        return err
    }
    setStatusReady() // set Ready=true & erase (error) message
    return nil
}
```

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+			for _, account := range mur.Spec.UserAccounts {
+				accountTmpls := account.Spec.NSTemplateSet
+				logger.Info("checking MasterUserRecord account", "name", mur.Name, "tier", accountTmpls.TierName)
+				if accountTmpls.TierName != tier.Name {
+					// skip if the user account is not based on the current Tier
+					continue
+				}
+				// compare all TemplateRefs in the current MUR.UserAccount vs NSTemplateTier tier
+				if !reflect.DeepEqual(templateRefsFor(accountTmpls), templateRefsFor(tier.Spec)) {
```
MatousJobanek, line 165: may I know why do you compare the NSTemplateSets in UserAccounts? if the MUR doesn't have the corresponding label, then it should be either updated or the label fixed

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,226 @@
...
+					logger.Info("creating a TemplateUpdateRequest to update the MasterUserRecord", "name", mur.Name, "tier", tier.Name)
+					tur := &toolchainv1alpha1.TemplateUpdateRequest{
```
MatousJobanek, line 167: You don't set the controller reference so the controller reconcile won't be triggered - there is no mapping

MatousJobanek: ok, now I noticed that you have added it here d566c2d while I was doing the review

xcoulon: yes, that's an issue I found during e2e tests ;)

pkg/controller/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -76,13 +80,15 @@ func TestChangeTierSuccess(t *testing.T) {
 		require.NoError(t, err)
 		murtest.AssertThatMasterUserRecord(t, "johny", cl).
 			UserAccountHasTier("another-cluster", *teamTier).
-			UserAccountHasTier(test.MemberClusterName, *basicTier)
+			HasLabel(nstemplatetier.TemplateTierHashLabelKey(teamTier.Name)).
+			UserAccountHasTier(test.MemberClusterName, murtest.DefaultNSTemplateTier).
+			HasLabel(nstemplatetier.TemplateTierHashLabelKey(murtest.DefaultNSTemplateTier.Name))
```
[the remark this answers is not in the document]
xcoulon: that can be refactored later

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -142,6 +148,33 @@ func (r *ReconcileMasterUserRecord) addFinalizer(logger logr.Logger, mur *toolch
 	return nil
 }
 
+// ensureTierTemplatesLabel checks all tiers in use, and make sure that there's a label with a hash of the templates for each of them.
+// this label will be needed to select master user record that need to be updated when tier templates changed.
+// NOTE: this is a one-time update in relation with https://issues.redhat.com/browse/CRT-693 (as part of https://issues.redhat.com/browse/CRT-665)
+func (r *ReconcileMasterUserRecord) ensureTierTemplatesLabel(logger logr.Logger, mur *toolchainv1alpha1.MasterUserRecord) error {
```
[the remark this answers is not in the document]
xcoulon: > In addition, you won't be able to check that the label has been really added - the update of the label won't change the generation, so it won't trigger a new reconcile

what do you mean by checking that the label has been added?

pkg/controller/changetierrequest/changetierrequest_controller_test.go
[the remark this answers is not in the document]
MatousJobanek: I guess that it's as easy as this:
```suggestion
			UserAccountHasTier(test.MemberClusterName, murtest.DefaultNSTemplateTier)
```
sorry for not making the suggestion earlier

MatousJobanek:
btw, you forgot to add an example of the TemplateUpdateRequest CRD

pkg/controller/masteruserrecord/masteruserrecord_controller.go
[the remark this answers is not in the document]
xcoulon: > I think that this shouldn't be here but in the UserSignup controller. I know that you modify "only" the labels and not the spec, but I would still keep the logic that changes MUR at one place if there isn't a big reason for not doing that.
> This is also why I created the function `migrateOrFixMurIfNecessary`
> https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/masteruserrecord.go#L8-L22 to migrate the existing MURs

done in 68033e5

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -124,6 +125,11 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 			return reconcile.Result{}, err
 		}
 	}
+	if err = r.ensureTierTemplatesLabel(logger, mur); err != nil {
+		r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToCheckLabelsReason), err, "unable to check labels on MasterUserRecord")
```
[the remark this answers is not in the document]
xcoulon: obsolete, since code was moved in usersignup controller (in 09e13e4)

pkg/controller/changetierrequest/changetierrequest_controller_test.go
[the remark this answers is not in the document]
xcoulon: right, I did not realize that the UserAccount assertion had access to the parent MasterUserRecord object.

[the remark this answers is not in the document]
xcoulon: done in 3c7702e

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: indeed, this check was superfluous since we already selected MasterUserRecords with an outdated label for the updated tier. I removed it in d12fa5f

xcoulon:
> btw, you forgot to add an example of the TemplateUpdateRequest CRD

done in 20430f3

pkg/controller/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -0,0 +1,575 @@
...
+			// when
+			res, err := r.Reconcile(req)
+			// then
+			require.NoError(t, err)
+			require.Equal(t, reconcile.Result{Requeue: true}, res) // expect a requeue to create more TemplateUpdateRequest resources
+			// check that a single TemplateUpdateRequest was created
+			actualTemplateUpdateRequests := toolchainv1alpha1.TemplateUpdateRequestList{}
+			err = cl.List(context.TODO(), &actualTemplateUpdateRequests)
+			require.NoError(t, err)
+			assert.Len(t, actualTemplateUpdateRequests.Items, 1)
```
MatousJobanek, line 335: it's just a suggestion for simplification of all tests - you could extract this part to:
```go
func assertNumberOfTemplateUpdateRequests(t *testing.T, cl *test.FakeClient, expectedNumber int) {
	actualTemplateUpdateRequests := toolchainv1alpha1.TemplateUpdateRequestList{}
	err := cl.List(context.TODO(), &actualTemplateUpdateRequests)
	require.NoError(t, err)
	assert.Len(t, actualTemplateUpdateRequests.Items, expectedNumber)
}
```
and then use it here (and at other places too):
```go
assertNumberOfTemplateUpdateRequests(t, cl, 1)
```

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+	}
+	logger.Info("done with creating TemplateUpdateRequest resources after update of NSTemplateTier", "tier", tier.Name)
+	return false, nil // no need to requeue
+}
+func (r *ReconcileNSTemplateTier) activeTemplateUpdateRequests(tier *toolchainv1alpha1.NSTemplateTier) (int, error) {
+	// fetch the list of TemplateUpdateRequest owned by the NSTemplateTier tier
+	templateUpdateRequests := toolchainv1alpha1.TemplateUpdateRequestList{}
+	if err := r.client.List(context.Background(), &templateUpdateRequests, client.MatchingLabels{
```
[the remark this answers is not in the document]
xcoulon: no idea, I should have used `TODO` as we do everywhere else

pkg/controller/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -0,0 +1,575 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "secret",
+			Namespace: "test-namespace",
+		},
+		Type: v1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"token": []byte("mycooltoken"),
+		},
+	}
+	initObjs = append(initObjs, secret)
```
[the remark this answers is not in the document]
xcoulon: just copy/pasting from another test

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,222 @@
...
+			}, &templateUpdateRequest); err == nil {
+				logger.Info("MasterUserRecord already has an associated TemplateUpdateRequest", "name", mur.Name)
+				continue
+			} else if !errors.IsNotFound(err) {
+				return false, errs.Wrapf(err, "unable to get TemplateUpdateRequest for MasterUserRecord '%s'", mur.Name)
+			}
+			for _, account := range mur.Spec.UserAccounts {
+				accountTmpls := account.Spec.NSTemplateSet
+				logger.Info("checking MasterUserRecord account", "name", mur.Name, "tier", accountTmpls.TierName)
+				if accountTmpls.TierName != tier.Name {
```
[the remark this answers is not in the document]
xcoulon: right, there's no need for that since I removed the previous `reflect.DeepEqual` statement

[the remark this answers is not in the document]
xcoulon: done in 17d64b4

pkg/controller/nstemplatetier/nstemplatetier_controller.go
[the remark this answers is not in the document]
xcoulon: replaced in 5acb67d

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,222 @@
...
+				if err != nil {
+					return false, err
+				}
+				if err = r.client.Create(context.TODO(), tur); err != nil {
+					return false, err
+				}
+				// the controller creates a single TemplateUpdateRequest resource per reconcile loop,
+				// so the request has to be requeued, in order to create other TemplateUpdateRequests
+				// in the next loops
+				return true, nil
```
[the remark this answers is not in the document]
xcoulon: refactored in bd040ce: no explicit `requeue=true` after creating a TemplateUpdateRequest, since the controller already watches such secondary resources.

pkg/controller/nstemplatetier/nstemplatetier_controller_test.go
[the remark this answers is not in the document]
xcoulon: removed in 91e68f6

pkg/controller/nstemplatetier/nstemplatetier_controller_test.go
xcoulon: yeah, fair point

xcoulon: refactored in 786f34a

pkg/controller/nstemplatetier/label_selector.go
```diff
@@ -0,0 +1,117 @@
...
+		return "", err
+	}
+	md5hash := md5.New()
+	// Ignore the error, as this implementation cannot return one
+	md5hash.Write(m)
+	hash := hex.EncodeToString(md5hash.Sum(nil))
+	return hash, nil
+}
+
+func templateRefsFor(obj interface{}) []string {
```
MatousJobanek, line 94: This is not used - you can remove it

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,207 @@
...
+		log.Error(err, "unable to ensure TemplateRequestUpdate resource after NSTemplateTier changed.")
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+// ensureTemplateUpdateRequest ensures that all relared MasterUserRecords are up-to-date with the NSTemplateTier that changed.
+// If not, then it creates a TemplateUpdateRequest resource for the first MasterUserRecord not up-to-date with the tier, and
+// return `true, nil` so the controller will requeue the request to create subsequent TemplateUpdateRequest resources, until the
+// `MaxPoolSize` threashold is reached.
```
MatousJobanek, line 116:
```suggestion
// ensureTemplateUpdateRequest ensures that all relared MasterUserRecords are up-to-date with the NSTemplateTier that changed.
// If not, then it creates a TemplateUpdateRequest resource for the first MasterUserRecord not up-to-date with the tier, and
// return `nil` so the controller will wait for the next reconcile to create subsequent TemplateUpdateRequest resources, until the
// `MaxPoolSize` threashold is reached.
```

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -0,0 +1,207 @@
...
+				return err
+			}
+			// the controller creates a single TemplateUpdateRequest resource per reconcile loop,
+			// and the creation of this TemplateUpdateRequest will trigger another reconcile loop
+			// since the controller watches TemplateUpdateRequests owned by the NSTemplateTier
+			return r.client.Create(context.TODO(), tur)
+		}
+	}
+	logger.Info("done with creating TemplateUpdateRequest resources after update of NSTemplateTier", "tier", tier.Name)
+	return nil // no need to requeue
```
MatousJobanek, line 180:
```suggestion
	return nil
```

pkg/controller/nstemplatetier/label_selector.go
xcoulon: I wonder why this was not reported by the `make lint` job 🤔

xcoulon: unless I missed it?

PR #226, opened 2020-06-16
notification controller with notification cleanup

tinakurian:
This PR creates a base notification controller which implements the logic to delete a notification once the duration_before_notification_deletion time has elapsed. 

Jira: https://issues.redhat.com/browse/CRT-508
Related PR: https://github.com/codeready-toolchain/api/pull/156

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
-replace github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d => github.com/tinakurian/api v0.0.0-20200616003334-d7a43a723f58
+replace github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d => github.com/tinakurian/api v0.0.0-20200616134103-dc3e9ed2d4af
 
 go 1.13
--- a/go.mod
+++ b/go.mod
@@ -33,4 +33,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d => github.com/tinakurian/api v0.0.0-20200616003334-d7a43a723f58
+
 go 1.13
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/host-operator
 
 require (
-	github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d
+	github.com/codeready-toolchain/api v0.0.0-20200616141104-ea135b5fc900
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20200608102500-f81c6b7e77cf
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v0.1.0
@@ -32,7 +32,5 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.17.4 // Required by prometheus-operator
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
-
-replace github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d => github.com/tinakurian/api v0.0.0-20200616134103-dc3e9ed2d4af
 
 go 1.13
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -85,7 +85,7 @@ func (r *ReconcileNotification) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// if is delivered, then check when status was changed and delete it if the requested duration has passed
-	completeCond, found := condition.FindConditionByType(notification.Status.Conditions, toolchainv1alpha1.NotificationDelivered)
+	completeCond, found := condition.FindConditionByType(notification.Status.Conditions, toolchainv1alpha1.NotificationSent)
 	if found && completeCond.Status == corev1.ConditionTrue {
 		deleted, requeueAfter, err := r.checkTransitionTimeAndDelete(reqLogger, notification, completeCond)
 		if deleted || err != nil {
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -136,8 +136,8 @@ func (r *ReconcileNotification) setStatusNotificationDelivered(notification *too
 	return r.updateStatusConditions(
 		notification,
 		toolchainv1alpha1.Condition{
-			Type:   toolchainv1alpha1.NotificationDelivered,
+			Type:   toolchainv1alpha1.NotificationSent,
 			Status: corev1.ConditionTrue,
-			Reason: toolchainv1alpha1.NotificationServiceDeliveredReason,
+			Reason: toolchainv1alpha1.NotificationServiceSentReason,
 		})
 }
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -1,23 +1,34 @@
 package notification
 
 import (
+	"context"
+	"time"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/pkg/configuration"
+	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 
+	"github.com/go-logr/logr"
+	errs "github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -33,4 +33,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api v0.0.0-20200608101253-f0555f65f63d => github.com/tinakurian/api v0.0.0-20200616003334-d7a43a723f58
```
xcoulon, line 36: reminder ;)

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+		AssertThatNotificationIsDeleted(t, cl, notification.Name)
+	})
+
+}
+
+func TestNotificationDeliveredFailure(t *testing.T) {
+	restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_DURATION_BEFORE_NOTIFICATION_DELETION", "10s")
+	defer restore()
+
+	t.Run("will return an error since it cannot delete the ChangeTierRequest after successfully delivering", func(t *testing.T) {
```
xcoulon, line 75:
```suggestion
	t.Run("will return an error since it cannot delete the Notification after successfully delivering", func(t *testing.T) {
```

tinakurian: fixed! 5916880fc5c597bfad16e21d9f68ad0f8bbbd321

go.mod
tinakurian: thanks Xavier :-)

pkg/controller/notification/notification_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// if is delivered, then check when status was changed and delete it if the requested duration has passed
+	completeCond, found := condition.FindConditionByType(notification.Status.Conditions, toolchainv1alpha1.NotificationSent)
+	if found && completeCond.Status == corev1.ConditionTrue {
+		deleted, requeueAfter, err := r.checkTransitionTimeAndDelete(reqLogger, notification, completeCond)
+		if deleted || err != nil {
+			return reconcile.Result{}, err
```
alexeykazakov, line 92: If err!=nil then we would want to wrap the status with the error (so it will set the error status and log the error).
But we probably need to create a separate condition for the deletion error status.
Imagine we have the following situation:
1. Notification created and sent. So the Sent condition is true.
2. We attempt to delete it but it fails with an error. We do not want to set the Sent status to false with an error because the notification has been sent. But we do want to reflect that error in the status so it's obvious why it's not deleted. Let's maybe create a new condition DeletionError?

pkg/controller/notification/notification_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return reconcile.Result{
+			Requeue:      true,
+			RequeueAfter: requeueAfter,
+		}, nil
+	}
+
+	reqLogger.Info("Notification has been delivered")
```
alexeykazakov, line 100: I would slightly refactor that code.

1. Log "Notification has been sent"
2. Set the status to Sent to true.
3. Then if step 2 returns an error, log the error and return it.
4. Requeue with DurationBeforeNotificationDeletion.

Otherwise you don't log the error if there is any errors while updating the status.

pkg/controller/notification/notification_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	var updated bool
+	notification.Status.Conditions, updated = condition.AddOrUpdateStatusConditions(notification.Status.Conditions, newConditions...)
+	if !updated {
+		// Nothing changed
+		return nil
+	}
+	return r.client.Status().Update(context.TODO(), notification)
+}
+
+func (r *ReconcileNotification) setStatusNotificationDelivered(notification *toolchainv1alpha1.Notification) error {
```
alexeykazakov, line 135: It says that notification has been (not) delivered everywhere in the code while it's actually about the notification being sent.

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+func TestNotificationSuccess(t *testing.T) {
+	// given
+	restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_DURATION_BEFORE_NOTIFICATION_DELETION", "10s")
+	defer restore()
+
+	t.Run("will not do anything and return requeue with shorter duration that 10s", func(t *testing.T) {
+		// given
+		mur := murtest.NewMasterUserRecord("jane")
```
alexeykazakov, line 34: What do you need mur for in these tests? The notification controller doesn't touch MURs. So you don't need them in the tests at all, do you?

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestNotificationSuccess(t *testing.T) {
+	// given
+	restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_DURATION_BEFORE_NOTIFICATION_DELETION", "10s")
+	defer restore()
+
+	t.Run("will not do anything and return requeue with shorter duration that 10s", func(t *testing.T) {
```
alexeykazakov, line 32: Will not do anything in case of what?
This is not very informative.
A good description would tell what use case you are testing here.
Something like "sent notification is requeued with the remaining deletion timeout"

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+		// then
+		require.NoError(t, err)
+		assert.True(t, result.Requeue)
+		assert.True(t, result.RequeueAfter < cast.ToDuration("10s"))
+		assert.True(t, result.RequeueAfter > cast.ToDuration("1s"))
+		murtest.AssertThatMasterUserRecord(t, "jane", cl)
+		AssertThatNotificationHasCondition(t, cl, notification.Name, toBeDelivered())
+	})
+
+	t.Run("will delete the notification as the notification duration before deletion will already pass", func(t *testing.T) {
```
alexeykazakov, line 51: "sent notification deleted when deletion timeout passed"

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+		// given
+		mur := murtest.NewMasterUserRecord("jane")
```
alexeykazakov, line 53: Same here. You don't need any mur.

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+}
+
+func TestNotificationDeliveredFailure(t *testing.T) {
+	restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_DURATION_BEFORE_NOTIFICATION_DELETION", "10s")
+	defer restore()
+
+	t.Run("will return an error since it cannot delete the Notification after successfully delivering", func(t *testing.T) {
+		// given
+		mur := murtest.NewMasterUserRecord("jane")
```
alexeykazakov, line 77: And here.

pkg/controller/notification/notification_controller_test.go
```diff
@@ -0,0 +1,156 @@
...
+		cl.MockDelete = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
+			return fmt.Errorf("error")
+		}
+
+		// when
+		_, err := controller.Reconcile(request)
+
+		// then
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "unable to delete Notification object 'notification-name'")
```
alexeykazakov, line 92: Check the whole error message here. Not just part of it. It's important to make sure we do include the original error (from line 84).

Also check the status. It has to be in an error state (DeletionError or whatever we call this condition).

pkg/controller/notification/notification_controller.go
tinakurian: @alexeykazakov yeah, i think Shane's PRs were using the term delivered but you had asked me to rename it to sent in api. However, there is code in host now that still uses the term "Deliver" or "Delivered" i'll change the code to use send and sent instead..

pkg/controller/notification/notification_controller_test.go
tinakurian: this is actually copy/pasted from an older pr for change tier requests... so i thought the test name was fine. i will fix it in both places..

pkg/controller/notification/notification_controller_test.go
tinakurian: this is actually copy/pasted from an older pr for change tier requests... so i thought the test name was fine. i will fix it in both places..

pkg/controller/notification/notification_controller_test.go
tinakurian: this is actually copy/pasted from an older pr for change tier requests... so i thought it would be fine to check part of the error. i'll make this change in both change tier requests and notifications

pkg/controller/notification/notification_controller.go
tinakurian: @alexeykazakov +1 i noticed that the change tier request handles this the same way right now. however, i suppose change tier request could also face this situation? the request could have been completed and not deleted. i guess it could also use the new condition?

pkg/controller/notification/notification_controller.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	reqLogger.Info("Notification has been delivered")
+	return reconcile.Result{
+		Requeue:      true,
+		RequeueAfter: r.config.GetDurationBeforeNotificationDeletion(),
+	}, r.setStatusNotificationDelivered(notification)
+}
+
+func (r *ReconcileNotification) checkTransitionTimeAndDelete(log logr.Logger, notification *toolchainv1alpha1.Notification, completeCond toolchainv1alpha1.Condition) (bool, time.Duration, error) {
```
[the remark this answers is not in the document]
tinakurian: this was copy/pasted from change tier requests - i'll add the description in both places..

pkg/controller/notification/notification_controller.go
alexeykazakov: Yes. I think we need to do the same it in the TierChangeRequest. Still not sure about DeletionError condition though.
@MatousJobanek any thoughts?

pkg/controller/notification/notification_controller.go
tinakurian: Sure, this will need to be fixed for both ChangeTierRequest and Notification

PR #227, opened 2020-06-16
Bug Fixes and Improvements for Metrics

alexeykazakov:
There are some bug fixes & improvements in operator metrics introduced in operator-sdk 0.16.
See https://sdk.operatorframework.io/docs/migration/version-upgrade-guide/#v016x


xcoulon:
is there a way we can verify that it works? (in e2e tests for example)

alexeykazakov:
> is there a way we can verify that it works? (in e2e tests for example)

Not sure. We don't have anything for testing metrics atm. This code is just a result of operator-sdk 0.17.1 generator.

xcoulon:
> > is there a way we can verify that it works? (in e2e tests for example)
> 
> Not sure. We don't have anything for testing metrics atm. This code is just a result of operator-sdk 0.17.1 generator.

I see. So should we add some tests now, then?

alexeykazakov:
> I see. So should we add some tests now, then?

We could. But let's to it later in a separate PR. I think it's good enough to make sure this PR doesn't break anything in the current tests. Don't want to keep this PR open for too long hoping waiting for me to find time for new e2e tests for metrics.

PR #233, opened 2020-06-24
Mailgun notification delivery service

sbryzak:
Adds the implementation for the mailgun notification delivery service.

Fixes https://issues.redhat.com/browse/CRT-675


pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,76 @@ package notification
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	// Send the message with a 10 second timeout
+	_, _, err = s.Mailgun.Send(ctx, message)
+	if err != nil {
+		return err
+	}
+
 	return nil
```
alexeykazakov, line 76: Or just:
```suggestion
	return err
```

sbryzak:
@alexeykazakov we had agreed that "live" mailgun testing with a real api key would be performed manually.  There will be a mock delivery service which can use mailgun's mock objects, I've discussed this with @tinakurian briefly but I don't have a good understanding of how it works yet.  From what Tina explained, I think it creates a local server against which you can send mailgun requests.  Tina have you looked into this deeper yet? I'm wondering if there's a way we can actually test the mailgun delivery service implementation this way.

alexeykazakov:
> @alexeykazakov we had agreed that "live" mailgun testing with a real api key would be performed manually.  There will be a mock delivery service which can use mailgun's mock objects, I've discussed this with @tinakurian briefly but I don't have a good understanding of how it works yet.  From what Tina explained, I think it creates a local server against which you can send mailgun requests.  Tina have you looked into this deeper yet? I'm wondering if there's a way we can actually test the mailgun delivery service implementation this way.

I wonder if we really need a mock service if we can mock the main service by providing some mock configuration or objects. But it depends on what mailgun supports for mocking.

tinakurian:
> > @alexeykazakov we had agreed that "live" mailgun testing with a real api key would be performed manually.  There will be a mock delivery service which can use mailgun's mock objects, I've discussed this with @tinakurian briefly but I don't have a good understanding of how it works yet.  From what Tina explained, I think it creates a local server against which you can send mailgun requests.  Tina have you looked into this deeper yet? I'm wondering if there's a way we can actually test the mailgun delivery service implementation this way.
> 
> I wonder if we really need a mock service if we can mock the main service by providing some mock configuration or objects. But it depends on what mailgun supports for mocking.

so from what i saw we could do something like this: 
```
	mNotificationDeliveryService := &MockNotificationDeliveryService{
		Mailgun: mailgun.NewMailgun("domain-test", "domainapikey-test"),
		server:  mailgun.NewMockServer(),
	}
```
which would allow us to have something like: 
```
func (m MockNotificationDeliveryService) Deliver() error {
	mailgunMessage := m.Mailgun.NewMessage("from@gmail.com", "test subject", "test message", "to@gmail.com")
	resp, id, err := m.Mailgun.Send(context.TODO(), mailgunMessage)
	return err
}
```

this seems to work... i think it just spoofs the server etc... i tested this and it returns the following: 
resp: Queded. Thank you
id: <ID-YTe9UVs68AcVscKa>
err: nil

pkg/controller/notification/notification_delivery_service_test.go
```diff
@@ -20,25 +20,41 @@ func NewMockNotificationDeliveryServiceConfig(service string) NotificationDelive
...
-		require.IsType(t, &MockNotificationDeliveryService{}, svc)
-	})
-
 	t.Run("factory configured with mailgun delivery service", func(t *testing.T) {
 		// when
 		factory := NewNotificationDeliveryServiceFactory(client,
-			NewMockNotificationDeliveryServiceConfig(configuration.NotificationDeliveryServiceMailgun))
+			NewMockNotificationDeliveryServiceConfig(configuration.NotificationDeliveryServiceMailgun),
+			NewMockMailgunConfiguration("foo.com", "12345", "sender@foo.com"))
 		svc, err := factory.CreateNotificationDeliveryService()
```
tinakurian, line 58: after creating the service, maybe we should have a test that actually calls `Send`?

go.sum
```diff
@@ -536,6 +541,9 @@ github.com/lovoo/gcloud-opentracing v0.3.0/go.mod h1:ZFqk2y38kMDDikZPAK7ynTTGuyt
 github.com/magiconair/properties v1.8.0/go.mod h1:PppfXfuXeibc/6YijjN8zIbojt8czPbwD3XqdrwzmxQ=
 github.com/magiconair/properties v1.8.1 h1:ZC2Vc7/ZFkGmsVC9KvOjumD+G5lXy2RtTKyzRKO2BQ4=
 github.com/magiconair/properties v1.8.1/go.mod h1:PppfXfuXeibc/6YijjN8zIbojt8czPbwD3XqdrwzmxQ=
+github.com/mailgun/mailgun-go v2.0.0+incompatible h1:0FoRHWwMUctnd8KIR3vtZbqdfjpIMxOZgcSa51s8F8o=
```
tinakurian, line 544: looks like we need to run `go mod tidy`

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,82 @@ package notification
...
-func (s *MailgunNotificationDeliveryService) Send(ctx context.Context, notificationCtx NotificationContext, templateName string) error {
-	// TODO implement
-	return nil
+type MailgunOption interface {
+	// ApplyToMailgun applies this configuration to the given mailgun instance.
+	ApplyToMailgun(mailgun.Mailgun)
+}
+
+// NewMailgunNotificationDeliveryService creates a delivery service that uses the Mailgun API to deliver email notifications
+func NewMailgunNotificationDeliveryService(client client.Client, config MailgunConfiguration,
```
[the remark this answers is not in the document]
sbryzak: Good point, removed.

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,82 @@ package notification
...
+	template, found, err := notificationtemplates.GetNotificationTemplate(templateName)
+	if err != nil {
+		return err
+	}
+
+	if !found {
+		return errors.New(fmt.Sprintf("notification template [%s] not found", templateName))
+	}
+
+	sender := s.SenderEmail
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,82 @@ package notification
...
+	if err != nil {
+		return err
+	}
+
+	body, err := s.base.GenerateContent(ctx, notificationCtx, template.Content)
+	if err != nil {
+		return err
+	}
+
+	recipient := notificationCtx.FullEmailAddress()
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/controller/notification/notification_context.go
```diff
@@ -2,6 +2,7 @@ package notification
 
 import (
 	"context"
+	"fmt"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
sbryzak: Fixed, also comments added to the notification context types.

pkg/controller/notification/notification_delivery_service_test.go
sbryzak: Test now added. :)

go.sum
sbryzak: Done

pkg/controller/notification/notification_delivery_service.go
```diff
@@ -15,24 +15,28 @@ type NotificationDeliveryServiceConfig interface {
...
+	}
 }
 
 func (f *NotificationDeliveryServiceFactory) CreateNotificationDeliveryService() (NotificationDeliveryService, error) {
 	switch f.Config.GetNotificationDeliveryService() {
-	case configuration.NotificationDeliveryServiceMock:
-		return NewMockNotificationDeliveryService(), nil
 	case configuration.NotificationDeliveryServiceMailgun:
-		return NewMailgunNotificationDeliveryService(), nil
+		return NewMailgunNotificationDeliveryService(f.Client, f.MailgunConfig), nil
```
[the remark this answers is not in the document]
sbryzak: You're right, I've removed it.

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,82 @@ package notification
...
+
+	// The message object allows you to add attachments and Bcc recipients
+	message := s.Mailgun.NewMessage(sender, subject, body, recipient)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	// Send the message with a 10 second timeout
+	_, _, err = s.Mailgun.Send(ctx, message)
```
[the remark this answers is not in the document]
sbryzak: I'm not quite sure what to do with these.  We could log them at this point here, but I don't know what the value of doing that would be.  We can't really return them because they're only meaningful within the mailgun delivery service implementation.  I would expect that if the response was something other than a successful send, then the Send() function would return an error anyway. @alexeykazakov any thoughts?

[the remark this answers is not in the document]
alexeykazakov: We need to check what errors `Send()` can return. If is it enough to relay on the `err` here or we really need to analyze the response too?
But anyway, in case of the error I think it could be helpful if we log the response body and the id (is it the request ID?) here. No need to return them IMO.

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,76 @@ package notification
...
+func NewMailgunNotificationDeliveryService(config MailgunConfiguration, templateLoader TemplateLoader,
+	opts ...MailgunOption) NotificationDeliveryService {
+
+	mg := mailgun.NewMailgun(config.GetMailgunDomain(), config.GetMailgunAPIKey())
+
+	for _, opt := range opts {
+		opt.ApplyToMailgun(mg)
+	}
+
+	svc := &MailgunNotificationDeliveryService{
```
tinakurian, line 38: we can probably omit the svc var and just return &MailgunNotificationDeliveryService{ ... } here, what do you think?

pkg/controller/notification/mailgun_notification_delivery_service.go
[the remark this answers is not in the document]
tinakurian: Sounds good. if we're going to log this error here should we also log the others above? 
```
template, found, err := s.base.TemplateLoader.GetNotificationTemplate(templateName)
subject, err := s.base.GenerateContent(notificationCtx, template.Subject)
body, err := s.base.GenerateContent(notificationCtx, template.Content)
```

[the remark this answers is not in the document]
alexeykazakov: what we can do actually is to wrap this error with out own custom own with additional information and return so it will be properly logged and added to the status

pkg/controller/notification/mailgun_notification_delivery_service.go
sbryzak: Good catch, fixed.

pkg/controller/notification/mailgun_notification_delivery_service.go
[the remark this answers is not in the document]
sbryzak: I've introduced a new custom error type and now return that if there's an error

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -2,17 +2,96 @@ package notification
...
+type MailgunOption interface {
+	// ApplyToMailgun applies this configuration to the given mailgun instance.
+	ApplyToMailgun(mailgun.Mailgun)
+}
+
+// NewMailgunNotificationDeliveryService creates a delivery service that uses the Mailgun API to deliver email notifications
+func NewMailgunNotificationDeliveryService(config MailgunConfiguration, templateLoader TemplateLoader,
+	opts ...MailgunOption) NotificationDeliveryService {
+
+	mg := mailgun.NewMailgun(config.GetMailgunDomain(), config.GetMailgunAPIKey())
```
tinakurian, line 50: not sure if we want to do a check here but if the config or templateLoader ends up being nil this will panic.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -0,0 +1,50 @@
...
+	templateLoader := NewMockTemplateLoader(&notificationtemplates.NotificationTemplate{
+		Subject: "foo",
+		Content: "bar",
+		Name:    "test",
+	})
+
+	t.Run("test mailgun notification delivery service send", func(t *testing.T) {
+		// when
+		mgds := NewMailgunNotificationDeliveryService(mgConfig, templateLoader, opt)
+		err := mgds.Send(context.Background(), notCtx, "test")
```
tinakurian, line 45: it might be worthwhile to also add in tests for the error conditions. for instance if a template is not found.

alexeykazakov: Yes, please add tests for errors too.

pkg/controller/notification/mailgun_notification_delivery_service.go
sbryzak: Not sure if it's worth doing a check here, we can guarantee these parameters are not nil elsewhere.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: Good idea, I had the same thought when looking at the coverage report.  By the way, is coverage broken right now?  I'm getting all sorts of stuff included in the coverage report that isn't part of this PR.

alexeykazakov: Yeah... codecov reports for PRs have not been very reliable for some time.

alexeykazakov: I wonder if there is anything in the mailgun server (mock server in this case) we can use to actually check that the api call was made?

pkg/controller/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	t.Run("test mailgun notification delivery service send fails", func(t *testing.T) {
+		// when
+		mgds := NewMailgunNotificationDeliveryService(mgConfig, templateLoader, invalidServerOption)
+		err := mgds.Send(context.Background(), notCtx, "test")
+
+		// then
+		require.Error(t, err)
+		require.IsType(t, MailgunDeliveryError{}, err)
```
alexeykazakov, line 71: Let's also check the error message.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	t.Run("test mailgun notification delivery service invalid template", func(t *testing.T) {
+		// when
+		mgds := NewMailgunNotificationDeliveryService(mgConfig, templateLoader, mockServerOption)
+		err := mgds.Send(context.Background(), notCtx, "bar")
+
+		// then
+		require.Error(t, err)
```
alexeykazakov, line 80: And here.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: Done.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: Done.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: I think there is (the mock API seems to store events) however it doesn't appear there is any programmatic access to these, instead they provide an events endpoint which can list the events, from this I guess they would need to be parsed and converted into objects before we could write any meaningful tests.  To me this cost/benefit seems like it wouldn't be worth it, just for the sake of confirming that the mailgun API actually does what it's supposed to.

tinakurian: It looks like we could actually filter on event types - https://documentation.mailgun.com/en/latest/api-events.html#filter-field. There is the event type `delivered` which we could filter for - https://documentation.mailgun.com/en/latest/api-events.html#event-types

PR #238, opened 2020-06-29
Implement notification delivery in notification controller

sbryzak:
This PR implements notification delivery within the notification controller.  The `newReconciler()` function now creates a new notification delivery service factory which uses the configuration to determine which type of delivery service to create.  Currently we support **mailgun** (for production use) and **mock** (for testing).  Depending on which is configured, the delivery service will be created as part of the NotificationReconciler.

Fixes https://issues.redhat.com/browse/CRT-678

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -58,6 +58,9 @@ const (
 
 	// varMailgunSenderEmail specifies the host operator mailgun senders email
 	varMailgunSenderEmail = "mailgun.sender.email"
+
+	// varRegistrationServiceURL is the URL used to access the registration service
+	varRegistrationServiceURL = "registration.service.url"
 )
 
 // Config encapsulates the Viper configuration registry which stores the
@@ -129,6 +132,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(VarDurationBeforeChangeRequestDeletion, "24h")
 	c.host.SetDefault(varNotificationDeliveryService, NotificationDeliveryServiceMailgun)
 	c.host.SetDefault(varDurationBeforeNotificationDeletion, "24h")
+	c.host.SetDefault(varRegistrationServiceURL, "https://registration.crt-placeholder.com")
 }
 
 // GetDurationBeforeChangeRequestDeletion returns the timeout before a complete TierChangeRequest will be deleted.
@@ -159,6 +163,11 @@ func (c *Config) GetMailgunAPIKey() string {
 // GetMailgunSenderEmail returns the host operator mailgun sender's email address
 func (c *Config) GetMailgunSenderEmail() string {
 	return c.host.GetString(varMailgunSenderEmail)
+}
+
+// GetRegistrationServiceURL returns the URL of the registration service
+func (c *Config) GetRegistrationServiceURL() string {
+	return c.host.GetString(varRegistrationServiceURL)
 }
 
 // GetAllRegistrationServiceParameters returns the map with key-values pairs of parameters that have REGISTRATION_SERVICE prefix
--- a/pkg/controller/notification/notification_context.go
+++ b/pkg/controller/notification/notification_context.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/host-operator/pkg/configuration"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -11,16 +12,17 @@ import (
 // NotificationContext is used to pass user-specific data to the notification template parser in order to generate
 // customised notification content.
 type NotificationContext struct {
-	UserID      string
-	FirstName   string
-	LastName    string
-	Email       string
-	CompanyName string
+	UserID          string
+	FirstName       string
+	LastName        string
+	UserEmail       string
+	CompanyName     string
+	RegistrationURL string
 }
 
 // NewNotificationContext creates a new NotificationContext by looking up the UserSignup with the specified userID
 // and using it to populate the context fields
-func NewNotificationContext(client client.Client, userID, namespace string) (*NotificationContext, error) {
+func NewNotificationContext(client client.Client, userID, namespace string, config *configuration.Config) (*NotificationContext, error) {
 	// Lookup the UserSignup resource with the specified userID
 	instance := &toolchainv1alpha1.UserSignup{}
 	err := client.Get(context.TODO(), types.NamespacedName{
@@ -40,7 +42,11 @@ func NewNotificationContext(client client.Client, userID, namespace string) (*No
 	}
 
 	if emailLbl, exists := instance.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
-		notificationCtx.Email = emailLbl
+		notificationCtx.UserEmail = emailLbl
+	}
+
+	if config != nil {
+		notificationCtx.RegistrationURL = config.GetRegistrationServiceURL()
 	}
 
 	return notificationCtx, nil
@@ -50,5 +56,5 @@ func (c *NotificationContext) FullEmailAddress() string {
 	return fmt.Sprintf("%s %s<%s>",
 		c.FirstName,
 		c.LastName,
-		c.Email)
+		c.UserEmail)
 }
--- a/pkg/controller/notification/notification_context.go
+++ b/pkg/controller/notification/notification_context.go
@@ -2,9 +2,12 @@ package notification
 
 import (
 	"context"
+	"errors"
 	"fmt"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/pkg/configuration"
+
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -45,9 +48,10 @@ func NewNotificationContext(client client.Client, userID, namespace string, conf
 		notificationCtx.UserEmail = emailLbl
 	}
 
-	if config != nil {
-		notificationCtx.RegistrationURL = config.GetRegistrationServiceURL()
+	if config == nil {
+		return nil, errors.New("configuration was not provided")
 	}
+	notificationCtx.RegistrationURL = config.GetRegistrationServiceURL()
 
 	return notificationCtx, nil
 }
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -56,7 +56,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager, config *configuration.Config) (reconcile.Reconciler, error) {
-	factory := NewNotificationDeliveryServiceFactory(mgr.GetClient(), config)
+	factory := NewNotificationDeliveryServiceFactory(mgr.GetClient(), config, config)
 
 	svc, err := factory.CreateNotificationDeliveryService()
 	if err != nil {
@@ -112,13 +112,14 @@ func (r *ReconcileNotification) Reconcile(request reconcile.Request) (reconcile.
 		}, nil
 	}
 
-	// Send the notification
-	notCtx, err := NewNotificationContext(context.Background(), r.client, notification.Spec.UserID, request.Namespace)
+	// Send the notification - first create the notification context
+	notCtx, err := NewNotificationContext(r.client, notification.Spec.UserID, request.Namespace)
 	if err != nil {
 		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
 			r.setStatusNotificationContextError, err, "failed to create notification context")
 	}
 
+	// Send the notification via the configured delivery service
 	err = r.deliveryService.Send(context.Background(), notCtx, notification.Spec.Template)
 	if err != nil {
 		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -119,6 +119,7 @@ func (r *ReconcileNotification) Reconcile(request reconcile.Request) (reconcile.
 			r.setStatusNotificationContextError, err, "failed to create notification context")
 	}
 
+	// if the environment is set to e2e do not attempt sending via mailgun
 	if r.config.GetEnvironment() != "e2e-tests" {
 		// Send the notification via the configured delivery service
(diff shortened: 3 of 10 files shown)


pkg/controller/notification/notification_controller.go
```diff
@@ -166,12 +186,46 @@ func (r *ReconcileNotification) setStatusNotificationDeletionFailed(notification
...
 		})
 }
+
+func (r *ReconcileNotification) setStatusNotificationDeliveryError(notification *toolchainv1alpha1.Notification, msg string) error {
+	return r.updateStatusConditions(
+		notification,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.NotificationSent,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.NotificationDeliveryErrorReason,
```
tinakurian, line 206: i guess we will need to raise a PR with these reasons added? looks like there is an old PR that can be re-purposed? https://github.com/codeready-toolchain/api/pull/129/files

pkg/controller/notification/notification_controller.go
```diff
@@ -28,7 +28,11 @@ var log = logf.Log.WithName("controller_notification")
 // Add creates a new Notification Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager, config *configuration.Config) error {
-	return add(mgr, newReconciler(mgr, config))
+	reconciler, err := newReconciler(mgr, config)
+	if err != nil {
```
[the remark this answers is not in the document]
sbryzak: It's done this way because the `newReconciler()` function in this controller can return an error due to misconfiguration.

pkg/controller/notification/notification_controller.go
```diff
@@ -51,8 +55,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, config *configuration.Config) reconcile.Reconciler {
-	return &ReconcileNotification{client: mgr.GetClient(), scheme: mgr.GetScheme(), config: config}
+func newReconciler(mgr manager.Manager, config *configuration.Config) (reconcile.Reconciler, error) {
+	factory := NewNotificationDeliveryServiceFactory(mgr.GetClient(), config)
+
+	svc, err := factory.CreateNotificationDeliveryService()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
sbryzak: It will be logged here: https://github.com/codeready-toolchain/host-operator/blob/master/cmd/manager/main.go#L148

pkg/controller/notification/notification_controller.go
```diff
@@ -100,17 +112,25 @@ func (r *ReconcileNotification) Reconcile(request reconcile.Request) (reconcile.
...
-		return reconcile.Result{}, err
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
+			r.setStatusNotificationContextError, err, "failed to create notification context")
 	}
 
+	// Send the notification via the configured delivery service
+	err = r.deliveryService.Send(context.Background(), notCtx, notification.Spec.Template)
+	if err != nil {
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
+			r.setStatusNotificationDeliveryError, err, "failed to deliver notification")
```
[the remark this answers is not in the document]
sbryzak: Good idea, changed.

pkg/controller/notification/notification_controller.go
```diff
@@ -166,12 +186,46 @@ func (r *ReconcileNotification) setStatusNotificationDeletionFailed(notification
...
+			Status:  corev1.ConditionTrue,
+			Reason:  toolchainv1alpha1.NotificationSentReason,
+			Message: msg,
+		})
+}
+
+func (r *ReconcileNotification) updateStatus(logger logr.Logger, notification *toolchainv1alpha1.Notification,
+	statusUpdater func(notification *toolchainv1alpha1.Notification, message string) error) error {
+
+	if err := statusUpdater(notification, ""); err != nil {
```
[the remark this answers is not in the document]
sbryzak: Hmm, good question... this function was copied from the UserSignupController.... @alexeykazakov do you know why we pass empty string for the message?

pkg/controller/notification/notification_controller.go
sbryzak: The API module already contains the new reasons as far as I'm aware.. i'm not sure what that old draft PR is, perhaps I should close it.

pkg/controller/notification/notification_controller.go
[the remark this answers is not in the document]
alexeykazakov: `message` is only used when we catch an error. It's for the error message. And since we handle errors by `wrapErrorWithStatusUpdate()` all other status updates don't need messages. So, I think we can remove `message` from statusUpdater function to avoid confusion.

pkg/controller/notification/notification_controller.go
tinakurian: ohhhh okay i see it :+1:

pkg/controller/notification/notification_controller.go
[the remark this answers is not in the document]
tinakurian: ohhhh okay, makes sense!

pkg/configuration/configuration.go
```diff
@@ -129,6 +132,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(VarDurationBeforeChangeRequestDeletion, "24h")
 	c.host.SetDefault(varNotificationDeliveryService, NotificationDeliveryServiceMailgun)
 	c.host.SetDefault(varDurationBeforeNotificationDeletion, "24h")
+	c.host.SetDefault(varRegistrationServiceURL, "https://registration.crt-placeholder.com")
```
alexeykazakov, line 135: Reg service URL will depend on the particular cluster/env. Not sure how it's best to setup it...
@tinakurian I guess we can add a config map name to the environments as we did for secrets: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/env/prod.yaml#L17
Something like this:
```
host-operator:
  config-map:
    name: host-operator-config
```
And load all keys/values from it when we loading the configuration. Similar to the secret loading.
Then we can create a config map with reg service URL in particular clusters.

tinakurian: sounds good, i can add this!

pkg/controller/notification/notification_controller.go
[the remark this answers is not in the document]
sbryzak: I'm not so sure it's easy to do this.  Each of the status updater functions accept a message parameter and uses it to set the message in the condition, like so:

````
func (r *ReconcileNotification) setStatusNotificationDeliveryError(notification *toolchainv1alpha1.Notification, msg string) error {
	return r.updateStatusConditions(
		notification,
		toolchainv1alpha1.Condition{
			Type:    toolchainv1alpha1.NotificationSent,
			Status:  corev1.ConditionFalse,
			Reason:  toolchainv1alpha1.NotificationDeliveryErrorReason,
			Message: msg,
		})
}
````

If I remove the `message` parameter from the `statusUpdater` function then I'll also need to remove this parameter from all of the status update functions.  Is that the intent here?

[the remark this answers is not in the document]
Comment (alexeykazakov) [reply pkg/controller/notification/notification_controller.go:228]: @sbryzak if `message` is actually used by some function then we have to keep it. If not then we can delete it.

File: pkg/configuration/configuration.go (hunk 6 above)
Comment (tinakurian) [reply pkg/configuration/configuration.go]: done here: https://github.com/codeready-toolchain/toolchain-infra/pull/21 and https://github.com/codeready-toolchain/host-operator/pull/244

File: pkg/controller/notification/notification_controller.go [hunk 7]
```diff
@@ -100,24 +112,37 @@ func (r *ReconcileNotification) Reconcile(request reconcile.Request) (reconcile.
...
+	notCtx, err := NewNotificationContext(r.client, notification.Spec.UserID, request.Namespace, r.config)
 	if err != nil {
-		reqLogger.Error(err, "unable to set sent status to Notification")
-		return reconcile.Result{}, err
+		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, notification,
+			r.setStatusNotificationContextError, err, "failed to create notification context")
+	}
+
+	// if the environment is set to e2e do not attempt sending via mailgun
+	if r.config.GetEnvironment() != "e2e-tests" {
```
alexeykazakov, line 123: Let's log an info message in case of `e2e-tests` saying that the notifications was skipped.

tinakurian: @alexeykazakov good idea, i've added it here: 09233c4dd5f3c2ed1b15defbc78c7f11eca29639

pkg/controller/notification/notification_context.go
```diff
@@ -40,7 +42,11 @@ func NewNotificationContext(client client.Client, userID, namespace string) (*No
 	}
 
 	if emailLbl, exists := instance.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
-		notificationCtx.Email = emailLbl
+		notificationCtx.UserEmail = emailLbl
+	}
+
+	if config != nil {
+		notificationCtx.RegistrationURL = config.GetRegistrationServiceURL()
```
xcoulon, line 49: shouldn't you return an error if `config==nil`?

pkg/controller/notification/notification_controller.go
```diff
@@ -51,8 +55,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, config *configuration.Config) reconcile.Reconciler {
-	return &ReconcileNotification{client: mgr.GetClient(), scheme: mgr.GetScheme(), config: config}
+func newReconciler(mgr manager.Manager, config *configuration.Config) (reconcile.Reconciler, error) {
+	factory := NewNotificationDeliveryServiceFactory(mgr.GetClient(), config, config)
```
xcoulon, line 59: it seems a bit weird that we have to pass the `config` twice in this call. Would it be possible to merge the `NotificationDeliveryServiceConfig` and `MailgunConfiguration` interfaces for example (or have a new interface composed of those 2 one)?

pkg/controller/notification/notification_controller.go
```diff
@@ -135,7 +162,8 @@ func (r *ReconcileNotification) checkTransitionTimeAndDelete(log logr.Logger, no
 	return false, diff, nil
 }
 
-func (r *ReconcileNotification) wrapErrorWithStatusUpdate(logger logr.Logger, notification *toolchainv1alpha1.Notification, statusUpdater func(notification *toolchainv1alpha1.Notification, message string) error, err error, format string, args ...interface{}) error {
+func (r *ReconcileNotification) wrapErrorWithStatusUpdate(logger logr.Logger, notification *toolchainv1alpha1.Notification,
+	statusUpdater func(notification *toolchainv1alpha1.Notification, message string) error, err error, format string, args ...interface{}) error {
```
xcoulon, line 166: you could create a `statusUpdater` type like this:
```
type statusUpdater func(notification *toolchainv1alpha1.Notification, message string) error
```
and your func signature would become:
```
func (r *ReconcileNotification) wrapErrorWithStatusUpdate(logger logr.Logger, notification *toolchainv1alpha1.Notification, statusUpdater statusUpdater, err error, format string, args ...interface{}) error
```
which is a little simpler to read

pkg/controller/notification/notification_controller.go
```diff
@@ -166,12 +194,46 @@ func (r *ReconcileNotification) setStatusNotificationDeletionFailed(notification
...
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.NotificationSent,
+			Status:  corev1.ConditionTrue,
+			Reason:  toolchainv1alpha1.NotificationSentReason,
+			Message: msg,
+		})
+}
+
+func (r *ReconcileNotification) updateStatus(logger logr.Logger, notification *toolchainv1alpha1.Notification,
+	statusUpdater func(notification *toolchainv1alpha1.Notification, msg string) error) error {
```
xcoulon, line 231: same here using the `statusUpdater` type suggested above

pkg/controller/notification/notification_controller_test.go
```diff
@@ -72,22 +88,224 @@ func TestNotificationSentFailure(t *testing.T) {
...
+		// Load the reconciled notification
+		key := types.NamespacedName{
+			Namespace: operatorNamespace,
+			Name:      notification.Name,
+		}
+		instance := &v1alpha1.Notification{}
+		err = client.Get(context.TODO(), key, instance)
+		require.NoError(t, err)
+
+		test.AssertConditionsMatch(t, instance.Status.Conditions,
```
xcoulon, line 202: you could probably refactor this assertion into something like https://github.com/codeready-toolchain/host-operator/blob/master/test/templateupdaterequest/assertion.go#L17-L44 to keep the test shorter. 
(It's just a suggestion, for the sake of readability ;))

pkg/controller/notification/notification_controller_test.go
```diff
@@ -72,22 +88,224 @@ func TestNotificationSentFailure(t *testing.T) {
...
+		require.Error(t, err)
+		require.False(t, result.Requeue)
+		require.Equal(t, "failed to send notification: delivery error", err.Error())
+
+		// Load the reconciled notification
+		key := types.NamespacedName{
+			Namespace: operatorNamespace,
+			Name:      notification.Name,
+		}
+		instance := &v1alpha1.Notification{}
```
xcoulon, line 272: same suggestion here with using a `AssertThatNotification`

pkg/controller/notification/notification_controller.go
tinakurian: good idea! changed as part of: 057eb87e2123a2edfa90a49c007140b845d59010

pkg/controller/notification/notification_controller.go
tinakurian: good idea! changed as part of: 057eb87e2123a2edfa90a49c007140b845d59010

pkg/controller/notification/notification_controller.go
tinakurian: changed here: 057eb87e2123a2edfa90a49c007140b845d59010

pkg/controller/notification/notification_controller_test.go
tinakurian: good idea, changed here: 057eb87e2123a2edfa90a49c007140b845d59010

pkg/controller/notification/notification_controller_test.go
tinakurian: changed here: 057eb87e2123a2edfa90a49c007140b845d59010

pkg/controller/notification/notification_context.go
tinakurian: good idea! changed as part of: 057eb87

pkg/controller/notification/notification_controller.go
tinakurian: done here: https://github.com/codeready-toolchain/host-operator/pull/251

PR #239, opened 2020-07-01
fix: handle errors correctly

MatousJobanek:
There were two cases when the errors weren't properly handled:
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/configuration/configuration.go#L88
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/sync.go#L121

As part of this PR I added tests that cover the error cases


pkg/controller/masteruserrecord/sync_test.go
```diff
@@ -159,24 +149,45 @@ func TestSynchronizeSpec(t *testing.T) {
 
 func TestSynchronizeStatus(t *testing.T) {
 	// given
-	logf.SetLogger(logf.ZapLogger(true))
```
xcoulon, line 162: oh, why did you remove these lines? I find it useful to have the `INFO` log messages.
(It's not a requirement to revert the change, though)

pkg/controller/masteruserrecord/sync_test.go
```diff
@@ -243,8 +300,48 @@ func TestSyncMurStatusWithUserAccountStatusWhenCompleted(t *testing.T) {
...
+		// given
+		hostClient := test.NewFakeClient(t, mur)
+		sync, memberClient := prepareSynchronizer(t, userAccount, mur, hostClient)
+
+		// when
+		err := sync.synchronizeStatus()
+
+		// then
+		require.NoError(t, err)
+		verifySyncMurStatusWithUserAccountStatus(t, memberClient, hostClient, userAccount, mur, toBeProvisionedNotificationCreated(), toBeProvisionedNotificationCreated())
```
xcoulon, line 313: just curious, why to you need `toBeProvisionedNotificationCreated()` twice?

pkg/controller/masteruserrecord/sync_test.go
MatousJobanek: because I moved it to `prepareSynchronizer` function - it was there before, so it was just a duplication:
https://github.com/codeready-toolchain/host-operator/blob/5361b85e3114bbb7972a88536f0486c159b7c9b4/pkg/controller/masteruserrecord/sync_test.go#L641-L649

pkg/controller/masteruserrecord/sync_test.go
MatousJobanek: wrong copy-paste :man_facepalming: 
good catch, thanks - fixed in 2368e4f

pkg/controller/masteruserrecord/sync_test.go
```diff
@@ -159,24 +149,45 @@ func TestSynchronizeSpec(t *testing.T) {
...
+		hostClient.MockStatusUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
+			return fmt.Errorf("some error")
+		}
+		sync, _ := prepareSynchronizer(t, userAccount, mur, hostClient)
+
+		// when
+		err := sync.synchronizeStatus()
+
+		// then
+		require.Error(t, err)
```
tinakurian, line 184: maybe we should also check the error itself?

pkg/configuration/configuration_test.go
```diff
@@ -82,15 +87,32 @@ func TestLoadFromSecret(t *testing.T) {
...
+		restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_SECRET_NAME", "test-secret")
+		defer restore()
+
+		cl := test.NewFakeClient(t)
+
+		// when
+		config, err := configuration.LoadConfig(cl)
+
+		// then
+		require.Error(t, err)
```
tinakurian, line 113: maybe we should check the error message here as well?

pkg/controller/masteruserrecord/sync_test.go
MatousJobanek: added 1a7b7de

pkg/configuration/configuration_test.go
MatousJobanek: added 1a7b7de

PR #240, opened 2020-07-08
feat: track updates of NSTemplateTiers

xcoulon:
Update the reconciler to let it take care of:
- inserting a new entry in the `status.updates` (and cleaning the 'failedAccounts` in the previous one)
- creating and delete the TemplateUpdateRequest to update the MAsterUserRecord associated with this tier
- updating the `Failed` counter in the `status.updates` when a MasterUserRecord failed to update
  and retrain the username in case of failure, for later diagnostics
- setting the `completionTime` when all MasterUserRecord have been processed

Also: move `TestComputeHash` in a separate file

See also 
- https://github.com/codeready-toolchain/api/pull/166
- https://github.com/codeready-toolchain/toolchain-e2e/pull/130

Fixes https://issues.redhat.com/browse/CRT-711


xcoulon:
@alexeykazakov @MatousJobanek PR is ready for review. As discussed during our meeting, I refactored the initial PR by using a single controller, and the status.updates do NOT include the totalProcessed counter because due to race conditions (I assume), the e2e tests fail to assert the exact value of this counter (I would sometimes have 12 instead of 11 for example). However, if we decide that the exact number of processed resources is less important than being able to track the progress, then I can easily restore this counter (or in a separate PR 😅 )

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
+	// create a new entry in the `status.history`
+	if added, err := r.ensureStatusUpdateRecord(logger, tier); err != nil {
+		logger.Error(err, "unable to insert a new entry in status.updates after NSTemplateTier changed.")
+		return reconcile.Result{}, err
+	} else if added {
+		logger.Info("Requeing after adding a new entry in tier.status.updates")
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if done, err := r.ensureTemplateUpdateRequest(logger, tier); err != nil {
 		logger.Error(err, "unable to ensure TemplateRequestUpdate resource after NSTemplateTier changed.")
```
alexeykazakov, line 135: Let's wrap this error with NotReady status. Similar to https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/masteruserrecord_controller.go#L214:37
We usually use `wrapErrorWithStatusUpdate()` function to:
- log the error
- Add Ready=False condition with the original error message to the resource status.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
 		return reconcile.Result{}, err
+	} else if done {
+		log.Info("No MasterUserRecord to update. Setting the completion timestamp immediately")
+		if err := r.markUpdateRecordAsCompleted(tier); err != nil {
+			logger.Error(err, "unable to mark latest status.update as complete.")
+			return reconcile.Result{}, err
+		}
+		logger.Info("No MasterUserRecord to update")
```
alexeykazakov, line 143: Already logged in line 138

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
+		return reconcile.Result{}, nil
```
alexeykazakov, line 144: Redundant return.
See line 146

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
+			Hash:      hash,
+		})
+		return true, r.client.Status().Update(context.TODO(), tier)
+	}
+	// check that last entry
+	if tier.Status.Updates[len(tier.Status.Updates)-1].Hash == hash {
+		logger.Info("current update still in progress")
+		return false, nil
+	}
+	tier.Status.Updates[len(tier.Status.Updates)-1].FailedAccounts = nil // reset
```
alexeykazakov, line 170: Why are we resetting the previous failed account count?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ah yes, indeed. Thanks! Removed in 0621d5b

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: Yes, I've seen this `wrapErrorWithStatusUpdate()` func being used in other controller, I don't really understand why the `NSTemplateTier` resource should be in a `NotReady` condition in this case. 
In other words, what does it mean for the The `NSTemplateTier` resource to be `Ready` or `NotReady`?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: removed in 0dc9fb0

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: that's something @MatousJobanek suggested when we discussed about storing the usernames for whom the update failed: these usernames are only worth keeping for the current update, so when we start a new one, we should reset the previous `FailedAccounts`, in order to avoid polluting the `Status.Updates` with too much data. 

I've clarified the comment in f9060c8

pkg/controller/nstemplatetier/nstemplatetier_controller.go
alexeykazakov: The point here is to have something in the status where we can see the error. Otherwise we will have to look up the errors in logs. It's not that convenient. We could use already available `Ready` condition here which could mean that the template tier is not ready to update templates. But we can introduce a new condition for that. `UpdateFailure` for example.
So, we use it for reporting errors and clean it up in the end of reconcile if everything is fine.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
alexeykazakov: > I've clarified the comment in f9060c8

Did you forget to commit this? I don't see f9060c8 commit here.

xcoulon: indeed, I forgot about to push it (I thought I did, though 🤨)

alexeykazakov: OK. It does make sense. And thanks for adding the comment. It helps.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ok, I get your point, but for that specific topic, may I suggest that we discuss about it when reviewing https://github.com/codeready-toolchain/host-operator/pull/243 for https://issues.redhat.com/browse/CRT-708?

alexeykazakov: Yes, I'm fine with addressing it in https://github.com/codeready-toolchain/host-operator/pull/243

xcoulon: thanks ;)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -190,23 +267,69 @@ func (r *ReconcileNSTemplateTier) activeTemplateUpdateRequests(logger logr.Logge
...
+	tier.Status.Updates[len(tier.Status.Updates)-1] = latest
+	if err := r.client.Status().Update(context.TODO(), tier); err != nil {
+		return err
+	}
+	logger.Info("incrementing counter after TemplateUpdateRequest completed", "name", tur.Name)
+	return nil
+}
+
+// markUpdateRecordAsCompleted looks-up the latest entry in the `status.updates` and sets the `CompletionTime` to `metav1.Now()`,
+// which means that the whole update process is done
```
alexeykazakov, line 325: Or maybe there was nothing to update, in this case the ComplitionTime means the time when we last checked if there is something to update. Not sure if it's worth reflecting somewhere... Here or/and in API.
Just thinking loud :)

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
+	} else if added {
+		logger.Info("Requeing after adding a new entry in tier.status.updates")
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if done, err := r.ensureTemplateUpdateRequest(logger, tier); err != nil {
 		logger.Error(err, "unable to ensure TemplateRequestUpdate resource after NSTemplateTier changed.")
 		return reconcile.Result{}, err
+	} else if done {
+		log.Info("No MasterUserRecord to update. Setting the completion timestamp immediately")
+		if err := r.markUpdateRecordAsCompleted(tier); err != nil {
```
alexeykazakov, line 139: Isn't already done in line 213 in `ensureTemplateUpdateRequest()`? Line 213 is the only place when that function returns `true` and it does `markUpdateRecordAsCompleted()` too.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -27,100 +29,164 @@ import (
...
+	if len(tier.Status.Updates) == 0 {
+		tier.Status.Updates = append(tier.Status.Updates, toolchainv1alpha1.NSTemplateTierHistory{
+			StartTime: metav1.NewTime(time.Now()),
+			Hash:      hash,
+		})
+		return true, r.client.Status().Update(context.TODO(), tier)
+	}
+	// check that last entry
+	if tier.Status.Updates[len(tier.Status.Updates)-1].Hash == hash {
+		logger.Info("current update still in progress")
```
alexeykazakov, line 165: And here too..
```suggestion
		logger.Info("current update/check if updated need is still in progress")
```

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -130,17 +196,22 @@ func (r *ReconcileNSTemplateTier) ensureTemplateUpdateRequest(logger logr.Logger
...
 			matchingLabels,
 		)
 		if err != nil {
-			return errs.Wrap(err, "unable to get MasterUserRecords to update")
+			return false, errs.Wrap(err, "unable to get MasterUserRecords to update")
 		}
 		logger.Info("listed MasterUserRecords", "count", len(murs.Items), "selector", matchingLabels)
+		if activeTemplateUpdateRequests == 0 && len(murs.Items) == 0 {
+			// we've reached the end: all MasterUserRecords are up-to-date
+			logger.Info("All MasterUserRecords have been updated. Setting the completion timestamp")
```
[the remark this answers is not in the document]
xcoulon: yes, the wording is better like you suggested, indeed.

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ah yes, indeed! So I'm keeping the call to the `makUpdateRecordAsComplete` in the `Reconcile` func so it's "on par" with the call to `ensureStatusUpdateRecord` on line 127. 
See 688820c

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: ok, let me rephrase a bit the comment ;)
See bbfac63

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: see 80d6ee7: using a slightly different sentence, which is easier to read, at least to me ;)

xcoulon:
> Great job!
> 
> Can we add some tests for not already covered errors? For example when we failed to list TURs?

sure! Let me check the test coverage to find the uncovered parts and see what I can do ;)

xcoulon:
> > Great job!
> > Can we add some tests for not already covered errors? For example when we failed to list TURs?
> 
> sure! Let me check the test coverage to find the uncovered parts and see what I can do ;)

with b32fa84 the NSTemplateTierController test coverage reaches 85% ;)

xcoulon:
> Nice tests! Thanks!

thanks! I've also started working on additional tests for the `TemplateUpdateRequest` controller ;)

xcoulon:
> > Nice tests! Thanks!
> 
> thanks! I've also started working on additional tests for the `TemplateUpdateRequest` controller ;)

see https://github.com/codeready-toolchain/host-operator/pull/240/commits/678cd7c44eb49c943349dd1d8d315400ea43a1fd

PR #241, opened 2020-07-10
Add overall toolchain status

rajivnathan:
## Description

This PR adds the `toolchainstatus` resource that provides an overall status of the hosted toolchain. It includes information about the registration-service, host operator as well as member clusters and the status of toolchain components on them. For more information and design details see the related issue linked below.

Related Issue: https://issues.redhat.com/browse/CRT-673

Related PRs:
https://github.com/codeready-toolchain/api/pull/168
https://github.com/codeready-toolchain/toolchain-common/pull/115
https://github.com/codeready-toolchain/member-operator/pull/178

To try it out run:
`oc get toolchainstatus toolchain-status -o yaml`

Example output:
```
kind: ToolchainStatus
metadata:
  annotations:
    toolchain.dev.openshift.com/last-applied-configuration: '{"metadata":{"name":"toolchain-status","namespace":"rsenthil-host-operator","creationTimestamp":null},"spec":{},"status":{"hostOperator":{"version":"","revision":"","buildTimestamp":"","deploymentName":""},"registrationService":{"deployment":{"name":""},"registrationServiceResources":{},"health":{"alive":"","buildTime":"","environment":"","revision":"","startTime":""}}}}'
  creationTimestamp: "2020-07-10T00:16:33Z"
  generation: 1
  name: toolchain-status
  namespace: rsenthil-host-operator
  resourceVersion: "2509898"
  selfLink: /apis/toolchain.dev.openshift.com/v1alpha1/namespaces/rsenthil-host-operator/toolchainstatuses/toolchain-status
  uid: 70ada6e3-9549-47c4-b378-8fbc9bb1d108
spec: {}
status:
  conditions:
  - lastTransitionTime: "2020-07-10T00:19:07Z"
    lastUpdatedTime: "2020-07-10T00:19:07Z"
    reason: AllComponentsReady
    status: "True"
    type: Ready
  hostOperator:
    buildTimestamp: "2020-07-10T02:39:45Z"
    conditions:
    - lastTransitionTime: "2020-07-10T00:19:07Z"
      lastUpdatedTime: "2020-07-10T00:19:07Z"
      reason: DeploymentReady
      status: "True"
      type: Ready
    deploymentName: host-operator
    revision: 7942ab8b31f15008eab3d134dfa40cf045130c78
    version: 0.0.1
  members:
  - clusterName: member-crc-m27h4
    memberStatus:
      conditions:
      - lastTransitionTime: "2020-07-10T00:19:06Z"
        lastUpdatedTime: "2020-07-10T00:19:06Z"
        reason: AllComponentsReady
        status: "True"
        type: Ready
  registrationService:
    deployment:
      conditions:
      - lastTransitionTime: "2020-07-10T00:19:07Z"
        lastUpdatedTime: "2020-07-10T00:19:07Z"
        reason: DeploymentReady
        status: "True"
        type: Ready
      name: registration-service
    health:
      alive: "true"
      buildTime: "0"
      conditions:
      - lastTransitionTime: "2020-07-10T00:19:07Z"
        lastUpdatedTime: "2020-07-10T00:19:07Z"
        reason: RegServiceReady
        status: "True"
        type: Ready
      environment: dev
      revision: 64af1be5c6011fae5497a7c35e2a986d633b3421
      startTime: "2020-07-10T00:16:50Z"
    registrationServiceResources:
      conditions:
      - lastTransitionTime: "2020-07-10T00:16:36Z"
        reason: Deployed
        status: "True"
        type: Ready
```


pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+	// bad response
+	if resp.StatusCode != http.StatusOK || resp.Body == nil {
+		err = errs.Wrap(fmt.Errorf("bad response from registration service: %+v", resp), errMsgRegistrationServiceNotReady)
+		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusReasonRegServiceNotReady, err.Error())
+		toolchainStatus.Status.RegistrationService.Health.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+		return err
+	}
+
+	// decode the response to JSON
+	defer resp.Body.Close()
```
alexeykazakov, line 407: It will still leak FDs. You have to read the body first then close it: `ioutil.ReadAll(response.Body)`
See https://github.com/fabric8-services/fabric8-auth/blob/975831c9c2bb8a15ea8c8a0620008a4bb7d461a1/rest/rest.go#L140 as an example.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+		members = append(members, noMemberClusterStatus)
+		return err
+	}
+	for _, cluster := range memberClusters {
+		memberClient := cluster.Client
+		if memberClient == nil {
+			// there is no client for the member cluster, create a status condition and add it to this member's status
+			err := fmt.Errorf("member cluster does not have a client assigned")
+			componentError = err
```
xcoulon, line 260:
```suggestion
			componentError = fmt.Errorf("member cluster does not have a client assigned")
```

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+		toolchainStatus,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.ToolchainStatusComponentsNotReadyReason,
+			Message: message,
+		})
+}
+
+func customMemberStatus(conditions ...toolchainv1alpha1.Condition) toolchainv1alpha1.MemberStatusStatus {
```
xcoulon, line 330: since you always pass a single `condition` in this func, you could rename it to:
```
func memberStatus(condition toolchainv1alpha1.Condition) 
```
and do something like:
```
members = append(members, memberResult(cluster, memberStatus(*memberStatusNotFoundCondition)))
``` 
above
(but that's just a suggestion to make code more compact)

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+			continue
+		}
+
+		// take only the overall conditions from the memberstatus (ie. don't include status of member operator, host connection, etc.) in order to keep readability.
+		// member status details can be viewed on the member cluster directly.
+		memberStatus := toolchainv1alpha1.MemberStatusStatus{Conditions: memberStatusObj.Status.Conditions}
+		for _, condition := range memberStatusObj.Status.Conditions {
+			if condition.Type == toolchainv1alpha1.ConditionReady && condition.Status != "True" {
+				// the existing memberstatus condition is added to the toolchainstatus but the condition is not ready so set the component error to bubble up the error to the overall toolchain status
+				componentError = errs.Wrap(fmt.Errorf(condition.Message), fmt.Sprintf("member cluster %s not ready", cluster.Name))
```
xcoulon, line 284: here `componentError` will only hold an error message for the last cluster which is not ready. Should we have a custom `ComponentError` type which may hold multiple cluster names, then return an error message with all the names at once for example?

pkg/controller/toolchainstatus/creation_test.go
```diff
@@ -0,0 +1,54 @@
...
+		cl.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+			return fmt.Errorf("creation failed")
+		}
+
+		// when
+		err = CreateOrUpdateResources(cl, s, HostOperatorNs, name)
+
+		// then
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "unable to create resource")
```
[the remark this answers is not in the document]
rajivnathan: In the unit tests the message is `unable to create resource of kind: , version: : creation failed` because we don't set the `TypeMeta` here: https://github.com/rajivnathan/host-operator/blob/f5bfffbea9adeeb7c5c6956b674f2be5b70314dc/pkg/controller/toolchainstatus/creation.go#L14

But we don't set the TypeMeta for other similar places. eg: https://github.com/rajivnathan/host-operator/blob/f5bfffbea9adeeb7c5c6956b674f2be5b70314dc/pkg/controller/registrationservice/creation.go#L24

If I add the following then it becomes `unable to create resource of kind: ToolchainStatus, version: v1alpha1: creation failed` but then we need to remember to explicitly update the APIVersion string when we update the API version in the future.
```
TypeMeta: v1.TypeMeta{
	Kind:             "ToolchainStatus",
	APIVersion: "v1alpha1",
},
```

So maybe better to leave it as is?

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+			}
+		}
```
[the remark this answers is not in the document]
rajivnathan: Done in 4fe2dd1087523aa210c2dea19fda210f0345bfb1

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusReasonRegServiceResourceNotFound, err.Error())
+		toolchainStatus.Status.RegistrationService.RegistrationServiceResources.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+		return err
+	}
+
+	// use the registrationservice resource directly in the toolchainstatus
+	toolchainStatus.Status.RegistrationService.RegistrationServiceResources.Conditions = registrationService.Status.Conditions
+	for _, condition := range registrationService.Status.Conditions {
+		if condition.Type == toolchainv1alpha1.ConditionReady && condition.Status != "True" {
+			return errs.Wrap(fmt.Errorf(condition.Message), "registration service resource not ready")
```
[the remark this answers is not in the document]
rajivnathan: Done in 4fe2dd1087523aa210c2dea19fda210f0345bfb1

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+		MemberStatus: memberStatus,
+	}
+}
+
+type regServiceSubstatusHandler struct {
+	httpClientImpl   httpClient
+	controllerClient client.Client
+}
+
+// handles the RegistrationService.RegistrationServiceResources part of the toolchainstatus
```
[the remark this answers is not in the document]
rajivnathan: Done in 4fe2dd1087523aa210c2dea19fda210f0345bfb1

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -0,0 +1,438 @@
...
+	toolchainStatus.Status.RegistrationService.RegistrationServiceResources.Conditions = registrationService.Status.Conditions
+	for _, condition := range registrationService.Status.Conditions {
+		if condition.Type == toolchainv1alpha1.ConditionReady && condition.Status != "True" {
+			return errs.Wrap(fmt.Errorf(condition.Message), "registration service resource not ready")
+		}
+	}
+	return nil
+}
+
+// handles the RegistrationService.Deployment part of the toolchainstatus
```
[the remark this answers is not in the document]
rajivnathan: Done in 4fe2dd1087523aa210c2dea19fda210f0345bfb1

pkg/controller/toolchainstatus/toolchainstatus_controller.go
rajivnathan: Ah, didn't know that. Thanks for the tip. Handled it in 4fe2dd1087523aa210c2dea19fda210f0345bfb1

pkg/controller/toolchainstatus/toolchainstatus_controller.go
rajivnathan: Nice catch, done in 4c61f7dc04914b4876dda060b2ae2c4460567717

pkg/controller/toolchainstatus/toolchainstatus_controller.go
rajivnathan: While it does make it more compact, conditions is still an array in the API and I figured we may want to add other conditions in the future.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
rajivnathan: I see your point but componentError is mainly used to indicate that there is a problem with a specific component, in this case the Members portion of the toolchainstatus. The details of the error will still be visible in the status condition for each member. I suppose we don't even need to really return an error, it could be a bool but we currently log the error using the reqLogger here: https://github.com/rajivnathan/host-operator/blob/a388dad6a7aa9ca73a51f3da87a1721bfa012b58/pkg/controller/toolchainstatus/toolchainstatus_controller.go#L172

But your point still stands in that only the error for the last cluster would be logged. Is there a good way to combine multiple errors into a single error and still have it clear and readable? Otherwise, maybe it's okay to only log the last error since each error is visible in the toolchainstatus?

pkg/controller/toolchainstatus/creation_test.go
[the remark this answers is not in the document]
alexeykazakov: IMO that's OK if we check that the message is `unable to create resource of kind: , version: : creation failed`. The main point here is to make sure we do include the original error `creation failed`.

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -0,0 +1,615 @@
...
+	response http.Response
+	err      error
+}
+
+func (f *fakeHTTPClient) Get(url string) (*http.Response, error) {
+	return &f.response, f.err
+}
+
+const respBodyGood = "{\"alive\":true,\"environment\":\"dev\",\"revision\":\"64af1be5c6011fae5497a7c35e2a986d633b3421\",\"buildTime\":\"0\",\"startTime\":\"2020-07-06T13:18:30Z\"}"
+const respBodyInvalid = "{\"not found\"}"
```
alexeykazakov, line 53: Just FYI. You can use `` instead:
```suggestion
const respBodyInvalid = `{"not found"}`
```

pkg/controller/toolchainstatus/creation_test.go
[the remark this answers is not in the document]
rajivnathan: Sounds good, I made the change here and for the other status-related unit tests as well

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
rajivnathan: I should have thought of that lol. I've updated it, thanks

rajivnathan:
@alexeykazakov 

> Looks good overall!
> Any e2e tests?

I put up the e2e tests here: https://github.com/codeready-toolchain/toolchain-e2e/pull/134

They're failing for me locally but in an unrelated area. Haven't determined why yet. Fails for me locally even when host-operator and toolchain-e2e are on master though so could be my environment. Will wait to see if it passes in CI.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
xcoulon: ok, sounds good to me ;)

pkg/controller/toolchainstatus/toolchainstatus_controller.go
xcoulon: I've seen functions that return an `Errors` instead of a standard `error`, for example.

rajivnathan: I think this is worth a bit more investigation and discussion so I'd like to handle it in a separate PR. I've included it in https://issues.redhat.com/browse/CRT-729 so we don't forget about it.

PR #242, opened 2020-07-10
fixing string replace to handle . and -

tinakurian:
Currently secret env vars are not properly created. The code replaces "-" instead of "." causing env var names such as `HOST_OPERATOR_MAILGUN.DOMAIN." Instead this should be `HOST_OPERATOR_MAILGUN_DOMAIN`

This PR is to handle replacing both "." and "-" with "_"


tinakurian:
> Can we add some tests for that?

@alexeykazakov good idea, i've added tests here: 50462149b881e583fc4b6cebd62962c6433c0ba6

PR #243, opened 2020-07-13
feat: support multiple attemps when updating MasterUserRecord

xcoulon:
Track each MasterUserRecord update failure as an individual condition in the
 `TemplateUpdateRequest.Status.Conditions` (thus allowing duplicates),
 but only retain the `complete=true/reason=success` condition if
 the update worked. Otherwise, requeue for another attempt
 In the NSTemplateTier controller, do not delete
 the `TemplateUpdateRequest` until the threshold
 is reached

Fixes https://issues.redhat.com/browse/CRT-708


xcoulon:
Notes:
- this PR "extends" https://github.com/codeready-toolchain/host-operator/pull/240 so only the last commit matters for the review here 
- No E2E tests, unless we have a "reliable" way to make `MasterUserRecord` update fail once or more times? (hahaha, looking for a reliable way to produce failures: the irony! 😂)

pkg/controller/templateupdaterequest/templateupdaterequest_controller.go
```diff
@@ -208,18 +238,18 @@ func namespaceType(templateRef string) string {
 	return parts[1]
 }
 
-// compareSyncIndexes compares the sync indexes in the given TemplateUpdateRequest status vs the given MasterUserRecord
-// returns `true` if ALL values are DIFFERENT, meaning that all user accounts were updated on the target clusters where the tier is in use
-func (r ReconcileTemplateUpdateRequest) compareSyncIndexes(logger logr.Logger, tur toolchainv1alpha1.TemplateUpdateRequest, mur toolchainv1alpha1.MasterUserRecord) bool {
+// sameSyncIndexes compares the sync indexes in the given TemplateUpdateRequest status vs the given MasterUserRecord
+// returns `false` if ALL values are DIFFERENT, meaning that all user accounts were updated on the target clusters where the tier is in use
```
alexeykazakov, line 242: I wonder why we consider the indexes be the same even if just one of them is the same and the rest are changed?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -301,6 +305,14 @@ func (r *NSTemplateTierReconciler) activeTemplateUpdateRequests(logger logr.Logg
 	return count, false, nil
 }
 
+// maxUpdateFailuresReached checks if the number of failure to update the MasterUserRecord is beyond the configured threshold
+func maxUpdateFailuresReached(tur toolchainv1alpha1.TemplateUpdateRequest, threshod int) bool {
+	return condition.Count(tur.Status.Conditions,
+		toolchainv1alpha1.TemplateUpdateRequestComplete,
+		corev1.ConditionFalse,
+		toolchainv1alpha1.TemplateUpdateRequestUnableToUpdateReason) > threshod
```
alexeykazakov, line 313: Should it be `>=`?
Otherwise we will actually make `maxAttemps` + 1 attempts, won't we?
If I'm not missing something I would set the default maxAttemps to 2 and use `>=` here.

pkg/controller/templateupdaterequest/templateupdaterequest_controller.go
xcoulon: I'm not sure I fully understand your question, but the point here is that we want to make sure that all syncIndexes changed, ie, that all user accounts associated with the updated tier have been updated. Does it clarify?

pkg/controller/nstemplatetier/nstemplatetier_controller.go
xcoulon: hmm, I thought my tests proved that the comparison was correct. 
Maybe the confusion is that the threshold is the number of failed attempts that allow for another retrial. So if `threshold=1`, then the controller may fail once, try again and then give up.
I understand it can be confusing, so I'm fine with changing the value to `2` and using `>=` instead. Just let me know ;)

alexeykazakov: Yes, exactly. You use this as a failure threshold. But IMO it can be confused with overall number of update attempts.
And I'm fine to keep it as a failure treshold but maybe we can rename it and add a comment to avoid confusion.
Something like:

```
// GetMasterUserRecordUpdateFailureThreshold returns the failure threshold for MUR updates. When a MUR update fails, we will re-try N times before giving up and stopping the entire tier update process.
func (c *Config) GetMasterUserRecordUpdateFailureThreshold() int {
	return c.host.GetInt(varMasterUserRecordUpdateFailureThreshold)
}
```

pkg/controller/templateupdaterequest/templateupdaterequest_controller.go
alexeykazakov: Then rename this function to `allSyncIndexesChanged()`? :)
Right now it's confusing. Right now `sameSyncIndexes()` returns true if one index is the same and another one is not. Which is confusing.

xcoulon: ok, let me do that

xcoulon: done in 011aabc

pkg/controller/nstemplatetier/nstemplatetier_controller.go
alexeykazakov: But this is minor.

xcoulon: done in b8cc04b

pkg/configuration/configuration.go
```diff
@@ -260,9 +260,9 @@ func (c *Config) GetEnvironment() string {
 	return c.host.GetString(varEnvironment)
 }
 
-// GetMasterUserRecordUpdateRetries returns the number of retries before failing to update a MasterUserRecord
-func (c *Config) GetMasterUserRecordUpdateRetries() int {
-	return c.host.GetInt(varMasterUserRecordUpdateRetries)
+// GetMasterUserRecordUpdateFailureThreshold returns the failure threshold for MUR updates. When a MUR update fails, we will re-try N times before giving up and stopping the entire tier update process.
```
alexeykazakov, line 263: This (my) comment is not good anymore. Sorry if I confused you but my suggestion was either clarify the language (so, I suggested this comment) or change the logic (what you did), so it's not the overall number of attempts and not the number of re-tries after the first failure (if any). So, your new comment "specifies the number allowed failures before stopping trying to update a MasterUserRecord" is what we need here too :)

Again, sorry for the confusion :(

```suggestion
// GetMasterUserRecordUpdateFailureThreshold returns the number of allowed failures before stopping trying to update a MasterUserRecord
```

xcoulon: no worries, it's also my fault. I made the changes in the evening and I was a bit in a hurry...

PR #245, opened 2020-07-14
Pod Idler

alexeykazakov:
This PR adds Idler resources to Tiers. The controller for Idlers has been added in https://github.com/codeready-toolchain/member-operator/pull/179

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/135


deploy/templates/nstemplatetiers/team/cluster.yaml
```diff
@@ -27,6 +27,18 @@ objects:
       annotations:
         openshift.io/requester: ${USERNAME}
       labels: null
+- apiVersion: toolchain.dev.openshift.com/v1alpha1
+  kind: Idler
+  metadata:
+    name: ${USERNAME}-dev
+  spec:
+    timeoutSeconds: 28800 # 8 hours
```
rajivnathan, line 35: Should we disable it for the team tier?

alexeykazakov: We don't have to disable idling for team pipelines because they are supposed to be short lived anyway. But we have couple of even listeners there which we don't wan't to kill...
@MatousJobanek should we maybe convert them to serverless services so they pods are short-lived too?

MatousJobanek: I have two comments on that:
1. I thought that serverless won't be available in the prod cluster - that the Serverless guys are not interested in maintaining/controlling the Serverless operator in hosted toolchain.
2. Even if the Serverless functionality was available then I'm still not sure if it would be possible to convert it to serverless service. The event listeners pods are owned by the EventListener CRs - this is pure Tekton functionality. However, maybe we don't have to care about these pods - if the pod is removed, then the EventListener controller will immediately create a new one.

alexeykazakov: Yep, you are right. I just scaled down the even listener deployment to 0 and it was immediately restored back to 1 by the controller. So, there is no problem :)

PR #248, opened 2020-07-15
adding environment variable

tinakurian:
Adding environment variable for host-operator in order to help write end-to-end tests for the notification service.

This env var will be used to determine whether or not we should even attempt sending a notification depending on the environment. In the e2e-environment it is not easily possible to test the notification service. Since we cannot expose the secret for end-to-end tests and setting up a mock server is not as easily do-able, we can instead set up an environment variable. If the environment variable is set to `e2e` then we will bypass sending the notification via mailgun and proceed onwards to set the notification status to `sent` and verify that the notification is deleted after the deletion timestamp has elapsed.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -61,6 +61,9 @@ const (
 
 	// varRegistrationServiceURL is the URL used to access the registration service
 	varRegistrationServiceURL = "registration.service.url"
+
+	// varEnvironment specifies the host-operator environment such as prod, stage, unit-tests, e2e-tests, dev, etc
+	varEnvironment = "environment"
 )
 
 // Config encapsulates the Viper configuration registry which stores the
@@ -187,6 +190,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(varNotificationDeliveryService, NotificationDeliveryServiceMailgun)
 	c.host.SetDefault(varDurationBeforeNotificationDeletion, "24h")
 	c.host.SetDefault(varRegistrationServiceURL, "https://registration.crt-placeholder.com")
+	c.host.SetDefault(varEnvironment, "dev")
 }
 
 // GetDurationBeforeChangeRequestDeletion returns the timeout before a complete TierChangeRequest will be deleted.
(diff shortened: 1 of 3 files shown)


pkg/configuration/configuration.go
```diff
@@ -187,6 +190,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(varNotificationDeliveryService, NotificationDeliveryServiceMailgun)
 	c.host.SetDefault(varDurationBeforeNotificationDeletion, "24h")
 	c.host.SetDefault(varRegistrationServiceURL, "https://registration.crt-placeholder.com")
+	c.host.SetDefault(varEnvironment, "dev")
```
alexeykazakov, line 193: We have to use `prod` by default or add `prod` to `deploy/env/prod.yaml`. I think it's less error prone to use `prod` here:
```suggestion
	c.host.SetDefault(varEnvironment, "prod")
```

pkg/configuration/configuration_test.go
```diff
@@ -60,6 +60,14 @@ func TestGetAllRegistrationServiceParameters(t *testing.T) {
 	})
 }
 
+func TestEnvironment(t *testing.T) {
+	restore := test.SetEnvVarAndRestore(t, "HOST_OPERATOR_ENVIRONMENT", "e2e-test")
+	defer restore()
+
+	config := getDefaultConfiguration(t)
+	assert.Equal(t, "e2e-test", config.GetEnvironment())
```
alexeykazakov, line 68: Also check that if no `HOST_OPERATOR_ENVIRONMENT` set then `prod` is used.

pkg/configuration/configuration.go
tinakurian: ohhh okay makes sense, i see that the registration service does the same. changed here: 9e8605c8ad04454dac98359c92913e9abdedc64d

pkg/configuration/configuration_test.go
tinakurian: good idea, updated here: 9e8605c8ad04454dac98359c92913e9abdedc64d

pkg/configuration/configuration.go
```diff
@@ -33,9 +33,12 @@ const (
 	// DefaultToolchainStatusName the default name for the toolchain status resource created during initialization of the operator
 	DefaultToolchainStatusName = "toolchain-status"
 
-	// VarDurationBeforeChangeRequestDeletion specificies the duration before a change request is deleted
+	// VarDurationBeforeChangeRequestDeletion specifies the duration before a change request is deleted
```
xcoulon, line 36: why is the `VarDurationBeforeChangeRequestDeletion ` variable public?

tinakurian: thanks for the review @xcoulon! it's used here: https://github.com/codeready-toolchain/host-operator/blob/master/cmd/manager/main.go#L285

PR #257, opened 2020-07-29
using configuration for console and che routes and namespaces

tinakurian:
We have to move
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/sync.go#L29-L30
and
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/sync.go#L178
to configuration so we can customize them via ConfigMap.

It's needed for OSD integration.

Jira: https://issues.redhat.com/browse/CRT-745
Related PR: https://github.com/codeready-toolchain/toolchain-infra/pull/28/files

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -84,6 +84,30 @@ const (
 	// defaultRegistrationServiceURL is the default location of the registration service
 	defaultRegistrationServiceURL = "https://registration.crt-placeholder.com"
 
+	// varConsoleNamespace is the console route namespace
+	varConsoleNamespace = "console.namespace"
+
+	// defaultConsoleNamespace is the default console route namespace
+	defaultConsoleNamespace = "openshift-console"
+
+	// varConsoleRouteName is the console route name
+	varConsoleRouteName = "console.route.name"
+
+	// defaultConsoleRouteName is the default console route name
+	defaultConsoleRouteName = "console"
+
+	// varCheNamespace is the che route namespace
+	varCheNamespace = "che.namespace"
+
+	// defaultCheNamespace is the default che route namespace
+	defaultCheNamespace = "toolchain-che"
+
+	// varCheRouteName is the che dashboard route
+	varCheRouteName = "che.route.name"
+
+	// defaultCheRouteName is the default che dashboard route
+	defaultCheRouteName = "che"
+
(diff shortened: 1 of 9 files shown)


pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -39,7 +39,10 @@ func TestAddFinalizer(t *testing.T) {
 		mur := murtest.NewMasterUserRecord(t, "john")
 		memberClient := test.NewFakeClient(t)
 		hostClient := test.NewFakeClient(t, mur)
-		cntrl := newController(hostClient, s, newGetMemberCluster(true, v1.ConditionTrue),
+		config, err := configuration.LoadConfig(hostClient)
+		require.NoError(t, err)
```
MatousJobanek, line 43: This could be moved into the `newController` function to not replicate it in all tests

pkg/configuration/configuration.go
```diff
@@ -270,6 +298,26 @@ func (c *Config) GetRegistrationServiceURL() string {
...
+
+// GetCheNamespace returns the Che route namespace
+func (c *Config) GetCheNamespace() string {
+	return c.host.GetString(varCheNamespace)
+}
+
+// GetCheRouteName returns the name of the Che dashboard route
+func (c *Config) GetCheRouteName() string {
+	return c.host.GetString(varCheRouteName)
+}
```
MatousJobanek, line 319: you didn't add tests for these methods

tinakurian: added here: fa2316815d63d4a7db87100bc4a77cd7235df3ef

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
tinakurian: changed here: fa2316815d63d4a7db87100bc4a77cd7235df3ef

pkg/configuration/configuration.go
tinakurian: and here: a77e5a89d165a6d496d8ef1a52de1a16f5109e0e

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
tinakurian: and here: 4e7bbf729a5194bb03e1bcc11216c9d964be35de :facepalm:

PR #262, opened 2020-08-12
fix: start metrics server at operator startup

xcoulon:
A couple of permissions were missing, which resulted
in the following errors:
```
{"level":"info","msg":"Could not create metrics Service","error":"failed to initialize service object for metrics: replicasets.apps \"host-operator-849b9cccc4\" is forbidden: User \"system:serviceaccount:xcoulon-host-12182550:host-operator\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"xcoulon-host-12182550\""}
```
and then
```
{"level":"info","msg":"Could not create ServiceMonitor object","error":"servicemonitors.monitoring.coreos.com \"host-operator-metrics\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/142

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -35,6 +35,7 @@ rules:
   - replicasets
   verbs:
   - "get"
+  - "list"
   - "create"
 - apiGroups:
   - apps
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -8,6 +8,7 @@ rules:
   resources:
   - pods
   - services
+  - services/finalizers
   - endpoints
   - persistentvolumeclaims
   - events
@@ -28,6 +29,13 @@ rules:
   - deployments
   verbs:
   - "*"
+- apiGroups:
+  - apps
+  resources:
+  - replicasets
+  verbs:
+  - "get"
+  - "create"
 - apiGroups:
   - apps
   resources:
--- a/deploy/olm-catalog/toolchain-host-operator/manifests/toolchain-host-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/toolchain-host-operator/manifests/toolchain-host-operator.clusterserviceversion.yaml
@@ -343,6 +343,7 @@ spec:
           resources:
           - pods
           - services
+          - services/finalizers
           - endpoints
           - persistentvolumeclaims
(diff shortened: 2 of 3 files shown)


xcoulon:
I'll see if I can add some e2e tests, for example by verifying that the pod is listening on port 8383

rajivnathan:
I guess this is linked to a previous PR? Maybe it's worth adding a link to the previous PR in the description for more context?

deploy/role.yaml
```diff
@@ -28,6 +29,13 @@ rules:
   - deployments
   verbs:
   - "*"
+- apiGroups:
+  - apps
+  resources:
+  - replicasets
+  verbs:
+  - "get"
+  - "create"
```
alexeykazakov, line 38: It's usually safer to add "list" as well if we add "get" just in case it's needed too.

xcoulon:
> I guess this is linked to a previous PR? Maybe it's worth adding a link to the previous PR in the description for more context? I think I missed the previous PR so I'm a bit out of the loop but I guess a link could also be helpful if we look back at this later :)

To be honest, I believe it's been failing for ages, so I don't think it's useful to link to a previous PR here.

deploy/role.yaml
xcoulon: ok, I'll add that

xcoulon:
> I was wondering why it was failing. Now we know why :) Thanks for fixing that!

my pleasure ;)

deploy/role.yaml
xcoulon: done in d47c571

alexeykazakov:
Forgot to re-generate?

xcoulon:
> Forgot to re-generate?

oh... 🤦‍♂️ 
done in 3cd9a4e. Thanks for catching this @alexeykazakov 😅

xcoulon:
> I'll see if I can add some e2e tests, for example by verifying that the pod is listening on port 8383

see https://github.com/codeready-toolchain/toolchain-e2e/pull/142

PR #263, opened 2020-08-17
registration-service configmap and secrets added

tinakurian:
This PR adds required configuration for twilio. 

This PR is also related to allowing registration-service to have it's configurations set via a configmap and set secrets (particularily for twilio). 

Related PR: https://github.com/codeready-toolchain/registration-service/pull/122, **PR for toolchain-infra coming soon...** 

Related Jira: https://issues.redhat.com/browse/CRT-775


cmd/manager/main.go
```diff
@@ -266,12 +266,6 @@ func serveCRMetrics(cfg *rest.Config, operatorNs string) error {
 }
 
 func printConfig(cfg *configuration.Config) {
-	logWithValuesRegServ := log
```
alexeykazakov, line 269: I think these could be useful to print. Let's maybe just mention that those variables can be overridden on the reg-service site and different values may be used by the reg-service instead.

deploy/env/dev.yaml
```diff
@@ -4,11 +4,19 @@ registration-service:
...
-    config-raw: '{
-                  "realm": "toolchain-public",
-                  "auth-server-url": "https://sso.prod-preview.openshift.io/auth",
-                  "ssl-required": "none",
-                  "resource": "crt",
-                  "clientId": "crt",
-                  "public-client": true
-                }'
\ No newline at end of file
+    config-raw:
```
alexeykazakov, line 7: Same here. Please do not change the raw configuration. Let's keep it as is.

deploy/env/dev.yaml
```diff
@@ -11,4 +11,13 @@ registration-service:
...
                   "clientId": "crt",
                   "public-client": true
-                }'
\ No newline at end of file
+                }'
+  verification.enabled: 'true'
+  verification.daily_limit: '3'
+  verification.attempts_allowed: '5'
+host-operator:
+  registration-service-url: 'https://sso.prod-preview.openshift.io/auth/realms/toolchain-public/protocol/openid-connect/certs'
```
alexeykazakov, line 19: This is not the registration-service-url. But since we can't actually provide the reg-service URL for the dev env I would stick with the default value we use in configuration.go and just skip it here.

deploy/env/dev.yaml
```diff
@@ -11,4 +11,13 @@ registration-service:
...
+  console-namespace: "openshift-console"
+  console-route-name: "console"
+  che-namespace: "toolchain-che"
+  che-route-name: "che"
```
alexeykazakov, line 23: Those are the default values we use in configuration.go, right? So, why adding them here?

tinakurian: we seem to have a few things added here that were already part of defaults in configuration.go such as the `config-raw` for instance. so i wasn't too sure if there was a reason for this and just followed what was already there. I will remove all the values that have defaults from here in that case

deploy/env/dev.yaml
```diff
@@ -11,4 +11,13 @@ registration-service:
                   "resource": "crt",
                   "clientId": "crt",
                   "public-client": true
-                }'
\ No newline at end of file
+                }'
+  verification.enabled: 'true'
+  verification.daily_limit: '3'
+  verification.attempts_allowed: '5'
```
MatousJobanek, line 17: This should be in a different format: 
```yaml
  verification:
    enabled: 'true'
    daily-limit: '3'
    attempts-allowed: '5'
```
Btw, why do you use the underscore in the param names? This can be confusing since the underscore is the default delimiter in the var names

deploy/env/dev.yaml
tinakurian: done

cmd/manager/main.go
tinakurian: removed here: 0bd59fa6669e3cf716c0411dd11adc2b9261ba8b

deploy/env/dev.yaml
tinakurian: removed fields that had defaults here: b03cfdd25a240c85b66225bceb6bdce25f774c59 and here: 0bd59fa6669e3cf716c0411dd11adc2b9261ba8b

deploy/env/dev.yaml
tinakurian: this doesn't need to be here since they are already set as defaults in configuration.go

deploy/env/dev.yaml
tinakurian: done as part of this: https://github.com/codeready-toolchain/host-operator/pull/263#discussion_r473209798

cmd/manager/main.go
tinakurian: done here: a2b8e79b0b05c4daed057b301cf845fe7469c6c6

tinakurian: @alexeykazakov i've also moved this to be part of the configuration (same as what you had suggested in the registration-service)

PR #264, opened 2020-08-18
test: replace 'assert' pkg

xcoulon:
replaced the use of `github.com/alecthomas/assert`
with `github.com/strechr/testify/assert`


xcoulon:
> Good catch.

I shall call it luck ;)

PR #266, opened 2020-08-19
Only provision user if verification is not required

sbryzak:
This PR adds a check to the UserSignup controller to ensure that users are only provisioned if they do not require a verification check.

Fixes https://issues.redhat.com/browse/CRT-766


alexeykazakov:
> I think we should also add some e2e tests for this - what do you think?

I think we can create (and we should!) some e2e tests when we cover the entire e2e flow: Reg-service adds VerificationRequired field. No MUR created. Ideally we mock verification somehow (or just set it to false if it's too tricky). MUR is created now.

sbryzak:
I've added an e2e test anyway, since it doesn't hurt.

PR #267, opened 2020-08-20
fix: recreate UserAccount when deleted (on purpose or by accident)

xcoulon:
When ensuring the UserAccounts for the MasterUserRecord exist, if one
is being deleted, then the controller will requeue (with a delay),
which should give time to the member cluster to finish deleting
the resources. Then, during the new reconcile loop, the
MasterUserRecord controller will recreate the UserAccount.

Also, small refactorings (func name and using `metav1.Now()`
as a shorthand)

E2E Tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/146

Fixes https://issues.redhat.com/browse/CRT-778

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -3,14 +3,17 @@ package masteruserrecord
 import (
 	"context"
 	"fmt"
+	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/pkg/configuration"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
+
 	"github.com/go-logr/logr"
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	errs "github.com/pkg/errors"
+	"github.com/redhat-cop/operator-utils/pkg/util"
 	coputil "github.com/redhat-cop/operator-utils/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -163,15 +170,20 @@ func (r *ReconcileMasterUserRecord) ensureUserAccount(logger logr.Logger, murAcc
...
 		}
 		// another/unexpected error occurred while trying to fetch the user account on the member cluster
-		return r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToGetUserAccountReason), err,
+		return false, r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToGetUserAccountReason), err,
 			"failed to get userAccount '%s' from cluster '%s'", mur.Name, murAccount.TargetCluster)
 	}
+	// if the UserAccount is being deleted (by accident?), then we should wait until is has been totally deleted, and this controller will recreate it again
+	if util.IsBeingDeleted(userAccount) {
+		logger.Info("UserAccount is being deleted. Waiting until deletion is complete", "member_cluster", memberCluster.Name)
+		return true, updateStatusConditions(logger, r.client, mur, toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, ""))
```
MatousJobanek, line 185: Wouldn't it be good to set also the message to something like:
```suggestion
		return true, updateStatusConditions(logger, r.client, mur, toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "recovering deleted UserAccount"))
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -132,6 +130,32 @@ func TestCreateMultipleUserAccountsSuccessful(t *testing.T) {
...
+		clusterClient(test.MemberClusterName, memberClient1),
+		clusterClient("member2-cluster", memberClient2),
+		clusterClient("member3-cluster", memberClient3))
+
+	// when reconciling
+	result, err := cntrl.Reconcile(newMurRequest(mur))
+	// then
+	require.NoError(t, err)
+	assert.True(t, result.Requeue)
+
```
MatousJobanek, line 156:
```suggestion
	assert.Equal(t, 3 * time.Second, result.RequeueAfter)
```

MatousJobanek (approved):
Nice solution :+1:

pkg/controller/masteruserrecord/masteruserrecord_controller.go
xcoulon: good idea!

PR #269, opened 2020-08-24
Generate NSTemplateTier from an openshift template

rajivnathan:
Updates the NSTemplateTier generator to create the NSTemplateTier resource based on an Openshift template provided in the new tier.yaml template file found alongside the other template files in `deploy/templates/nstemplatetiers/`.

Example template:
```
kind: Template
apiVersion: v1
metadata:
  name: advanced-tier
objects:
- kind: NSTemplateTier
  metadata:
    name: advanced
    namespace: ${NAMESPACE}
  spec:
    clusterResources:
      templateRef: ${CLUSTER_TEMPL_REF}
    namespaces:
      - templateRef: ${CODE_TEMPL_REF}
      - templateRef: ${DEV_TEMPL_REF}
      - templateRef: ${STAGE_TEMPL_REF}
parameters:
- name: NAMESPACE
- name: CLUSTER_TEMPL_REF
- name: CODE_TEMPL_REF
- name: DEV_TEMPL_REF
- name: STAGE_TEMPL_REF
```

Example of template once it's processed:
```
kind: NSTemplateTier
metadata:
  labels:
    toolchain.dev.openshift.com/provider: codeready-toolchain
  name: advanced
  namespace: host-operator
spec:
  clusterResources:
    templateRef: advanced-clusterresources-07cac69
  namespaces:
  - templateRef: advanced-code-cb6fbd2
  - templateRef: advanced-dev-4d49fe0
  - templateRef: advanced-stage-4d49fe0
```

Also updated the unit tests.

Related issue: https://issues.redhat.com/browse/CRT-643


pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -23,74 +24,110 @@ import (
...
+	rawTemplates   *templates
+	tierTemplates  []*toolchainv1alpha1.TierTemplate
+	nstmplTierObjs []commonclient.ToolchainObject
+}
+
+// templates: namespaces and other cluster-scoped resources belonging to a given tier ("advanced", "basic", "team", etc.) and the NSTemplateTier that combines them
 type templates struct {
 	namespaceTemplates map[string]template // namespace templates (including roles, etc.) indexed by type ("dev", "code", "stage")
 	clusterTemplate    *template           // other cluster-scoped resources, in a single template file
+	nsTemplateTier     *template           // NSTemplateTier resource with tier-scoped configuration and references to namespace and cluster templates in its spec, in a single template file
```
xcoulon, line 69: `clusterTemplate` is a pointer to `template` because it's optional, but I believe that the new `tier.yaml` template must exist for each tier, so I guess you don't need a pointer to `template` here?

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -23,74 +24,110 @@ import (
...
+
+	c := &tierGenerator{
+		client:          client,
+		namespace:       namespace,
+		scheme:          s,
+		templatesByTier: templatesByTier,
+	}
+
+	// process tierTemplates
+	err = c.newTierTemplates()
```
xcoulon, line 94: IMO it's a bit confusing to have these calls to `c.newTierTemplates()` and `c.newNSTemplateTiers()` here in the `newTierGenerator(...)` func, and then calls to `generator.createTierTemplates()` and `generator.createNSTemplateTiers()` in the `CreateOrUpdateResources` func above.
Would it be possible to merge the `c.newTierTemplates()` with `generator.createTierTemplates()` funcs and `c.newNSTemplateTiers()` with `generator.createNSTemplateTiers()`? 

Or maybe more simply, have `generator.createTierTemplates()` call `c.newTierTemplates()` and `generator.createNSTemplateTiers()` call `c.newNSTemplateTiers()` ?

xcoulon (commented):
I like the idea of having a new `tier.yaml` template which is easier to maintain compared to the current, programmatic way of generating the `NSTemplateTier` resources 🙌 
I left some comments to refactor the code and hopefully make it easier to understand.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
rajivnathan: That's a good point. I can try changing that.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
rajivnathan: The reason I call `c.newTierTemplates()` and `c.newNSTemplateTiers()` in the `newTierGenerator(...)` is so that if there were a problem with one of the templates we process all of the templates and error out before any resources are created on the system. But I can see how the current way is a bit confusing. How about if I merge `c.newTierTemplates()` and `c.newNSTemplateTiers()` into 1 function eg. `initGenerator()` and that is called when `generator.createTierTemplates()` or `generator.createNSTemplateTiers()` are both called?

xcoulon: > The reason I call `c.newTierTemplates()` and `c.newNSTemplateTiers()` in the `newTierGenerator(...)` is so that if there were a problem with one of the templates we process all of the templates and error out before any resources are created on the system.

it's a fair point, indeed. But I believe that the resources potentially created before an error occurs will be "recreated" at the next deployment (well, updated with probably no change is more accurate)

> How about if I merge `c.newTierTemplates()` and `c.newNSTemplateTiers()` into 1 function eg. `initGenerator()` and that is called when `generator.createTierTemplates()` or `generator.createNSTemplateTiers()` are both called?

yes, that could be an option, indeed. Or that could remain in the `newTierGenerator(...)` func, but maybe named differently to avoid the confusion with the names?

rajivnathan: So I tried a few different things and ended up renaming `c.newTierTemplates()` to `c.initTierTemplates()` and `c.newNSTemplateTiers()` to `c.initNSTemplateTiers()` and then creating a new `c.initTemplateData()` function that calls both of them. The `newTierGenerator()` func now calls `c.initTemplateData()`. You can see the change in https://github.com/codeready-toolchain/host-operator/pull/269/commits/01de1ec7c2c308aa3bd0d0647b2b283d01126623
Do you think it's less confusing now?

xcoulon: yes, it's cleaner like this indeed, thanks!

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/269/commits/b571580f5c26cb418c69a9ee10f09565bdd033ca

xcoulon (approved):
Thanks for undertaking this refactoring @rajivnathan 🙌 
It's going to make our lives simpler when we have to make changes in the generated NSTemplateTier templates!

rajivnathan:
Thanks for the review @xcoulon and for the helpful suggestions!

PR #275, opened 2020-09-03
chore: refactor ToolchainStatus controller

MatousJobanek:
This PR refactors the ToolchainStatus controller:
* the handlers doesn't return error but boolean saying if the components are ready or not
   * as a side effect, we will log all errors, not only the latest one (as it was done before)
* it checks if any ToolchainCluster that was previously present was removed - if so, then it sets the `ready = false` with reason `MemberToolchainClusterRemoved`
    * in the next PR, it will be combined with the conter - it will verify if the removed member cluster contains any provisioned UserAccounts or not to check if the removal was safe.
* removed the check if the member client is nill or not - this was unreal use-case - all member clusters that are in cache has the client initialized

All these changes are required to make it working with the counter that will come in the next PR.
* sligtly modified the message handling
* removed unused function

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.6.0
 )
 
+replace github.com/codeready-toolchain/api => github.com/matousjobanek/api v0.0.0-20200903135031-26fcc19ae417
+
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -30,6 +30,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.6.0
 )
 
+replace github.com/codeready-toolchain/api => github.com/matousjobanek/api v0.0.0-20200903135031-26fcc19ae417
```
alexeykazakov, line 33: A reminder :)

alexeykazakov (approved):

rajivnathan (approved):
Nice changes! 👍

go.mod
MatousJobanek: updated 75f6fdc

PR #276, opened 2020-09-04
feat: add MUR and UserAccount counter

MatousJobanek:
This PR adds a counter that counts all created/deleted MasterUserRecords in host cluster and UserAccounts in member clusters.
It keeps the counts in a cached object and let ToolchainStatus synchronize with the object.

As part of doing the necessary changes in the UserSignup/MUR tests I added missing comments + added some subtests to make it easier to read.

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/159

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -186,6 +186,7 @@ func (r *ReconcileMasterUserRecord) ensureUserAccount(logger logr.Logger, murAcc
 		logger.Info("UserAccount is being deleted. Waiting until deletion is complete", "member_cluster", memberCluster.Name)
 		deletionTimestamp := userAccount.GetDeletionTimestamp()
 
+		// this code block makes sure that we will requeue but only 3s after the deletion timestamp, and we should not update the counter twice
 		requeueTime := 3 * time.Second
 		timeUntilDeletion := time.Until(deletionTimestamp.Time)
 		if timeUntilDeletion+3*time.Second >= 0 {
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -288,6 +288,7 @@ func (r *ReconcileMasterUserRecord) manageCleanUp(logger logr.Logger, mur *toolc
 		return r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToRemoveFinalizerReason), err,
 			"failed to update MasterUserRecord while deleting finalizer")
 	}
+	counter.DecrementMasterUserRecordCount(logger)
 	log.Info("Finalizer removed from MasterUserRecord")
 	return nil
 }
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -117,12 +117,12 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		}
 		logger.Info("ensuring user accounts")
 		for _, account := range mur.Spec.UserAccounts {
-			requeue, err := r.ensureUserAccount(logger, account, mur)
+			requeueTime, err := r.ensureUserAccount(logger, account, mur)
 			if err != nil {
 				logger.Error(err, "unable to synchronize with member UserAccount")
 				return reconcile.Result{}, err
-			} else if requeue {
-				return reconcile.Result{Requeue: true, RequeueAfter: 3 * time.Second}, err // waiting for a few seconds to give time to the member cluster to finish its deletions
+			} else if requeueTime > 0 {
+				return reconcile.Result{Requeue: true, RequeueAfter: requeueTime}, err // waiting for a few seconds to give time to the member cluster to finish its deletions
 			}
 		}
 		// If the UserAccount is being deleted, delete the UserAccounts in members.
@@ -154,12 +154,12 @@ func (r *ReconcileMasterUserRecord) addFinalizer(logger logr.Logger, mur *toolch
 // ensureUserAccount ensures that there's a UserAccount resource on the member cluster for the given `murAccount`.
 // If the UserAccount resource already exists, then this latter is synchronized using the given `murAccount` and the associated `mur` status is also updated to reflect
 // the UserAccount specs.
-// Returns `true, nil` if there is a need for requeing (eg, if the remote UserAccount is being deleted and the controller should wait until the deletion is complete)
-func (r *ReconcileMasterUserRecord) ensureUserAccount(logger logr.Logger, murAccount toolchainv1alpha1.UserAccountEmbedded, mur *toolchainv1alpha1.MasterUserRecord) (bool, error) {
+// Returns non-zero duration as the first argument if there is a need for requeing (eg, if the remote UserAccount is being deleted and the controller should wait until the deletion is complete)
+func (r *ReconcileMasterUserRecord) ensureUserAccount(logger logr.Logger, murAccount toolchainv1alpha1.UserAccountEmbedded, mur *toolchainv1alpha1.MasterUserRecord) (time.Duration, error) {
 	// get & check member cluster
 	memberCluster, err := r.getMemberCluster(murAccount.TargetCluster)
 	if err != nil {
-		return false, r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordTargetClusterNotReadyReason), err,
+		return 0, r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordTargetClusterNotReadyReason), err,
 			"failed to get the member cluster '%s'", murAccount.TargetCluster)
 	}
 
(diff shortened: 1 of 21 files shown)


alexeykazakov (commented):
Nice!

But what difficulties you face with the e2e tests?

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// If the ToolchainStatus doesn't contain any MUR counts, then it lists all existing
+// MURs, counts them and stores in both cache and ToolchainStatus object
+//
+// If the cached counter is initialized and ToolchainStatus contains already some numbers
+// then it updates the ToolchainStatus numbers with the one taken from the cached counter
+func Synchronize(cl client.Client, toolchainStatus *v1alpha1.ToolchainStatus) error {
+	cachedCounts.Lock()
+	defer cachedCounts.Unlock()
+	if toolchainStatus.Status.HostOperator == nil || toolchainStatus.Status.HostOperator.CapacityUsage.MasterUserRecordCount == 0 {
```
xcoulon, line 113: why do you need to check if `toolchainStatus.Status.HostOperator.CapacityUsage.MasterUserRecordCount == 0`? Isn't `toolchainStatus.Status.HostOperator==nil` enough? (and we could have 0 MUR when the platform is just started)

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+		if err := loadCurrentNumbers(cl, toolchainStatus.Namespace); err != nil {
+			return err
+		}
+	}
+	if toolchainStatus.Status.HostOperator == nil {
+		toolchainStatus.Status.HostOperator = &v1alpha1.HostOperatorStatus{}
+	}
+	if !cachedCounts.initialized {
+		cachedCounts.MasterUserRecordCount += toolchainStatus.Status.HostOperator.CapacityUsage.MasterUserRecordCount
```
xcoulon, line 122: why `cachedCounts.MasterUserRecordCount += ` and not just `cachedCounts.MasterUserRecordCount = ` ?

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+		for _, memberStatus := range toolchainStatus.Status.Members {
+			cachedCounts.UserAccountsPerClusterCounts[memberStatus.ClusterName] += memberStatus.CapacityUsage.UserAccountCount
```
xcoulon, line 124: same question here: why incrementing a counter value instead of assigning it?

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		cachedCounts.initialized = true
+	}
+
+	toolchainStatus.Status.HostOperator.CapacityUsage.MasterUserRecordCount = cachedCounts.MasterUserRecordCount
+CachedCountsPerCluster:
+	for clusterName, count := range cachedCounts.UserAccountsPerClusterCounts {
+		for index, member := range toolchainStatus.Status.Members {
```
xcoulon, line 132: the code here could be a bit simpler if `toolchainStatus.Status.Members` was a `map[string]Member` instead of a `[]Member` 🤔

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+	})
+}
+
+// GetCounts returns Counts struct containing number of MURs and number of UserAccounts per member cluster.
+// If the counter is not yet initialized, then it returns error
+func GetCounts() (Counts, error) {
+	cachedCounts.RLock()
+	defer cachedCounts.RUnlock()
+	if !cachedCounts.initialized {
+		return cachedCounts.Counts, fmt.Errorf("counter is not initialized")
```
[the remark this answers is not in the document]
MatousJobanek: This is slightly tricky - we need to initialize with the existing `ToolchainStatus`. I should not retrieve the `ToolchainStatus` before having the client cache synced, but waiting for the cache being synced can be slower than the execution in controllers. Thus, I decided to let the `ToolchainStatus` initialize the cache during the first reconcile, which should be immediate.
Theoretically, when the `UserSignup` controller will read the cache, it will probably also have the `ToolchainStatus` retrieved, so it will be possible to pass it together with the resource. On the other hand, the only entity who should update and synchronize the `ToolchainStatus` resource is its controller and not `UserSignup`. So, I would really let the `ToolchainStatus` initialize the cache - there won't be almost not problem with it - in the worst case, the controller will rerun the reconciliation for unapproved `UserSignups` based on the returned error one more time.

pkg/counter/cache.go
```diff
@@ -0,0 +1,162 @@
...
+			ClusterName: clusterName,
+			CapacityUsage: v1alpha1.CapacityUsageMember{
+				UserAccountCount: count,
+			},
+		})
+	}
+	return nil
+}
+
+func loadCurrentNumbers(cl client.Client, namespace string) error {
```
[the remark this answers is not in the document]
MatousJobanek: Well, it doesn't load only MUR counts, but also number of UserSignups. `loadCurrentNumbers` sounds good to me, but if you don't like it, then I could offer you something like: `countMursAndUserAccounts` WDYT?

pkg/counter/cache.go
MatousJobanek: It's not enough, the `toolchainStatus.Status.HostOperator` will be set by `ToolchainStatus` controller  - I'm checking if it's nil just for my own paranoia - I don't like such panics :grimacing: 
Anyway, I need to check the case when this feature is deployed to the cluster for the first time - in that case the count is zero and I need to list all present MURs and count them.

pkg/counter/cache.go
MatousJobanek: because I want to cover the case when it could happen that there is a new MUR or UserAccount created in the small tiny gap before the cache is initialized. Then the cache would already contain some number with need to add to the one in `ToolchainStatus`.

pkg/counter/cache.go
MatousJobanek: and the same answer :sweat_smile:

pkg/counter/cache.go
MatousJobanek: yeah, but I don't want to refactor the `ToolchainStatus` CRD just because of this for cycle :man_shrugging: 
I remember that there was some discussion between @rajivnathan and @alexeykazakov about using map, and they decided for using the array. I like the array more.

alexeykazakov (approved):
Did only very brief review. Looks good! :)

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -170,19 +171,31 @@ func (r *ReconcileMasterUserRecord) ensureUserAccount(logger logr.Logger, murAcc
...
 	// if the UserAccount is being deleted (by accident?), then we should wait until is has been totally deleted, and this controller will recreate it again
 	if util.IsBeingDeleted(userAccount) {
 		logger.Info("UserAccount is being deleted. Waiting until deletion is complete", "member_cluster", memberCluster.Name)
-		return true, updateStatusConditions(logger, r.client, mur, toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "recovering deleted UserAccount"))
+		deletionTimestamp := userAccount.GetDeletionTimestamp()
+
+		requeueTime := 3 * time.Second
+		timeUntilDeletion := time.Until(deletionTimestamp.Time)
+		if timeUntilDeletion+3*time.Second >= 0 {
+			counter.DecrementUserAccountCount(logger, murAccount.TargetCluster)
```
xcoulon, line 192: ok, so if I understand correctly, the code here makes sure that we will requeue but only 3s after the deletion timestamp, but we should not update the counter twice? 
Maybe you could add a comment to explain the reasoning? (for later, when we come back to this code)

xcoulon, line 192: `counter.DecrementUserAccountCount(logger, murAccount.TargetCluster)` is also called on line 315, is it normal?

pkg/counter/cache.go
```diff
@@ -0,0 +1,166 @@
...
+func IncrementMasterUserRecordCount() {
+	write(func() {
+		cachedCounts.MasterUserRecordCount++
+	})
+}
+
+// DecrementMasterUserRecordCount decreases the number of MasterUserRecord in the cached counter
+func DecrementMasterUserRecordCount(log logr.Logger) {
+	write(func() {
+		if cachedCounts.MasterUserRecordCount != 0 || !cachedCounts.initialized {
```
xcoulon, line 61: shouldn't it be 
```suggestion
		if cachedCounts.MasterUserRecordCount != 0 && cachedCounts.initialized {
```
instead?

pkg/counter/cache.go
```diff
@@ -0,0 +1,166 @@
...
+func IncrementUserAccountCount(clusterName string) {
+	write(func() {
+		cachedCounts.UserAccountsPerClusterCounts[clusterName]++
+	})
+}
+
+// DecrementUserAccountCount decreases the number of UserAccount fo the given member cluster in the cached counter
+func DecrementUserAccountCount(log logr.Logger, clusterName string) {
+	write(func() {
+		if cachedCounts.UserAccountsPerClusterCounts[clusterName] != 0 || !cachedCounts.initialized {
```
xcoulon, line 80: same question here
```suggestion
		if cachedCounts.UserAccountsPerClusterCounts[clusterName] != 0 && cachedCounts.initialized {
```

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: Yes, you are right with the understanding. 
Ok, I'll add the explanation there - maybe I'll copy-paste your comment :-)

MatousJobanek: Yes, it is - this covers the case when the UserAccount is deleted manually, the second one is for the standard deletion.
BTW, I think that you wrote this code :-)

xcoulon: yes, I remember about that. That's because we needed to restore a `MUR` in case of manual deletion (cc: @alexeykazakov 😛)

xcoulon: sure, feel free to reuse my prose! 😉

pkg/counter/cache.go
MatousJobanek: nope, my version is correct. 
I don't want to let the counter decrement below zero if the cache is initialized, but I can let go into negative numbers if the counter is not initialized. This cover a small tiny use-case when MUR would be deleted before it is initialized by the content of the ToolchainStatus

pkg/counter/cache.go
MatousJobanek: and the same answer, but this time it applies on UserAccounts :-)

pkg/counter/cache.go
xcoulon: I'm still not sure why we have to cover such a tiny gap during the startup. I mean, when the counters are initialized, shouldn't we have the exact/absolute values?

pkg/counter/cache.go
xcoulon: sure, and same concern ;)

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: done 7623326

pkg/counter/cache.go
MatousJobanek: Probably because I'm too paranoid :man_shrugging: 
I was thinking about this use case:
1. there is some bug why users are not properly removed, or the host-operator pod is dead and meantime some users should be deleted
2. we deploy a new version of host operator
3. at the very beginning, the MUR controller goes through the MUR to be deleted and decrements the counter
4. the ToolchainStatus controller synchronizes with the cache

xcoulon: what if we synchronize the cached counters with the database on a regular basis (ie, using some requeuing or some go rountine)? In that case, even if we miss a creation or deletion, the counter would only remain "off by 1" for a certain time, not until the next deployment.

MatousJobanek: > what if we synchronize the cached counters with the database on a regular basis (ie, using some requeuing or some go rountine)?

all the possible approaches were mentioned and discussed in the related doc - this was one of the options
https://docs.google.com/document/d/1W--tiikHRFiiQ3LHiWfY9OXytP8klmT3Xk2JvFSA26k/edit#heading=h.dm34odbbr6ln
but I guess that listing all 3k or 10k MURs every 5 seconds is not very optimal. To avoid such listing I came with the cached counter. 

> In that case, even if we miss a creation or deletion, the counter would only remain "off by 1" for a certain time, not until the next deployment.

just to be clear, the current implementation doesn't synchronize with etcd after each deployment - it lists all existing MURs only once when this feature is deployed to the running cluster. The reason is that we need to count the already provisioned users.
One of the options I mentioned in the doc was that we would synchronize the counter with etcd every 24 hours or something like that. We (me and @alexeykazakov) decided not to do that right now and introduce the synchronization when we figure out that it's needed.

xcoulon (approved):

PR #281, opened 2020-09-14
Deactivation controller

rajivnathan:
Adds a new deactivation controller that watches MUR resources and triggers deactivation of users after the deactivation period expires. The deactivation period is determined by their associated tier.

Adds a configurable domain exclusion list that allows user accounts that match any of the domains in the list to be excluded from automatic deactivation

- The deactivation controller checks the NSTemplateTier associated with the MUR to discover the deactivation timeout value
- If the time since the MUR was provisioned exceeds the Deactivation Period then deactivate the user
- Otherwise, it will requeue the reconcile to a time when the user should be deactivated
- Added unit tests

Note: We cannot activate auto deactivation until we have an API to delete users from Che so this PR introduces the controller but it will not deactivate any users because none of the tiers have `DeactivationTimeoutDays` set. Once the Che changes are in we can put up a separate PR to update the tiers.

Related issue:
https://issues.redhat.com/browse/CRT-755

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -122,6 +122,12 @@ const (
 
 	// defaultToolchainStatusRefreshTime is the default refresh period for ToolchainStatus
 	defaultToolchainStatusRefreshTime = "5s"
+
+	// varDeactivationDomainsExcluded is a string of space-separated domains that should be excluded from automatic user deactivation
+	varDeactivationDomainsExcluded = "deactivation.domains.excluded"
+
+	// defaultDeactivationDomainsExcluded is the default set of domains that should be excluded from automatic user deactivation
+	defaultDeactivationDomainsExcluded = "@redhat.com"
 )
 
 // Config encapsulates the Viper configuration registry which stores the
@@ -194,6 +200,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(varEnvironment, defaultEnvironment)
 	c.host.SetDefault(varMasterUserRecordUpdateFailureThreshold, 2) // allow 1 failure, try again and then give up if failed again
 	c.host.SetDefault(varToolchainStatusRefreshTime, defaultToolchainStatusRefreshTime)
+	c.host.SetDefault(varDeactivationDomainsExcluded, defaultDeactivationDomainsExcluded)
 }
 
 // GetToolchainStatusName returns the configured name of the member status resource
@@ -290,4 +297,9 @@ func (c *Config) GetAllRegistrationServiceParameters() map[string]string {
 		}
 	}
 	return vars
+}
+
+// GetDeactivationDomainsExcludedList returns a string of space-separated domains that should be excluded from automatic user deactivation
+func (c *Config) GetDeactivationDomainsExcludedList() string {
+	return c.host.GetString(varDeactivationDomainsExcluded)
 }
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -123,11 +123,9 @@ const (
 	// defaultToolchainStatusRefreshTime is the default refresh period for ToolchainStatus
 	defaultToolchainStatusRefreshTime = "5s"
 
-	// varDeactivationDomainsExcluded is a string of space-separated domains that should be excluded from automatic user deactivation
+	// varDeactivationDomainsExcluded is a string of comma-separated domains that should be excluded from automatic user deactivation
+	// For example: "@redhat.com,@ibm.com"
 	varDeactivationDomainsExcluded = "deactivation.domains.excluded"
-
-	// defaultDeactivationDomainsExcluded is the default set of domains that should be excluded from automatic user deactivation
-	defaultDeactivationDomainsExcluded = "@redhat.com"
 )
 
 // Config encapsulates the Viper configuration registry which stores the
@@ -200,7 +198,6 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(varEnvironment, defaultEnvironment)
 	c.host.SetDefault(varMasterUserRecordUpdateFailureThreshold, 2) // allow 1 failure, try again and then give up if failed again
 	c.host.SetDefault(varToolchainStatusRefreshTime, defaultToolchainStatusRefreshTime)
-	c.host.SetDefault(varDeactivationDomainsExcluded, defaultDeactivationDomainsExcluded)
 }
 
 // GetToolchainStatusName returns the configured name of the member status resource
@@ -299,7 +296,9 @@ func (c *Config) GetAllRegistrationServiceParameters() map[string]string {
 	return vars
 }
 
-// GetDeactivationDomainsExcludedList returns a string of space-separated domains that should be excluded from automatic user deactivation
-func (c *Config) GetDeactivationDomainsExcludedList() string {
-	return c.host.GetString(varDeactivationDomainsExcluded)
+// GetDeactivationDomainsExcludedList returns a string of comma-separated user email domains that should be excluded from automatic user deactivation
+func (c *Config) GetDeactivationDomainsExcludedList() []string {
+	return strings.FieldsFunc(c.host.GetString(varDeactivationDomainsExcluded), func(c rune) bool {
+		return c == ','
+	})
 }
--- a/pkg/controller/deactivation/deactivation_controller.go
+++ b/pkg/controller/deactivation/deactivation_controller.go
@@ -122,7 +122,7 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
-	// The tier a mur belongs to is defined as part of its user account, since murs can in theory have multiple user accounts we will only consider the first one
+	// The tier is defined as part of its user account, since murs can in theory have multiple user accounts we will only consider the first one
 	account := mur.Spec.UserAccounts[0]
 
 	// Get the tier associated with the user account, we'll observe the deactivation timeout period from the tier spec
--- a/pkg/controller/deactivation/deactivation_controller.go
+++ b/pkg/controller/deactivation/deactivation_controller.go
@@ -110,7 +110,7 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
 
 	// Check the domain exclusion list, if the user's email matches then they cannot be automatically deactivated
 	if emailLbl, exists := usersignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
-		for _, domain := range strings.Split(r.config.GetDeactivationDomainsExcludedList(), " ") {
+		for _, domain := range r.config.GetDeactivationDomainsExcludedList() {
 			if strings.HasSuffix(emailLbl, domain) {
 				logger.Info("user cannot be automatically deactivated because they belong to the exclusion list", "domain", domain)
 				return reconcile.Result{}, nil
--- a/pkg/controller/deactivation/deactivation_controller.go
+++ b/pkg/controller/deactivation/deactivation_controller.go
@@ -2,6 +2,7 @@ package deactivation
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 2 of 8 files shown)


pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,421 @@
...
+			expectedDeactivationTimeout := 30
+			userSignup := &toolchainv1alpha1.UserSignup{
+				ObjectMeta: newObjectMeta(username, "foo@bar.com"),
+				Spec: toolchainv1alpha1.UserSignupSpec{
+					Username:      "foo@bar.com",
+					Approved:      true,
+					TargetCluster: "east",
+				},
+			}
+			basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates, tiertest.WithCurrentUpdateInProgress())
```
MatousJobanek, line 53: you use the subtests, but you don't use the main functionality of it - that you could share the test setup. This line could be moved to the main test, since it's used in other subtests.
Similarly, for the other tiers...

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,421 @@
...
+			// given
+			expectedDeactivationTimeout := 30
+			userSignup := &toolchainv1alpha1.UserSignup{
+				ObjectMeta: newObjectMeta(username, "foo@bar.com"),
+				Spec: toolchainv1alpha1.UserSignupSpec{
+					Username:      "foo@bar.com",
+					Approved:      true,
+					TargetCluster: "east",
+				},
+			}
```
MatousJobanek, line 52: it looks like the `expectedDeactivationTimeout := 30` as well as the `userSignup` is used in other subtests as well - could you please extract it?

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,421 @@
...
+			basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates, tiertest.WithCurrentUpdateInProgress())
+			murProvisionedTime := &metav1.Time{Time: time.Now()}
+			mur := murtest.NewMasterUserRecord(t, username, murtest.Account("cluster1", *basicTier), murtest.ProvisionedMur(murProvisionedTime), murtest.UserIDFromUserSignup(userSignup))
+			initObjs := []runtime.Object{basicTier, mur, userSignup}
+			r, req, cl := prepareReconcile(t, mur.Name, initObjs...)
```
MatousJobanek, line 57: why do we need to store it into an array? we can use it without it...
```suggestion
			r, req, cl := prepareReconcile(t, mur.Name, basicTier, mur, userSignup)
```

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Annotations: map[string]string{
+			toolchainv1alpha1.UserSignupUserEmailAnnotationKey: email,
+		},
+		Labels: map[string]string{
+			toolchainv1alpha1.UserSignupUserEmailHashLabelKey: emailHash,
+		},
+	}
+}
+
+func isUserSignupDeactivated(cl *test.FakeClient, name string) (bool, error) {
```
MatousJobanek, line 414: we could make it an assertion method 
`func assertThatUserSignupDeactivated(t *testing.T, cl *test.FakeClient, name string, expectedValue bool)`
it would make the tests shorter

pkg/controller/deactivation/deactivation_controller_test.go
MatousJobanek: applies for all other tests

MatousJobanek (commented):
I like the logic - it's clear and easy to follow :+1: I would maybe try to split the `Reconcile` method - it's pretty long, but since the logic is simple then it's probably fine :-).

And btw, don't you want to set some domain exclusions for e2e tests?

pkg/configuration/configuration.go
```diff
@@ -122,6 +122,12 @@ const (
 
 	// defaultToolchainStatusRefreshTime is the default refresh period for ToolchainStatus
 	defaultToolchainStatusRefreshTime = "5s"
+
+	// varDeactivationDomainsExcluded is a string of space-separated domains that should be excluded from automatic user deactivation
+	varDeactivationDomainsExcluded = "deactivation.domains.excluded"
+
+	// defaultDeactivationDomainsExcluded is the default set of domains that should be excluded from automatic user deactivation
+	defaultDeactivationDomainsExcluded = "@redhat.com"
```
alexeykazakov, line 130: Let's not use any domains by default. I would rather deactivate everything by default.

pkg/configuration/configuration.go
```diff
@@ -291,3 +298,8 @@ func (c *Config) GetAllRegistrationServiceParameters() map[string]string {
 	}
 	return vars
 }
+
+// GetDeactivationDomainsExcludedList returns a string of space-separated domains that should be excluded from automatic user deactivation
```
alexeykazakov, line 302: Also, maybe it's worth clarifying that we are talking about **user email** domains here?

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -0,0 +1,166 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("deactivation-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource MasterUserRecord
+	err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.MasterUserRecord{}}, &handler.EnqueueRequestForObject{})
```
alexeykazakov, line 47: We need to watch `mur.Status.ProvisionedTime` only, right? Not interested in Spec? If so then we could use some predicate which filters only Status changes. I think we use it some controllers...

alexeykazakov (commented):
Nice work!

alexeykazakov (commented):
Btw, what about e2e tests?
Could be tricky to use deactivation timeouts in days for that though...

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		Namespace: mur.Namespace,
+		Name:      mur.Spec.UserID,
+	}, usersignup); err != nil {
+		// Error getting usersignup - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Check the domain exclusion list, if the user's email matches then they cannot be automatically deactivated
+	if emailLbl, exists := usersignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
+		for _, domain := range strings.Split(r.config.GetDeactivationDomainsExcludedList(), " ") {
```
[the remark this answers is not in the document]
rajivnathan: I believe @alexeykazakov mentioned space-separated list. I don't really have a preference so I'll let you two decide. 🙂

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -0,0 +1,166 @@
...
+			if strings.HasSuffix(emailLbl, domain) {
+				logger.Info("user cannot be automatically deactivated because they belong to the exclusion list", "domain", domain)
+				return reconcile.Result{}, nil
+			}
+		}
+	}
+
+	if len(mur.Spec.UserAccounts) == 0 {
+		logger.Error(err, "cannot determine deactivation timeout period because the mur has no associated user accounts")
```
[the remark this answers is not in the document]
rajivnathan: 🤦  - good catch!

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
alexeykazakov: I don't have a preference either :)
Fine with both options.

pkg/controller/deactivation/deactivation_controller_test.go
rajivnathan: Good point, fixed in aa87f8eda85f8c5c07982f59f310afe925a115bd

pkg/controller/deactivation/deactivation_controller_test.go
rajivnathan: Good point, fixed in aa87f8eda85f8c5c07982f59f310afe925a115bd

pkg/controller/deactivation/deactivation_controller_test.go
rajivnathan: Good point, fixed in aa87f8eda85f8c5c07982f59f310afe925a115bd

pkg/controller/deactivation/deactivation_controller_test.go
rajivnathan: Also a good point, done in 990b83d5fa5e4071eb93d641b014e07a6d006004

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
rajivnathan: Changed to comma-separate list for consistency with the phone verification code. done in 3d97e57209f2a8ff7f5039f53906370909e9c60a

pkg/configuration/configuration.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/commit/3d97e57209f2a8ff7f5039f53906370909e9c60a

pkg/configuration/configuration.go
rajivnathan: Done

pkg/controller/deactivation/deactivation_controller.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/281/commits/e07f94318ac6b69ffc0e2e9b68e18c5df87a18ce

alexeykazakov (approved):

rajivnathan:
e2e tests are here: https://github.com/codeready-toolchain/toolchain-e2e/pull/162

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	userSignupFoobar := userSignupWithEmail(username, "foo@bar.com")
+	userSignupRedhat := userSignupWithEmail(username, "foo@redhat.com")
+
+	t.Run("controller should not deactivate user", func(t *testing.T) {
+
+		// the time since the mur was provisioned is within the deactivation timeout period for the 'basic' tier
+		t.Run("usersignup should not be deactivated - basic tier (30 days)", func(t *testing.T) {
+			// given
+			murProvisionedTime := &metav1.Time{Time: time.Now()}
```
MatousJobanek, line 54: it's just a detail, but you can use a shorter version:
```
murProvisionedTime := metav1.Now()
```

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -0,0 +1,305 @@
...
+			mur := murtest.NewMasterUserRecord(t, username, murtest.Account("cluster1", *noDeactivationTier), murtest.ProvisionedMur(murProvisionedTime), murtest.UserIDFromUserSignup(userSignupFoobar))
+			r, req, cl := prepareReconcile(t, mur.Name, noDeactivationTier, mur, userSignupFoobar)
+			// when
+			res, err := r.Reconcile(req)
+			// then
+			require.NoError(t, err)
+			require.False(t, res.Requeue, "requeue should not be set")
+			require.True(t, res.RequeueAfter == 0, "requeue should not be set")
+			assertThatUserSignupDeactivated(t, cl, username, false)
+		})
```
MatousJobanek, line 197: this test doesn't cover any failure and looks like a duplication of https://github.com/codeready-toolchain/host-operator/pull/281/files#diff-61785d47f3d452a843fb690e18cb6eefR88-R100

MatousJobanek (commented):
@rajivnathan Looks good - great job. :+1: :1st_place_medal: 
I have just two minor comments.

PR #282, opened 2020-09-14
Refactor user signup

MatousJobanek:
This PR slightly refactors the UserSignup controller - the code was long with long methods and hard to maintain and read.
I moved:
* predicate related code into `predicate.go` (the related tests were moved into `predicate_test.go`)
* mapper related code into `mapper.go` (the related tests were moved into `mapper_test.go`)
* functions taking care of status update into `status.go`. There was no test that would verify the status update functionality, but I added a new one that verifies the method `wrapErrorWithStatusUpdate` into `status_test.go`

All above is just copy-paste operations

Then I split the `Reconcile` method into separated methods. Newly, none one of the new methods returns the `reconcile.Result{}` - they return only values related to their functionality.

Newly, when the UserSignup is invalid (eg. missing label or annotation) then the controller returns an error. Previously, the information was only set in status but no error was returned.

I also removed the `SignupError` type and replaced the usage with a string error because there was no real usage of the type.

### DIFF
--- a/pkg/controller/usersignup/usersignup_controller.go
+++ b/pkg/controller/usersignup/usersignup_controller.go
@@ -92,20 +92,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 var _ reconcile.Reconciler = &ReconcileUserSignup{}
 
-// NewSignupError returns a new Signup error
-func NewSignupError(msg string) SignupError {
-	return SignupError{message: msg}
-}
-
-// SignupError an error that occurs during user signup
-type SignupError struct {
-	message string
-}
-
-func (err SignupError) Error() string {
-	return err.message
-}
-
 // ReconcileUserSignup reconciles a UserSignup object
 type ReconcileUserSignup struct {
 	*statusUpdater
@@ -219,7 +205,7 @@ func (r *ReconcileUserSignup) ensureMurIfAlreadyExists(reqLogger logr.Logger, us
 	murs := murList.Items
 	// If we found more than one MasterUserRecord, then die
 	if len(murs) > 1 {
-		err := NewSignupError("multiple matching MasterUserRecord resources found")
+		err := fmt.Errorf("multiple matching MasterUserRecord resources found")
 		return false, r.wrapErrorWithStatusUpdate(reqLogger, userSignup, r.setStatusInvalidMURState, err, "Multiple MasterUserRecords found")
 	} else if len(murs) == 1 {
 		mur := &murs[0]
@@ -300,8 +286,7 @@ func (r *ReconcileUserSignup) ensureNewMurIfApproved(reqLogger logr.Logger, user
 					return statusError
 				}
 
-				err = NewSignupError("no target clusters available")
-				return err
+				return fmt.Errorf("no target clusters available")
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


alexeykazakov (approved):

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -247,136 +176,135 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
 
 			hashIsValid := validateEmailHash(emailLbl, emailHashLbl)
 			if !hashIsValid {
-				return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusInvalidEmailHash)
+				err := fmt.Errorf("hash is invalid")
```
xcoulon, line 179: I wonder why we check if the email hash is valid at this point and not before?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -247,136 +176,135 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+	if err := r.client.List(context.TODO(), murList, opts); err != nil {
+		return false, r.wrapErrorWithStatusUpdate(reqLogger, userSignup, r.setStatusInvalidMURState, err, "Failed to list MasterUserRecords")
 	}
 
 	murs := murList.Items
 	// If we found more than one MasterUserRecord, then die
 	if len(murs) > 1 {
-		err = NewSignupError("multiple matching MasterUserRecord resources found")
-		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, instance, r.setStatusInvalidMURState, err, "Multiple MasterUserRecords found")
+		err := fmt.Errorf("multiple matching MasterUserRecord resources found")
```
xcoulon, line 208: can't we just use the `usersignup.Status.CompliantUsername` to look-up the `MasterUserRecord` (by name)?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -247,136 +176,135 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
-		reqLogger.Info("missing annotation on usersignup", "annotation", toolchainv1alpha1.UserSignupUserEmailAnnotationKey)
-		return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusInvalidMissingUserEmailAnnotation)
+		err := fmt.Errorf("missing annotation at usersignup")
+		return banned, r.wrapErrorWithStatusUpdate(reqLogger, userSignup, r.setStatusInvalidMissingUserEmailAnnotation, err,
+			"the required annotation '%s' is not present", toolchainv1alpha1.UserSignupUserEmailAnnotationKey)
 	}
+	return banned, nil
+}
 
+func (r *ReconcileUserSignup) ensureMurIfAlreadyExists(reqLogger logr.Logger, userSignup *toolchainv1alpha1.UserSignup, banned bool) (bool, error) {
```
xcoulon, line 196: could you add a comment to explain what the returned `bool` stands for?

xcoulon (commented):
great work! code is much cleaner 🙌 
Just a few comments and also questions but not due to your refactoring (so can be addressed later)

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: no clue :man_shrugging:  that's maybe a question for @sbryzak

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: no, since we don't always know which compliantUsername was used for creation - don't forget that we add a number as a suffix when the username is already taken.

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: added here ab81d26

PR #284, opened 2020-09-16
Add missing config params to reg service deployment

sbryzak:
This PR adds support for a few missing configuration parameters for phone verification in the registration service deployment.

Fixes https://issues.redhat.com/browse/CRT-811

### DIFF
--- a/deploy/env/e2e-tests.yaml
+++ b/deploy/env/e2e-tests.yaml
@@ -1,6 +1,13 @@
 registration-service:
   environment: 'e2e-tests'
   replicas: 2
+  verification:
+    enabled: 'true'
+    excluded-email-domains: 'redhat.com,acme.com'
+  twilio:
+    account-sid: ''
+    auth-token: ''
+    from-number: ''
 host-operator:
   duration-before-change-request-deletion: '5s'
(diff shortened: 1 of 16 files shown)


alexeykazakov (approved):

tinakurian (approved):

deploy/env/e2e-tests.yaml
```diff
@@ -1,6 +1,13 @@
 registration-service:
   environment: 'e2e-tests'
   replicas: 2
+  verification:
+    enabled: 'true'
+    excluded-email-domains: 'redhat.com,acme.com'
+  twilio:
+    account-sid: ''
```
tinakurian, line 8: do we need to add these here if the default is empty?

tinakurian (changes requested):
i think we shouldn't add the twilio fields here if they are just set to empty? empty can be the default and we currently don't specify fields in the yamls that contain defaults because configuration will take care of that

deploy/env/e2e-tests.yaml
sbryzak: I didn't intend to add these changes yet.. I've commented them out until the e2e tests are running successfully.

PR #288, opened 2020-09-22
feat: automatic approval with capacity management

MatousJobanek:
Introduces automatic approval connected to capacity management.
The logic checks if the user can be approved and provisioned to any member cluster. It the user can be approved then it picks a suitable member cluster the user will be provisioned to.
If the user is approved manually then if the target cluster is not already specified for UserSignup then it just gets a member cluster with enough capacity.
If the user is not approved manually, then it loads HostOperatorConfig to check if automatic approval is enabled or not. If it is then it checks capacity thresholds and the actual use to check if there is any suitable member cluster available. 
If there is no suitable member cluster (but the user can be approved), then it returns an error and waits for the next reconcile.

Builds on top of https://github.com/codeready-toolchain/host-operator/pull/290 so the only commit to be reviewed is: 47aee03

TODO: 
- [x] e2e tests

Postponed for separated PR:
when automatic approval is enabled then it should approve and provision the oldest unapproved UserSignup.

### DIFF
--- a/pkg/controller/usersignup/approval.go
+++ b/pkg/controller/usersignup/approval.go
@@ -1,0 +1,120 @@
+package usersignup
+
+import (
+	"context"
+	"fmt"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	crtCfg "github.com/codeready-toolchain/host-operator/pkg/configuration"
+	"github.com/codeready-toolchain/host-operator/pkg/counter"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// getClusterIfApproved checks if the user can be approved and provisioned to any member cluster.
+// It the user can be approved then the function returns true as the first returned value, the second value contains a cluster name the user should be provisioned to.
+// If there is no suitable member cluster, then it returns empty string.
+//
+// If the user is approved manually then it tries to get member cluster with enough capacity if the target cluster is not already specified for UserSignup.
+// If the user is not approved manually, then it loads HostOperatorConfig to check if automatic approval is enabled or not. If it is then it checks
+// capacity thresholds and the actual use if there is any suitable member cluster.
+func getClusterIfApproved(cl client.Client, crtConfig *crtCfg.Config, userSignup *toolchainv1alpha1.UserSignup, getMemberClusters cluster.GetMemberClustersFunc) (bool, string, error) {
+	config := &toolchainv1alpha1.HostOperatorConfig{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: userSignup.Namespace, Name: "config"}, config); err != nil {
+		if apierrors.IsNotFound(err) {
+			if userSignup.Spec.Approved {
+				return true, getOptimalTargetCluster(userSignup, getMemberClusters), nil
+			}
+			return false, "", nil
+		}
+		return false, "", errors.Wrapf(err, "unable to read HostOperatorConfig resource")
+	}
+	if !userSignup.Spec.Approved && !config.Spec.AutomaticApproval.Enabled {
+		return false, "", nil
+	}
+
+	status := &toolchainv1alpha1.ToolchainStatus{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: userSignup.Namespace, Name: crtConfig.GetToolchainStatusName()}, status); err != nil {
+		return false, "", errors.Wrapf(err, "unable to read ToolchainStatus resource")
+	}
+	counts, err := counter.GetCounts()
+	if err != nil {
+		return false, "", errors.Wrapf(err, "unable to get the number of provisioned users")
+	}
+
+	targetCluster := getOptimalTargetCluster(userSignup, getMemberClusters, hasNotReachedMaxNumberOfUsersThreshold(config, counts), hasEnoughResources(config, status))
+	if targetCluster == "" {
+		return false, "", fmt.Errorf("no suitable member cluster found")
+	}
+
+	// approved manually
+	if userSignup.Spec.Approved {
+		return true, targetCluster, nil
+	}
+
+	if config.Spec.AutomaticApproval.MaxNumberOfUsers.Overall != 0 {
+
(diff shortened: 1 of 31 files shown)


pkg/controller/usersignup/approval.go
```diff
@@ -0,0 +1,120 @@
...
+// getClusterIfApproved checks if the user can be approved and provisioned to any member cluster.
+// It the user can be approved then the function returns true as the first returned value, the second value contains a cluster name the user should be provisioned to.
+// If there is no suitable member cluster, then it returns empty string.
+//
+// If the user is approved manually then it tries to get member cluster with enough capacity if the target cluster is not already specified for UserSignup.
+// If the user is not approved manually, then it loads HostOperatorConfig to check if automatic approval is enabled or not. If it is then it checks
+// capacity thresholds and the actual use if there is any suitable member cluster.
+func getClusterIfApproved(cl client.Client, crtConfig *crtCfg.Config, userSignup *toolchainv1alpha1.UserSignup, getMemberClusters cluster.GetMemberClustersFunc) (bool, string, error) {
+	config := &toolchainv1alpha1.HostOperatorConfig{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: userSignup.Namespace, Name: "config"}, config); err != nil {
```
alexeykazakov, line 26: Well.. I guess we don't want to explicitly load the host operator config in every place where we need the configuration.
Can we abstract access to the config?
Something like `config.HostOperatorConfig()` function somewhere. And reuse this function everywhere we need access to the config. The function for now could load the corresponding "config" resource but later we can convert it to a controller which will watch the "config" resource and update the internal cache if it's changed. Not sure if we really need such cache though. Maybe the go client cache is already good enough. But we can think about that cache later.

alexeykazakov (approved):
Overall looks good!
But I have one suggestion regarding the config.

pkg/controller/usersignup/approval.go
MatousJobanek: Yes yes yes, I had the very same concern when I was writing the code. 
Actually, I started working on a controller that does exactly the same what you proposed with the cache :smile:, I just want to submit it as a separated PR and when this automatic approval is done. See this: https://github.com/MatousJobanek/host-operator/tree/hostoperatorconfig-controller/pkg/controller/hostoperatorconfig
For now, I decided to load it here and then replace it with calling the cache.

alexeykazakov: OK. Makes sense. Let's keep it for a separate PR then!

pkg/controller/usersignup/approval.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/codeready-toolchain/host-operator/pkg/counter"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// getClusterIfApproved checks if the user can be approved and provisioned to any member cluster.
+// It the user can be approved then the function returns true as the first returned value, the second value contains a cluster name the user should be provisioned to.
```
tinakurian, line 18:
```suggestion
// If the user can be approved then the function returns true as the first returned value, the second value contains a cluster name the user should be provisioned to.
```

pkg/controller/usersignup/approval.go
```diff
@@ -0,0 +1,120 @@
...
+	return func(cluster *cluster.CachedToolchainCluster) bool {
+		threshold, found := config.Spec.AutomaticApproval.ResourceCapacityThreshold.SpecificPerMemberCluster[cluster.Name]
+		if !found {
+			threshold = config.Spec.AutomaticApproval.ResourceCapacityThreshold.DefaultThreshold
+		}
+		if threshold == 0 {
+			return true
+		}
+		for _, memberStatus := range status.Status.Members {
+			if memberStatus.ClusterName == cluster.Name {
```
tinakurian, line 91: could we refactor this to avoid some of the nesting by any chance?

pkg/controller/usersignup/approval.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		return false
+	}
+}
+
+func getOptimalTargetCluster(userSignup *toolchainv1alpha1.UserSignup, getMemberClusters cluster.GetMemberClustersFunc, conditions ...cluster.Condition) string {
+	// If a target cluster hasn't been selected, select one from the members
+	if userSignup.Spec.TargetCluster != "" {
+		return userSignup.Spec.TargetCluster
+	} else {
```
tinakurian, line 111: this else clause can probably be removed?

MatousJobanek: good catch 7c8e20e

pkg/controller/usersignup/approval.go
MatousJobanek: hmmm... do you have any proposal?

rajivnathan: How about if we combine the first two if statements?

```suggestion
			if memberStatus.ClusterName == cluster.Name && len(memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole) > 0 {
```

And then move the following to a separate function named `clusterHasEnoughResources` or something?
```
for _, usagePerNode := range memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole {
	if usagePerNode >= threshold {
		return false
	}
}
return true
```

I guess that would help with readability?

rajivnathan (approved):
Nice, really like the use of cluster condition functions and status creator functions. 👍 

Added a suggestion for reducing the nesting in the `hasEnoughResources` function. Not sure if you think it makes it more readable so feel free to ignore if you disagree.

pkg/controller/usersignup/approval.go
MatousJobanek: > How about if we combine the first two if statements?

I technically could, but it would go through all member statuses no matter if the correct one was already found or not (for the case when the memory usage is `0`). I know that we won't probably have many of member clusters with such number, but I would like to stay with the proper solution if you don't mind.
In addition, I'm not a fan of such a long if statements - for me it's harder to read.

If it helps, I can separate the inner part - see: 8462fab
is it better @tinakurian @rajivnathan

rajivnathan: Looks good for me. Thanks for adjusting it Matous!

tinakurian: looks good @MatousJobanek!

tinakurian (approved):
looks great @MatousJobanek thanks for addressing my comments!

PR #289, opened 2020-09-22
Deactivation notification

rajivnathan:
This PR adds code to send an email to the user's account when their account has been deactivated. This applies to both when their account is manually or automatically deactivated.

Related issue: https://issues.redhat.com/browse/CRT-644

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/165


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -813,6 +819,33 @@ func (r *ReconcileUserSignup) updateStatusConditions(userSignup *toolchainv1alph
...
+	}
+
+	err := controllerutil.SetControllerReference(userSignup, notification, r.scheme)
+	if err != nil {
+		logger.Error(err, "Failed to set owner reference for deactivation notification")
+		return err
+	}
+
+	if err := r.client.Create(context.TODO(), notification); err != nil {
+		logger.Error(err, "Failed create deactivation notification resource")
```
alexeykazakov, line 841: Any tests for this error?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -324,7 +326,11 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
 	// If there is no MasterUserRecord created, yet the UserSignup is marked as Deactivated, simply set the status
 	// and return
 	if instance.Spec.Deactivated {
-		return reconcile.Result{}, r.updateStatus(reqLogger, instance, r.setStatusDeactivated)
+		err := r.updateStatus(reqLogger, instance, r.setStatusDeactivated)
+		if err == nil {
+			err = r.sendDeactivatedNotification(reqLogger, instance)
```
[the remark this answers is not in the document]
rajivnathan: Oh right, forgot about that. Will add it.

pkg/controller/usersignup/usersignup_controller.go
rajivnathan: Included in https://github.com/codeready-toolchain/host-operator/pull/289/commits/ba8ad1c44e6ef9e9e86ac3c84d5da84be6ef3300

pkg/controller/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
rajivnathan: Included in https://github.com/codeready-toolchain/host-operator/pull/289/commits/ba8ad1c44e6ef9e9e86ac3c84d5da84be6ef3300

rajivnathan:
@alexeykazakov I updated the controller to set a new UserSignup status called `UserDeactivatedNotificationCreated` that has 3 states: `NotificationCRCreated`, `NotificationCRNotCreated` and `NotificationCRCreationFailed`. 
`NotificationCRCreated` is set when UserSignup.Spec.Deactivated = `true` and the notification resource was successfully created
`NotificationCRCreationFailed` is set when UserSignup.Spec.Deactivated = `true` and we failed to create the notification resource
`NotificationCRNotCreated` is set when UserSignup.Spec.Deactivated = `false`

I also added some tests for the reactivating a UserSignup after deactivation scenario.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -128,6 +125,13 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
 
+	// If the usersignup is deactivated then ensure the deactivated notification status is set to false, this is especially important for cases when a user is deactivated and then reactivated
+	if !instance.Spec.Deactivated {
+		if err := r.updateStatus(reqLogger, instance, r.setStatusDeactivationNotificationNotCreated); err != nil {
```
alexeykazakov, line 130: Hm.. yeah. I didn't think about that but we do need to cover that case indeed. Good catch.
The only thing is that I would probably use a different reason. The CR is not created because it's not created doesn't sound super informative :)
I would use "UserIsActive" or "UserIsNotDeactivated" as the reason for the "NotificationCreated/False" condition :)

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1332,6 +1423,212 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
+			},
+			v1alpha1.Condition{
+				Type:    v1alpha1.UserSignupUserDeactivatedNotificationCreated,
+				Status:  v1.ConditionFalse,
+				Reason:  "NotificationCRCreationFailed",
+				Message: "unable to create deactivation notification",
+			})
+		AssertThatCounterHas(t, 2)
+
+		// A deactivated notification should have been created
```
alexeykazakov, line 1500: It should or it shouldn't? You assert that it's not created.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1332,6 +1423,212 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
+				Status: v1.ConditionTrue,
+			},
+			v1alpha1.Condition{
+				Type:   v1alpha1.UserSignupUserDeactivatedNotificationCreated,
+				Status: v1.ConditionFalse,
+				Reason: "NotificationCRNotCreated",
+			})
+		AssertThatCounterHas(t, 2)
+
+		// A deactivation notification should not be created because this is the reactivation case
```
alexeykazakov, line 1626: I'm a bit lost what you are actually testing here?
Is it a test for covering status update case? Then what does the reactivation have with it? I don't even see that the user is reactivated in the test.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1332,6 +1423,212 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
+			},
+			v1alpha1.Condition{
+				Type:   v1alpha1.UserSignupUserDeactivatedNotificationCreated,
+				Status: v1.ConditionFalse,
+				Reason: "NotificationCRNotCreated",
+			})
+		AssertThatCounterHas(t, 3)
+
+		// There should not be a notification created because the user was reactivated
+		notifications := &v1alpha1.NotificationList{}
```
alexeykazakov, line 1576: Same here. It doesn't seem to be testing what you state. We do need to test the reactivation flow but it seems to be missing.

pkg/controller/usersignup/usersignup_controller.go
rajivnathan: Haha, yeah I guess it's not super informative. I changed it to `UserIsActive`

included in https://github.com/codeready-toolchain/host-operator/pull/289/commits/6bf0e589a480fce21f3bd1e4c69414ff0c73b23f

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: Oops, should be "shouldn't". Fixed in https://github.com/codeready-toolchain/host-operator/pull/289/commits/448e505672962faf3eda38deaa43deb1de12d8c7

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: When a usersignup is deactivated (`Deactivated: true`) then it should get status condition `NotificationCRCreated`. So the test starts with a usersignup has `Deactivated: false` but it has condition `NotificationCRCreated` so this is the scenario where a user was deactivated and was just reactivated. Then the reconcile should update the status to `UserIsActive`. I added comments and initialized the status conditions in the individual test functions to make it a bit more clear. Does it help? Any other suggestions?

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: Same as above comment

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1315,8 +1407,251 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
+		Status: v1alpha1.UserSignupStatus{
+			CompliantUsername: "john-doe",
+		},
+	}
+	userSignup.Labels["toolchain.dev.openshift.com/approved"] = "true"
+	key := test.NamespacedName(operatorNamespace, userSignup.Name)
+
+	t.Run("when reactivating the usersignup successfully", func(t *testing.T) {
+		// given
+		// start with a usersignup that has the Notification Created status which signals an active user
```
alexeykazakov, line 1525:
```suggestion
		// start with a usersignup that has the Notification Created status set to "true" but Spec.Deactivated is set to "false" which signals a user which has been just reactivated.
```
Right?

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1315,8 +1407,251 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
+		// There should not be a notification created because the user was reactivated
+		notifications := &v1alpha1.NotificationList{}
+		err = r.client.List(context.TODO(), notifications)
+		require.NoError(t, err)
+		require.Len(t, notifications.Items, 0)
+	})
+
+	t.Run("when resetting the usersignup deactivation notification status fails", func(t *testing.T) {
+		// given
+		// start with a usersignup that has the Notification Created status which signals an active user
```
alexeykazakov, line 1589:
```suggestion
		// start with a usersignup that has the Notification Created status set to "true" but Spec.Deactivated is set to "false" which signals a user which has been just reactivated.
```

alexeykazakov (approved):
Looks good! Just a couple of suggestions to improve code readability.

pkg/controller/usersignup/usersignup_controller_test.go
alexeykazakov: OK. Makes sense. The comments helped but I think we can make them even clearer :) See my suggestions.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1384,6 +1704,39 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+			v1alpha1.Condition{
+				Type:   v1alpha1.UserSignupComplete,
+				Status: v1.ConditionTrue,
+				Reason: "Deactivated",
+			},
+			v1alpha1.Condition{
+				Type:   v1alpha1.UserSignupUserDeactivatedNotificationCreated,
+				Status: v1.ConditionTrue,
+				Reason: "NotificationCRCreated",
+			})
```
MatousJobanek, line 1739: Could you please move this into a subtest? This verifies another functionality based on the previous test - it's much easier to read when it's used as a subtest.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1679,6 +2037,14 @@ func TestUserSignupDeactivatedButMURDeleteFails(t *testing.T) {
...
 		})
 	AssertThatCounterHas(t, 1)
+
+	// There should not be a notification created since the mur deletion failed even if reconciled again
+	_, err = r.Reconcile(req)
+	require.Error(t, err)
+	notifications := &v1alpha1.NotificationList{}
+	err = r.client.List(context.TODO(), notifications)
+	require.NoError(t, err)
+	require.Len(t, notifications.Items, 0)
```
MatousJobanek, line 2047: I believe that these kinds of checks could be extracted to a separated function - it makes the code more compact and shorter.

MatousJobanek (approved):
Good job. I really like the solution where the deactivation notification is managed by UserSignup controller based on the fact that the MUR was deleted :+1:

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: Sure, makes sense!

rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/289/commits/a546d2527c71eb6962eb1428bc6ec22e8f6ceca9

pkg/controller/usersignup/usersignup_controller_test.go
rajivnathan: Good suggestion, created a new function in the test/notification package for that. Done in https://github.com/codeready-toolchain/host-operator/pull/289/commits/c53daf95d071e6c1317f7f9e1ad78d0fd163e927

PR #291, opened 2020-09-25
Fixes for phone verification e2e tests

sbryzak:
This PR fixes a few minor issues that were blocking the phone verification e2e tests from passing.

### DIFF
--- a/pkg/controller/toolchainstatus/toolchainstatus_controller.go
+++ b/pkg/controller/toolchainstatus/toolchainstatus_controller.go
@@ -415,7 +415,7 @@ func (s regServiceSubstatusHandler) addRegistrationServiceDeploymentStatus(reqLo
 // addRegistrationServiceHealthStatus handles the RegistrationService.Health part of the toolchainstatus
 func (s regServiceSubstatusHandler) addRegistrationServiceHealthStatus(reqLogger logr.Logger, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
 	// get the JSON payload from the health endpoint
-	registrationServiceHealthURL := "http://registration-service/" + registrationServiceHealthEndpoint
+	registrationServiceHealthURL := "http://registration-service" + registrationServiceHealthEndpoint
 	resp, err := s.httpClientImpl.Get(registrationServiceHealthURL)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -415,7 +415,7 @@ func (s regServiceSubstatusHandler) addRegistrationServiceDeploymentStatus(reqLo
 // addRegistrationServiceHealthStatus handles the RegistrationService.Health part of the toolchainstatus
 func (s regServiceSubstatusHandler) addRegistrationServiceHealthStatus(reqLogger logr.Logger, toolchainStatus *toolchainv1alpha1.ToolchainStatus) bool {
 	// get the JSON payload from the health endpoint
-	registrationServiceHealthURL := "http://registration-service/" + registrationServiceHealthEndpoint
+	registrationServiceHealthURL := "http://registration-service" + registrationServiceHealthEndpoint
```
MatousJobanek, line 418: Just don't forget to fix also the test https://github.com/codeready-toolchain/host-operator/blob/b42cfc858d73ce013962693828472c3147987dcf/pkg/controller/toolchainstatus/toolchainstatus_controller_test.go#L354

MatousJobanek (approved):

MatousJobanek:
@sbryzak btw, I've just noticed that you are trying to pair e2e PR with both host-operator PR and registration-service PR - that's not allowed - you should pair one e2e PR with only one operator PR.

sbryzak:
@MatousJobanek you're right, however this PR for host-operator is basically a "flip the switch" change that enables the phone verification feature, because unfortunately the configuration for the e2e tests is stored in this repo and not in toolchain-e2e.  If I committed this PR first, then that would break all of the existing tests due to the now active verification requirement.  

Ideally there would have been just this tiny PR plus the new tests in toolchain-e2e, however in the process of writing the tests I uncovered numerous bugs in registration service which required fixing for the tests to pass also.  Perhaps the better approach might have been to simply enable the verification feature first in conjunction with fixing the existing tests as a separate set of PRs, followed by the addition of the new verification e2e tests and reg service fixes in a separate set of PRs.  I guess it's not too late to go this route actually, it would just require a rebase after making the changes.

PR #293, opened 2020-09-30
Enable phone verification in e2e tests

sbryzak:
This PR simply enables the phone verification feature for e2e tests.

This is part of the changes required for https://issues.redhat.com/browse/CRT-792.

### DIFF
--- a/deploy/env/e2e-tests.yaml
+++ b/deploy/env/e2e-tests.yaml
@@ -2,13 +2,10 @@ registration-service:
   environment: 'e2e-tests'
   replicas: 2
   verification:
-    #enabled: 'true'
-    enabled: 'false'
+    enabled: 'true'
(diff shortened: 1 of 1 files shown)


deploy/env/e2e-tests.yaml
```diff
@@ -2,13 +2,10 @@ registration-service:
   environment: 'e2e-tests'
   replicas: 2
   verification:
-    #enabled: 'true'
-    enabled: 'false'
+    enabled: 'true'
     excluded-email-domains: 'redhat.com,acme.com'
```
alexeykazakov, line 6: I wonder if we use something more obvious for the excluded email domains in our tests. Something like:
```suggestion
    excluded-email-domains: 'redhat-exclude-from-verification.com,exclude.com'
```
WDYT?

alexeykazakov (approved):

PR #286, opened 2020-09-17
feat: expose user signups metrics

xcoulon:
- WIP - custom usersignup metrics
- feat: expose user signups metrics

PR #296, opened 2020-10-07
Gather and expose some custom metrics

rajivnathan:
This PR adds to https://github.com/codeready-toolchain/host-operator/pull/286 which exposes the metrics endpoint.

## Metrics 
Metric | Type | Description 
--- | --- | ---
user_signups_total | Counter | Total number of unique User Signups
user_signups_provisioned_total | Counter | Total number of Provisioned User Signups
user_signups_banned_total | Counter | Total number of Banned User Signups
user_signups_deactivated_total | Counter | Total number of Deactivated User Signups
user_signups_auto_deactivated_total | Counter | Total number of Automatically Deactivated User Signups
master_user_record_current | Gauge | Current number of Master User Records

### Scenario

1. single usersignup created
1. deactivate
1. reactivated
1. deactivated again
1. reactivated again

### Metrics snippets from endpoint
```
# HELP user_signups_auto_deactivated_total Total number of Automatically Deactivated User Signups
# TYPE user_signups_auto_deactivated_total counter
user_signups_auto_deactivated_total 0
# HELP user_signups_banned_total Total number of Banned User Signups
# TYPE user_signups_banned_total counter
user_signups_banned_total 0
# HELP user_signups_deactivated_total Total number of Deactivated User Signups
# TYPE user_signups_deactivated_total counter
user_signups_deactivated_total 2
# HELP user_signups_provisioned_total Total number of Provisioned User Signups
# TYPE user_signups_provisioned_total counter
user_signups_provisioned_total 2
# HELP user_signups_total Total number of unique User Signups
# TYPE user_signups_total counter
user_signups_total 1
```

```
# HELP master_user_record_current Current number of Master User Records
# TYPE master_user_record_current gauge
master_user_record_current 1
```

### Work In Progress
- [x] ensure counters aren't incremented after operator restart (using Metrics status)
- [x] ensure provisioned and deactivated counts are still handled correctly after reactivation (should still increment)
- [ ] add metric for user accounts per cluster
- [ ] unit tests
- [ ] e2e tests

### DIFF
--- a/deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
+++ b/deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
@@ -160,7 +160,7 @@ spec:
                 description: Metrics is used to track metrics-related data of this
                   usersignup
                 properties:
-                  metricKeyValues:
+                  metricsRegistry:
                     additionalProperties:
                       type: string
                     description: MetricsRegistry uses metric names as keys. The values
--- a/deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
+++ b/deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
@@ -156,6 +156,19 @@ spec:
                 x-kubernetes-list-map-keys:
                 - type
                 x-kubernetes-list-type: map
+              metrics:
+                description: Metrics is used to track metrics-related data of this
+                  usersignup
+                properties:
+                  metricKeyValues:
+                    additionalProperties:
+                      type: string
+                    description: MetricsRegistry uses metric names as keys. The values
+                      indicate whether this resource has been counted for the associated
+                      metric.
+                    type: object
+                    x-kubernetes-map-type: atomic
+                type: object
             type: object
         type: object
         x-kubernetes-preserve-unknown-fields: true
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/codeready-toolchain/host-operator/pkg/controller"
 	"github.com/codeready-toolchain/host-operator/pkg/controller/registrationservice"
 	"github.com/codeready-toolchain/host-operator/pkg/controller/toolchainstatus"
+	"github.com/codeready-toolchain/host-operator/pkg/metrics"
 	"github.com/codeready-toolchain/host-operator/pkg/templates/assets"
 	"github.com/codeready-toolchain/host-operator/pkg/templates/nstemplatetiers"
 	"github.com/codeready-toolchain/host-operator/version"
@@ -23,7 +24,7 @@ import (
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
+	sdkmetrics "github.com/operator-framework/operator-sdk/pkg/metrics"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
@@ -147,6 +148,8 @@ func main() {
 	// Add the Metrics Service
 	addMetrics(ctx, cfg)
 
+	metrics.RegisterCustomMetrics()
+	defer metrics.UnregisterCustomMetrics()
 	stopChannel := signals.SetupSignalHandler()
 
 	go func() {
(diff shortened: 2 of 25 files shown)


deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
```diff
@@ -156,6 +156,19 @@ spec:
...
+                properties:
+                  metricKeyValues:
+                    additionalProperties:
+                      type: string
+                    description: MetricsRegistry uses metric names as keys. The values
+                      indicate whether this resource has been counted for the associated
+                      metric.
+                    type: object
+                    x-kubernetes-map-type: atomic
+                type: object
```
MatousJobanek: I thought that we agreed on using the `state` label for indicating that the UserSignup was counted. Why do you introduce the `metric` in the status? I there anything you are missing when using the `state` label?

MatousJobanek (commented):
I like these metrics, but I'm curious why do you want to use the `metric` part of the `status` section. If I'm not mistaken, we can easily use the `state` label - see this PR: https://github.com/codeready-toolchain/host-operator/pull/298 - if we slightly change the logic, we will achieve the very same functionality but with less code and without introducing addition status part.

deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
rajivnathan: Yes, I initially was going to do that but the issue is to change a label/annotation we need to update the UserSignup resource but in most cases I'm doing this from the usersignup controller so that breaks the pattern of only modifying secondary resources. It's probably not such a big deal because it's only label/annotation but it also causes additional reconciles. By using the status we can avoid breaking the pattern and the additional reconciles so it sounded like a good idea.

Also, I didn't think we wanted to list resources by these metrics labels but rethinking about that it could actually be useful because we can then list all resources that have the `user_signups_deactivated_total ` label to get the current list of users that were deactivated and things like that. What do you think?

cc @alexeykazakov

alexeykazakov: Right, @rajivnathan.
Pros for using status:
- No need to modify the primary resource from the controller (the controller modifies the status sub-resource only). Which leads to the second pro:
- No additional reconciles when the status is changed (annotation update would cause one extra reconcile).
Cons:
- Impossible to use it as a selector. But if we don't need it then it's fine.

MatousJobanek: > No need to modify the primary resource from the controller (the controller modifies the status sub-resource only). Which leads to the second pro:

But we are going to change the `toolchain.dev.openshift.com/state` label from the UserSignup controller anyway - you saw my PR right?

> No additional reconciles when the status is changed (annotation update would cause one extra reconcile).

We use the `UserSignupChangedPredicate` as a predicate for `UserSignup` kind which triggers reconcile loops only when the generation has been changed. IIRC the `generation` is a sequence number representing changes in the desired state - ie. in the `spec`. That means that the number is not incremented for the changes in labels, which means: no additional reconcile loops.

> Also, I didn't think we wanted to list resources by these metrics labels but rethinking about that it could actually be useful because we can then list all resources that have the user_signups_deactivated_total label to get the current list of users that were deactivated and things like that.

You probably misunderstood what I meant - I didn't want to introduce new labels for the metrics. I was proposing using the `toolchain.dev.openshift.com/state` label that I added here: https://github.com/codeready-toolchain/toolchain-e2e/pull/175
The logic that sets the label values is here: https://github.com/codeready-toolchain/host-operator/pull/298 Now, look at the implementation - you could very easily sneak in and do what you are doing here by just checking the label values:
* if the label hasn't been set at all, and we are going to set the `not-ready` value => increment `user_signups_total`
* if we are going to set `approved` value and it wasn't set before => increment `user_signups_provisioned_total`
* if we are going to set `banned` value and it wasn't set before => increment `user_signups_banned_total`
* if we are going to set `deactivated` value and it wasn't set before => increment `user_signups_deactivated_total`
* if we are going to change the `spec.deactivated` value from the deactivation controller => increment `user_signups_auto_deactivated_total`

For the future
* if we are going to set `pending` value and it wasn't set before => increment `user_signups_on_waitinglist`
* if we are going to change label from `pending` to `approved` => decrement `user_signups_on_waitinglist`

rajivnathan: Ok, yes I did misunderstood what you meant 🙂  and you are right about the generation field only being updated for changes to the `spec`, see https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/customresources-subresources.md#status-behavior so that's less of a concern.

With this method we can also determine the user reactivation scenario as well because he mentioned it could be handled with the use of the state label in https://github.com/codeready-toolchain/host-operator/pull/298 where in the `setStateLabel` function we can check the old value against the new value and then if the old value was deactivated and the new value is any of `not-ready`, `pending` or `approved` then we know that is a reactivation. I'm ok to change this PR (or create a new PR) to use this method.

MatousJobanek: > I'm ok to change this PR (or create a new PR) to use this method.

Do you mean to merge this PR and then create a new one changing it, or just close this one and create a brand-new one...? I'm asking because I really don't want to merge this one since it introduces a new status part that would be dropped in the next PR anyway. So, I would suggest changing this one.

rajivnathan: Wasn't suggesting to merge this one as is. I meant update this PR or close this PR and create a new one since I don't think I'd reuse much from this PR it's probably faster to just start with a fresh PR.

MatousJobanek: ok, then it's up to you - do whatever is easier for you ;-)

rajivnathan:
new PR: https://github.com/codeready-toolchain/host-operator/pull/309

PR #299, opened 2020-10-09
[CRT-836] Fix registration link in notification email

rajivnathan:
The link to the registration URL is broken in the notification emails to users. Opening an html file with the output from this line works https://github.com/codeready-toolchain/host-operator/blob/2f6affe669a9f8efed9e425d486cd68bd98675a4/pkg/controller/notification/mailgun_notification_delivery_service.go#L80 using firefox and chrome shows the link fine but in the email it cannot be clicked.

In the email that doesn't work the href value is not quoted but in other emails that do have working links they all have them quoted, so although quotes may not be strictly needed let's see if it makes a difference.


alexeykazakov (approved):
I guess it's too tricky to test it how it will look in a email? I wonder if we could use some test mailgun account for testing it localy...

xcoulon (approved):

MatousJobanek (approved):

MatousJobanek:
> I guess it's too tricky to test it how it will look in a email? I wonder if we could use some test mailgun account for testing it localy...

for local testing, it would be better to change the code and at least verify the generation of the email content - we are not doing it right now. But it wouldn't catch this mistake anyway. Any HTML parser wouldn't help either I'm afraid - we will just need to pay attention to it in the future and maybe test it manually in staging/prod with some dummy user-account.

tinakurian (approved):

PR #301, opened 2020-10-13
feat: introduce HostOperatorConfig controller & cache

MatousJobanek:
This PR adds the first version of `HostOperatorConfig` controller and related cache. The controller stores(updates) the `HostOperatorConfig` resource in the cache for every "reconcile loop".
The cache then provides a function `GetConfig` that returns the cached instance of the `HostOperatorConfig` resource. If no instance is stored in the cache, then it retrieves it from the cluster (using the provided client and namespace) and stores in the cache. If the resource is not found, then returns the default config (empty for now).

In the future, the cache/controller will have additional logic for proper loading default values for params that are not specified in the config.


pkg/controller/hostoperatorconfig/cache.go
```diff
@@ -0,0 +1,72 @@
...
+
+func updateConfig(config *v1alpha1.HostOperatorConfig) {
+	configCache.set(config)
+}
+
+func loadLatest(cl client.Client, namespace string) error {
+	config := &v1alpha1.HostOperatorConfig{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: "config"}, config); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
```
alexeykazakov, line 40: Maybe worth logging such not found error at least?

pkg/controller/hostoperatorconfig/hostoperatorconfig_controller.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Fetch the HostOperatorConfig instance
+	hostOperatorConfig := &toolchainv1alpha1.HostOperatorConfig{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, hostOperatorConfig)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
```
alexeykazakov, line 71: Worth logging this error I think.

alexeykazakov (approved):
Nice!

pkg/controller/hostoperatorconfig/cache.go
MatousJobanek: added fd64a4e

pkg/controller/hostoperatorconfig/hostoperatorconfig_controller.go
MatousJobanek: added fd64a4e

PR #310, opened 2020-10-23
chore: take the console & che url from ToolchainStatus

MatousJobanek:
This PR drops the direct retrieval of console & che routes from member clusters and takes the URL from the ToolchainStatus. 
It also drops unnecessary configuration parameters that were moved to member operator.

This PR depends on https://github.com/codeready-toolchain/member-operator/pull/213

In the future, we will change the registration-service logic so the URLs are not taken from `MUR.status` but from `ToolchainStatus`. Then we will be able to drop this logic from host-operator completely.


alexeykazakov (approved):
Nice!

tinakurian (approved):

PR #315, opened 2020-11-06
Implement alerts for toolchain status problems

sbryzak:
This PR modifies the notification system to allow ad-hoc notifications to be sent to an admin email address, without requiring a specific template.  Specifically for this PR, a notification is now sent to the configured admin email address when the ToolchainStatus status has been in an "unready" state for an extended period of time.

Fixes https://issues.redhat.com/browse/CRT-829


pkg/controller/notification/user_notification_context.go
```diff
@@ -0,0 +1,64 @@
...
+		FirstName:   instance.Spec.GivenName,
+		LastName:    instance.Spec.FamilyName,
+		CompanyName: instance.Spec.Company,
+	}
+
+	if emailLbl, exists := instance.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]; exists {
+		notificationCtx.UserEmail = emailLbl
+	}
+
+	if config == nil {
```
alexeykazakov, line 51: Shouldn't we check it first before even trying to retrieve the usersignup?

pkg/controller/notification/notification_controller_test.go
```diff
@@ -166,6 +166,50 @@ func TestNotificationDelivery(t *testing.T) {
...
+		require.True(t, iter.First(context.Background(), &events))
+		require.True(t, iter.Last(context.Background(), &events))
+		require.Len(t, events, 1)
+		e := events[0]
+		require.IsType(t, &events2.Accepted{}, e)
+		accepted := e.(*events2.Accepted)
+		require.Equal(t, "sandbox-admin@developers.redhat.com", accepted.Recipient)
+		require.Equal(t, "developers.redhat.com", accepted.RecipientDomain)
+		require.Equal(t, "Alert", accepted.Message.Headers.Subject)
+		require.Equal(t, "noreply@foo.com", accepted.Message.Headers.From)
```
alexeykazakov, line 210: Is it possible to also verify the sent email content?

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -176,6 +184,35 @@ func (r *ReconcileToolchainStatus) aggregateAndUpdateStatus(reqLogger logr.Logge
...
+					}
+
+					if err := r.setStatusToolchainStatusUnreadyNotificationCreated(toolchainStatus); err != nil {
+						reqLogger.Error(err, "Failed to update notification created status")
+						return err
+					}
+				}
+			}
+		}
+
```
alexeykazakov, line 215: This is cosmetic but... we could move it to a separate function to improve code readability. But feel free to keep it as it is if you prefer it this way.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -176,6 +184,35 @@ func (r *ReconcileToolchainStatus) aggregateAndUpdateStatus(reqLogger logr.Logge
...
 	if len(unreadyComponents) > 0 {
+		// If the current ToolchainStatus:
+		// a) Is currently not ready,
+		// b) has not been ready for longer than the configured threshold, and
+		// c) no notification has been already sent, then
+		// send a notification to the admin mailing list
+		c, found := condition.FindConditionByType(toolchainStatus.Status.Conditions, toolchainv1alpha1.ConditionReady)
+		if found && c.Status == corev1.ConditionFalse {
+			threshold := time.Now().Add(-minutesAfterUnready * time.Minute)
+			if c.LastUpdatedTime.Before(&metav1.Time{threshold}) {
```
alexeykazakov, line 195: We wanna check `LastTransitionTime` here. Not `LastUpdatedTime`, do we?
`LastUpdatedTime` is updated every 5 seconds no matter if there are any real updates in the status. It's used as a "ping" timestamp, so we know if it's not updated then there is no running controller to monitor/update the status.
So, if we have our status controller running then you will never hit this 10 second treshshold. 

`LastTransitionTime` is updated only if there is a some real change in the condition (Ready True -> Ready False or Ready False -> Ready True). And this is exactly what we want to use to see if we stuck with Readh/False for more than 10 seconds, right?

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -176,6 +184,35 @@ func (r *ReconcileToolchainStatus) aggregateAndUpdateStatus(reqLogger logr.Logge
...
+				c, found = condition.FindConditionByType(toolchainStatus.Status.Conditions,
+					toolchainv1alpha1.ToolchainStatusUnreadyNotificationCreated)
+				if !found {
```
alexeykazakov, line 198: Unless I'm missing something but this code will allow a single notification only. If it's ever sent once the notification condition is created and if later the status is going to the not-ready state again the notification will be skipped.
We need to reset/delete this condition when the status is ready again.
Also, just checking if the condition exists is not quite correct. Theoretically the condition can be in FALSE state.
So, it will be cleaner if we check that the condition is found and true.
Use https://github.com/codeready-toolchain/toolchain-common/blob/66539924b7b695797f8601631d69f4180ba2691d/pkg/condition/condition.go#L51 `condition.IsTrue()` instead.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -176,6 +184,35 @@ func (r *ReconcileToolchainStatus) aggregateAndUpdateStatus(reqLogger logr.Logge
 
 	// if any components were not ready then set the overall status to not ready
 	if len(unreadyComponents) > 0 {
+		// If the current ToolchainStatus:
```
alexeykazakov, line 187: I don't see any tests for this toolchain status controller notification change. Can we add some please?

pkg/controller/toolchainstatus/toolchainstatus_controller.go
sbryzak: Remember that the `updateStatusConditions` function in this controller totally overwrites all existing conditions - keeping this in mind, when `setStatusReady()` is called it will as a result remove the notification condition:

```func (r *ReconcileToolchainStatus) setStatusReady(toolchainStatus *toolchainv1alpha1.ToolchainStatus) error {
	return r.updateStatusConditions(
		toolchainStatus,
		toolchainv1alpha1.Condition{
			Type:   toolchainv1alpha1.ConditionReady,
			Status: corev1.ConditionTrue,
			Reason: toolchainv1alpha1.ToolchainStatusAllComponentsReadyReason,
		})
}
```

Good point about using `condition.IsTrue()`, I'll switch to using it instead.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
sbryzak: Good catch, fixed.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
sbryzak: Sure thing, moved.

pkg/controller/notification/user_notification_context.go
sbryzak: Yes, you're right.  Fixed.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
alexeykazakov: Yeah I missed that. I would rename `updateStatusConditions()` to `replaceStatusConditions()` then. You are not changing this method here but it should be pretty trivial change.
We use `updateStatusConditions()` in almost every controller and it always updates or adds the given conditions and doesn't touch the existing ones which are not explicitly listed when this method is called. The only exception is our Toolchain (and probably member and host statuses too) Status controller. So, I would rename it to avoid confusions.

pkg/controller/notification/notification_controller_test.go
sbryzak: There doesn't seem to be any way to read that via the mailgun api.

pkg/controller/notification/mailgun_notification_delivery_service.go
```diff
@@ -60,29 +61,37 @@ func NewMailgunNotificationDeliveryService(config NotificationDeliveryServiceFac
...
+		}
+
+		body, err = s.base.GenerateContent(notificationCtx, template.Content)
+		if err != nil {
+			return err
+		}
+	} else {
+		// If there is no template specified then simply use the subject and content provided by the notification
+		subject = notification.Spec.Subject
+		body = notification.Spec.Content
```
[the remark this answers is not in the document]
sbryzak: I think it makes sense to return an error if a template is set also - if there is no subject or content, then it definitely seems to be a problem to me whether a template is set or not.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
sbryzak: Test added

pkg/controller/toolchainstatus/toolchainstatus_controller.go
sbryzak: Agreed, the function name is misleading.  I've renamed it.

alexeykazakov (approved):
Looks good!
We still could improve the toolchain status tests though :)
Right now you test only the case when status is not ready for long enough to generate the notificatoin.
Would be also nice to test the entire flow:
- status is ready: no notification created
- status is not ready for some period shorter then the threshold: no notification created
- status is now not ready for long enough: notification created
- status is ready again: no notification created
- status is not ready again for long enough: new notification created

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -636,6 +582,169 @@ func TestToolchainStatusConditions(t *testing.T) {
...
+			HasMemberStatus(memberClusterSingleReady()).
+			HasRegistrationServiceStatus(registrationServiceReady())
+
+		// Lookup the notification
+		var notification toolchainv1alpha1.Notification
+		err = fakeClient.Get(context.Background(), test.NamespacedName(test.HostOperatorNs, "toolchainstatus-unready"),
+			&notification)
+		require.Error(t, err)
+		require.IsType(t, &errors.StatusError{}, err)
+		require.Equal(t, string(err.(*errors.StatusError).ErrStatus.Reason), "NotFound")
```
alexeykazakov, line 635: It's `Equal(expected, actual)`. Not `Equal(actual, expected)`
Also you can use `k8s.io/apimachinery/pkg/api/errors.IsNotFound(err)`
```suggestion
		require.True(t, errors.IsNotFound(err))
```

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -636,6 +582,169 @@ func TestToolchainStatusConditions(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, requeueResult, res)
+
+		// Lookup the notification
+		var notification toolchainv1alpha1.Notification
+		err = fakeClient.Get(context.Background(), test.NamespacedName(test.HostOperatorNs, "toolchainstatus-unready"),
+			&notification)
+		require.Error(t, err)
+		require.IsType(t, &errors.StatusError{}, err)
+		require.Equal(t, string(err.(*errors.StatusError).ErrStatus.Reason), "NotFound")
```
alexeykazakov, line 683: It's a duplicate assertion from a previous test. Can we move it to `AssertToolchainStatusNotificationNotCreated()` function so we can reuse it everywhere?

PR #321, opened 2020-11-19
Improve error message for username generation

alexeykazakov:
It may happen that we try to generate a username for an existing MUR account. It's not a problem but we need to return an error in this case so the next usersignup reconcile loop can pick it up properly. So, this PR just improves the error text so it's clear now that there is nothing wrong with the account. We just need to tiger another reconcile loop.


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -433,7 +433,7 @@ func (r *ReconcileUserSignup) generateCompliantUsername(instance *toolchainv1alp
 		} else if mur.Labels[toolchainv1alpha1.MasterUserRecordUserIDLabelKey] == instance.Name {
 			// If the found MUR has the same UserID as the UserSignup, then *it* is the correct MUR -
 			// Return an error here and allow the reconcile() function to pick it up on the next loop
-			return "", fmt.Errorf(fmt.Sprintf("could not generate compliant username as MasterUserRecord [%s] already exists", mur.Name))
+			return "", fmt.Errorf(fmt.Sprintf("INFO: could not generate compliant username as MasterUserRecord with the same name [%s] and user id [%s] already exists. The next reconcile loop will pick it up.", mur.Name, instance.Name))
```
xcoulon, line 436: we could even add a `keep calm` :)

xcoulon (approved):

MatousJobanek (approved):
:+1:

PR #324, opened 2020-11-24
Add Network Policy to allow ingress traffic from CRW namespace

alexeykazakov:
E2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/198

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
@@ -107,6 +107,20 @@ objects:
     podSelector: {}
     policyTypes:
     - Ingress
+- apiVersion: networking.k8s.io/v1
+  kind: NetworkPolicy
+  metadata:
+    name: allow-from-codeready-workspaces-operator
+    namespace: ${USERNAME}-code
+  spec:
+    ingress:
+      - from:
+          - namespaceSelector:
+              matchLabels:
+                network.openshift.io/policy-group: codeready-workspaces
+    podSelector: {}
+    policyTypes:
+      - Ingress
 parameters:
 - name: USERNAME
   required: true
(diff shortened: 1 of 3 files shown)


MatousJobanek (approved):
The formatting of the indentation in the yaml files is a bit weird, but since the e2e tests passed and the resulting JSON is correct, then let's fix it later. :-)

MatousJobanek (approved):
Nice :+1:

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
```diff
@@ -107,6 +107,20 @@ objects:
...
+  kind: NetworkPolicy
+  metadata:
+    name: allow-from-codeready-workspaces-operator
+    namespace: ${USERNAME}-code
+  spec:
+    ingress:
+      - from:
+          - namespaceSelector:
+              matchLabels:
+                network.openshift.io/policy-group: codeready-workspaces
```
ibuziuk, line 120: this label is expected to be added manually by the admin?

MatousJobanek: yes - this has to be added manually. 
I can incorporate it into our setup logic so it will be added automatically for any future executions. 
But for the existing namespace, yes, please add it there manually :-).

ibuziuk (commented):
@alexeykazakov I would recommend adding the network policy for managed addons based on the `hive.openshift.io/managed=true`. This will work in the general case not pnly for CRW but for the future addons that will land in the Sandbox

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
ibuziuk: ok, +1 if this will be part of the setup, but I also think that supporting managed addons case, in general, would be really nice to have (based on the `hive.openshift.io/managed=true`)

PR #325, opened 2020-11-27
Add handling for forbidden username prefixes

sbryzak:
This PR introduces a change to `UserSignupController` in order to check that usernames don't begin with a forbidden prefix, such as "openshift" or "kubernetes".  If the username does start with one of these strings, it will be prefixed with "crt-".

Fixes https://issues.redhat.com/browse/CRT-783

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -93,13 +93,20 @@ const (
 	// varDeactivationDomainsExcluded is a string of comma-separated domains that should be excluded from automatic user deactivation
 	// For example: "@redhat.com,@ibm.com"
 	varDeactivationDomainsExcluded = "deactivation.domains.excluded"
+
+	// varForbiddenUsernamePrefixes defines the prefixes that a username may not have when signing up.  If a
+	// username has a forbidden prefix, then the username compliance prefix is added to the username
+	varForbiddenUsernamePrefixes = "username.forbidden.prefixes"
+
+	DefaultForbiddenUsernamePrefixes = "openshift,kubernetes"
 )
 
 // Config encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
 type Config struct {
-	host         *viper.Viper
-	secretValues map[string]string
+	host                      *viper.Viper
+	secretValues              map[string]string
+	forbiddenUsernamePrefixes []string
 }
 
 // initConfig creates an initial, empty configuration.
@@ -113,6 +120,12 @@ func initConfig(secret map[string]string) *Config {
 	c.host.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 	c.host.SetTypeByDefaultValue(true)
 	c.setConfigDefaults()
+
+	if c.host.GetString(varForbiddenUsernamePrefixes) != "" {
+		c.forbiddenUsernamePrefixes = strings.FieldsFunc(c.host.GetString(varForbiddenUsernamePrefixes), func(c rune) bool {
+			return c == ','
+		})
+	}
 
 	return &c
 }
@@ -161,6 +174,7 @@ func (c *Config) setConfigDefaults() {
 	c.host.SetDefault(varEnvironment, defaultEnvironment)
 	c.host.SetDefault(varMasterUserRecordUpdateFailureThreshold, 2) // allow 1 failure, try again and then give up if failed again
 	c.host.SetDefault(varToolchainStatusRefreshTime, defaultToolchainStatusRefreshTime)
+	c.host.SetDefault(varForbiddenUsernamePrefixes, DefaultForbiddenUsernamePrefixes)
 }
 
 // GetToolchainStatusName returns the configured name of the member status resource
@@ -249,4 +263,8 @@ func (c *Config) GetDeactivationDomainsExcludedList() []string {
 	return strings.FieldsFunc(c.host.GetString(varDeactivationDomainsExcluded), func(c rune) bool {
 		return c == ','
 	})
+}
+
+func (c *Config) GetForbiddenUsernamePrefixes() []string {
+	return c.forbiddenUsernamePrefixes
 }
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -98,7 +98,7 @@ const (
 	// username has a forbidden prefix, then the username compliance prefix is added to the username
 	varForbiddenUsernamePrefixes = "username.forbidden.prefixes"
 
-	DefaultForbiddenUsernamePrefixes = "openshift,kubernetes"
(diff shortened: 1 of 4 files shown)


pkg/configuration/configuration.go
```diff
@@ -114,6 +121,12 @@ func initConfig(secret map[string]string) *Config {
 	c.host.SetTypeByDefaultValue(true)
 	c.setConfigDefaults()
 
+	if c.host.GetString(varForbiddenUsernamePrefixes) != "" {
+		c.forbiddenUsernamePrefixes = strings.FieldsFunc(c.host.GetString(varForbiddenUsernamePrefixes), func(c rune) bool {
+			return c == ','
+		})
+	}
```
MatousJobanek, line 128: could you please add a test for this part?

MatousJobanek (approved):
Nice and simple solution :+1:

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -412,6 +412,15 @@ func getNsTemplateTier(cl client.Client, tierName, namespace string) (*toolchain
 
 func (r *ReconcileUserSignup) generateCompliantUsername(instance *toolchainv1alpha1.UserSignup) (string, error) {
 	replaced := transformUsername(instance.Spec.Username)
+
+	// Check for any forbidden prefixes
+	for _, prefix := range r.crtConfig.GetForbiddenUsernamePrefixes() {
```
alexeykazakov, line 417: It's very minor but I thought that it would be more consistent to keep that transformation in transformUsername() function. But feel free to keep it here if you prefer that.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -2174,6 +2174,35 @@ func assertName(t *testing.T, expected, username string) {
...
+	require.NoError(t, err)
+
+	// Lookup the MUR
+	murs := &v1alpha1.MasterUserRecordList{}
+	err = r.client.List(context.TODO(), murs, client.InNamespace(test.HostOperatorNs))
+	require.NoError(t, err)
+	require.Len(t, murs.Items, 1)
+	mur := murs.Items[0]
+	require.Equal(t, userSignup.Name, mur.Labels[v1alpha1.MasterUserRecordUserIDLabelKey])
+	require.Equal(t, "crt-openshift-Bob", mur.Name)
```
alexeykazakov, line 2203: Please add a test for kubernetes too.

alexeykazakov (commented):
Looks good but please add a test for kubernetes and the configuration part mentioned by Matous.

pkg/configuration/configuration.go
```diff
@@ -93,13 +93,20 @@ const (
 	// varDeactivationDomainsExcluded is a string of comma-separated domains that should be excluded from automatic user deactivation
 	// For example: "@redhat.com,@ibm.com"
 	varDeactivationDomainsExcluded = "deactivation.domains.excluded"
+
+	// varForbiddenUsernamePrefixes defines the prefixes that a username may not have when signing up.  If a
+	// username has a forbidden prefix, then the username compliance prefix is added to the username
+	varForbiddenUsernamePrefixes = "username.forbidden.prefixes"
+
+	DefaultForbiddenUsernamePrefixes = "openshift,kubernetes"
```
xcoulon, line 101: there are namespaces prefixed with `kube-`, we should include it in the list of forbidden prefixes

pkg/configuration/configuration.go
```diff
@@ -93,13 +93,20 @@ const (
...
 )
 
 // Config encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
 type Config struct {
-	host         *viper.Viper
-	secretValues map[string]string
+	host                      *viper.Viper
+	secretValues              map[string]string
+	forbiddenUsernamePrefixes []string
```
xcoulon, line 109: why do we need this var, and why can't we just rely on the config loaded by viper (and still apply the default if the config is explicitly empty, or even better, always ensure that the default values are set)?

xcoulon (changes requested):

pkg/configuration/configuration.go
sbryzak: Done.

pkg/configuration/configuration.go
sbryzak: Ah you're right, I wasn't aware that viper supported array values now.  I've updated to use the viper config directly.

pkg/configuration/configuration.go
sbryzak: Added test to explicitly check this.

pkg/configuration/configuration.go
xcoulon: I can't see the change in your PR 😳 
Also, would it be simpler if the var was an `[]string` instead of a list of value joined with commas?

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: Included tests for all prefixes.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: Technically it seems that `transformUsername()` is restricted to transformation of invalid characters, and to be honest I don't see much value in having this as a separate function since it's not used anywhere else.  Since we are still appending to the username in `generateCompliantUsername()` (in the vacant mur name check) then it felt ok to prefix the forbidden usernames here also.

pkg/configuration/configuration.go
sbryzak: Here's the line:

https://github.com/codeready-toolchain/host-operator/pull/325/files#diff-a0e28739a3c85fb2099ec7cdf95f0579f5f84488cca12aee25fb2e62702db5d6R101

Arrays can't be constants in Go, so we need the comma-separated string.

pkg/configuration/configuration.go
```diff
@@ -93,13 +93,20 @@ const (
...
```
MatousJobanek, line 109: you forgot to remove this field - this is not used anywhere

pkg/configuration/configuration.go
```diff
@@ -250,3 +260,7 @@ func (c *Config) GetDeactivationDomainsExcludedList() []string {
 		return c == ','
 	})
 }
+
+func (c *Config) GetForbiddenUsernamePrefixes() []string {
+	return c.host.GetStringSlice(varForbiddenUsernamePrefixes)
```
MatousJobanek, line 265: Since you didn't add a test for the case when it would be set using a variable, could you please clarify what string format is expected in this case? I tried comma-separated string, but that doesn't work.

PR #335, opened 2020-12-04
Increase CPU limit to 10 cores

alexeykazakov:
After long discussion with Ilya and OpenShift folks we decided to increase user's CPU limit to 10 cores.
I'm also increasing the default CPU limit for -dev and -stage namespaces accordingly.

We can adjust the limits later.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/210

### DIFF
--- a/deploy/templates/nstemplatetiers/team/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/team/ns_dev.yaml
@@ -64,7 +64,7 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 150m
+        cpu: 500m
         memory: 1Gi
       defaultRequest:
         cpu: 10m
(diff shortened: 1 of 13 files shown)


MatousJobanek (approved):

deploy/templates/nstemplatetiers/team/ns_dev.yaml
```diff
@@ -64,7 +64,7 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 150m
+        cpu: 500m
```
MatousJobanek, line 67: Hopefully, it won't break our pipelines. 
Luckily the new pipelines don't create so many containers, so we should be fine

MatousJobanek: ignore my previous comment - the 10cores will ensure that it will run fine :-)

PR #337, opened 2020-12-04
Drop cpu/memory limits for host operator & add requests for reg-service

alexeykazakov:
Same as here: https://github.com/codeready-toolchain/member-operator/pull/225

This PR also adds cpu/memory requests to the reg-service which we can adjust later.


rajivnathan (approved):
👍

PR #342, opened 2020-12-06
Move twilio phone number to secret

alexeykazakov:
We can't keep it in env/prod.yaml because we need to use different numbers in prod and stage. So, let's keep it in the host operator secret as we do for the rest of twilio configuration.
The secret is already setup correctly: https://github.com/codeready-toolchain/sandbox-sre/blob/7e5a36466d2b5be02b6067834d5f9eb31e5e11a1/config/sandbox.x8i5.p1.openshiftapps.com/host/operator_secret.yaml#L18
We just need to drop the reg-service env var from the deployment and start loading it from the secret.
Also see https://github.com/codeready-toolchain/registration-service/pull/152

Will fail until https://github.com/codeready-toolchain/registration-service/pull/152 and https://github.com/codeready-toolchain/toolchain-e2e/pull/213 are merged.

### DIFF
--- a/deploy/env/e2e-tests.yaml
+++ b/deploy/env/e2e-tests.yaml
@@ -4,8 +4,6 @@ registration-service:
   verification:
     enabled: 'true'
     excluded-email-domains: 'redhat.com,acme.com'
-  twilio:
-    from-number: '+15005550006'
 host-operator:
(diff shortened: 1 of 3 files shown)


rajivnathan (approved):
👍

PR #344, opened 2020-12-07
fix: deploy all reg-service resources at once

MatousJobanek:
currently, there are issues with setting the correct status of `RegistrationService` CR - maybe, our check, if the resource was updated or not, doesn't work in the very same way as the k8s check (for the generation number). As a result, we can wait for a reconciliation that will not come.
As a solution, the controller will create/update all resources at once - if it seems that any resource was updated/created, then it adds the information into the message and sets to re-queue after 1 second (just as a safety in case it wouldn't be reconciled by the change)


alexeykazakov (approved):
Yep. Let's try this :)

PR #348, opened 2020-12-08
fix: keep member status list empty when no ToolchainCluster CR is present

MatousJobanek:
fix: https://issues.redhat.com/browse/CRT-886

keep member status list empty when no member ToolchainCluster CR is present in the host cluster

### DIFF
--- a/pkg/controller/toolchainstatus/toolchainstatus_controller.go
+++ b/pkg/controller/toolchainstatus/toolchainstatus_controller.go
@@ -312,9 +312,6 @@ func (r *ReconcileToolchainStatus) membersHandleStatus(reqLogger logr.Logger, to
 	if len(memberClusters) == 0 {
 		err := fmt.Errorf("no member clusters found")
 		reqLogger.Error(err, "number of member clusters is zero")
-		memberStatusNotFoundCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusNoClustersFoundReason, err.Error())
-		noMemberStatus := customMemberStatus(*memberStatusNotFoundCondition)
-		members[""] = noMemberStatus
 		ready = false
 	}
 	for _, memberCluster := range memberClusters {
@@ -399,8 +396,8 @@ func compareAndAssignMemberStatuses(reqLogger logr.Logger, toolchainStatus *tool
 		if ok {
 			toolchainStatus.Status.Members[index].MemberStatus = newMemberStatus
 			delete(members, memberStatus.ClusterName)
-		} else if memberStatus.ClusterName != "" && memberStatus.UserAccountCount > 0 {
-			err := fmt.Errorf("toolchainCluster not found for member cluster %s that was previously registered in the host", memberStatus.ClusterName)
+		} else if memberStatus.UserAccountCount > 0 {
+			err := fmt.Errorf("ToolchainCluster CR wasn't found for member cluster `%s` that was previously registered in the host", memberStatus.ClusterName)
 			reqLogger.Error(err, "the member cluster seems to be removed")
 			memberStatusNotFoundCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberToolchainClusterRemovedReason, err.Error())
 			toolchainStatus.Status.Members[index].MemberStatus = customMemberStatus(*memberStatusNotFoundCondition)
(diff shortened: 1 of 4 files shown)


pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -433,8 +433,29 @@ func TestToolchainStatusConditions(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, requeueResult, res)
+			AssertThatToolchainStatus(t, req.Namespace, requestName, fakeClient).
+				HasConditions(componentsNotReady(string(memberConnectionsTag))).Exists().
+				HasHostOperatorStatus(hostOperatorStatusReady()).
+				HasMemberStatus().
+				HasRegistrationServiceStatus(registrationServiceReady())
+		})
+
+		t.Run("ToolchainClustre CR of member clusters was removed", func(t *testing.T) {
```
rajivnathan, line 451:
```suggestion
		t.Run("ToolchainCluster CR of member clusters was removed", func(t *testing.T) {
```

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -399,8 +396,8 @@ func compareAndAssignMemberStatuses(reqLogger logr.Logger, toolchainStatus *tool
 		if ok {
 			toolchainStatus.Status.Members[index].MemberStatus = newMemberStatus
 			delete(members, memberStatus.ClusterName)
-		} else if memberStatus.ClusterName != "" && memberStatus.UserAccountCount > 0 {
-			err := fmt.Errorf("toolchainCluster not found for member cluster %s that was previously registered in the host", memberStatus.ClusterName)
+		} else if memberStatus.UserAccountCount > 0 {
+			err := fmt.Errorf("ToolchainCluster CR wasn't found for member cluster `%s` that was previously registered in the host", memberStatus.ClusterName)
 			reqLogger.Error(err, "the member cluster seems to be removed")
 			memberStatusNotFoundCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberToolchainClusterRemovedReason, err.Error())
```
rajivnathan, line 402: So we don't really have a removed condition anymore right? Maybe we should change `ToolchainStatusMemberToolchainClusterRemovedReason` to be `ToolchainStatusMemberToolchainClusterMissingReason`?

rajivnathan (approved):
Nice! This change makes it a lot more understandable, at least for me. 🙂  Approved but added a couple optional suggestions if you wish to take them.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
MatousJobanek: > So we don't really have a removed condition anymore right?

It's not a condition but a reason, but I probably don't understand the question. We still use it as you can see in the code.

>  Maybe we should change ToolchainStatusMemberToolchainClusterRemovedReason to be ToolchainStatusMemberToolchainClusterMissingReason?

yes, that would probably make more sense

rajivnathan: Sorry, I probably could have worded that better. I didn't see `ToolchainStatusMemberToolchainClusterRemovedReason` used anywhere else. I mainly just meant we could rename the reason to `ToolchainStatusMemberToolchainClusterMissingReason` since with the changed approach we aren't going to have a condition in the Members section for a cluster that is removed successfully.

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
MatousJobanek: :man_facepalming:

pkg/controller/toolchainstatus/toolchainstatus_controller.go
MatousJobanek: Didn't you mean that we can remove `ToolchainStatusMemberStatusNoClustersFoundReason`? And just rename the `ToolchainStatusMemberToolchainClusterRemovedReason` to `ToolchainStatusMemberToolchainClusterMissingReason`?
I'll do that

alexeykazakov (approved):

pkg/controller/toolchainstatus/toolchainstatus_controller.go
MatousJobanek: https://github.com/codeready-toolchain/api/pull/207

PR #349, opened 2020-12-08
disable CR metrics

xcoulon:
by default, this function wants to use all out CRDs, but when we have a "host-only"
cluster, this metrics server fails to load the CRDs from etcd, which causes CrashLoops.
If we really need this metrics server, then we should taylor the list of CRDs to expose.

Keeping the `serveCRMetrics` for now, but with a comment to avoid linter issue,
since it's not used.

Also, address pending lint issues with fixes or comments

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -213,10 +213,6 @@ func addMetrics(ctx context.Context, cfg *rest.Config) error {
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


alexeykazakov (approved):

xcoulon:
thanks @alexeykazakov!

PR #352, opened 2020-12-11
Wait for UserAccount to be deleted before deleting MUR

rajivnathan:
Fix for https://issues.redhat.com/browse/CRT-892

The MasterUserRecord was being deleted even if there was a failure deleting the related UserAccount(s). This change updates the logic related to deleting user accounts to verify the deletion is completed before continuing with the MasterUserRecord deletion.

With the change the MasterUserRecord would not be deleted and would propagate the error correctly.
eg.
```
status:
  conditions:
  - lastTransitionTime: "2020-12-11T01:36:17Z"
    message: waiting for UserAccount deletion
    reason: UnableToDeleteUserAccounts
    status: "False"
    type: Ready
```

Related e2e PR to make tests pass:
https://github.com/codeready-toolchain/toolchain-e2e/pull/218

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -128,7 +128,7 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		// If the UserAccount is being deleted, delete the UserAccounts in members.
 	} else if coputil.HasFinalizer(mur, murFinalizerName) {
 		if err = r.manageCleanUp(logger, mur); err != nil {
-			logger.Error(err, "unable to clean up MasterUserRecord as part of deletion")
+			logger.Error(err, "unable to clean up MasterUserRecord as part of deletion, sometimes this can happen because the deletion has not completed yet")
 			return reconcile.Result{}, err
 		}
 	}
@@ -303,18 +303,31 @@ func (r *ReconcileMasterUserRecord) deleteUserAccount(logger logr.Logger, target
 	// Get the User associated with the UserAccount
 	userAcc := &toolchainv1alpha1.UserAccount{}
 	namespacedName := types.NamespacedName{Namespace: memberCluster.OperatorNamespace, Name: name}
-	err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc)
-	if err != nil {
+	if err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc); err != nil {
 		if errors.IsNotFound(err) {
 			return nil
 		}
 		return err
 	}
+
+	if coputil.IsBeingDeleted(userAcc) {
+		return fmt.Errorf("UserAccount deletion in progress")
+	}
+
 	if err := memberCluster.Client.Delete(context.TODO(), userAcc); err != nil {
 		return err
 	}
 	counter.DecrementUserAccountCount(logger, targetCluster)
-	return nil
+
+	// Verify the UserAccount is deleted
+	userAcc = &toolchainv1alpha1.UserAccount{}
+	if err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	return fmt.Errorf("waiting for UserAccount deletion") // return error to reconcile again and wait for deletion
 }
 
 func toBeProvisioned() toolchainv1alpha1.Condition {
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -296,7 +296,7 @@ func (r *ReconcileMasterUserRecord) manageCleanUp(logger logr.Logger, mur *toolc
 			"failed to update MasterUserRecord while deleting finalizer")
 	}
 	counter.DecrementMasterUserRecordCount(logger)
-	log.Info("Finalizer removed from MasterUserRecord")
+	logger.Info("Finalizer removed from MasterUserRecord")
 	return 0, nil
 }
 
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -127,9 +127,12 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		}
 		// If the UserAccount is being deleted, delete the UserAccounts in members.
 	} else if coputil.HasFinalizer(mur, murFinalizerName) {
-		if err = r.manageCleanUp(logger, mur); err != nil {
-			logger.Error(err, "unable to clean up MasterUserRecord as part of deletion, sometimes this can happen because the deletion has not completed yet")
+		requeueTime, err := r.manageCleanUp(logger, mur)
+		if err != nil {
+			logger.Error(err, "unable to clean up MasterUserRecord as part of deletion")
 			return reconcile.Result{}, err
+		} else if requeueTime > 0 {
+			return reconcile.Result{Requeue: true, RequeueAfter: requeueTime}, err
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rajivnathan:
@MatousJobanek let me know if you're good with this approach or had something else in mind. Once we're settled on the approach I can complete the unit and e2e tests.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -303,18 +303,31 @@ func (r *ReconcileMasterUserRecord) deleteUserAccount(logger logr.Logger, target
...
+
+	// Verify the UserAccount is deleted
+	userAcc = &toolchainv1alpha1.UserAccount{}
+	if err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	return fmt.Errorf("waiting for UserAccount deletion") // return error to reconcile again and wait for deletion
```
MatousJobanek, line 330: I don't think that this is really useful - the deletion won't be done for sure at this point in the time. The reason is that the member controllers have to delete many other resources which takes some time.
I would just return with postponed reconciliation for example by 5-10 seconds.

TBH, there will be some additional reconciliations triggered by `UserAccount_Status_Controller` based on the changed status in UserAccount - the longer delay for the next reconciliation is just to be sure that we don't miss the final deletion.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -303,18 +303,31 @@ func (r *ReconcileMasterUserRecord) deleteUserAccount(logger logr.Logger, target
...
-	if err != nil {
+	if err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc); err != nil {
 		if errors.IsNotFound(err) {
 			return nil
 		}
 		return err
 	}
+
+	if coputil.IsBeingDeleted(userAcc) {
+		return fmt.Errorf("UserAccount deletion in progress")
```
MatousJobanek, line 314: This state is not an error, I would just return here with postponed reconciliation. WDYT?

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Yeah, I was thinking about this too but wasn't sure whether it could be deleted quickly or not. If it won't in most cases then I agree, we can just remove it.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Yeah, I thought about this too but the main reason I wanted to return it as an error is because if there's a problem deleting the UserAccount and it's stuck deleting then returning the error would make use of the controller's backoff mechanism so that we're not reconciling every x seconds when there's a persistent failure.

MatousJobanek: ok, how about doing this - let's check the deletion timestamp and if it's older than 1 minute, then start returning an error, until then, just reconcile every 10 seconds.

rajivnathan: That's a good suggestion! 🙂

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/352/commits/e9a3765ce783f8eb8561e95d79636723ef9fc666

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/352/commits/e9a3765ce783f8eb8561e95d79636723ef9fc666

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -276,45 +279,61 @@ func (r *ReconcileMasterUserRecord) useExistingConditionOfType(condType toolchai
...
 }
 
-func (r *ReconcileMasterUserRecord) manageCleanUp(logger logr.Logger, mur *toolchainv1alpha1.MasterUserRecord) error {
+func (r *ReconcileMasterUserRecord) manageCleanUp(logger logr.Logger, mur *toolchainv1alpha1.MasterUserRecord) (time.Duration, error) {
 	for _, ua := range mur.Spec.UserAccounts {
-		if err := r.deleteUserAccount(logger, ua.TargetCluster, mur.Name); err != nil {
-			return r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToDeleteUserAccountsReason), err,
+		requeueTime, err := r.deleteUserAccount(logger, ua.TargetCluster, mur.Name)
+		if err != nil {
+			logger.Error(err, "unable to clean up MasterUserRecord as part of deletion")
```
alexeykazakov, line 286: This error is logged in reconcile() (line 132), correct? We do not have to duplicate it here.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -276,45 +279,61 @@ func (r *ReconcileMasterUserRecord) useExistingConditionOfType(condType toolchai
...
 	}
 	// Get the User associated with the UserAccount
 	userAcc := &toolchainv1alpha1.UserAccount{}
 	namespacedName := types.NamespacedName{Namespace: memberCluster.OperatorNamespace, Name: name}
-	err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc)
-	if err != nil {
+	if err = memberCluster.Client.Get(context.TODO(), namespacedName, userAcc); err != nil {
 		if errors.IsNotFound(err) {
-			return nil
+			log.Info("UserAccount deleted")
```
alexeykazakov, line 316:
```suggestion
			logger.Info("UserAccount deleted")
```

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: yeah, I would remove this one and keep only the one at line 132

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -721,33 +727,109 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
...
-		ClusterClient(test.MemberClusterName, memberClient))
+		cntrl := newController(t, hostClient, s, NewGetMemberCluster(true, v1.ConditionTrue),
+			ClusterClient(test.MemberClusterName, memberClient))
 
-	// when
-	_, err := cntrl.Reconcile(newMurRequest(mur))
+		// when
+		result1, err1 := cntrl.Reconcile(newMurRequest(mur))
+		require.NoError(t, err1)
+		assert.True(t, result1.Requeue)
```
MatousJobanek, line 749: could you please also check that the finalizer is still there after the first reconcile? You could also move the second reconcile-case into a subtest of this test to make it easier to read :-)

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -721,33 +727,109 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
...
+		memberClient := test.NewFakeClient(t, userAcc)
+
+		cntrl := newController(t, hostClient, s, NewGetMemberCluster(true, v1.ConditionTrue),
+			ClusterClient(test.MemberClusterName, memberClient))
+
+		// when
+		result1, err := cntrl.Reconcile(newMurRequest(mur))
+		require.NoError(t, err)
+		assert.True(t, result1.Requeue)
+		assert.Equal(t, int64(result1.RequeueAfter), int64(10*time.Second))
```
MatousJobanek, line 787: the same also here

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Nice catch! I missed that

MatousJobanek (approved):
This looks good :+1: 
Thanks for adding the extra logic that checks the deletion timestamp

pkg/controller/masteruserrecord/masteruserrecord_controller.go
rajivnathan: Done in eefde512ed3508572559a09206194139afb80df6

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -276,45 +279,61 @@ func (r *ReconcileMasterUserRecord) useExistingConditionOfType(condType toolchai
...
+		deletionTimestamp := userAcc.GetDeletionTimestamp()
+		if time.Since(deletionTimestamp.Time) > 60*time.Second {
+			return 0, fmt.Errorf("UserAccount deletion has not completed in over 1 minute")
+		}
+		return requeueTime, nil
+	}
+
 	if err := memberCluster.Client.Delete(context.TODO(), userAcc); err != nil {
-		return err
+		return requeueTime, err
```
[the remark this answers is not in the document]
rajivnathan: I thought you said the deletion wouldn't be done right away anyway so requeueing after 10 seconds should be fine?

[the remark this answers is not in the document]
rajivnathan: Oh, nevermind 🤦  this is the error scenario so yes you're right

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
rajivnathan: Done in d203faf29d69fbe2b0a700a162d64ac3ba223dea

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
rajivnathan: Done in d203faf29d69fbe2b0a700a162d64ac3ba223dea

alexeykazakov (approved):
Looks good!

PR #355, opened 2020-12-18
Generated files for che memberstatus condition

rajivnathan:
Related PRs:
https://github.com/codeready-toolchain/member-operator/pull/231
https://github.com/codeready-toolchain/api/pull/210

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -167,6 +167,10 @@ objects:
                     configMapKeyRef:
                       name: registration-service
                       key: verification.message_template
+              resources:
+                requests:
+                  cpu: "50m"
+                  memory: "100M"
   - kind: Service
     apiVersion: v1
     metadata:
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -167,10 +167,6 @@ objects:
                     configMapKeyRef:
                       name: registration-service
                       key: verification.message_template
-              resources:
-                requests:
-                  cpu: "50m"
-                  memory: "100M"
   - kind: Service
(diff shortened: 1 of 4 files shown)


deploy/registration-service/registration-service.yaml
```diff
@@ -167,10 +167,6 @@ objects:
                     configMapKeyRef:
                       name: registration-service
                       key: verification.message_template
-              resources:
-                requests:
-                  cpu: "50m"
-                  memory: "100M"
```
alexeykazakov, line 173: Why it's being deleted?

rajivnathan: I'm not sure, I don't see anything in my changes that would cause that.

alexeykazakov: Then please bring it back :)

MatousJobanek: The reason is that the resource requests are not defined in the registration-service repo.

MatousJobanek: fixed in https://github.com/codeready-toolchain/registration-service/pull/163

MatousJobanek (approved):

deploy/registration-service/registration-service.yaml
alexeykazakov: Good catch. BTW I think we don't really need that much resources (there are three replicas so the total would be 150m/300M) but let's leave it for another PR :)

rajivnathan: Thanks @MatousJobanek. @alexeykazakov it's back now after pulling in Matous' change and regenerating.

alexeykazakov (approved):

PR #356, opened 2021-01-08
Fix: Adding GenerateName to Notification Name

tinakurian:
Relates to: https://github.com/codeready-toolchain/toolchain-e2e/pull/219
Jira: https://issues.redhat.com/browse/CRT-893

As a user, if I get deactivated twice in less than 24hrs, my account gets stuck in a deactivating status and the usersignup has the following status condition:

- lastTransitionTime: "2020-12-09T13:23:24Z"
    message: notifications.toolchain.dev.openshift.com "xcoulon-deactivated" already exists
We should add a timestamp suffix on the notification resource name to avoid such conflicts.

To solve this problem, I have changed the Name field to GenerateName. I have also added two new labels to the api repo for notification_types.go. Subsequently, I had to also change GET notifications to LIST notifications since more than one notification of a specific type is now permitted.


pkg/controller/masteruserrecord/sync.go
```diff
@@ -185,7 +186,7 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
 	if condition.IsNotTrue(s.record.Status.Conditions, toolchainv1alpha1.MasterUserRecordUserProvisionedNotificationCreated) {
 		notification := &toolchainv1alpha1.Notification{
 			ObjectMeta: v1.ObjectMeta{
-				Name:      s.record.Name + "-provisioned",
+				Name:      s.record.Name + "-provisioned" + "-" + strconv.FormatInt(v1.Now().UnixNano(), 10),
```
sbryzak, line 189: I think it would be preferable here to use `fmt.Sprintf()` here instead of a multi-string concatenation.

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -533,7 +534,7 @@ func (r *ReconcileUserSignup) DeleteMasterUserRecord(mur *toolchainv1alpha1.Mast
 func (r *ReconcileUserSignup) sendDeactivatedNotification(logger logr.Logger, userSignup *toolchainv1alpha1.UserSignup) error {
 	notification := &toolchainv1alpha1.Notification{
 		ObjectMeta: v1.ObjectMeta{
-			Name:      userSignup.Status.CompliantUsername + "-deactivated",
+			Name:      userSignup.Status.CompliantUsername + "-deactivated" + "-" + strconv.FormatInt(v1.Now().UnixNano(), 10),
```
sbryzak, line 537: Same as above.

sbryzak (approved):
Looks good! Just a very minor suggestion.

pkg/controller/masteruserrecord/sync.go
alexeykazakov, line 189: You can use `GenerateName` instead of `Name` to generate a random resource name. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#idempotency

alexeykazakov (changes requested):
Can we also add a test which actually tests that multiple notifications can be created?

pkg/controller/masteruserrecord/sync.go
```diff
@@ -185,8 +186,11 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
...
 				Namespace: s.record.Namespace,
+				Labels: map[string]string{
+					toolchainv1alpha1.NotificationUserNameLabelKey: s.record.Name,
```
alexeykazakov, line 192: This label is not used anywhere, is it?

tinakurian: it's used in the e2e tests now. @xcoulon suggested that we add a label to look up the notifications for a specific user in the tests. without the label, we need to check if the name contains the user name and the reason we are notifying (deactivated for instance). instead, it may be better to use auto-generated names for the notification name and use the label to understand who the notification is for. perhaps we can add a label for the reason we are notifying too? what do you think?

alexeykazakov: OK. Adding username and notification reason/type labels for lookups is probably a good idea... But:
- let's keep the overall notification name template for generated names as it is now: `GeneratedName: <username>-<type>-` so it's a human friendly name.
- let's add comments to the code specifying that the labels atm are used only for easier notification lookup during debugging and in e2e tests.

tinakurian: Okay, sounds good!

pkg/controller/masteruserrecord/sync.go
MatousJobanek: Yes, I agree with Alexey, we should rather use the `GenerateName` instead of adding a timestamp suffix

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1335,10 +1339,13 @@ func TestUserSignupFailedToCreateDeactivationNotification(t *testing.T) {
...
 		// A deactivated notification should not have been created
-		notification := &v1alpha1.Notification{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Status.CompliantUsername + "-deactivated", Namespace: userSignup.Namespace}, notification)
-		require.Error(t, err)
-		require.Equal(t, "notifications.toolchain.dev.openshift.com \"john-doe-deactivated\" not found", err.Error())
+		labels := map[string]string{v1alpha1.NotificationUserNameLabelKey: userSignup.Status.CompliantUsername, v1alpha1.NotificationTypeLabelKey: "deactivated"}
+		opts := client.MatchingLabels(labels)
+
+		notificationList := &v1alpha1.NotificationList{}
+		err = r.client.List(context.TODO(), notificationList, opts)
```
MatousJobanek, line 1346: It's just a detail, but I would not use the labels in this test - in this way you also verify that it wasn't created without the labels:
```suggestion
		notificationList := &v1alpha1.NotificationList{}
		err = r.client.List(context.TODO(), notificationList)
```

MatousJobanek (commented):
Looks good, just please don't forget to add a comment to the labels as was asked by Alexey: https://github.com/codeready-toolchain/host-operator/pull/356/files#r555954628

I'm also wondering if it could be possible to add constants for the notification type... WDYT?

pkg/controller/usersignup/usersignup_controller_test.go
tinakurian: fixed here: eebffef30151239dd963cccf246d3c2dda86d070

pkg/controller/masteruserrecord/sync.go
tinakurian: This is removed in favour of using GenerateName

pkg/controller/usersignup/usersignup_controller.go
tinakurian: This is removed in favour of using GenerateName

pkg/controller/masteruserrecord/sync.go
tinakurian: updated to use GenerateName here: 1ed72f7ee9951af397509be0b324422c81ce6ad4

tinakurian:
> Can we also add a test which actually tests that multiple notifications can be created?

Just updated the PR with an offline discussion. @alexeykazakov and I have agreed to not add tests for multiple notifications due to the potential of adding a flaky test. 

Background:
The e2e-tests are currently set to delete notifications after 5s. In order to ensure the tests pass we would need to increase this timeout. However, the amount we would need to increase this timeout in order to ensure that there is no flakiness, would probably need to be upwards of a minute. This will cause our e2e-tests to slow down.

tinakurian:
> Looks good, just please don't forget to add a comment to the labels as was asked by Alexey: https://github.com/codeready-toolchain/host-operator/pull/356/files#r555954628
> 
> I'm also wondering if it could be possible to add constants for the notification type... WDYT?

Yeah, good idea. I'll go ahead and add constants!

go.mod
```diff
@@ -33,6 +33,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/api v0.0.0-20201208173147-3fa9c0601d08 => github.com/tinakurian/api v0.0.0-20210114191541-6f2eed8aebca
```
alexeykazakov, line 36: A reminder to remove that.

pkg/controller/masteruserrecord/sync.go
```diff
@@ -185,8 +185,14 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
 	if condition.IsNotTrue(s.record.Status.Conditions, toolchainv1alpha1.MasterUserRecordUserProvisionedNotificationCreated) {
 		notification := &toolchainv1alpha1.Notification{
 			ObjectMeta: v1.ObjectMeta{
-				Name:      s.record.Name + "-provisioned",
+				GenerateName:      s.record.Name + "-" + toolchainv1alpha1.NotificationTypeProvisioned + "-",
```
alexeykazakov, line 188:
```suggestion
				GenerateName:      fmt.Stringf("%s-%s-", s.record.Name, toolchainv1alpha1.NotificationTypeProvisioned,
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -533,8 +533,14 @@ func (r *ReconcileUserSignup) DeleteMasterUserRecord(mur *toolchainv1alpha1.Mast
 func (r *ReconcileUserSignup) sendDeactivatedNotification(logger logr.Logger, userSignup *toolchainv1alpha1.UserSignup) error {
 	notification := &toolchainv1alpha1.Notification{
 		ObjectMeta: v1.ObjectMeta{
-			Name:      userSignup.Status.CompliantUsername + "-deactivated",
+			GenerateName:      userSignup.Status.CompliantUsername + "-" + toolchainv1alpha1.NotificationTypeDeactivated + "-",
```
alexeykazakov, line 536:
```suggestion
			GenerateName:      fmt.Stringf("%s-%s-", userSignup.Status.CompliantUsername, toolchainv1alpha1.NotificationTypeDeactivated,
```

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -712,12 +712,15 @@ func TestSyncMurStatusWithUserAccountStatuses(t *testing.T) {
...
+		notifications := &toolchainv1alpha1.NotificationList{}
+		err = hostClient.List(context.TODO(), notifications)
+		require.NoError(t, err)
+		require.Len(t, notifications.Items, 1)
+		notification := notifications.Items[0]
 		require.NoError(t, err)
 		assert.Equal(t, userAccount.Spec.UserID, notification.Spec.UserID)
 		assert.Equal(t, "userprovisioned", notification.Spec.Template)
-		assert.Equal(t, userAccount.Name+"-provisioned", notification.Name)
+		assert.Contains(t, notification.Name, userAccount.Name+"-provisioned")
```
alexeykazakov, line 723:
```suggestion
		assert.Contains(t, notification.Name, userAccount.Name+"-provisioned-")
		assert.True(t, len(notification.Name) > len(userAccount.Name+"-provisioned-"))
```

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1248,10 +1248,12 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
...
 		// A deactivated notification should have been created
-		notification := &v1alpha1.Notification{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Status.CompliantUsername + "-deactivated", Namespace: userSignup.Namespace}, notification)
+		notifications := &v1alpha1.NotificationList{}
+		err = r.client.List(context.TODO(), notifications)
 		require.NoError(t, err)
-		require.Equal(t, "john-doe-deactivated", notification.Name)
+		require.Len(t, notifications.Items, 1)
+		notification := notifications.Items[0]
+		require.Contains(t, notification.Name, "john-doe-deactivated")
```
alexeykazakov, line 1256:
```suggestion
		require.Contains(t, notification.Name, "john-doe-deactivated-")
		assert.True(t, len(notification.Name) > len("john-doe-deactivated-"))

```

tinakurian: Fixed as part of: 21c323901080723156e76ab6563614f16880baf3

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
tinakurian: Updated as part of: 21c323901080723156e76ab6563614f16880baf3

pkg/controller/usersignup/usersignup_controller.go
tinakurian: Fixed as part of: 21c323901080723156e76ab6563614f16880baf3

pkg/controller/masteruserrecord/sync.go
tinakurian: Fixed as part of: 21c323901080723156e76ab6563614f16880baf3. I used Sprintf here - I think Stringf may be a typo.

alexeykazakov: Yep. Thanks :)

alexeykazakov (approved):
Looks good!
Please don't forget to update the api dep link before merging.

alexeykazakov:
Also, please do not forget to update the description and the title. We are not using timestamps anymore in this PR. (in the e2e PR too).

PR #357, opened 2021-01-13
fix: limit number of other resources in users' namespaces

MatousJobanek:
lower limits of the core resources + add limitations to non-core ones

paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/221


rajivnathan (approved):

alexeykazakov (approved):
Nice!

MatousJobanek:
I had to change the format of the resource count limits. When I was testing it manually, it didn't limit the non-core resources - I have to add the `count/` prefix to it and use dot-notation format for defining the resource. That works and limits the number of resources.

To keep some consistency, I also added the same `count/` prefix to the core resources - it works as well - verified manually against most of the resources.

MatousJobanek:
For more information see: https://kubernetes.io/docs/concepts/policy/resource-quotas/#object-count-quota

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -491,7 +491,7 @@ func limitRangeObj(kind, cpuLimit, memoryLimit, cpuRequest, memoryRequest string
...
+	return fmt.Sprintf(`{"apiVersion":"quota.openshift.io/v1","kind":"ClusterResourceQuota","metadata":{"name":"for-${USERNAME}"},"spec":{"quota":{"hard":{"count/buildconfigs.build.openshift.io":"10","count/configmaps":"100","count/cronjobs.batch":"30","count/daemonsets.apps":"30","count/deploymentconfigs.apps":"30","count/deployments.apps":"30","count/ingresses.extensions":"10","count/jobs.batch":"30","count/persistentvolumeclaims":"5","count/pods":"30","count/replicasets.apps":"30","count/replicationcontrollers":"30","count/routes.route.openshift.io":"10","count/secrets":"100","count/services":"10","count/statefulsets.apps":"30","limits.cpu":"%[1]s","limits.ephemeral-storage":"7Gi","limits.memory":"%[3]s","requests.cpu":"%[2]s","requests.ephemeral-storage":"7Gi","requests.memory":"%[3]s","requests.storage":"15Gi"}},"selector":{"annotations":{"openshift.io/requester":"${USERNAME}"},"labels":null}}}`, cpuLimit, cpuRequest, memoryLimit)
```
xcoulon, line 494: minor suggestion: would it be more readable if we had a JSON struct which we would marshal in order to return a `string`?

xcoulon (approved):

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
MatousJobanek: That way is used in e2e tests, I think that it's fine using both approaches - here the JSON content and struct in e2e tests

PR #358, opened 2021-01-13
User namespace networking

rajivnathan:
Adds network policies that allow connections between all namespaces belonging to a user but not from other user's namespaces.

- Adds a new `name` label for each user namespace that will be used for the network selector in the NetworkPolicy
- Adds a NetworkPolicy for each of the user namespaces to allow connections from each of the other namespaces belonging to the same user

Note: I manually verified connectivity between each of the namespace types and also verified other namespaces cannot access the user's namespace.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/222

Related issue: https://issues.redhat.com/browse/CRT-819

### DIFF
--- a/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
@@ -130,12 +130,12 @@ objects:
     name: allow-from-dev-namespace
     namespace: ${USERNAME}-code
   spec:
-    podSelector: {}
     ingress:
       - from:
         - namespaceSelector:
             matchLabels:
               name: ${USERNAME}-dev
+    podSelector: {}
     policyTypes:
       - Ingress
 - apiVersion: networking.k8s.io/v1
@@ -144,12 +144,12 @@ objects:
     name: allow-from-stage-namespace
     namespace: ${USERNAME}-code
   spec:
-    podSelector: {}
     ingress:
       - from:
         - namespaceSelector:
             matchLabels:
               name: ${USERNAME}-stage
+    podSelector: {}
     policyTypes:
       - Ingress
 parameters:
--- a/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/advanced/ns_code.yaml
@@ -127,7 +127,7 @@ objects:
 - apiVersion: networking.k8s.io/v1
   kind: NetworkPolicy
   metadata:
-    name: allow-from-dev-namespace
+    name: allow-from-other-user-namespaces
     namespace: ${USERNAME}-code
   spec:
     ingress:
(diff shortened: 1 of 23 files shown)


deploy/templates/nstemplatetiers/advanced/ns_code.yaml
```diff
@@ -122,6 +124,34 @@ objects:
...
+  spec:
+    ingress:
+      - from:
+        - namespaceSelector:
+            matchLabels:
+              name: ${USERNAME}-dev
+    podSelector: {}
+    policyTypes:
+      - Ingress
+- apiVersion: networking.k8s.io/v1
```
alexeykazakov, line 141: Does it have to be a separate policy (one policy per namespace)? Or it's possible to define one policy which would accept traffic from namespaces with either one label **or** another? Basically is it possible to have some kind of **or** operation in the `- from`?

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -410,10 +410,22 @@ func assertNamespaceTemplate(t *testing.T, decoder runtime.Decoder, actual templ
 	require.NoError(t, err)
 	assert.Equal(t, expected, actual)
 	// Assert expected objects in the template
-	// Each template should have one Namespace, one RoleBinding, one LimitRange and three NetworkPolicy objects
+	// Each template should have one Namespace, one RoleBinding, one LimitRange and a varying number of NetworkPolicy objects depending on the namespace kind
+
+	// Template objects count
+	if tier == "team" {
```
alexeykazakov, line 416: Let's use switch/case instead.

alexeykazakov (commented):
Overall looks good but I have a question.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
rajivnathan: Done in 95ca667b15c6c242c200cfb797b9eecb577d188c

deploy/templates/nstemplatetiers/advanced/ns_code.yaml
rajivnathan: I don't see way to do that. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#networkpolicyingressrule-v1-networking-k8s-io

Even the label selector only supports ANDing things together: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#labelselector-v1-meta

alexeykazakov: Hm..
> from | List of sources which should be able to access the pods selected for this rule. Items in this list are combined using a logical OR operation. If this field is empty or missing, this rule matches all sources (traffic not restricted by source). If this field is present and contains at least one item, this rule allows traffic only if the traffic matches at least one item in the from list.

alexeykazakov: So sounds like this is what we need:
```
  spec:
    ingress:
      - from:
        - namespaceSelector:
            matchLabels:
              name: ${USERNAME}-dev
        - namespaceSelector:
            matchLabels:
              name: ${USERNAME}-stage
```
right?

rajivnathan: Oops! Yes you're right. I misread it. I'll make the change.

rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/358/commits/34599fc68836ec12a22b9887db21bf808a7e0423

alexeykazakov (approved):
Nice!

MatousJobanek (approved):
Good job :+1:

PR #360, opened 2021-01-14
fix: prevent redhat and default prefixes

MatousJobanek:
prevent from creating `redhat` & `default` prefixes
```
Customer workloads should be placed in customer namespaces, and should not match this regular expression: (^kube.*|^openshift.*|^default$|^redhat.*)
```
I know that the `default` is not a prefix, but it's easier to implement and doesn't hurt us :-) 

paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/223


alexeykazakov (approved):
+1

xcoulon (approved):

PR #361, opened 2021-01-19
CRT-915

sbryzak:
Update UserSignup operator to use new MUR owner label

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -110,11 +110,17 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 	}
 
 	// TODO remove this code block once all MUR resources have had their labels updated
-	mur, err = r.migrateLabels(mur)
+	migrated, err := r.migrateLabels(mur)
 	if err != nil {
 		// I think it's overkill to define a new status reason here?  At least log the error for now
 		logger.Error(err, "unable to migrate deprecated MUR labels")
 		return reconcile.Result{}, err
+	}
+
+	if migrated {
+		return reconcile.Result{
+			Requeue: true,
+		}, nil
 	}
 
 	// If the UserAccount is not being deleted, create or synchronize UserAccounts.
@@ -150,18 +156,19 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 
 // migrateLabels is a temporary function that will replace deprecated labels with non-deprecated labels
 // Deprecated: Remove this function once all deprecated labels have been migrated
-func (r *ReconcileMasterUserRecord) migrateLabels(mur *toolchainv1alpha1.MasterUserRecord) (*toolchainv1alpha1.MasterUserRecord, error) {
+func (r *ReconcileMasterUserRecord) migrateLabels(mur *toolchainv1alpha1.MasterUserRecord) (bool, error) {
 	if val, ok := mur.Labels[toolchainv1alpha1.MasterUserRecordUserIDLabelKey]; ok {
 		mur.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey] = val
 		delete(mur.Labels, toolchainv1alpha1.MasterUserRecordUserIDLabelKey)
 
 		err := r.client.Update(context.TODO(), mur)
 		if err != nil {
-			return nil, err
+			return false, err
 		}
+		return true, nil
 	}
 
-	return mur, nil
+	return false, nil
 }
 
 func (r *ReconcileMasterUserRecord) addFinalizer(logger logr.Logger, mur *toolchainv1alpha1.MasterUserRecord, finalizer string) error {
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -109,20 +109,6 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, err
 	}
 
-	// TODO remove this code block once all MUR resources have had their labels updated
-	migrated, err := r.migrateLabels(mur)
-	if err != nil {
-		// I think it's overkill to define a new status reason here?  At least log the error for now
-		logger.Error(err, "unable to migrate deprecated MUR labels")
-		return reconcile.Result{}, err
-	}
-
-	if migrated {
-		return reconcile.Result{
-			Requeue: true,
-		}, nil
-	}
-
 	// If the UserAccount is not being deleted, create or synchronize UserAccounts.
 	if !coputil.IsBeingDeleted(mur) {
 		// Add the finalizer if it is not present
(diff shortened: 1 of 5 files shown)


alexeykazakov (approved):
Looks good!

sbryzak:
@alexeykazakov thanks for the quick approval, but apologies as it occurred to me we need more than this.  I believe we will also require a similar migration function for UserSignup, to copy the UserSignup.Name to UserSignup.Spec.UserID where it hasn't already been set.. does that seem right to you also?

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -108,6 +108,15 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 		logger.Error(err, "unable to get MasterUserRecord")
 		return reconcile.Result{}, err
 	}
+
+	// TODO remove this code block once all MUR resources have had their labels updated
+	mur, err = r.migrateLabels(mur)
```
alexeykazakov, line 113: Shouldn't we return here if the mur was migrated/updated? So the next reconcile is triggered when the actual update happens?

alexeykazakov (commented):
Right. But we can deal with UserSignup.Name -> UserSignup.Spec.UserID migration in a separate PR, right?

I also have one more question/comment I missed before.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
sbryzak: Sure, that makes sense.  I'll have it return a boolean also.  As for the `UserSignup.Name` -> `Spec.UserID` migration, I *think* we could possibly do this in a separate PR...  actually though, looking at the code it's not really clear where we will use the `UserSignup.Spec.UserID` value.  We're now setting the `owner` label on MUR to point to the UserSignup name, but I'm not certain about the purpose for UserID.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -108,6 +108,21 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+	migrated, err := r.migrateLabels(mur)
+	if err != nil {
+		// I think it's overkill to define a new status reason here?  At least log the error for now
+		logger.Error(err, "unable to migrate deprecated MUR labels")
+		return reconcile.Result{}, err
+	}
+
+	if migrated {
+		return reconcile.Result{
+			Requeue: true,
```
alexeykazakov, line 122: I don't think we have to explicitly request a requeue. We watch MUR anyway: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L81-L88
And UserSignup reconcile will be called again as soon as MUR is updated.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -139,6 +154,23 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 	return reconcile.Result{}, nil
 }
 
+// migrateLabels is a temporary function that will replace deprecated labels with non-deprecated labels
+// Deprecated: Remove this function once all deprecated labels have been migrated
+func (r *ReconcileMasterUserRecord) migrateLabels(mur *toolchainv1alpha1.MasterUserRecord) (bool, error) {
```
MatousJobanek, line 159: Btw, there is already a function for migrating MURs: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/masteruserrecord.go#L9
but it's fine keeping it as it is, since you are going to remove it anyway...

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -77,7 +77,7 @@ func TestUserSignupCreateMUROk(t *testing.T) {
 	require.Len(t, murs.Items, 1)
 	mur := murs.Items[0]
 	require.Equal(t, test.HostOperatorNs, mur.Namespace)
-	require.Equal(t, userSignup.Name, mur.Labels[v1alpha1.MasterUserRecordUserIDLabelKey])
+	require.Equal(t, userSignup.Name, mur.Labels[v1alpha1.MasterUserRecordOwnerLabelKey])
```
MatousJobanek, line 80: There are still many cases where the `MasterUserRecordUserIDLabelKey` is used in this test, but you don't change it to `MasterUserRecordOwnerLabelKey` nor add a check for the existence of the new label. I know that these are the cases for already existing MURs, but I believe that it would be good to migrate all to the new label, wouldn't it?

alexeykazakov (commented):
Shouldn't we also start using the new label in UserSignup controller?
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L262
https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L442

Regarding UserSignup.Spec.UserID.
We should use it when creating MUR: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L490
But as I mentioned above we can do it in a separate PR. This one could handle the label thing only and the next one could handle the UserSignup.Name -> UserSignup.Spec.UserID migration.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: This case a bit different - this inside of `masteruserrecord_controller`, not in `usersignup_controller`, so if you would want to rely on usersignup controller functionality, then it should be moved there.
The `masteruserrecord_controller` watches changes in MURs only when the generation number is changed which is not the case for changes in the labels.

MatousJobanek: To be honest, the cleaner solution for this would be moving the migration stuff to usersignup functionality - as I mentioned before: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/masteruserrecord.go#L9

But since we are "only" changing the label, then it's not a big deal. On the other hand, this requeue is a bit confusing here, since we don't requeue when we change something in the `MUR.status` either - after all, it has more or less the same impact on the definition of the MUR resource.

alexeykazakov: Ah right! I missed the fact that this migration code in the MUR controller. Not in the UserSignup one. And yeah.. I think it should be in the UserSignup controller.

sbryzak:
> Shouldn't we also start using the new label in UserSignup controller?
> https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L262
> https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L442

In both of these places it's using the new label, however it's also checking for the old label to ensure that updates aren't missed.

> Regarding UserSignup.Spec.UserID.
> We should use it when creating MUR: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L490
> But as I mentioned above we can do it in a separate PR. This one could handle the label thing only and the next one could handle the UserSignup.Name -> UserSignup.Spec.UserID migration.

This parameter value is assigned to the new owner label, which as we discussed should continue to reference the UserSignup resource name.  How else is the actual User ID (the original user's `subject` claim value) be used in relation to MasterUserRecord?

pkg/controller/masteruserrecord/masteruserrecord_controller.go
sbryzak: I can move the migration over to the UserSignup controller, however it would require a specific list operation on the MUR to find any of the older labels, an operation that would have to be run every single time we reconcile, whereas in the MUR controller the resource is given to you.

alexeykazakov:
> In both of these places it's using the new label, however it's also checking for the old label to ensure that updates aren't missed.

Right. Not sure why I missed that.. :)

> This parameter value is assigned to the new owner label, which as we discussed should continue to reference the UserSignup resource name. How else is the actual User ID (the original user's subject claim value) be used in relation to MasterUserRecord?

Currently UserSignup.Name is used for two different things:
- it's propagated to mur.labels["userid"] to be used as the owner reference. You are changing it here to use a different label for the same purpose.
- it's used to set mur.Spec.UserID when we are creating a MUR in UserSignup controller. See https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/masteruserrecord.go#L71 This one is propagated by the MUR controller to the UserAccount which uses it to create a user and identity in the member. We need to set mur.Spec.UserID to UserSignup.Spec.UserID but I think we can do it in a separate PR.

pkg/controller/usersignup/usersignup_controller_test.go
sbryzak: I must have missed these extra instances, hopefully they're all covered now.

PR #365, opened 2021-02-01
chore: prevent from creating sandbox* users

MatousJobanek:
Since we have dedicated admin users with `sandbox-` prefix, we need to prevent from creating such users via provisioning flow.

paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/230

### DIFF
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -294,9 +294,10 @@ func TestForbiddenUsernamePrefixesHaveCorrectDefaults(t *testing.T) {
 	defer restore()
 
 	config := getDefaultConfiguration(t)
-	require.Len(t, config.GetForbiddenUsernamePrefixes(), 4)
+	require.Len(t, config.GetForbiddenUsernamePrefixes(), 5)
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):

pkg/configuration/configuration_test.go
```diff
@@ -294,9 +294,10 @@ func TestForbiddenUsernamePrefixesHaveCorrectDefaults(t *testing.T) {
 	defer restore()
 
 	config := getDefaultConfiguration(t)
-	require.Len(t, config.GetForbiddenUsernamePrefixes(), 4)
+	require.Len(t, config.GetForbiddenUsernamePrefixes(), 5)
```
xcoulon, line 297: we could use the `ElementsMatch` func instead:
```
require.ElementsMatch(t, config.GetForbiddenUsernamePrefixes(), []string{"openshift", "kube", "default", "redhat", "sandbox"})
```

xcoulon (approved):

pkg/configuration/configuration_test.go
MatousJobanek: Thanks for the recommendation - will change it next time :-)

PR #368, opened 2021-02-02
Cleanup temporary code

sbryzak:
This PR removes the temporary code that was required as part of https://issues.redhat.com/browse/CRT-915.  This is a work in progress, and is first waiting upon https://github.com/codeready-toolchain/api/pull/213.

### DIFF
--- a/pkg/controller/masteruserrecord/sync.go
+++ b/pkg/controller/masteruserrecord/sync.go
@@ -195,7 +195,7 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
 				},
 			},
 			Spec: toolchainv1alpha1.NotificationSpec{
-				UserID:   s.record.Spec.UserID,
+				UserID:   s.record.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey],
 				Template: notificationtemplates.UserProvisioned.Name,
 			},
 		}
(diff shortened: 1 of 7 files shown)


alexeykazakov (approved):

pkg/controller/masteruserrecord/sync.go
```diff
@@ -195,7 +195,7 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
 				},
 			},
 			Spec: toolchainv1alpha1.NotificationSpec{
-				UserID:   s.record.Spec.UserID,
+				UserID:   s.record.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey],
```
MatousJobanek, line 198: The UserID in notification doesn't actually need the `UserID`, but a name of the UserSignup, right?
Could you please add some comment to it, so it is  clear why it is set in this way? It would really help

MatousJobanek (approved):
Nice :+1:

pkg/controller/masteruserrecord/sync.go
alexeykazakov: Correct. Notification.Spec.UserID is a misleading name. It should be UserSignupName or something. We have discussed it with Shane.
We should ether rename it (in a separate PR) or wait until we refactor the entire notification mechanism. We don't need UserID or UserSignupName in the new design anyway.

But for now let's add comment as Matous is suggesting.

sbryzak: Comment added.

MatousJobanek: Thanks for adding the comment.
I would keep the Notification CRD as it is for now and wait for the new Notification structure - this parameter will be removed there anyway.

PR #371, opened 2021-02-10
Introducing Test tier

alexeykazakov:
There is a need for the CRW team to test some different namespace settings for user namespaces. See https://issues.redhat.com/browse/CRT-940?focusedCommentId=15762779&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15762779 for details.
But I think it would be useful to have some `test` tier which we could use for such research/tests. Ideally we would want to have some ability to create temporal tiers directly in the cluster but it's not that easy right now.

So, for now let's have a permanent special tier we can use for that. If we need to test something else we just modify this test tier.
We do not care much about detailed unit tests for this tier content besides testing the the tier is present and parseible. Also I don't think we have to include it into our e2e tests.

For now this test tier is a copy of the basic tier but -code namespace has the same limit range as -dev and -stage.

**NOTE:** This PR also fixes -stage templates in all tiers to align the default memory limits with -dev namespaces.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/236

### DIFF
--- a/deploy/templates/nstemplatetiers/test/ns_code.yaml
+++ b/deploy/templates/nstemplatetiers/test/ns_code.yaml
@@ -1,0 +1,146 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: test-code
+objects:
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      che.eclipse.org/openshift-username: ${USERNAME}
+      openshift.io/description: ${USERNAME}-code
+      openshift.io/display-name: ${USERNAME}-code
+      openshift.io/requester: ${USERNAME}
+    labels:
+      name: ${USERNAME}-code
+    name: ${USERNAME}-code
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: rbac-edit
+    namespace: ${USERNAME}-code
+  rules:
+    - apiGroups:
+        - authorization.openshift.io
+        - rbac.authorization.k8s.io
+      resources:
+        - roles
+        - rolebindings
+      verbs:
+        - get
+        - list
+        - watch
+        - create
+        - update
+        - patch
+        - delete
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: user-rbac-edit
+    namespace: ${USERNAME}-code
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: Role
+    name: rbac-edit
+  subjects:
+    - kind: User
+      name: ${USERNAME}
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: user-edit
+    namespace: ${USERNAME}-code
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: edit
+  subjects:
+  - kind: User
+    name: ${USERNAME}
+- apiVersion: v1
+  kind: LimitRange
+  metadata:
+    name: resource-limits
+    namespace: ${USERNAME}-code
+  spec:
+    limits:
+    - type: "Container"
+      default:
+        cpu: 500m
+        memory: 750Mi
+      defaultRequest:
+        cpu: 10m
+        memory: 64Mi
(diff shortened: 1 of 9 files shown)


xcoulon (approved):

deploy/templates/nstemplatetiers/test/ns_stage.yaml
```diff
@@ -0,0 +1,131 @@
...
+              name: ${USERNAME}-code
+        - namespaceSelector:
+            matchLabels:
+              name: ${USERNAME}-dev
+    podSelector: {}
+    policyTypes:
+      - Ingress
+parameters:
+- name: USERNAME
+  required: true
```
ibuziuk, line 131: no new line in the end

deploy/templates/nstemplatetiers/test/ns_dev.yaml
```diff
@@ -0,0 +1,131 @@
...
+              name: ${USERNAME}-code
+        - namespaceSelector:
+            matchLabels:
+              name: ${USERNAME}-stage
+    podSelector: {}
+    policyTypes:
+      - Ingress
+parameters:
+- name: USERNAME
+  required: true
```
ibuziuk, line 131: no new line in the end

deploy/templates/nstemplatetiers/test/ns_code.yaml
```diff
@@ -0,0 +1,146 @@
...
+              name: ${USERNAME}-dev
+        - namespaceSelector:
+            matchLabels:
+              name: ${USERNAME}-stage
+    podSelector: {}
+    policyTypes:
+      - Ingress
+parameters:
+- name: USERNAME
+  required: true
```
ibuziuk, line 146: no new line in the end

deploy/templates/nstemplatetiers/test/ns_stage.yaml
alexeykazakov: Does it matter?

deploy/templates/nstemplatetiers/test/ns_stage.yaml
```diff
@@ -0,0 +1,131 @@
...
+  kind: LimitRange
+  metadata:
+    name: resource-limits
+    namespace: ${USERNAME}-stage
+  spec:
+    limits:
+    - type: "Container"
+      default:
+        cpu: 500m
+        memory: 512Mi
```
ibuziuk, line 70: I thought it would be  `750Mi`

ibuziuk (commented):
It looks like limits are not identical across the namespaces e.g.  `memory: 750Mi` is not set for stage

rajivnathan (approved):

deploy/templates/nstemplatetiers/test/ns_stage.yaml
alexeykazakov: Good catch!
It turned out that all our tiers had 512Mi in -stage namespaces. I guess we changed it in -dev to 750 at some point and forgot to update -stage...
Anyway it's fixed now in https://github.com/codeready-toolchain/host-operator/pull/371/commits/5184da0bfb69168d17f2ece417b50deb673dc830

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -432,18 +435,17 @@ func assertNamespaceTemplate(t *testing.T, decoder runtime.Decoder, actual templ
...
-	memoryLimit := "512Mi"
+	memoryLimit := "750Mi"
 	memoryRequest := "64Mi"
 	cpuRequest := "10m"
 	if kind == "code" {
 		cpuLimit = "1000m"
 		cpuRequest = "60m"
 		memoryRequest = "307Mi"
+		memoryLimit = "512Mi"
 	}
```
ibuziuk, line 446: this test checks the current default tier limits right? if we would apply the same limits across all namespaces as part of https://issues.redhat.com/browse/CRT-940 this if kind == "code" block will be removed, right?

alexeykazakov: If we apply the same limits for all NSs for all tiers then yes, that's correct.
But in this PR we apply the same limits only in the `test` tier which is not covered by this test.

ibuziuk (approved):

PR #376, opened 2021-02-23
refactor metrics

xcoulon:
remove the wrappers, refactor the init/reset
and update the tests accordingly: ince the `Reset()` func
now resets all metrics including the gauge(s), it's important
to set the initial values after preparing the reconciler and
resetting the metrics

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -11,28 +11,24 @@ import (
 
 func TestInitCounter(t *testing.T) {
 	// given
-	m := initCounter("test_counter", "test counter description")
+	m := newCounter("test_counter", "test counter description")
 
 	// when
-	m.Increment()
+	m.Inc()
 
 	// then
-	assert.Equal(t, "sandbox_test_counter", m.name)
-	assert.Equal(t, "test counter description", m.help)
-	assert.Equal(t, float64(1), promtestutil.ToFloat64(m.prometheusCounter))
+	assert.Equal(t, float64(1), promtestutil.ToFloat64(m))
 }
 
 func TestInitGauge(t *testing.T) {
 	// given
-	m := initGauge("test_gauge", "test gauge description")
+	m := newGauge("test_gauge", "test gauge description")
 
 	// when
 	m.Set(22)
 
 	// then
-	assert.Equal(t, "sandbox_test_gauge", m.name)
-	assert.Equal(t, "test gauge description", m.help)
-	assert.Equal(t, float64(22), promtestutil.ToFloat64(m.prometheusGauge))
+	assert.Equal(t, float64(22), promtestutil.ToFloat64(m))
 }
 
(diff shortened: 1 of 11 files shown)


xcoulon:
Note: no new feature here, but this refactoring is worth its own
PR/commit, so the upcoming work on the "user signups per cluster" 
metric will be easier to review ;)

pkg/metrics/metrics_test.go
```diff
@@ -42,25 +38,26 @@ func TestRegisterCustomMetrics(t *testing.T) {
...
 	}
 
 	for _, m := range allGauges {
-		assert.True(t, k8smetrics.Registry.Unregister(m.prometheusGauge))
+		assert.True(t, k8smetrics.Registry.Unregister(m))
 	}
+
+	// for _, m := range allGaugeVecs {
+	// 	assert.True(t, k8smetrics.Registry.Unregister(m))
+	// }
```
alexeykazakov, line 50: Zombie?

alexeykazakov (approved):

tinakurian (approved):
Nice!

pkg/metrics/metrics_test.go
xcoulon: oh thanks, I missed that! Removed in 263f9c5 ;)

rajivnathan:
The main reason for the wrapper was to be able to access the metric name/help for testing purposes because we use our own prefix for the names and I don't think prometheus api gives access to the name/help after creating the metrics. It looks like you're removing the unit tests that test that the naming is expected. I guess we're okay with that?

xcoulon:
> The main reason for the wrapper was to be able to access the metric name/help for testing purposes because we use our own prefix for the names and I don't think prometheus api gives access to the name/help after creating the metrics. It looks like you're removing the unit tests that test that the naming is expected. I guess we're okay with that?

that's right, once we initialized the Promtheus metric, we don't get access to its name/help anymore (or not in a simpler manner). My understanding is that we want to retain those values so we could reset the underlying Prometheus metrics in our tests. But with this refactoring there's a more straightforward way to proceed, I believe. 

Please let me know if you're ok with that @rajivnathan

rajivnathan (approved):
@xcoulon I may be misremembering but I think the purpose of the wrappers was to retain the name/help strings for unit testing purposes. But I guess it's not really too important to unit test those since prometheus API doesn't even provide access to them so I think it's fine. The changes look good overall!

xcoulon:
> @xcoulon I may be misremembering but I think the purpose of the wrappers was to retain the name/help strings for unit testing purposes. But I guess it's not really too important to unit test those since prometheus API doesn't even provide access to them so I think it's fine. The changes look good overall!

yes, indeed, I think that the whole purpose of the wrappers was for unit testing, so I guess we can get rid of them, they did not bring any real value (not even for the tests, I believe).

PR #381, opened 2021-02-25
add tests on toolchain status with multiple members

xcoulon:
also, refactor the func to initialize the `cluster.GetMemberClustersFunc`
func by passing the name(s) of the expected member clusters.

### DIFF
--- a/pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
+++ b/pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
@@ -59,7 +59,7 @@ const respBodyInvalid = `{"not found"}`
 const respBodyBad = `{"alive":false,"environment":"dev","revision":"64af1be5c6011fae5497a7c35e2a986d633b3421","buildTime":"0","startTime":"2020-07-06T13:18:30Z"}`
 
 func prepareReconcile(t *testing.T, requestName string, httpTestClient *fakeHTTPClient,
-	getMemberClustersFunc func(fakeClient client.Client) cluster.GetMemberClustersFunc, initObjs ...runtime.Object) (*ReconcileToolchainStatus, reconcile.Request, *test.FakeClient) {
+	getMemberClustersFunc getMemberClustersFunc, initObjs ...runtime.Object) (*ReconcileToolchainStatus, reconcile.Request, *test.FakeClient) {
 	s := scheme.Scheme
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1101,19 +1162,25 @@ func newDeploymentWithConditions(deploymentName string, deploymentConditions ...
...
 }
 
-func newGetMemberClustersFuncEmpty(_ client.Client) cluster.GetMemberClustersFunc {
-	return func(conditions ...cluster.Condition) []*cluster.CachedToolchainCluster {
-		var clusters []*cluster.CachedToolchainCluster
-		return clusters
+func noMemberClusters() getMemberClustersFunc {
+	return func(cl client.Client) cluster.GetMemberClustersFunc {
+		return func(conditions ...cluster.Condition) []*cluster.CachedToolchainCluster {
+			return make([]*cluster.CachedToolchainCluster, 0)
```
xcoulon, line 1182: the return of the return of the return! 
Like in a low-budget movie sequel 😂

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1101,19 +1162,25 @@ func newDeploymentWithConditions(deploymentName string, deploymentConditions ...
...
+			return clusters
 		}
 	}
 }
 
-func newGetMemberClustersFuncEmpty(_ client.Client) cluster.GetMemberClustersFunc {
-	return func(conditions ...cluster.Condition) []*cluster.CachedToolchainCluster {
-		var clusters []*cluster.CachedToolchainCluster
-		return clusters
+func noMemberClusters() getMemberClustersFunc {
```
MatousJobanek, line 1179: do you really need this function - couldn't you just use `memberClusters()` as you did here: https://github.com/codeready-toolchain/host-operator/blob/761645b91eeeb0ea513c2879734fe85cea63e22e/pkg/controller/toolchainstatus/toolchainstatus_controller_test.go#L433

MatousJobanek (approved):
Looks good, but I guess that the function `noMemberClusters()` is not really needed

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
xcoulon: you're right. I thought that this `noMemberCluster()` func would be more expressive than something like `memberClusters()` (and I did not realize I used it elsewhere).
That being said, I'm fine with dropping `noMemberClusters()` and using `memberClusters()` instead. 
Let me know what you prefer.

MatousJobanek: I prefer having just `memberClusters()` and dropping `noMemberCluster()`

xcoulon: ok, sounds good to me! 
done in 2c92725

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -77,8 +77,8 @@ func prepareReconcile(t *testing.T, requestName string, httpTestClient *fakeHTTP
 	return r, reconcile.Request{NamespacedName: test.NamespacedName(test.HostOperatorNs, requestName)}, fakeClient
 }
 
-func prepareReconcileWithStatusConditions(t *testing.T, requestName string, conditions []toolchainv1alpha1.Condition, initObjs ...runtime.Object) (*ReconcileToolchainStatus, reconcile.Request, *test.FakeClient) {
-	reconciler, req, fakeClient := prepareReconcile(t, requestName, newResponseGood(), newGetMemberClustersFuncReady, initObjs...)
+func prepareReconcileWithStatusConditions(t *testing.T, requestName string, getMemberClustersFunc getMemberClustersFunc, conditions []toolchainv1alpha1.Condition, initObjs ...runtime.Object) (*ReconcileToolchainStatus, reconcile.Request, *test.FakeClient) {
```
rajivnathan, line 80: May be worth renaming either the variable or the type `getMemberClustersFunc` for readability

rajivnathan (approved):
Nice! Made a minor suggestion but it's only a suggestion.

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
xcoulon: fair point, I'll see what I can do about it ;)

xcoulon: done in d37a2cd: code should be much simpler now:no more func with 3 levels of nested returns! 😅

MatousJobanek (approved):
Nice :+1:

rajivnathan (approved):
Awesome!

xcoulon:
thanks for the feedback @MatousJobanek and @rajivnathan! 😊

PR #383, opened 2021-03-01
Member user limit

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-949

While updating the e2e tests for automatic approval to test multiple members, I discovered that the automatic approval was actually allowing 1 more user than the specified limit. For example, if you configure the HostOperatorConfig for a member to have a user limit of 1 then 2 users would be approved to be provisioned on the cluster. This is because in the `hasNotReachedMaxNumberOfUsersThreshold` func `numberOfUserAccounts <= threshold` will cause it to return true if the number of user accounts has reached the specified limit when it should really be returning false.

Also found that the unit test for the case 'two clusters where the first one reaches max number of UserAccounts' was misleading because it was actually hitting the resource limit, instead of the user limit. So I updated that test as well by setting the max user limit for member1 to `800` which is equal to the initialized counter for member1.


MatousJobanek (approved):
Good catch :1st_place_medal: :rocket: 
Thanks a lot for fixing it  :+1:

alexeykazakov (approved):
Good catch indeed!

PR #384, opened 2021-03-01
Support configuration of ReplyTo email address for notifications

sbryzak:
This PR introduces a new configuration parameter `mailgun.replyto.email` which is read from the host operator configuration secret.  If set, the reply-to email address will be set to this value when notifications are sent to users.

Fixes https://issues.redhat.com/browse/CRT-910

### DIFF
--- a/pkg/controller/notification/mailgun_notification_delivery_service_test.go
+++ b/pkg/controller/notification/mailgun_notification_delivery_service_test.go
@@ -28,7 +28,8 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
 	mockServerOption := NewMailgunAPIBaseOption(mgs.URL())
 	invalidServerOption := NewMailgunAPIBaseOption("https://127.0.0.1:60000/v3")
 
-	config := NewNotificationDeliveryServiceFactoryConfig("mg.foo.com", "abcd12345", "noreply@foo.com", "mailgun")
+	config := NewNotificationDeliveryServiceFactoryConfig("mg.foo.com", "abcd12345",
+		"noreply@foo.com", "", "mailgun")
 	notCtx := &UserNotificationContext{
 		UserID:      "jsmith123",
 		FirstName:   "John",
@@ -126,5 +127,22 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
 		// then
 		require.Error(t, err)
 		require.Equal(t, "template: template:1: function \"invalid_expression\" not defined", err.Error())
+	})
+
+	t.Run("test mailgun notification delivery service send with reply-to address", func(t *testing.T) {
+		// when
+		config := NewNotificationDeliveryServiceFactoryConfig("mg.foo.com", "abcd12345",
+			"noreply@foo.com", "info@foo.com", "mailgun")
+
+		mgds := NewMailgunNotificationDeliveryService(config, templateLoader, mockServerOption)
+		err := mgds.Send(notCtx, &v1alpha1.Notification{
+			Spec: v1alpha1.NotificationSpec{
+				Subject: "test",
+				Content: "abc",
+			},
+		})
+
+		// then
+		require.NoError(t, err)
 	})
 }
--- a/pkg/controller/notification/mailgun_notification_delivery_service_test.go
+++ b/pkg/controller/notification/mailgun_notification_delivery_service_test.go
@@ -144,5 +144,6 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
 
 		// then
 		require.NoError(t, err)
+
 	})
 }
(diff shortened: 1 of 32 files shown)


pkg/controller/notification/notification_delivery_service_test.go
```diff
@@ -40,7 +41,11 @@ func (c *MockNotificationDeliveryServiceFactoryConfig) GetMailgunSenderEmail() s
...
 }
 
-func NewNotificationDeliveryServiceFactoryConfig(domain, apiKey, senderEmail, service string) NotificationDeliveryServiceFactoryConfig {
+func (c *MockNotificationDeliveryServiceFactoryConfig) GetMailgunReplyToEmail() string {
+	return c.Mailgun.ReplyToEmail
+}
+
+func NewNotificationDeliveryServiceFactoryConfig(domain, apiKey, senderEmail, replyToEmail, service string) NotificationDeliveryServiceFactoryConfig {
 	return &MockNotificationDeliveryServiceFactoryConfig{
 		Mailgun: MockMailgunConfiguration{Domain: domain, APIKey: apiKey, SenderEmail: senderEmail},
```
MatousJobanek, line 50: the `replyToEmail` param is not used anywhere - it should be assigned to `MockMailgunConfiguration`:
```suggestion
func NewNotificationDeliveryServiceFactoryConfig(domain, apiKey, senderEmail, replyToEmail, service string) NotificationDeliveryServiceFactoryConfig {
	return &MockNotificationDeliveryServiceFactoryConfig{
		Mailgun: MockMailgunConfiguration{Domain: domain, APIKey: apiKey, SenderEmail: senderEmail, ReplyToEmail:replyToEmail},
```

pkg/controller/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -127,4 +128,21 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
...
+		mgds := NewMailgunNotificationDeliveryService(config, templateLoader, mockServerOption)
+		err := mgds.Send(notCtx, &v1alpha1.Notification{
+			Spec: v1alpha1.NotificationSpec{
+				Subject: "test",
+				Content: "abc",
+			},
+		})
+
+		// then
+		require.NoError(t, err)
```
MatousJobanek, line 146: I'm afraid that the test doesn't verify anything. It's apparent because - as I stated in the previous comment - the reply-to email wasn't used anywhere. 
Could you please verify it either via the mock or using [Gock](https://github.com/h2non/gock)

MatousJobanek (commented):
Could you please take a look at the test?

pkg/controller/notification/notification_delivery_service_test.go
sbryzak: Fixed, good catch.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: Unfortunately the mock doesn't provide us the capability to examine events created, it instead provides an events API that returns a random array of various Mailgun events - see https://github.com/mailgun/mailgun-go/blob/master/mock_events.go#L12.  This means the best we can do without actually testing with a real Mailgun API key is to just ensure that the Mailgun service doesn't return an error when using a mock server.

MatousJobanek (approved):
You could try to use the Gock mocking library to check the outcoming requests, but if you feel confident with verifying it via the mocked Mailgun service, then I'm fine with it.

pkg/controller/notification/mailgun_notification_delivery_service_test.go
sbryzak: > You could try to use the Gock mocking library to check the outcoming requests, but if you feel confident with verifying it via the mocked Mailgun service, then I'm fine with it.

All the existing Mailgun features were verified this way and tested in staging.  @alexeykazakov will we have the opportunity to test this in staging also before it gets deployed to production?

MatousJobanek: No, it won't - everything what is deployed to staging is at the same time also deployed to prod.
I'm not afraid of breaking it now (in the sense that I don't think that we would break it), I was more thinking about adding a test to verify that it won't be unintentionally changed in a wrong way in the future. However, if the effort/time to implement such a test would be much bigger than the benefit from checking that it's sending the correct request, then it makes sense to omit it.

sbryzak: I'm going to do some local testing with this today before I merge, using the live Mailgun API keys.  This will at least confirm that the ReplyTo is working as intended before we deploy to prod.

PR #387, opened 2021-03-05
Add user cleanup controller

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-895

Signed-off-by: sbryzak <sbryzak@gmail.com>


sbryzak:
Tests still to come.

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	// Watch for changes to the secondary resource UserSignup and requeue the owner UserSignup
+	if err := c.Watch(
+		&source.Kind{Type: &toolchainv1alpha1.UserSignup{}},
+		&handler.EnqueueRequestForOwner{
+			IsController: true,
+			OwnerType:    &toolchainv1alpha1.UserSignup{},
+		}); err != nil {
```
MatousJobanek, line 54: correct me if I'm wrong - the owner of the `UserSignup` cannot be a `UserSignup`
you should use `handler.EnqueueRequestForObject{}` instead

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileUserCleanup{}
+
+// ReconcileUserCleanup cleans up old UserSignup resources
+type ReconcileUserCleanup struct {
```
MatousJobanek, line 64: Could we name the file, directory and the controller `UserSignupCleanup` to make clear that it cleans just the `UserSignups` and not the entire user accounts.

pkg/controller/usercleanup/user_cleanup_controller.go
MatousJobanek: In addition, you could use a predicate to filter `UserSignups` with the label `toolchain.dev.openshift.com/state` equal to either `not-ready` or `deactivated`

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+			if cond.Type == toolchainv1alpha1.UserSignupComplete &&
+				cond.Reason == toolchainv1alpha1.UserSignupUserDeactivatedReason {
+				deactivatedStatusCondition = &cond
+				break
+			}
+		}
+
+		if deactivatedStatusCondition == nil {
+			// We cannot find the status condition
+			// TODO how should we handle this if we can't find the condition?
```
MatousJobanek, line 129: you could use the last timestamp of the update taken from `metadata.managedFields`

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
+
+	if instance.Spec.VerificationRequired && !instance.Spec.Approved {
+
+		createdTime := instance.ObjectMeta.CreationTimestamp
+
+		unverifiedThreshold := time.Now().Add(-time.Duration(r.crtConfig.GetUserSignupUnverifiedRetentionDays()*24) * time.Hour)
+
+		if createdTime.Time.Before(unverifiedThreshold) {
+			reqLogger.Info("UserSignup deleted due to exceeding unverified retention period")
```
MatousJobanek, line 101: since this is printed before the deletion is triggered, then it should contain something like:
```suggestion
			reqLogger.Info("Deleting UserSignup due to exceeding unverified retention period")
```

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
+
+	if instance.Spec.VerificationRequired && !instance.Spec.Approved {
```
MatousJobanek, line 94: When could it happen that the `VerificationRequired` is true and at the same time the `Approved` is true as well?

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+	if instance.Spec.Deactivated {
+		// Find the UserSignupComplete condition
+		var deactivatedStatusCondition *toolchainv1alpha1.Condition
+		for _, cond := range instance.Status.Conditions {
+			if cond.Type == toolchainv1alpha1.UserSignupComplete &&
+				cond.Reason == toolchainv1alpha1.UserSignupUserDeactivatedReason {
+				deactivatedStatusCondition = &cond
+				break
+			}
+		}
```
MatousJobanek, line 125: You could use the functions from toolchain-common repo:
```
condition.FindConditionByType(instance.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
```
and/or also:
```
condition.HasConditionReason(instance.Status.Conditions, toolchainv1alpha1.UserSignupComplete,toolchainv1alpha1.UserSignupUserDeactivatedReason)
```

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+			// TODO how should we handle this if we can't find the condition?
+			return reconcile.Result{}, nil
+		}
+
+		// If the LastTransitionTime of the deactivated status condition is older than the configured threshold,
+		// then delete the UserSignup
+
+		deactivatedThreshold := v1.Time{time.Now().Add(-time.Duration(r.crtConfig.GetUserSignupDeactivatedRetentionDays()*24) * time.Hour)}
+
+		if deactivatedStatusCondition.LastTransitionTime.Before(&deactivatedThreshold) {
```
MatousJobanek, line 138: You can use the actual time from the condition instead of initializing the meta.Time:
```suggestion
		deactivatedThreshold := time.Now().Add(-time.Duration(r.crtConfig.GetUserSignupDeactivatedRetentionDays() * 24) * time.Hour)

		if deactivatedStatusCondition.LastTransitionTime.Time.Before(deactivatedThreshold) {
```

pkg/controller/usercleanup/user_cleanup_controller.go
```diff
@@ -0,0 +1,166 @@
...
+			reqLogger.Info("UserSignup deleted due to exceeding deactivated retention period")
```
MatousJobanek, line 139: The same with the message:
```suggestion
			reqLogger.Info("Deleting UserSignup due to exceeding deactivated retention period")
```

pkg/controller/usercleanup/user_cleanup_controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+						Type:   v1alpha1.UserSignupApproved,
+						Status: v1.ConditionTrue,
+						Reason: "ApprovedAutomatically",
+					},
+				},
+				CompliantUsername: "abigail-thompson",
+			},
+		}
+		userSignup.Labels["toolchain.dev.openshift.com/approved"] = "true"
+		userSignup.Labels[v1alpha1.UserSignupStateLabelKey] = "approved"
```
MatousJobanek, line 48: Please, for the creation of the UserSignups use the functions from the `test` package: https://github.com/codeready-toolchain/host-operator/blob/master/test/usersignup.go

pkg/controller/usercleanup/user_cleanup_controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+		r, req, _ := prepareReconcile(t, userSignup.Name, userSignup)
+
+		_, err := r.Reconcile(req)
+		require.NoError(t, err)
+
+		// Confirm the UserSignup still exists
+		key := test.NamespacedName(test.HostOperatorNs, userSignup.Name)
+		require.NoError(t, r.client.Get(context.Background(), key, userSignup))
+		require.NotNil(t, userSignup)
```
MatousJobanek, line 99: Could you please verify that the `Requeue` is `true` and the `RequeueAfter` is approx within the range of the expected time?

MatousJobanek (commented):
Could you please add tests that will verify also the deletion of the UserSignups?

pkg/controller/usercleanup/user_cleanup_controller.go
alexeykazakov: It probably can't if everything works as expected but I would keep that check just to be on the safe side. I don't want us accidentally delete an active signup.

pkg/controller/usercleanup/user_cleanup_controller.go
alexeykazakov: I think we should ignore signups without the complete condition.

pkg/controller/usercleanup/user_cleanup_controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
+	"time"
+
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func TestUserCleanup(t *testing.T) {
+	t.Run("test that user cleanup doesn't delete an active UserSignup", func(t *testing.T) {
+		userSignup := &v1alpha1.UserSignup{
```
alexeykazakov, line 25: Maybe I missed that but you don't set the creation timestamp, do you? Let's set the timestamp to something old enough to make sure an active signup is not deleted even if it's older than both thresholds (deactivated and unverified).

pkg/controller/usercleanup/user_cleanup_controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+		_, err := r.Reconcile(req)
+		require.NoError(t, err)
+
+		// Confirm the UserSignup still exists
+		key := test.NamespacedName(test.HostOperatorNs, userSignup.Name)
+		require.NoError(t, r.client.Get(context.Background(), key, userSignup))
+		require.NotNil(t, userSignup)
+	})
+
+	t.Run("test that user cleanup doesn't delete a recently deactivated UserSignup", func(t *testing.T) {
```
alexeykazakov, line 61: Please also add tests for:
- old deactivated signup deleted
- old unverified signup deleted
- old (it has to be old) verified but unapproved is not deleted

alexeykazakov (commented):
I wonder if we can add some e2e tests...
It should be possible if we can update an existing UserSignup to set an older creation time. I doubt it's possible though.
Setting the thresholds to something very short is not a good option since it may affect other tests.

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Are `not-ready` and `deactivated` the only states we should be interested in?  What about `banned` ?

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Done

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: It now simply returns here.

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Done

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: I included this check as insurance, just to be safe as Alexey mentions.

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Done

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Done

pkg/configuration/configuration.go
```diff
@@ -102,6 +102,21 @@ const (
...
 
 	DefaultForbiddenUsernamePrefixes = "openshift,kube,default,redhat,sandbox"
+
+	// varUserSignupUnverifiedRetentionDays is used to configure how many days we should keep unverified (i.e. the user
+	// hasn't completed the user verification process via the registration service) UserSignup resources before deleting
+	// them.  It is intended for this parameter to define an aggressive cleanup schedule for unverified user signups,
+	// and the default configuration value for this parameter reflects this.
+	varUserSignupUnverifiedRetentionDays = "usersignup.unverified.retention.days"
+
+	defaultUserSignupUnverifiedRetentionDays = 2
```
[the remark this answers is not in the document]
sbryzak: Changed to 7 days

sbryzak:
@MatousJobanek thanks for the detailed review, I appreciate it :)

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: Fixed, thanks for the tip.

pkg/controller/usercleanup/user_cleanup_controller.go
alexeykazakov: I would avoid relying on the state label in this predicate. If we add more states or rename existing ones we can forget updating this. It's not critical but less error prone if our predicate is simple.

pkg/controller/usercleanup/user_cleanup_controller_test.go
sbryzak: Done.  I've added a check here to confirm the `RequeueAfter` is within a 2 day range of the default expected value.  We might want to tighten this up though, what do you think?

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			RequeueAfter: requeueAfter,
+		}, nil
+	}
+
+	if instance.Spec.Deactivated {
+		// Find the UserSignupComplete condition
+		cond, found := condition.FindConditionByType(instance.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
+
+		if !found || cond.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason {
+			// We cannot find the status condition, simply return
```
alexeykazakov, line 118:
```suggestion
			// We cannot find the status condition with "deactivated" reason, simply return
```

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: In this case do we even need a predicate?

pkg/controller/usercleanup/user_cleanup_controller_test.go
sbryzak: I've looked at these functions.  It seems there's no simple way to create a `UserSignup` with the same spec values (including `Deactivated`), conditions and labels, or am I missing something?

pkg/controller/usercleanup/user_cleanup_controller_test.go
sbryzak: I've set the creation timestamp to 3 years in the past.

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Fixed.

pkg/controller/usercleanup/user_cleanup_controller_test.go
sbryzak: Done

pkg/controller/usercleanup/user_cleanup_controller.go
alexeykazakov: That would work. I'm fine with both options (do not have a predicate or filter out state == approved)

pkg/controller/usersignupcleanup/predicate.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	controllerPredicate "sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type UserSignupCleanupPredicate struct {
+	controllerPredicate.Funcs
+}
```
MatousJobanek, line 11: You can remove it if you don't want to use it ;-)

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileUserCleanup{}
+
+// ReconcileUserCleanup cleans up old UserSignup resources
+type ReconcileUserCleanup struct {
```
MatousJobanek, line 61: Thanks for renaming the files, but you forgot to rename the struct :-)

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
+		require.NoError(t, err)
+
+		// Confirm the UserSignup has been deleted
+		key := test.NamespacedName(test.HostOperatorNs, userSignup.Name)
+		err = r.client.Get(context.Background(), key, userSignup)
+		require.Error(t, err)
+		require.IsType(t, &errors.StatusError{}, err)
+		statusErr := err.(*errors.StatusError)
+		require.Equal(t, metav1.StatusReasonNotFound, statusErr.Status().Reason)
+		require.Equal(t, fmt.Sprintf("usersignups.toolchain.dev.openshift.com \"%s\" not found", key.Name), statusErr.Error())
```
MatousJobanek, line 160: I guess that you could just do:
```
require.Error(t, err)
require.True(t, errors.IsNotFound(err))
```

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
```
MatousJobanek, line 199: The same here

pkg/controller/usercleanup/user_cleanup_controller_test.go
MatousJobanek: you can use the modifier functions for that https://github.com/codeready-toolchain/host-operator/blob/master/test/usersignup.go#L16-L56
and create new ones if needed. Usage of these functions makes the test shorter, cleaner, and it's easier to focus on what is important to be configured - at least from my point of view :wink: .

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			Requeue:      true,
+			RequeueAfter: requeueAfter,
+		}, nil
+	}
+
+	if instance.Spec.Deactivated {
+		// Find the UserSignupComplete condition
+		cond, found := condition.FindConditionByType(instance.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
+
+		if !found || cond.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason {
```
alexeykazakov, line 117: I guess we also need to check if the condition is set to true. We should not have Condition Complete set to False with Deactivated reason but just to be explicit here...

```suggestion
		if !found || cond.Status != apiv1.ConditionTrue || cond.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason {
```

with import:
```
apiv1 "k8s.io/api/core/v1"
```

alexeykazakov (commented):
Looks good! Just one last comment from me and please also address the Matous's comments.

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Fixed

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
sbryzak: Fixed

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Oops, fixed.

pkg/controller/usersignupcleanup/predicate.go
sbryzak: I've removed the predicate altogether.

pkg/controller/usercleanup/user_cleanup_controller_test.go
sbryzak: No problem, I've replaced `UserSignup` creation with the functions and added new modifier functions where required.

pkg/controller/usercleanup/user_cleanup_controller.go
sbryzak: I've removed the predicate for now.

alexeykazakov (approved):
Looks good! Thanks for addressing the comments.

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func prepareReconcile(t *testing.T, name string, initObjs ...runtime.Object) (*ReconcileUserSignupCleanup, reconcile.Request, *test.FakeClient) {
+	metrics.Reset()
+
+	s := scheme.Scheme
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+
+	initObjs = append(initObjs)
+
```
MatousJobanek, line 152: I'm not sure what is the purpose of this append, but it actually doesn't do anything, does it? 
```suggestion
```

MatousJobanek (approved):
Awesome, thanks a lot for taking extra time and changing the tests :+1: :1st_place_medal: 

btw, could you please run `make format-go-code` to format the code?

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
sbryzak: This was a copy/paste from another test.  I've removed it.

PR #388, opened 2021-03-05
chore: add GH Actions workflow for releasing new operator versions

MatousJobanek:
adds GH Actions workflow that releases a new version of host-operator on every commit to master

depends on codeready-toolchain/toolchain-cicd#22


alexeykazakov (approved):
Great!

PR #391, opened 2021-03-11
CRDs updated after adding a new property ClusterApiEndpoint to MemberStatusStatus

ranakan19:
CRDs updated by running `make generate` after adding new property ClusterApiEndpoint to MemberStatusStatus

Fixes related to: https://issues.redhat.com/browse/CRT-971


alexeykazakov (commented):
Please create a PR against the api repo too and link it here.
There are unrelated changes (userid) which we should not change but let's discuss it in the api pr.

ranakan19:
> inst the api repo too and link it here.
> There are unrelated changes (userid) which we should not change but let's discuss it in the api pr.

Yes, was doing that, found an issue, so closing this PR. will create all of them once.

PR #393, opened 2021-03-12
Notify users that their account is going to be deactivated soon

sbryzak:
This PR will send a notification to users 3 days before their account is due to be deactivated.

Fixes https://issues.redhat.com/browse/CRT-994


pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -58,6 +73,7 @@ var _ reconcile.Reconciler = &ReconcileDeactivation{}
 
 // ReconcileDeactivation reconciles a Deactivation object
 type ReconcileDeactivation struct {
+	*usersignup.StatusUpdater
```
MatousJobanek, line 76: TBH, I have a bad feeling that this will break the "primary/secondary resource" approach
The status of the `UserSignup` should be modified only by the `UserSignup_controller` (`UserSignup` is a primary resource there). The spec should be modified only by the clients - which is also the `Deactivation_controller`. 
However, with this change the deactivation controller would modify both - the status and the spec.

MatousJobanek: What I was trying to say was that the `UserSignup_controller` should take care of sending the notification and updating the status. By that it will notify the clients (`deactivation_controller`) that the notification was already sent and based on that information the `deactivation_controller` can start counting the last 3 days before setting the `UserSignup.spec.deactivated` to `true`

MatousJobanek: Another approach would be prolonging the lifetime of the `Notification` CR to 4 days. Then the `deactivation_controller` could check if it exists or not. In the future, we could add a parameter to the `Notification` CRD that would set the lifetime of the CRs. As a downside I see that we will lose the information from the `UserSignup.status`, which could be a complication in a case we would need to troubleshoot something :-/.
Or, we could create a new CRD that would keep the track of the UserSignup deactivation.

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,6 +165,30 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+			return reconcile.Result{}, r.WrapErrorWithStatusUpdate(logger, usersignup, r.SetStatusDeactivatingNotificationCreationFailed, err, "Failed to create user deactivating notification")
+		}
+
+		if err := r.UpdateStatus(logger, usersignup, r.SetStatusDeactivatingNotificationCreated); err != nil {
+			logger.Error(err, "Failed to update notification created status")
+			return reconcile.Result{}, err
+		}
+	}
+
 	if timeSinceProvisioned < deactivationTimeout {
```
MatousJobanek, line 192: As I already mentioned in the slack channel, we should count 3 days from the time when the deactivating notification was sent and then deactivate.

pkg/controller/deactivation/deactivation_controller.go
MatousJobanek: So we discussed it during our sync meeting, and we agreed that prolonging `Notification` CR is a bad idea - so let's not consider this option anymore.
We discussed two more options:
1. instead of storing the information in `UserSignup.status` we could store it as an annotation of the `UserSignup` CR. In that case the deactivation controller would manage the whole deactivation flow and would also update the `UserSignup` CR.
2. second options was splitting the notification status info into `UserSignup.status` & `MUR.status`. `UserSignup.status` would still keep the information that the "deactivated notification" was sent and `MUR.status`would keep the information about "deactivating notification". That would mean that MUR controller would need to monitor the 27 days before sending the "deactivating notification" and also would be responsible for creating the notification CR.

Personally, I would vote for the first options - storing the information as a `UserSignup` annotation, but it depends on the amount of work & refactoring. Looking at the code, I believe that it will require even less work that the second option.

alexeykazakov: I'm also for the first option (annotations) if it doesn't require significantly more work than the second one.

pkg/controller/deactivation/deactivation_controller.go
alexeykazakov: Yeah it would be a better UX if we deactivate in 3 days after the actual email/notification. So imagine we have an account which is to be deactivated in 4 days. And then we have an outage for 4 days. When the service is up again we would like to avoid the situation when we send a notification saying the account will be deactivated in 3 days and then immediately deactivate it. It would be OK to postpone the actual deactivation even if it's past due.
But I'm a bit concerned about suck accounts if for some reason we failed to create a notification resource... Not sure if we should really be concerned about that since we would have a bigger problem if we can't/don't create notification resources for extended period of time.

pkg/controller/deactivation/deactivation_controller.go
sbryzak: Out of those two options I prefer the first one also, however I've been contemplating whether there might be another option besides these two, however the ideas I'm working through would require a redesign.  This might end up being dependent on the notification subsystem redesign too, @alexeykazakov will probably discuss this with you tomorrow during our weekly call.

sbryzak: Guys, I'm going to implement this using the new states property, however this doesn't mean we require a full refactor of all the existing usage of `Approved`, `Deactivated`, `VerificationRequired` properties right now.  We can use a hybrid approach here, making use of the `states` property only for the new `Deactivating` state, then followup after this with separate refactorings for the other properties.  I'll create a new story in Jira to track this.

pkg/controller/deactivation/deactivation_controller.go
sbryzak: Agreed we should start the deactivation timer from when the pre notification is sent out.  I'm currently working on the changes for this now, it's going to break a lot of our tests.

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +165,47 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+			return reconcile.Result{}, err
+		}
+
+		// Requeue so that the deactivation due time can be calculated after the notification has been sent
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	deactivatingCondition, found := condition.FindConditionByType(usersignup.Status.Conditions,
+		toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated)
+	if !found || deactivatingCondition.Reason != toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason {
```
alexeykazakov, line 195:
```suggestion
	if !found || deactivatingCondition.Status != corev1.ConditionTrue || deactivatingCondition.Reason != toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason {
```

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +165,47 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+		// If the UserSignup has been marked as deactivating, however the deactivating notification hasn't been
+		// created yet, then requeue - the notification should be created shortly
+		return reconcile.Result{Requeue: true}, nil
```
alexeykazakov, line 198: First. I think it's worth mentioning that the notification is created by the UserSignup controller when we set the state to "deactivating".
Second. I'm concerned about always requeuing without any delay if the notification condition is not in the expected state. What if there is some global problem with notification creation? Will we "DDoS" API server by requeuing requests in this case?
Should we either add some delay or even return an error so, the controller manager takes care of properly delaying the reconcile requests?
WDYT?

pkg/controller/usersignup/status.go
```diff
@@ -257,7 +258,38 @@ func (u *statusUpdater) setStatusDeactivationNotificationCreationFailed(userSign
...
+		})
+}
+
+func (u *statusUpdater) setStatusDeactivatingNotificationUserIsActive(userSignup *toolchainv1alpha1.UserSignup, _ string) error {
+	return u.updateStatusConditions(
+		userSignup,
+		toolchainv1alpha1.Condition{
+			Type:   toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated,
+			Status: corev1.ConditionFalse,
+			Reason: toolchainv1alpha1.UserSignupDeactivatingNotificationUserIsActiveReason,
```
alexeykazakov, line 277: This reason looks a bit confusing to me...
User is still active even if he/she is in the deactivating state until we actually deactivate the account. Isn't he?

pkg/controller/usersignup/status.go
```diff
@@ -285,17 +317,35 @@ func (u *statusUpdater) updateCompleteStatus(compliantUsername string) func(user
 				Message: message,
 			})
 
+		logger.Info(fmt.Sprintf("### userNameUpdated: %t  conditionUpdated: %t", usernameUpdated, conditionUpdated))
```
alexeykazakov, line 320: Zombie.
```suggestion
```

pkg/controller/usersignup/status.go
```diff
@@ -285,17 +317,35 @@ func (u *statusUpdater) updateCompleteStatus(compliantUsername string) func(user
...
+		logger.Info(fmt.Sprintf("### ResourceVersion before updating status in updateCompleteStatus: %s", userSignup.ResourceVersion))
+
+		/*err := u.Client.Get(context.TODO(), client.ObjectKey{
+			Namespace: userSignup.Namespace,
+			Name:      userSignup.Name,
+		}, userSignup)
+
+		if err != nil {
+			logger.Error(err, "### Error while reloading userSignup in updateCompleteStatus")
+		}*/
```
alexeykazakov, line 335: Another one.
```suggestion
```

pkg/controller/usersignup/status.go
```diff
@@ -285,17 +317,35 @@ func (u *statusUpdater) updateCompleteStatus(compliantUsername string) func(user
...
+
+		err := u.Client.Status().Update(context.TODO(), userSignup)
+		if err != nil {
+			logger.Error(err, "### Error while updating status in updateCompleteStatus")
+		}
+		logger.Info(fmt.Sprintf("### ResourceVersion after updating status in updateCompleteStatus: %s", userSignup.ResourceVersion))
+		return err
```
alexeykazakov, line 342:
```suggestion
		return u.Client.Status().Update(context.TODO(), userSignup)
```

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -158,12 +161,42 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+
+		if err := r.UpdateStatus(reqLogger, instance, r.setStatusDeactivatingNotificationCreated); err != nil {
+			reqLogger.Error(err, "Failed to update notification created status")
 			return reconcile.Result{}, err
 		}
 	}
 
 	if exists, err := r.ensureMurIfAlreadyExists(reqLogger, instance, banned); exists || err != nil {
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Error while ensuring MasterUserRecord exists, Error: %s", err))
```
alexeykazakov, line 198:
```suggestion
			reqLogger.Error(err, "Error while ensuring MasterUserRecord exists")
```

alexeykazakov (commented):
Overall looks good!
I have a few comments to address though.
And I don't see any new tests.. Or maybe I'm missing something?
Could you please carefully review the tests to make sure the new code is fully covered? Please  generate a test coverage report locally (codecov is unfortunately useless) to make sure it's all covered.

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -151,43 +151,123 @@ func TestReconcile(t *testing.T) {
...
+
+				// deactivating state should still be true
+				require.True(t, states.Deactivating(userSignupFoobar))
+
+				// deactivated state should still be false
+				require.False(t, userSignupFoobar.Spec.Deactivated)
+
+				t.Run("usersignup requeued after deactivating notification created for user", func(t *testing.T) {
+					// Set the notification status condition as sent
+					userSignupFoobar.Status.Conditions = []toolchainv1alpha1.Condition{
```
alexeykazakov, line 199: Shouldn't this condition be already true for `userSignupFoobar` set by the parent test?

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -151,43 +151,123 @@ func TestReconcile(t *testing.T) {
...
+
+						// Reload the userSignup
+						require.NoError(t, cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
+
+						// deactivating state should still be true
+						require.True(t, states.Deactivating(userSignupFoobar))
+
+						// deactivated state should now be true also
+						require.True(t, userSignupFoobar.Spec.Deactivated)
+
```
alexeykazakov, line 254: We can also check that the deactivation notification is created and the corresponding condition is true, right?

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +165,47 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+	if !states.Deactivating(usersignup) {
+		states.SetDeactivating(usersignup, true)
+
+		logger.Info("setting usersignup state to deactivating")
+		if err := r.client.Update(context.TODO(), usersignup); err != nil {
+			logger.Error(err, "failed to update usersignup")
+			return reconcile.Result{}, err
+		}
+
+		// Requeue so that the deactivation due time can be calculated after the notification has been sent
```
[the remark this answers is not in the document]
sbryzak: The sequence of events here are:

1. Deactivation controller sets the UserSignup state to deactivating if it's not already set (in line 181)
2. Reconciliation is requeued (line 190)
3. UserSignup controller picks up that the deactivating state has been set, and responds by a) creating a pre-deactivating notification for the user, and b) setting the "deactivating notification created" status condition to true.
4. Deactivation controller reconciles again, and if it doesn't find the notification created status condition as expected, requeues again, otherwise:
5. Deactivation controller calculates the amount of time that has passed since the deactivating notification was sent, based on the LastTransitionTime of the condition.  If enough time has now passed, it sets the UserSignup to deactivated.

For step 2) and 4), we could possibly introduce a delay here, perhaps something like 5 seconds might be sufficient.  The reason for the requeue though is correct according to the comment; the reconciliation cannot proceed until the status condition has been set and the time passed since the notification was sent can be calculated.

pkg/controller/deactivation/deactivation_controller.go
sbryzak: No problem, I've updated the comment as such.  As I mentioned above, it might be worth (and certainly wouldn't hurt) to put in a small delay between reconciles.  I suggested 5 seconds, however there's really no reason why this couldn't be much more conservative, even as high as 10 minutes wouldn't hurt here in my opinion.

pkg/controller/usersignup/status.go
sbryzak: That's correct, I was wondering if there might be a better reason string to use here, or perhaps just an empty string even.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
alexeykazakov: Yeah.. I actually got this flow when continued to read the code below after leaving this comment.
The flow should work but it's quite complicated and not hard to grok from the first glance. Can you please add this comment with detailed flow to the code (I think right here would be a good place)?

pkg/controller/deactivation/deactivation_controller.go
alexeykazakov: It's deactivation flow which normally happens once in 30 days. So, I think a 10 second delay would be totally fine :)

pkg/controller/usersignup/status.go
alexeykazakov: If you don't have a better name (I don't) then let's make sure this reason const has detailed comment explaining what this reason means (in API where we declare the const).

pkg/controller/deactivation/deactivation_controller.go
sbryzak: Adjusted delay to 10 seconds.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
sbryzak: Comment with detailed workflow has been added.

pkg/controller/usersignup/status.go
sbryzak: I've been trying to think of a better name... let's discuss during our call tomorrow.

pkg/controller/deactivation/deactivation_controller_test.go
sbryzak: As discussed, this is set by the UserSignup controller.

pkg/controller/deactivation/deactivation_controller_test.go
sbryzak: As above.

pkg/controller/usersignup/status.go
sbryzak: Resolved as discussed.

alexeykazakov (approved):

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -178,12 +181,40 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
 		if err := r.updateStatus(logger, userSignup, r.setStatusDeactivationNotificationUserIsActive); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
 
+	// If the usersignup is not banned and not within the pre-deactivation period then ensure the deactivation notification
+	// status is set to false. This is especially important for cases when a user is deactivated and then reactivated
+	// because the status is used to trigger sending of the notification. If a user is reactivated a notification should
+	// be sent to the user again.
+	if !banned && !states.Deactivating(userSignup) {
```
MatousJobanek, line 195: Shouldn't we also check that the user is not deactivated? Otherwise, we will reset the status also for deactivated users, which we don't really need, do we?

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -178,12 +181,40 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+		if err := r.updateStatus(logger, userSignup, r.setStatusDeactivatingNotificationNotInPreDeactivation); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	if states.Deactivating(userSignup) && condition.IsNotTrue(userSignup.Status.Conditions,
+		toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated) {
```
MatousJobanek, line 202: Based on our experience with the outdated values in the `status` part of the `UserSignup` resource I'm wondering if this is a safe way to do - to decide what to do based on the value in `UserSignup.status`.
Anyway, I don't think that it's so critical because:
* we saw the issue with the outdated values during provisioning when there are a lot of changes in both resources `UserSignup` and `MUR` which triggers many reconciles. But this is not the case - there should be so many changes, so I believe that we should not get the outdated value so often.
* in the worst case, we will send the notification twice, which is not a big deal :-)

MatousJobanek (changes requested):
Good job :+1: I like the code - it's clear and easy to follow. It also nicely solve the problem :1st_place_medal: 
I added a few comments.
Apart from what I wrote there, I really miss some unit test for the part in UserSignup controller - to verify that it creates the "Deactivating" notification CR when needed. Sorry if I it.

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +167,60 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+		// 2. Reconciliation is requeued
+		// 3. UserSignup controller picks up that the deactivating state has been set, and responds by:
+		//		a) creating a pre-deactivating notification for the user, and
+		//		b) setting the "deactivating notification created" status condition to true.
+		// 4. This  controller reconciles again, and if it doesn't find the notification created status condition as
+		//    expected, requeues again, otherwise:
+		// 5. This controller calculates the amount of time that has passed since the deactivating notification was sent,
+		//  based on the LastTransitionTime of the condition. If enough time has now passed, it sets the UserSignup to deactivated.
+
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(10) * time.Second}, nil
```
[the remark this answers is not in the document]
MatousJobanek: stupid me :man_facepalming: I just realized that we don't watch the UserSignup CRD in the deactivation controller

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +167,60 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+	}
+
+	deactivatingCondition, found := condition.FindConditionByType(usersignup.Status.Conditions,
+		toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated)
+	if !found || deactivatingCondition.Status != corev1.ConditionTrue ||
+		deactivatingCondition.Reason != toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason {
+		// If the UserSignup has been marked as deactivating, however the deactivating notification hasn't been
+		// created yet, then requeue - the notification should be created shortly by the UserSignup controller
+		// once the "deactivating" state has been set
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(10) * time.Second}, nil
```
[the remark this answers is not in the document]
MatousJobanek: I can ignore this comment https://github.com/codeready-toolchain/host-operator/pull/393#discussion_r623835974

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
MatousJobanek: But this actually brings a question - since we modify the `UserSignup` resource several times, and we also rely on the status of the `UserSignup`, shouldn't we start watching that CRD as well?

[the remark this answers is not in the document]
MatousJobanek: Thinking about it more and more, I believe that we should start watching the `UserSignup` CRD as well (as the secondary resources). This controller relies on the changes in the CRs, and it is actually an important part of the complex lifecycle of the `UserSignup` CRD. 
This would however mean that we would need to add a mapper between `UserSignup` & `MUR` to the watcher, but I really believe that it's worth it. In addition, such a mapper should be easy - just read the `UserSignup.Status.CompliantUsername`
WDYT @alexeykazakov @sbryzak

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: Now I understand - the deactivating status will be always set even when the `UserSignup` is deactivated. But it's not really clear from the code. It looks like two different states.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
alexeykazakov: This is exactly what is bothering me here (I mentioned it to you Matous when we were talking about this): not obvious relation between UserSignup and deactivation controllers.
But yes, I think it would make sense to start watching UserSignup CRD here too. However maybe it's better to merge this PR first and then create a separate one to switch to a watcher instead relying on re-queuing. But I would leave it to Shane to decide (one PR or two). We might face unexpected issues with the watcher so in that case we would already have a working solution and could just abandon that idea in worst case scenario.

pkg/controller/usersignup/usersignup_controller.go
alexeykazakov: I think when we move `Spec.Deactivated` to `Spec.State = [...]` we will need to auto delete `deactivating` from the state when setting it to `deactivated`.

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +167,60 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
 	logger.Info("user account time values", "deactivation timeout duration", deactivationTimeout, "provisionedTimestamp", provisionedTimestamp)
 
 	timeSinceProvisioned := time.Since(provisionedTimestamp.Time)
-	if timeSinceProvisioned < deactivationTimeout {
+
+	deactivatingNotificationTimeout := time.Duration((deactivationTimeoutDays-config.Deactivation.DeactivatingNotificationDays)*24) * time.Hour
```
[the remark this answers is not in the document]
sbryzak: The default is 3, set here:

https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/hostoperatorconfig_types.go#L66

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
sbryzak: We are not waiting for a change in `UserSignup.Spec` though, we are waiting for a change to `UserSignup.Status` - will a status change still trigger another reconcile?

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
sbryzak: Same question as above.

pkg/controller/deactivation/deactivation_controller.go
```diff
@@ -149,10 +167,60 @@ func (r *ReconcileDeactivation) Reconcile(request reconcile.Request) (reconcile.
...
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(10) * time.Second}, nil
+	}
+
+	deactivationDueTime := deactivatingCondition.LastTransitionTime.Time.Add(time.Duration(config.Deactivation.DeactivatingNotificationDays*24) * time.Hour)
+
+	if time.Now().Before(deactivationDueTime) {
 		// It is not yet time to deactivate so requeue when it will be
-		requeueAfterExpired := deactivationTimeout - timeSinceProvisioned
-		logger.Info("requeueing request", "RequeueAfter", requeueAfterExpired, "Expected deactivation date/time", time.Now().Add(requeueAfterExpired).String())
+		requeueAfterExpired := time.Until(deactivationDueTime) + time.Duration(1*time.Minute)
```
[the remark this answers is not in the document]
sbryzak: I can't recall why I added this, I've removed it and hopefully the tests will still be fine. :)

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -308,9 +412,21 @@ func userSignupWithEmail(username, email string) *toolchainv1alpha1.UserSignup {
 	}
 }
 
+func assertThatUserSignupDeactivating(t *testing.T, cl *test.FakeClient, name string, expected bool) {
+	userSignup := &toolchainv1alpha1.UserSignup{}
+	err := cl.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: operatorNamespace}, userSignup)
+	require.NoError(t, err)
+	require.Equal(t, expected, states.Deactivating(userSignup))
+}
+
```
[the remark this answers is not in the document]
sbryzak: I had used it previously, but then decided to do things differently.  Removed for now.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: This is the same flow that we use for deactivation also, further down in the code.  I'm concerned about the outdated resource issue too, and I think we should look at our general methodology for how we update resources within our controllers and possibly refactor to reduce the number of distinct updates.  However for now I don't have an easy fix for this, and as you say in the worst case we will send the notification twice.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
alexeykazakov: It depends on the watcher you use. We can watch for new generations only (GenerationChangedPredicate) or for status only when the Status changed (OnlyUpdateWhenGenerationNotChanged: https://github.com/codeready-toolchain/toolchain-common/blob/a2981455c932c8d35d3106711f3ab60600f723d9/pkg/predicate/predicate.go#L17) or for everything if we want.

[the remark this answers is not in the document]
sbryzak: Let's please get this one merged first, before looking at further refactorings.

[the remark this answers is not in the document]
sbryzak: By the way, I agree with both of you that we should probably watch UserSignup also, instead of relying on the requeue mechanism.

pkg/controller/usersignup/usersignup_controller.go
sbryzak: @alexeykazakov will include this in the state manager when I work on [CRT-1006](https://issues.redhat.com/browse/CRT-1006)

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
MatousJobanek: I would rather keep that in host-operator repo than in api. Or do you really want to store all default configurations in api repo?

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
MatousJobanek: As I see it:
* it would simplify the flow
* make the code more clear
* follow the convention that we should watch secondary resources
* maybe also improve the stability of the flow/controller.

Is there any Jira issue to track this?

pkg/controller/usersignup/usersignup_controller.go
MatousJobanek: Yeah, I agree with you. 
My guess is that we should not rely on the values set in status part inside the controller - to decide what to do next. Maybe, we should rather use annotations instead.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
sbryzak: I've created an issue for this: https://issues.redhat.com/browse/CRT-1070

pkg/controller/deactivation/deactivation_controller_test.go
```diff
@@ -314,3 +418,8 @@ func assertThatUserSignupDeactivated(t *testing.T, cl *test.FakeClient, name str
 	require.NoError(t, err)
 	require.Equal(t, expected, userSignup.Spec.Deactivated)
 }
+
+func newHostOperatorConfigWithReset(t *testing.T, options ...test.HostOperatorConfigOption) *toolchainv1alpha1.HostOperatorConfig {
+	t.Cleanup(hostoperatorconfig.Reset)
+	return test.NewHostOperatorConfig(options...)
+}
```
MatousJobanek, line 425: btw, there is already one function doing the very same thing https://github.com/codeready-toolchain/host-operator/blob/master/test/config.go#L11-L14 is there any reason for adding it here?

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1770,6 +1843,84 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+			},
+			CompliantUsername: "freja-johanssen",
+		},
+	}
+
+	// given
+	mur := murtest.NewMasterUserRecord(t, "edward-jones", murtest.MetaNamespace(test.HostOperatorNs))
+	mur.Labels = map[string]string{
+		v1alpha1.MasterUserRecordOwnerLabelKey: userSignup.Name,
+		v1alpha1.UserSignupStateLabelKey:       "approved",
```
MatousJobanek, line 1874: Why do you add the `UserSignupStateLabelKey` label to `MUR` - that label belongs to `UserSignup` CRs

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1770,6 +1843,84 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+	}
+
+	userSignup.Labels["toolchain.dev.openshift.com/approved"] = "true"
```
MatousJobanek, line 1877: We don't use this label anywhere

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1770,6 +1843,84 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+				},
+				{
+					Type:   v1alpha1.UserSignupApproved,
+					Status: v1.ConditionTrue,
+					Reason: "ApprovedAutomatically",
+				},
+			},
+			CompliantUsername: "freja-johanssen",
+		},
+	}
```
MatousJobanek, line 1868: you could use functions from https://github.com/codeready-toolchain/host-operator/blob/master/test/usersignup.go to simplify it

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1770,6 +1843,84 @@ func TestUserSignupDeactivatingWhenMURExists(t *testing.T) {
...
+	assert.Equal(t, "approved", userSignup.Labels[v1alpha1.UserSignupStateLabelKey])
+
+	notifications := &v1alpha1.NotificationList{}
+	err = r.client.List(context.TODO(), notifications)
+
+	require.Len(t, notifications.Items, 1)
+
+	require.Equal(t, "userdeactivating", notifications.Items[0].Spec.Template)
+	require.Equal(t, userSignup.Name, notifications.Items[0].Spec.UserID)
+
```
MatousJobanek, line 1899: it would be also cool to verify that the MUR hasn't been deleted yet.

pkg/controller/usersignup/usersignup_controller_test.go
```diff
@@ -1562,6 +1630,11 @@ func TestUserSignupReactivateAfterDeactivated(t *testing.T) {
 				Status: v1.ConditionTrue,
 				Reason: "Deactivated",
 			},
+			v1alpha1.Condition{
+				Type:   v1alpha1.UserSignupUserDeactivatingNotificationCreated,
+				Status: v1.ConditionFalse,
+				Reason: "UserNotInPreDeactivation",
+			},
```
MatousJobanek, line 1637: Here you verify that the status condition is set when the UserSignup is reactivated, but you don't set the condition to 
```
                {
			Type:   toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated,
			Status: corev1.ConditionTrue,
			Reason: toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason,
		}
```
in the `// given` part - analogically as it's done for `UserSignupUserDeactivatedNotificationCreated` condition:
https://github.com/sbryzak/host-operator/blob/6cb26692b97bd178314c1b0afabe12ae6d219d54/pkg/controller/usersignup/usersignup_controller_test.go#L1596-L1600 
to verify that the condition will be reset

MatousJobanek (commented):
Thanks for adding the unit test that verifies the creation of the Notification CR. It would be also cool to add a test that checks that the status is not set when the creation of the CR fails. But it's probably just a detail.

pkg/controller/deactivation/deactivation_controller.go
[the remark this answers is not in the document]
MatousJobanek: Thanks :+1:

MatousJobanek:
@sbryzak in case you missed my last review https://github.com/codeready-toolchain/host-operator/pull/393#pullrequestreview-650157648

MatousJobanek:
@sbryzak pinging you again :-)

PR #394, opened 2021-03-17
Assigning apiEndpoint in MemberStatusStatus 

ranakan19:
Assigning apiEndpoint in MemberStatusStatus for Members. 

fix wrt: https://issues.redhat.com/browse/CRT-971
related PR: https://github.com/codeready-toolchain/toolchain-common/pull/155

end-to-test updated for apiEndpoint - https://github.com/codeready-toolchain/toolchain-e2e/pull/265


pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -323,6 +323,7 @@ func (r *ReconcileToolchainStatus) membersHandleStatus(logger logr.Logger, toolc
 			logger.Error(err, fmt.Sprintf("cannot find memberstatus resource in namespace %s in cluster %s", memberCluster.OperatorNamespace, memberCluster.Name))
 			memberStatusNotFoundCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusNotFoundReason, err.Error())
 			memberStatus := customMemberStatus(*memberStatusNotFoundCondition)
+			memberStatus.ApiEndpoint = memberCluster.APIEndpoint
```
MatousJobanek, line 326: Since the APIEndpoint is taken from  `ToolchainCluster` CR here in host-operator, then I guess we should move the parameter from `MemberStatus` CRD into `Member` type in `ToolchainCluster` CRD: https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go#L180-L190
The reason is that `MemberStatus` CRD is used in member operator to give information about the state of the member cluster to the host operator - that information is then copied into `ToolchainStatus` CR. But because we don't need that parameter in member operator at all, we should remove it from the `MemberStatus` CRD

MatousJobanek (commented):
Looks good, I'm just afraid that we used an inappropriate place for the `APIEndpoint` parameter.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
alexeykazakov: It's not like we don't need it in member (it could be a good place to store it there but we don't have to) we rather don't have it in member. I didn't realize we do a ToolchainStatus.Status.Member[] which wraps individual MemberStatuses though.. So, yeah, +1 to moving it to https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go#L180-L190

ranakan19: moved apiEndpoint to Member in https://github.com/codeready-toolchain/api/pull/221

ranakan19: added the change now, @alexeykazakov and @MatousJobanek

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -347,9 +347,21 @@ func (r *ReconcileToolchainStatus) membersHandleStatus(logger logr.Logger, toolc
...
 }
 
+func assignApiEndpoint(toolchainStatus *toolchainv1alpha1.ToolchainStatus, memberClusters []*cluster.CachedToolchainCluster) {
+	for _, memberCluster := range memberClusters {
+		for index, member := range toolchainStatus.Status.Members{
+			if member.ClusterName == memberCluster.Name {
+				toolchainStatus.Status.Members[index].ApiEndpoint = memberCluster.APIEndpoint
+				break
+			}
+		}
```
MatousJobanek, line 362: It's just a proposal, but do you think that we could move this part into `compareAndAssignMemberStatuses` method and do the assignment there? https://github.com/codeready-toolchain/host-operator/blob/57ec214efcf9269277c617dfc714784fb6edc862/pkg/controller/toolchainstatus/toolchainstatus_controller.go#L424-L431

It's pretty hard to follow the association and the source of names of the member clusters

MatousJobanek (commented):
Cool, this looks good.
Could you please also modify the e2e tests to check that the ApiEndpoint parameter is set...?

pkg/controller/toolchainstatus/toolchainstatus_controller.go
ranakan19: i was worried that it doesn't add to the confusion to move to this function, considering the function name says to update memberStatuses, while apiEndpoint is in Member

go.mod
```diff
@@ -30,6 +35,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common => github.com/ranakan19/toolchain-common v0.0.0-20210318182646-b0141bf2d375 // using 'github.com/ranakan19/codeready-toolchain@CRT971
```
alexeykazakov, line 38: Just a reminder to remove this before merging.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -345,11 +345,19 @@ func (r *ReconcileToolchainStatus) membersHandleStatus(logger logr.Logger, toolc
 		members[memberCluster.Name] = memberStatus
 	}
 
-	// add member cluster statuses to toolchainstatus
-	ready = compareAndAssignMemberStatuses(logger, toolchainStatus, members) && ready
+	// add member cluster statuses to toolchainstatus, and assign apiEndpoint in members
+	ready = compareAndAssignMemberStatuses(logger, toolchainStatus, members,memberClusters) && ready
```
alexeykazakov, line 349: What IDE are you using? Configure it to use `go fmt` to automatically format the code :)

alexeykazakov (commented):
Looks good! Just need to add the API check to e2e tests as Matous mentioned.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
MatousJobanek: You are right, it's probably more appropriate at the place where you put it. In addition, the code that I referenced is only for the creation of the new Members that are not part of the status yet and doesn't update of the already existing ones. Your solution will update both of them. :+1:

pkg/controller/toolchainstatus/toolchainstatus_controller.go
ranakan19: i am using goLand, will do

ranakan19:
> Looks good! Just need to add the API check to e2e tests as Matous mentioned.

@alexeykazakov @MatousJobanek e2e tests updated here - https://github.com/codeready-toolchain/toolchain-e2e/pull/265

alexeykazakov (approved):
I'm approving the PR but please format the code (`gofmt -s -w .`) and switch to common master in go.mod.

MatousJobanek:
>  please format the code (gofmt -s -w .) 

or you can use `make format-go-code`

MatousJobanek (approved):
Cool :+1:

go.mod
```diff
@@ -1,14 +1,16 @@
...
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20210224173547-fe6930d7b5df
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v0.1.0
 	github.com/gofrs/uuid v3.3.0+incompatible
-	github.com/golangci/golangci-lint v1.37.1 // indirect
+	github.com/google/go-cmp v0.5.4 // indirect
 	github.com/mailgun/mailgun-go/v4 v4.1.3
+	github.com/onsi/ginkgo v1.14.2 // indirect
+	github.com/onsi/gomega v1.10.4 // indirect
```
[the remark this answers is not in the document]
ranakan19: done

PR #375, opened 2021-02-18
Split ClusterResourceQuotas into different groups

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-941

Related e2e PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/238

This PR splits up our ClusterResourceQuotas into different groups as a workaround for an issue where 409 conflict errors occur when creating apps via dev console. It should make 409 situations less likely to happen.

Checklist:
- [x] Manually test that the 409 errors are less likely to occur with these changes
- [x] Negligible performance difference when comparing with and without this PRs changes
- [x] Unit tests updated
- [x] e2e tests updated

PR #398, opened 2021-03-25
Create base tier

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-985

Due to multiple problems with different OpenShift operators not working well in clusters with large numbers of namespaces we want to reduce the namespace count per user. We need to get rid of the -code namespace and keep -dev and -stage only. This PR introduces a new tier `base` that is the same as the `basic` tier except for these changes:

- 'code' namespace removed
~~- CRQ requests.storage: 25Gi~~
- CRW NetworkPolicy exists in 'dev' namespace

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/268


deploy/templates/nstemplatetiers/base/cluster.yaml
```diff
@@ -0,0 +1,142 @@
...
+    name: for-${USERNAME}-compute
+  spec:
+    quota:
+      hard:
+        limits.cpu: 20000m
+        limits.memory: 7Gi
+        limits.ephemeral-storage: 7Gi
+        requests.cpu: 1750m
+        requests.memory: 7Gi
+        requests.storage: 25Gi
```
alexeykazakov, line 18: We need to set 25Gi for the existing `basic` tier only. Not for the new one.
Users provisioned to the new tier won't have workspaces in two namespaces.
But existing users in basic tier may have two namespaces with che workspaces.

And one more thing. I just realized that we actually have che test accounts in basicdeactivationdisabled tier too.
So, let's migrate this tier too:
- create a new tier `basedeactivationdisabled`
- set the storage quota to 25 in the existing `basicdeactivationdisabled`
- then we will migrate the users and kill the existing `basicdeactivationdisabled` as well as `basic`.

rajivnathan: Ah, that makes more sense. I'll make the changes.

rajivnathan: Done in a9ddde029fb06e08c35476a48b3a4f8d03976116

alexeykazakov (approved):
Looks good!

make/go.mk
```diff
@@ -39,6 +39,7 @@ clean-metadata:
 generate-metadata: clean-metadata
 	@echo "generating templates metadata for manifests in $(NSTEMPLATES_BASEDIR)" 
 	@$(foreach tmpl,$(wildcard $(NSTEMPLATES_FILES)),$(call git_commit,$(tmpl),$(NSTEMPLATES_BASEDIR),metadata.yaml))
+	@cat $(NSTEMPLATES_BASEDIR)/metadata.yaml
```
MatousJobanek, line 42: Do we still need that?

MatousJobanek (approved):
Nice :+1:

make/go.mk
rajivnathan: Nope, I was debugging an error I saw but I cannot reproduce it anymore so I will just remove it now.

rajivnathan: Done https://github.com/codeready-toolchain/host-operator/pull/398/commits/8175aa7d049e469f54424f339896ffb252f1ade9

MatousJobanek:
> CRW NetworkPolicy exists in 'dev' namespace

as was mentioned in the slack channel - the NetworkPolicy needs to be added to the dev namespace also in the current `basic` tier - feel free to address it either in this PR or in the following one.

MatousJobanek (approved):
Good job :+1: :1st_place_medal:

deploy/templates/nstemplatetiers/base/cluster.yaml
```diff
@@ -0,0 +1,142 @@
...
+        openshift.io/requester: ${USERNAME}
+      labels: null
+- apiVersion: quota.openshift.io/v1
+  kind: ClusterResourceQuota
+  metadata:
+    name: for-${USERNAME}-bc
+  spec:
+    quota:
+      hard:
+        count/buildconfigs.build.openshift.io: "30"
```
xcoulon, line 99: I wonder if we need a specific ClusterResourceQuota for each kind of resource? (secrets, buildconfigs, configmaps, etc.). Could we use a single cluster quota resource to specify all the counts at once instead?

xcoulon (approved):

deploy/templates/nstemplatetiers/base/cluster.yaml
rajivnathan: We had to split the ClusterResourceQuotas because of issue https://issues.redhat.com/browse/CRT-941. (https://github.com/codeready-toolchain/host-operator/pull/375)

PR #400, opened 2021-03-31
refactor counter tests

xcoulon:
use func to specify all the base values, including
the number of master user records.

this change will allow for specifying more values/metrics
in the future ;)


pkg/counter/cache.go
```diff
@@ -16,14 +17,20 @@ var log = logf.Log.WithName("counter_cache")
...
 }
 
 // Counts is type that contains number of MURs and number of UserAccounts per member cluster
 type Counts struct {
-	MasterUserRecordCount        int
+	// Total number of Master User Accounts
+	MasterUserRecordCount int
+	// User Accounts per Clusters (indexed by cluster name)
 	UserAccountsPerClusterCounts map[string]int
+	// Activations per Users (number of users indexed by their number of activations)
```
alexeykazakov, line 32: It's not activations per user. It's number of users per activation count. Correct?

alexeykazakov (approved):
Nice improvement. But it's not only about tests, is it?

pkg/counter/cache.go
xcoulon: ah right, I cherry-picked a change to make a separate PR and:
- I did not realize this file was affected in this PR (it should not)
- I forgot to edit the commend (the initial var name was misleading, as you noted)

xcoulon: removed in f9a2b08

PR #403, opened 2021-04-09
add restored check and send notification

ranakan19:
fix wrt: https://issues.redhat.com/browse/CRT-1002


pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -229,6 +234,29 @@ func (r *ReconcileToolchainStatus) notificationCheck(reqLogger logr.Logger, tool
...
+func (r *ReconcileToolchainStatus) restoredCheck(reqLogger logr.Logger, toolchainStatus *toolchainv1alpha1.ToolchainStatus) error {
+	// If the current ToolchainStatus:
+	// a) Was not ready before,
+	// b) notification was sent out due prolonged not ready status, and
+	// c) status is now restored
+	// send a notification to the admin mailing list
+	c, found := condition.FindConditionByType(toolchainStatus.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	if found && c.Status == corev1.ConditionFalse {
+		c, found = condition.FindConditionByType(toolchainStatus.Status.Conditions, toolchainv1alpha1.ToolchainStatusUnreadyNotificationCreated)
+		if found && c.Status == corev1.ConditionTrue {
```
alexeykazakov, line 246: Let's add `IsFalse(conditions []toolchainv1alpha1.Condition, conditionType toolchainv1alpha1.ConditionType) bool` next to https://github.com/codeready-toolchain/toolchain-common/blob/91c4f9716eff40a84239948f1c13e4f477146ce4/pkg/condition/condition.go#L63
so we can use it here as:
```suggestion
	if condition.isFalse(toolchainStatus.Status.Conditions, toolchainv1alpha1.ConditionReady) &&
	   condition.isTrue(toolchainStatus.Status.Conditions, toolchainv1alpha1.ToolchainStatusUnreadyNotificationCreated) {
```

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -400,6 +428,48 @@ func (r *ReconcileToolchainStatus) sendToolchainStatusUnreadyNotification(logger
...
+
+	notification := &toolchainv1alpha1.Notification{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("toolchainstatus-ready-%s", tsValue),
+			Namespace: toolchainStatus.Namespace,
+		},
+		Spec: toolchainv1alpha1.NotificationSpec{
+			Recipient: r.config.GetAdminEmail(),
+			Subject:   adminReadyNotificationSubject,
+			Content:   "<div><pre><code>" + string(statusYaml) + "</code></pre></div>", // wrap with div/pre/code tags so the formatting remains intact in the delivered mail
```
alexeykazakov, line 455: We don't need to send the status body. Just a message stating that the status is back to Ready is enough.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -400,6 +428,48 @@ func (r *ReconcileToolchainStatus) sendToolchainStatusUnreadyNotification(logger
...
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(toolchainStatus, notification, r.scheme); err != nil {
+		logger.Error(err, "Failed to set owner reference for toolchain status unready notification resource")
```
alexeykazakov, line 460:
```suggestion
		logger.Error(err, "Failed to set owner reference for toolchain status restored notification resource")
```

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -229,6 +234,29 @@ func (r *ReconcileToolchainStatus) notificationCheck(reqLogger logr.Logger, tool
...
+	if found && c.Status == corev1.ConditionFalse {
+		c, found = condition.FindConditionByType(toolchainStatus.Status.Conditions, toolchainv1alpha1.ToolchainStatusUnreadyNotificationCreated)
+		if found && c.Status == corev1.ConditionTrue {
+			if err := r.sendToolchainStatusRestoredNotification(reqLogger, toolchainStatus); err != nil {
+				reqLogger.Error(err, "Failed to create toolchain status restored notification")
+
+				// set the failed to create notification status condition
+				return r.wrapErrorWithStatusUpdate(reqLogger, toolchainStatus,
+					r.setStatusReadyNotificationCreationFailed, err,
+					"Failed to create user deactivation notification")
```
alexeykazakov, line 253:
```suggestion
					"Failed to create toolchain status restored notification")
```

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -516,6 +586,20 @@ func (r *ReconcileToolchainStatus) setStatusUnreadyNotificationCreationFailed(re
...
+func (r *ReconcileToolchainStatus) setStatusReadyNotificationCreationFailed(reqLogger logr.Logger,
+	toolchainStatus *toolchainv1alpha1.ToolchainStatus, message string) error {
+
+	return r.updateStatusConditions(
+		reqLogger,
+		toolchainStatus,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionTrue,
+			Reason:  "Status Ready Notification creation failed",
```
alexeykazakov, line 598: 1. The ready status should be set to false if there is a problem. Not to true.
2. Reason is a machine readable string. It's like a type. For human friendly details about the error we use messages:
```suggestion
			Status:  corev1.ConditionFalse,
			Reason:  toolchainv1alpha1.ToolchainStatusRestoredNotificationCRCreationFailedReason,
```

alexeykazakov (commented):
The overall approach is correct! There are a few comments to address though.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -400,6 +428,48 @@ func (r *ReconcileToolchainStatus) sendToolchainStatusUnreadyNotification(logger
 	return nil
 }
 
+func (r *ReconcileToolchainStatus) sendToolchainStatusRestoredNotification(logger logr.Logger,
```
rajivnathan, line 431: It looks like there's some opportunity for reuse between this function and `sendToolchainStatusUnreadyNotification`. It looks like only the name, subject and content are different so maybe we can create a new function eg. `sendToolchainStatusNotification` that accepts those as params and then both `sendToolchainStatusUnreadyNotification` and `sendToolchainStatusRestoredNotification` can use this function. The logic for validating the email address, removing the mangedfields, etc. can be in the common `sendToolchainStatusNotification` function.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -195,6 +196,10 @@ func (r *ReconcileToolchainStatus) aggregateAndUpdateStatus(reqLogger logr.Logge
 
 		return r.setStatusNotReady(reqLogger, toolchainStatus, fmt.Sprintf("components not ready: %v", unreadyComponents))
 	}
+	err := r.restoredCheck(reqLogger, toolchainStatus)
```
rajivnathan, line 199: I initially thought this would be a good place to do this check, but maybe it'd be better at the beginning of the `setStatusReady` function instead since it's basically tied to that function?

rajivnathan (commented):
Nice work! Added a couple of comments for consideration

pkg/controller/toolchainstatus/toolchainstatus_controller.go
ranakan19: good point, i was initially setting the condition type ready's status as true even if sending the notification failed. Because the notification is sent only when the status is back to ready. But @alexeykazakov also commented saying not to set the status to true [here](https://github.com/codeready-toolchain/host-operator/pull/403#discussion_r611270771), so then it makes even more sense to move this within setStatusReady

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -373,30 +402,41 @@ func (r *ReconcileToolchainStatus) sendToolchainStatusUnreadyNotification(logger
 	}
```
alexeykazakov, line 402: Move it to the line 408 where you initialize the contentString with the actual value in case of unready status. We do not need to marshal the entire status in case of ready status. It's not used there.

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1035,9 +1055,9 @@ func assertToolchainStatusNotificationCreated(t *testing.T, fakeClient *test.Fak
 	return &notifications.Items[0]
 }
 
-func assertToolchainStatusNotificationNotCreated(t *testing.T, fakeClient *test.FakeClient) {
+func assertToolchainStatusNotificationNotCreated(t *testing.T, fakeClient *test.FakeClient, notificationType string) {
 	var notification toolchainv1alpha1.Notification
-	err := fakeClient.Get(context.Background(), test.NamespacedName(test.HostOperatorNs, "toolchainstatus-unready"),
+	err := fakeClient.Get(context.Background(), test.NamespacedName(test.HostOperatorNs, notificationType),
```
alexeykazakov, line 1060: It seems that we actually had a bug here. We can't just check if there is a notification with such name (== notificationType) because the actual notification resource name is in `<notificationType>-<timestamp>` format.
Instead of using `Get` we have to use `List` to list all the notifications and check their names if there any notification with the name which start with: `<notificationType>-`. There should not be any notification with such a name.
Otherwise this assertion will always pass.

alexeykazakov (commented):
Looks good overall! I have a few comments to address though.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
ranakan19: i don't understand what you are suggesting, i removed the function "sendToolchainStatusUnreadyNotification" to have one merged function sendToolchainStatusNotification.

alexeykazakov: I'm suggesting to move the code from lines 397-402:
```
	toolchainStatus = toolchainStatus.DeepCopy()
	toolchainStatus.ManagedFields = nil // we don't need these managed fields in the notification
	statusYaml, err := yaml.Marshal(toolchainStatus)
	if err != nil {
		return err
	}
```
to:
```
	if status == unreadyStatus {
		toolchainStatus = toolchainStatus.DeepCopy()
		toolchainStatus.ManagedFields = nil // we don't need these managed fields in the notification
		statusYaml, err := yaml.Marshal(toolchainStatus)
		if err != nil {
			return err
		}
		contentString = "<div><pre><code>" + string(statusYaml) + "</code></pre></div>"
		subjectString = adminUnreadyNotificationSubject
	} else if status == restoredStatus {
		contentString = "<div><pre>ToolchainStatus is back to ready status.</pre></div>"
		subjectString = adminRestoredNotificationSubject
	} else {
		return fmt.Errorf("invalid ToolchainStatusNotification status type: %s", status)
	}
```
or even:
```
	switch status {
	case unreadyStatus:
		toolchainStatus = toolchainStatus.DeepCopy()
		toolchainStatus.ManagedFields = nil // we don't need these managed fields in the notification
		statusYaml, err := yaml.Marshal(toolchainStatus)
		if err != nil {
			return err
		}
		contentString = "<div><pre><code>" + string(statusYaml) + "</code></pre></div>"
		subjectString = adminUnreadyNotificationSubject
	case restoredStatus:
		contentString = "<div><pre>ToolchainStatus is back to ready status.</pre></div>"
		subjectString = adminRestoredNotificationSubject
	default:
		return fmt.Errorf("invalid ToolchainStatusNotification status type: %s", status)
	}
```
So, we don't marshal the status yaml when we don't need it.

alexeykazakov (approved):
Nice work!

ranakan19:
@alexeykazakov added the environment in the restored notification body. Was mentioned in the story, i had missed it before.

pkg/controller/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -406,10 +407,11 @@ func (r *ReconcileToolchainStatus) sendToolchainStatusNotification(logger logr.L
 		if err != nil {
 			return err
 		}
-		contentString = "<div><pre><code>" + string(statusYaml) + "</code></pre></div>"
+		contentString = "<div><pre><code>" + string(statusYaml) + "</code></pre></div>" // wrap with div/pre/code tags so the formatting remains intact in the delivered mail
 		subjectString = adminUnreadyNotificationSubject
 	case restoredStatus:
-		contentString = "<div><pre>ToolchainStatus is back to ready status.</pre></div>" // wrap with div/pre/code tags so the formatting remains intact in the delivered mail
+		environment = toolchainStatus.Status.RegistrationService.Health.Environment
```
alexeykazakov, line 413: `toolchainStatus.Status.RegistrationService.Health.Environment` is useless here.
The name of this field is a bit misleading. It's actually not "environment" but rather "mode". So both our stage and prod is running in "prod" mode. Vs. "developer" or "test" modes.
What we actually need is the host cluster api server URL or the host cluster name...

alexeykazakov: Also, if we add this environment thing then we need to add it to both notifications. Unready and restored.

ranakan19: i believe member apiEndpoint would be fine as well, right?

alexeykazakov: There can be multiple member clusters. We are interested in the host cluster here but not sure if it's easy to get it from this code...
I would leave it as is (without mentioning the env) and think about improving it later.

PR #406, opened 2021-04-20
refactor counter test setup

xcoulon:
Initializing the counters from a `ToolchainStatus` resource, which
can be created on-the-fly or passed as-is (and used elsewhere in the test)


test/counter.go
```diff
@@ -19,6 +19,16 @@ import (
...
 
+func ToolchainStatus(baseValues ...BaseValue) *v1alpha1.ToolchainStatus {
+	toolchainStatus := &v1alpha1.ToolchainStatus{
+		Status: v1alpha1.ToolchainStatusStatus{},
+	}
+	for _, apply := range baseValues {
+		apply(toolchainStatus)
+	}
+	return toolchainStatus
+}
```
MatousJobanek, line 30: TBH, this is very confusing - there is already `NewToolchainStatus` https://github.com/codeready-toolchain/host-operator/blob/master/test/toolchainstatus.go#L65-L76 function accepting different option type 
Also, when you take a look at your changes, sometimes you use:
```
NewToolchainStatus(WithHost(WithMasterUserRecordCount(20)))
``` 
but in other cases
```
ToolchainStatus(MasterUserRecords(2), UserAccountsForCluster(test.MemberClusterName, 2))
```
For me, it was hard to understand the difference and why do we have two similar ways of creating `ToolchainStatus` CR

MatousJobanek (commented):
I agree that the original functions:
```
InitializeCounter
InitializeCounterWithBaseValues
InitializeCounterWithoutReset
```
weren't clear/transparent, but maybe it would be enough to just rename them and to not require the client to be passed when it's not needed

pkg/controller/usersignup/approval_test.go
```diff
@@ -36,6 +36,8 @@ func TestGetClusterIfApproved(t *testing.T) {
 				MaxUsersNumber(2000, PerMemberCluster("member1", 1000)).
 				ResourceCapThreshold(80, PerMemberCluster("member1", 70)))
 		fakeClient := NewFakeClient(t, toolchainStatus, hostOperatorConfig)
+		InitializeCounters(t, fakeClient, toolchainStatus)
+
```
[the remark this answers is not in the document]
xcoulon: in most of the cases, we need to initialize yet another client to use in the `initializeCounters` so we can call `counter.Synchronize(cl, toolchainStatus)`

test/counter.go
xcoulon: ok, so maybe by merging those 2 funcs it would be clearer?

pkg/controller/usersignup/approval_test.go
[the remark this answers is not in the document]
MatousJobanek: I know, but it's not needed for the evaluation of the test. The initialization is needed just to make the counter happy, so I wanted to make the initialization as easy as possible.

I forgot to mention, that I didn't want to reject this particular change - I just shared the reason why I did what I did: to not require the client when it's not needed and create the instance silently inside the function (it's just a implementation detail :smile:) . Sorry for not being clear.

test/counter.go
MatousJobanek: Either that or keeping the original version.

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -39,11 +39,11 @@ func TestAddFinalizer(t *testing.T) {
 
 	t.Run("ok", func(t *testing.T) {
 		// given
-		InitializeCounter(t, MasterUserRecords(1), UserAccountsForCluster(test.MemberClusterName, 1))
 		defer counter.Reset()
 		mur := murtest.NewMasterUserRecord(t, "john")
 		memberClient := test.NewFakeClient(t)
 		hostClient := test.NewFakeClient(t, mur)
+		InitializeCounters(t, hostClient, ToolchainStatus(MasterUserRecords(1), UserAccountsForCluster(test.MemberClusterName, 1)))
```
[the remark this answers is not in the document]
xcoulon: done in cb6d5dd

test/counter.go
xcoulon: One of the goal of this PR is to show that counters are initialized from a `ToolchainStatus` resource (this was hidden within the `initializeCounters` func), so I'd rather merge the 2 funcs and move on, instead of moving backward

MatousJobanek: ok, then feel free to merge them.

xcoulon: done in b93458a. Is it better for you like that?

MatousJobanek: Yes, thank you

pkg/controller/usersignup/approval_test.go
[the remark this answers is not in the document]
MatousJobanek: btw, the second part of my comment is still valid:

> In addition, you don't reset the counter, so it will be initialized only once - for the rest of the cases the initialization will be ignored.

so, if you want to make it consistent, you should also copy the reset to all test-cases

MatousJobanek:
btw, when you are already doing these changes - wouldn't make sense to drop all the `defer counter.Reset()` and replace them by calling:
```
t.Cleanup(counter.Reset)
``` 
inside `InitializeCounters` function?

pkg/counter/cache_test.go
```diff
@@ -18,7 +20,7 @@ var logger = logf.Log.WithName("cache_test")
 
 func TestAddMurToCounter(t *testing.T) {
 	// given
-	InitializeCounter(t)
+	InitializeCounters(t, commontest.NewFakeClient(t), ToolchainStatus())
```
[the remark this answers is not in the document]
xcoulon: changed in https://github.com/codeready-toolchain/host-operator/pull/406/commits/b93458aa6f7ecbfebae3790bb277fe3ad8932c8b

pkg/controller/usersignup/approval_test.go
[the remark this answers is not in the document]
xcoulon: addressed in https://github.com/codeready-toolchain/host-operator/pull/406/commits/cb6d5ddea48dbaa8f2cc23737a2bb788758785ab

test/counter.go
```diff
@@ -17,54 +17,37 @@ import (
...
-		apply(toolchainStatus)
-	}
-	require.NotNil(t, toolchainStatus.Status.HostOperator)
-	assert.Equal(t, toolchainStatus.Status.HostOperator.MasterUserRecordCount, counts.MasterUserRecordCount)
-	AssertMetricsGaugeEquals(t, toolchainStatus.Status.HostOperator.MasterUserRecordCount, metrics.MasterUserRecordGauge)
-
-	assert.Len(t, counts.UserAccountsPerClusterCounts, len(toolchainStatus.Status.Members))
-	for _, member := range toolchainStatus.Status.Members {
-		assert.Equal(t, member.UserAccountCount, counts.UserAccountsPerClusterCounts[member.ClusterName])
-		AssertMetricsGaugeEquals(t, member.UserAccountCount, metrics.UserAccountGaugeVec.WithLabelValues(member.ClusterName))
```
MatousJobanek, line 66: could you please point me to the place in the new code where you verify the gauge metrics?

test/toolchainstatus.go
```diff
@@ -10,19 +10,50 @@ import (
...
+}
+
+func EmptyToolchainStatus() *toolchainv1alpha1.ToolchainStatus {
+	return &toolchainv1alpha1.ToolchainStatus{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configuration.DefaultToolchainStatusName,
+			Namespace: test.HostOperatorNs,
+		},
+	}
+}
```
MatousJobanek, line 33: I guess that there is no need for this function - we can just simply call `NewToolchainStatus()`

test/toolchainstatus.go
```diff
@@ -10,19 +10,50 @@ import (
...
+
+func DefaultToolchainStatus() *toolchainv1alpha1.ToolchainStatus {
+	return &toolchainv1alpha1.ToolchainStatus{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configuration.DefaultToolchainStatusName,
+			Namespace: test.HostOperatorNs,
+		},
+	}
+}
+
```
MatousJobanek, line 43: Not used?

pkg/controller/masteruserrecord/sync_test.go
```diff
@@ -542,6 +542,7 @@ func TestRoutes(t *testing.T) {
 	t.Run("routes are set", func(t *testing.T) {
 		// given
 		toolchainStatus := NewToolchainStatus(
+			WithHost(),
```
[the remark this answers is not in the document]
xcoulon: it's a left-over from a refactoring, which I later reverted. I removed it in 4dd4023

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -155,7 +155,7 @@ func TestToolchainStatusConditions(t *testing.T) {
 		hostOperatorDeployment := newDeploymentWithConditions(defaultHostOperatorName, status.DeploymentAvailableCondition(), status.DeploymentProgressingCondition())
 		registrationServiceDeployment := newDeploymentWithConditions(registrationservice.ResourceName, status.DeploymentAvailableCondition(), status.DeploymentProgressingCondition())
 		memberStatus := newMemberStatus(ready())
-		toolchainStatus := NewToolchainStatus()
+		toolchainStatus := EmptyToolchainStatus()
```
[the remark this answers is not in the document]
xcoulon: Same as above: it's a left-over from a refactoring, which I later reverted. I removed it in 4dd4023

test/toolchainstatus.go
xcoulon: yes, I removed it in 4dd4023

test/toolchainstatus.go
xcoulon: yes, it's also a left-over from an older refactoring. I removed it in 7bc517e

test/counter.go
```diff
@@ -17,54 +17,37 @@ import (
...
+	counts, err := counter.GetCounts()
+	require.NoError(t, err)
+	return &CounterAssertion{
+		t:      t,
+		counts: counts,
 	}
 }
 
-func AssertThatUninitializedCounterHas(t *testing.T, baseValues ...BaseValue) {
+func AssertThatUnitializedCounters(t *testing.T) *CounterAssertion {
```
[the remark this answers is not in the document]
xcoulon: renamed in eb6a472

test/counter.go
xcoulon: oops, good catch! I was too eager when removing the `BaseValue`-related stuff

xcoulon: assertions restored in f71a52d

MatousJobanek (approved):
Awesome :+1: 
Thanks for making all these changes to make the code better and also for taking an extra time to address all my comments :wink:

PR #408, opened 2021-04-22
update usersignup retention to 180 days

xcoulon:
deactivated usersignup resources will be retained for a longer
period, so we can keep track of returning users even later

fixes https://issues.redhat.com/browse/CRT-1021


MatousJobanek (approved):
:+1:

PR #409, opened 2021-04-22
update linter and address issues

xcoulon:
install without altering `go.mod` and `go.sum`
address most of the rule violation, or use `//nolint` comments


deploy/crds/toolchain_v1alpha1_masteruserrecord_cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: toolchain.dev.openshift.com/v1alpha1
 kind: MasterUserRecord
 metadata:
-  name: johnsmith #username
+  name: johnsmith # username
```
alexeykazakov, line 4: Those are generated yamls. We should not change them manually.

xcoulon: ah right. So let me exclude the YAML files in `deploy/crds`, then

xcoulon: reverted in d20fe7f

make/lint.mk
```diff
@@ -10,6 +10,10 @@ lint-yaml: ${YAML_FILES}
...
-	$(Q)go get github.com/golangci/golangci-lint/cmd/golangci-lint
-	$(Q)${GOPATH}/bin/golangci-lint ${V_FLAG} run
\ No newline at end of file
+lint-go-code: generate install-golangci-lint
+	$(Q)${GOPATH}/bin/golangci-lint ${V_FLAG} run -E gofmt,golint,megacheck,misspell ./...
+
+.PHONY: install-golangci-lint
+## Install development tools.
+install-golangci-lint:
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.39.0
```
alexeykazakov, line 19: We should be very careful with running any scripts from master branches from some github repos we can't 100% trust. It's not safe.

alexeykazakov (changes requested):
The changes in the code looks good but I really concerned about adding the golint installation script to our make file.

make/lint.mk
MatousJobanek: yeah, we should fix that also in our sandbox-sre repo

xcoulon: I understand your concern. In sandbox-sre, we use the GitHub action, and this script is only used locally.
There are a few alternatives to installing and running locally, via `brew` on mac or via a Docker image: https://golangci-lint.run/usage/install/
I'm fine with discussing what's the best solution for us. In particular, I was wondering if we could start using the GH action for all our repositories, instead of OpenShift CI jobs? WDYT?

alexeykazakov: Yes, I think we can/should move all jobs but e2e tests to GitHub Actions.
And yes, I noticed that the linter is used locally only. We still should not run any scripts from 3rd party master branches in our make targets.
Let's remove it from this PR and commit the fixes found by the linter only. We can discuss our options with installing the linter separatly.

xcoulon: reverted in e801db8, and we can discuss about next steps later

alexeykazakov (approved):
Looks good. Thanks!

PR #412, opened 2021-04-29
remove minishift support from makefiles, update readme

ranakan19:
fix wrt https://issues.redhat.com/browse/CRT-904


alexeykazakov (approved):
Looks good!

xcoulon (approved):

MatousJobanek (commented):
I believe that we cal also drop all these targets: 
https://github.com/codeready-toolchain/host-operator/blob/cbbf78d6805a5ef01f910ce9a6174c164a404e60/make/dev.mk#L12-L56
they were used only in a combination with minishift/OS3

Together with that you can also drop related parts from README

README.adoc
```diff
@@ -16,21 +16,9 @@ This repository uses https://github.com/golang/go/wiki/Modules[Go modules]. You
...
-$ minishift start --profile host
-```
-
-Then you can run the operator locally with the help of `operator-sdk` (you need version v0.19.2):
-
-```bash
-$ make up-local
-```
 
 That Makefile target takes care of additional several steps which can be executed separately:
```
rajivnathan, line 20: Same comment as in the member-operator repo. Seems like this sentence is referring to the make target up-local that we're removing above. Maybe it's better to point people to the e2e repo for development? They can use the dev- targets in the e2e repo. Not sure what we want to do with the bullet points below, maybe keep them for informational purposes only?

rajivnathan (commented):
Looks good, just had 1 suggestion. Same as member repo

README.adoc
ranakan19: @rajivnathan Thanks, that makes sense. Added a line saying toolchain-e2e can be used for starting instead, lemme know if you think of better wordings.

README.adoc
```diff
@@ -16,31 +16,10 @@ This repository uses https://github.com/golang/go/wiki/Modules[Go modules]. You
 
 == Development
 
-To run this operator locally you need to have at least one Minishift profile started:
+toolchain-e2e repo makefile can be used as a starting point for setting up host and member operators for development.
```
rajivnathan, line 19: I think it would be helpful to provide a link as well.
```suggestion
The dev targets in the toolchain-e2e repository can be used to build and deploy the host and member operators for development: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/dev.mk
```

README.adoc
```diff
@@ -16,31 +16,10 @@ This repository uses https://github.com/golang/go/wiki/Modules[Go modules]. You
...
-$ make up-local
-```
-
-That Makefile target takes care of additional several steps which can be executed separately:
-
-* logging as system:admin user: `$ make login-as-admin`
-* creating local test namespace: `$ make create-namespace`
-* deploy CRDs: `$ make deploy-crd`
+Following Makefile targets takes care of additional several steps which can be executed separately:
 * building the project: `$ make build`
```
rajivnathan, line 22: I guess this is probably not very useful to keep anymore? If we're pointing to the dev targets in toolchain-e2e it's probably good enough? So we can just remove these 2 lines as well.
```suggestion
```

rajivnathan (commented):
Couple wording suggestions for the Development section

README.adoc
ranakan19: yeah, sure

rajivnathan (approved):
Looks great! Awesome work! 🙌

README.adoc
```diff
@@ -16,31 +16,8 @@ This repository uses https://github.com/golang/go/wiki/Modules[Go modules]. You
 
 == Development
 
-To run this operator locally you need to have at least one Minishift profile started:
+The dev targets in the toolchain-e2e repository can be used to build and deploy the host and member operators for development: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/make/dev.mk
```
MatousJobanek, line 19: The same as for member operator - I would rather reference the guide https://github.com/codeready-toolchain/toolchain-e2e/blob/master/dev_install.adoc

MatousJobanek (approved):
Nice. Thanks for cleaning it up :+1:

PR #416, opened 2021-05-05
lint: do not install golangci-lint from source

xcoulon:
check if the `golangci-lint` executable is in the path
if not, fail with a error message including a link
to the install instructions.


xcoulon:
@MatousJobanek @alexeykazakov once we agreed on the way to install/run `golangci-lint` locally, I will apply the same settings on other repos.

xcoulon:
@alexeykazakov @MatousJobanek also, should we consider replacing the `ci/prow/lint` job with a GitHub Action, as we do on sandbox-sre? I believe the integration with GitHub will be better and we won't have to deal with installing `golangci-lint` when running the job on `ci/prow`

alexeykazakov (approved):

alexeykazakov:
Yes let's move it to GH actions

make/lint.mk
```diff
@@ -11,5 +11,7 @@ lint-yaml: ${YAML_FILES}
 .PHONY: lint-go-code
 ## Checks the code with golangci-lint
 lint-go-code: generate
-	$(Q)go get github.com/golangci/golangci-lint/cmd/golangci-lint
-	$(Q)${GOPATH}/bin/golangci-lint ${V_FLAG} run
\ No newline at end of file
+ ifeq (, $(shell which golangci-lint))
```
MatousJobanek, line 14: This will make the output more clear
```suggestion
 ifeq (, $(shell which golangci-lint 2>/dev/null))
```

MatousJobanek (approved):
Nice :+1:

xcoulon:
> Yes let's move it to GH actions

see https://github.com/openshift/release/pull/18329 to remove the `ci/prow/lint` job

make/lint.mk
xcoulon: thanks! done in e9c760c

PR #418, opened 2021-05-05
Refactor to use UserSignup states instead of VerificationRequired

sbryzak:
This PR refactors the host operator to use the `UserSignup.Spec.States` property to track signup state instead of the individual `UserSignup.Spec.VerificationRequired` field.

Fixes https://issues.redhat.com/browse/CRT-1005

Associated PRs:

- Registration Service https://github.com/codeready-toolchain/registration-service/pull/184
- E2E tests https://github.com/codeready-toolchain/toolchain-e2e/pull/290
- Sandbox SRE https://github.com/codeready-toolchain/sandbox-sre/pull/223


alexeykazakov (commented):
We need to add some migration steps.
You can't update both reg-service and host in the same time. If we merge host first then we need to take into account that:
1) there are still existing resources with Spec.VerficationRequired
2) reg-service is still using Spec.VerficationRequired

Also, when don't forget to double check our sandbox-sre cli to make sure it manipulates the states correctly when approving the user.

alexeykazakov:
Also, please make sure you have links to all related and coupled PRs across all the repos in the PR description.

sbryzak:
@alexeykazakov links to other PRs have been added.  I'm still thinking about how we might deploy this change cleanly, I agree that registration service should be updated first however there may still be a small window where things could get messy if any particular users are currently active in the verification process.  I'll think over our options here and the possible ramifications.

PR #419, opened 2021-05-05
feat: add annotation for email address on MUR

xcoulon:
add annotation when creating a MUR
also, migrate existing MURs

Fixes https://issues.redhat.com/browse/CRT-1075

### DIFF
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller.go
@@ -116,6 +116,24 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 			logger.Error(err, "unable to add finalizer to MasterUserRecord")
 			return reconcile.Result{}, err
 		}
+		// migration for CRT-1075: add an annotation with the email address (same as for associated UserSignup resource)
+		usersignup := &toolchainv1alpha1.UserSignup{}
+		if r.client.Get(context.TODO(), types.NamespacedName{
+			Namespace: mur.Namespace,
+			Name:      mur.Spec.UserID,
+		}, usersignup); err != nil {
+			logger.Error(err, "unable to get associated UserSignup")
+			return reconcile.Result{}, err
+		}
+		if mur.Annotations == nil {
+			mur.Annotations = map[string]string{}
+		}
+		mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey] = usersignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]
+		if r.client.Update(context.TODO(), mur); err != nil {
+			logger.Error(err, "unable to update the MUR with the email address annotation")
+			return reconcile.Result{}, err
+		}
+
 		logger.Info("ensuring user accounts")
 		for _, account := range mur.Spec.UserAccounts {
 			requeueTime, err := r.ensureUserAccount(logger, account, mur)
--- a/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
+++ b/pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
@@ -31,6 +31,31 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// TODO: to be removed once CRT-1075 is done
+func TestMigrationAddEmailAddressAnnotation(t *testing.T) {
+	// given
+	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+	s := apiScheme(t)
+	usersignup := NewUserSignup() // email is `foo@redhat.com`
+	mur := murtest.NewMasterUserRecord(t, "john", murtest.UserID(usersignup.Name))
+	hostClient := test.NewFakeClient(t, usersignup, mur)
+	memberClient := test.NewFakeClient(t)
+	cntrl := newController(t, hostClient, s, NewGetMemberCluster(true, v1.ConditionTrue),
+		ClusterClient(test.MemberClusterName, memberClient))
+
+	// when
+	result, err := cntrl.Reconcile(newMurRequest(mur))
+
+	// then
+	require.NoError(t, err)
+	require.False(t, result.Requeue)
+
+	murtest.AssertThatMasterUserRecord(t, "john", hostClient).
+		HasConditions(toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "")).
+		HasFinalizer()
+
+	// then
+}
 func TestAddFinalizer(t *testing.T) {
 	// given
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
--- a/go.mod
+++ b/go.mod
@@ -46,4 +46,8 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace (
+	github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20210504131614-5439aaf86248
+	github.com/codeready-toolchain/toolchain-common => github.com/xcoulon/toolchain-common v0.0.0-20210504131210-9c21774a1af1
+)
 go 1.14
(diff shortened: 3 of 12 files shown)


pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -116,6 +116,24 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 			logger.Error(err, "unable to add finalizer to MasterUserRecord")
 			return reconcile.Result{}, err
 		}
+		// migration for CRT-1075: add an annotation with the email address (same as for associated UserSignup resource)
```
alexeykazakov, line 119: Don't forget to add a task (sub-task in JIRA?) to clean up this migration step when done.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -116,6 +116,24 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+			Namespace: mur.Namespace,
+			Name:      mur.Spec.UserID,
+		}, usersignup); err != nil {
+			logger.Error(err, "unable to get associated UserSignup")
+			return reconcile.Result{}, err
+		}
+		if mur.Annotations == nil {
+			mur.Annotations = map[string]string{}
+		}
+		mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey] = usersignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]
```
alexeykazakov, line 131: Check if the annotation is not already set before executing the migration step.

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -31,6 +31,31 @@ import (
...
+	// then
+	require.NoError(t, err)
+	require.False(t, result.Requeue)
+
+	murtest.AssertThatMasterUserRecord(t, "john", hostClient).
+		HasConditions(toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "")).
+		HasFinalizer()
+
+	// then
+}
```
alexeykazakov, line 58: It looks like you forgot to check if the annotation is set?

alexeykazakov (commented):
Let's add the annotation check in our e2e tests. In the function when we verify provisioned resources.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
xcoulon: already done: https://issues.redhat.com/browse/CRT-1074 😜

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
xcoulon: oops 😬 
done in 646e3f1

pkg/controller/masteruserrecord/masteruserrecord_controller.go
xcoulon: good point! done in 646e3f1

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -116,6 +116,25 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
...
+			}
+			if mur.Annotations == nil {
+				mur.Annotations = map[string]string{}
+			}
+			mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey] = usersignup.Annotations[toolchainv1alpha1.UserSignupUserEmailAnnotationKey]
+			if r.client.Update(context.TODO(), mur); err != nil {
+				logger.Error(err, "unable to update the MUR with the email address annotation")
+				return reconcile.Result{}, err
+			}
+		}
```
MatousJobanek, line 137: This migration shouldn't be in `MUR` controller, but in `UserSignup` controller - that controller is responsible for setting the proper annotation in MUR. We have this function: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/masteruserrecord.go#L10-L41 for the purpose of migrating MUR - you can move the code there

xcoulon: IMO it's a bit counter-intuitive to migrate the MasterUserRecords while processing the UserSignups, but since there's already a function that deals with that, then I will move the block above there 🤷‍♂️

pkg/controller/masteruserrecord/masteruserrecord_controller.go
alexeykazakov: It's not done until it's in the PR description ;)

deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
```diff
@@ -45,8 +45,8 @@ spec:
       name: ApprovedBy
       priority: 1
       type: string
-    - jsonPath: .spec.deactivated
-      name: Deactivated
+    - jsonPath: .spec.states
```
alexeykazakov, line 48: Please rebase.

make/lint.mk
```diff
@@ -6,10 +6,15 @@ YAML_FILES := $(shell find . -type f -regex ".*y[a]ml" -print)
 .PHONY: lint-yaml
 ## runs yamllint on all yaml files
 lint-yaml: ${YAML_FILES}
+ifeq (, $(shell which yamllint))
```
alexeykazakov, line 9: Is it something from another PR?

pkg/controller/masteruserrecord/masteruserrecord_controller.go
alexeykazakov: Not really. MUR is a secondary resource of the UserSignup controller. It means UserSignup controller actually manages MURs (creates, updates, deletes them). And MUR controller only watches MURs changes to manage its own secondary resources: UserAccounts. So it makes sense to migrate MURs in UserSignup controller.
But in practice I think it's fine to migrate resources in the place where it's easier to do.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -116,6 +116,25 @@ func (r *ReconcileMasterUserRecord) Reconcile(request reconcile.Request) (reconc
 			logger.Error(err, "unable to add finalizer to MasterUserRecord")
 			return reconcile.Result{}, err
 		}
+		// migration for CRT-1075: add an annotation with the email address (same as for associated UserSignup resource)
+		if mur.Annotations == nil || mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey] == "" {
+			usersignup := &toolchainv1alpha1.UserSignup{}
+			if r.client.Get(context.TODO(), types.NamespacedName{
+				Namespace: mur.Namespace,
+				Name:      mur.Spec.UserID,
```
alexeykazakov, line 124: `UserID` doesn't necessarily match `UserSignup.Name`.
It used to but not anymore. We have to get the usersignup name from either `toolchain.dev.openshift.com/owner` label or `ownerReference`.

alexeykazakov: Actually, `MUR.Spec.UserID` is actually the corresponding `UserSignup.Name`. The UserID name of this field is confusing. We even have a JIRA to rename it: https://issues.redhat.com/browse/CRT-942
So, While actual UserID is not longer necessarily the same as the UserSignup name (atm for most resources it's the same but for some it's not) this particular field refers to UserSignup.Name.

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: If I'm not mistaken, we do process every MUR in UserSignup controller as well.
The controller that manages secondary resources should ensure that the meta & spec information in the resources are up-to-date and matches the desired state - that's their purpose. 
In addition, to finish this particular the migration we need both resources: MUR and UserSignup - this is something we already have in UserSignup controller. Getting the UserSignup resource which is an owner of the MUR in MUR controller to make the MUR up-to-date looks like anti-pattern to me.

The only exception when we need to do the migration in the controller that watches the resources as the primary resource is UserSignup - the reason is that there is no controller that would manage UsesrSignups as secondary resources, so we need to do it in its controller.

deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
xcoulon: yes, I'm not sure what happened here. My apologies, I'll fix/clean that

pkg/controller/masteruserrecord/masteruserrecord_controller.go
xcoulon: well, I missed your last comments while I was refactoring the code (see 9145b69), so please let me know if it's ok for you now. Basically, I added a block in the `migrateOrFixMurIfNecessary` to set the email address on the MUR if it's missing. And for sure, that having all the migration logic in one place is better. It's just that I missed that in my first PR 😬

deploy/crds/toolchain_v1alpha1_usersignup_crd.yaml
xcoulon: it's fixed now 😅

make/lint.mk
xcoulon: yeah, same as above, not sure what happened here 😬 , but it's outdated now

pkg/controller/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: yes, it's perfect :-)  - thank you :+1:

MatousJobanek (approved):
Thanks a lot for moving the migration code to the "usersignup controller logic"

alexeykazakov (approved):

PR #422, opened 2021-05-07
Migrate UserSignup VerificationRequired property to states 

sbryzak:
This PR refactors the host operator to use the `UserSignup.Spec.States` property to track signup state instead of the individual `UserSignup.Spec.VerificationRequired` field.

Fixes https://issues.redhat.com/browse/CRT-1005

Associated PRs:

- E2E tests https://github.com/codeready-toolchain/toolchain-e2e/pull/293
- Sandbox SRE https://github.com/codeready-toolchain/sandbox-sre/pull/223


pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -146,6 +146,22 @@ func (r *ReconcileUserSignup) Reconcile(request reconcile.Request) (reconcile.Re
...
+	// TODO remove this after all UserSignup instances have been migrated
+	//
+	// ###################################################################################
+	if userSignup.Spec.VerificationRequired && !states.VerificationRequired(userSignup) {
+		states.SetVerificationRequired(userSignup, true)
+		if err := r.client.Update(context.TODO(), userSignup); err != nil {
+			return reconcile.Result{}, err
+		}
+		// Requeue the reconciliation if the UserSignup was migrated
+		return reconcile.Result{}, nil
```
alexeykazakov, line 162: Weill, an empty Result won't trigger a requeue, correct? But that's fine because the spec update will trigger reconcile. So the code is fine but the comment is not quite correct.

alexeykazakov (approved):
Looks good, besides one minor comment!

PR #431, opened 2021-05-12
Use constant for toolchainstatus name

rajivnathan:
No need for toolchainstatus name to be configurable, we'll just hardcode it to "toolchain-status"


alexeykazakov (approved):

MatousJobanek (approved):
:+1:

PR #432, opened 2021-05-12
Added migration for UserSignup Approved property

sbryzak:
This PR adds migration to move UserSignup.Spec.Approved to the states property instead.  It has been compiled with an updated API (with Approved property removed - see https://github.com/sbryzak/api/tree/CRT-1004) in order to identify all usages.

Fixes https://issues.redhat.com/browse/CRT-1004

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/298


alexeykazakov (commented):
I'm a bit confused with this PR. Shouldn't we support/verify both APIs in the migration PR? And why are you deleting the old API from the CRD?

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -91,7 +91,7 @@ func (r *ReconcileUserSignupCleanup) Reconcile(request reconcile.Request) (recon
 	}
 	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
 
-	if states.VerificationRequired(instance) && !instance.Spec.Approved {
+	if states.VerificationRequired(instance) && !states.Approved(instance) {
```
alexeykazakov, line 94: This is a different controller and we can't guarantee that the UserSignup is already migrated by the usersignup controller. We should either add a migration step here too or check both `states.Approved()` and `instance.Spec.Approved` in case it's not migrated yet.

alexeykazakov (commented):
Looks good overall but I think we need to fix the usersignup cleanup controller.
And please also add migration tests for both usersignup and usersignupcleanup controllers.

MatousJobanek:
btw, there is one weird thing I just realized. What does the `states` parameter contain when the user is automatically approved. That means:
* the user is not deactivated
* it either doesn't have to pass the phone verification or the user already did it
* the user is not approved manually

If I understand it correctly, the `states` parameter should be empty, correct? So, it would mean when there is no state required, then it's supposed to be automatically approved (which doesn't have to be always true when there is no space in the clusters, but that's just a detail). Anyway, this doesn't match with the content of this function: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/states/state_manager.go#L5-L9

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: The migration block has been added to the UserSignup cleanup controller also, and tests added for both controllers.

sbryzak:
@MatousJobanek that's a good point, the states property is empty under the conditions you described.  It seems as though the `Active()` function isn't actually being used outside of tests, possibly we can just remove it for now - @alexeykazakov what do you think?

alexeykazakov:
I'm fine with both options:
- modify `states.Active()` so it returns true if `Spec.States` is empty.
- remove `states.Active()` for now.

alexeykazakov (approved):

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
MatousJobanek: But this controller should not contain any migration - the migration should be only in `usersignup_controller.go`.
@alexeykazakov proposed modifying the `if` statement, so it would check both the `states.Approved()` as well as  `instance.Spec.Approved` and not copy-pasting the migration code.

MatousJobanek (changes requested):
Please, remove the migration code from `usersignup_cleanup_controller.go`

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: The if check won't work here, the migration needs to set the approved spec property to false otherwise it breaks our tests.

MatousJobanek: Why? What's wrong on:
```
if states.VerificationRequired(instance) && !(states.Approved(instance) || instance.Spec.Approved)
```
?

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -79,7 +79,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	// TODO remove this after migration complete
 	// Migrate the Approved property
-	if instance.Spec.Approved && !states.Approved(instance) {
+	/*if instance.Spec.Approved && !states.Approved(instance) {
```
alexeykazakov, line 82: We can delete this block now instead of putting it in a comment, right?

sbryzak: Done

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Changes made as suggested.

alexeykazakov (approved):

pkg/controller/usersignup/usersignup_controller.go
```diff
@@ -105,6 +105,48 @@ type Reconciler struct {
...
+}
+
+func remove(s []toolchainv1alpha1.UserSignupState, state toolchainv1alpha1.UserSignupState) []toolchainv1alpha1.UserSignupState {
+	for i, v := range s {
+		if v == state {
+			return append(s[:i], s[i+1:]...)
+		}
+	}
+	return s
+}
```
MatousJobanek, line 144: I'm confused? Why not using the state manager https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/states/state_manager.go#L58-L84

MatousJobanek (approved):
Thanks for removing the duplicated migration code

pkg/controller/usersignup/usersignup_controller.go
sbryzak: We don't use the state manager here because using it to set Approved will cause the Deactivating and Deactivated states to be set to false.  This may result in already-deactivated users becoming reactivated again.

MatousJobanek: I thought that this was solved by: 
https://github.com/sbryzak/host-operator/blob/38ac9b78b3876b0b9e58ab54ea8ed6c1dc626ad3/pkg/controller/usersignup/usersignup_controller.go#L176

PR #437, opened 2021-05-17
feat: metric to measure MasterUserRecords by email address domain

xcoulon:
New `sandbox_master_user_records` metric with a `domain` label
and associated `masterUserRecordsPerDomain` counter persisted in
the `ToolchainStatus.Status.Metrics`.
Expected values are `Red Hat`, `IBM` and `Other`

Also: trigger counter and metric initialization during operator startup
(also affect unit tests setup)

Fixes https://issues.redhat.com/browse/CRT-1072

### Associated PRs:
- https://github.com/codeready-toolchain/api/pull/235
- https://github.com/codeready-toolchain/toolchain-common/pull/174
- https://github.com/codeready-toolchain/toolchain-e2e/pull/302


alexeykazakov (approved):
Looks great! But I wonder if we really need three domains. Shouldn't we combine RH and IBM?

xcoulon:
> Looks great! But I wonder if we really need three domains. Shouldn't we combine RH and IBM?

done in a4da9af: there are now 2 domains: `internal` and `external`. The former is for `@redhat.com`, `@ibm.com` and subdomains, the latter is for all other email address domains.

pkg/metrics/metrics.go
```diff
@@ -67,6 +69,7 @@ func initMetrics() {
 	// Gauges with labels
 	UserAccountGaugeVec = newGaugeVec("user_accounts_current", "Current number of User Accounts (per member cluster)", "cluster_name")
 	UsersPerActivationGaugeVec = newGaugeVec("users_per_activations", "Number of users per activation", []string{"activations"}...)
+	MasterUserRecordGaugeVec = newGaugeVec("master_user_records", "Number of Master User Records (per domain)", "domain")
```
xcoulon, line 72: I wonder if we should use `domain` or `email_domain`. The latter is more specific, so it would avoid conflict/confusion if we have another kind of "domain" in the future. WDYT?

alexeykazakov (approved):

pkg/metrics/metrics.go
alexeykazakov: I think both email_domain or domain would work. I'm OK to keep it shorter as "domain" but feel free to rename it if you like email_domain better.

pkg/counter/cache.go
```diff
@@ -55,42 +58,56 @@ func Reset() {
...
+
+		cachedCounts.MasterUserRecordPerDomainCounts[string(domain)]++
+		log.Info("incrementing MasterUserRecordGaugeVec", "value", cachedCounts.MasterUserRecordPerDomainCounts[string(domain)], "domain", domain)
+		metrics.MasterUserRecordGaugeVec.WithLabelValues(string(domain)).Set(float64(cachedCounts.MasterUserRecordPerDomainCounts[string(domain)]))
 	})
 }
 
 // DecrementMasterUserRecordCount decreases the number of MasterUserRecord in the cached counter
-func DecrementMasterUserRecordCount(logger logr.Logger) {
+func DecrementMasterUserRecordCount(domain metrics.Domain) {
```
rajivnathan, line 86: Is there a reason to remove the logger passed into the function? I think it's passed in so that we know which controller was calling the function?

The `DecrementUserAccountCount` function still has logger as a param, so whatever we decide, we should probably make them consistent.

pkg/metrics/metrics.go
```diff
@@ -28,17 +28,19 @@ var (
 
 // gauges
 var (
-	// DEPRECATED - See UserAccountGaugeVec
+	// DEPRECATED - See UserAccountGaugeVec and MasterUserRecordGaugeVec
```
rajivnathan, line 31:
```suggestion
	// DEPRECATED - See MasterUserRecordGaugeVec
```
I guess the reference to `UserAccountGaugeVec` was when we had a `UserAccountGauge` in the past?

pkg/metrics/metrics.go
```diff
@@ -28,17 +28,19 @@ var (
...
 
 // gauge with labels
 var (
-	// UserAccountGaugeVec should reflect the current number of master user records in the system, with a label to partition per member cluster
+	// UserAccountGaugeVec reflects the current number of master user records in the system, with a label to partition per member cluster
 	UserAccountGaugeVec *prometheus.GaugeVec
-
+	// UsersPerActivationGaugeVec reflects the number of users labelled with on their current number of activations
 	UsersPerActivationGaugeVec *prometheus.GaugeVec
+	// MasterUserRecordGaugeVec reflects the current number of MasterUserRecords, labelled with their email address domain (`@redhat.com` vs `@*.ibm.com` vs others)
```
rajivnathan, line 42:
```suggestion
	// MasterUserRecordGaugeVec reflects the current number of MasterUserRecords, labelled with their email address domain (`internal` vs `external`)
```
I guess this should now refer to internal and external?

rajivnathan (commented):
Looks great! Added a couple minor comments for consideration.

pkg/counter/cache.go
xcoulon: yes, that's a fair point. I'll restore that

pkg/metrics/metrics.go
xcoulon: right, that's probably a quick copy/paste error of mine

pkg/counter/cache.go
xcoulon: restored in 0fd0a08

pkg/metrics/metrics.go
xcoulon: Ok, thanks for your feedback. I've kept the label short (`domain`), but I clarified its purpose in the metric description in c0d0906

pkg/counter/cache.go
```diff
@@ -219,6 +240,11 @@ func initialize(cl client.Client, toolchainStatus *toolchainv1alpha1.ToolchainSt
 		toolchainStatus.Status.HostOperator.MasterUserRecordCount == 0 {
 		return initializeFromResources(cl, toolchainStatus.Namespace)
 	}
+	// Migration for CRT-1072: force initialization from existing UserSignups/MasterUserRecords
+	// TODO: remove as part of CRT-1074
+	if _, exists := toolchainStatus.Status.Metrics[toolchainv1alpha1.MasterUserRecordsPerDomainMetricKey]; !exists {
+		return initializeFromResources(cl, toolchainStatus.Namespace)
+	}
```
MatousJobanek, line 247: I'm curious - why should we remove it?
I would expect to be part of the previous if statement with the fact that it should stay there forever. 
https://github.com/codeready-toolchain/host-operator/blob/f21acd30e5bc453ddacd144b1c4f24cc936806d3/pkg/counter/cache.go#L239-L242
And we could do the same also for the other metric `toolchainv1alpha1.UsersPerActivationMetricKey`

pkg/metrics/email_domain.go
```diff
@@ -0,0 +1,26 @@
...
+
+const Internal Domain = "internal"
+const External Domain = "external"
+
+var internalDomainPattern *regexp.Regexp
+
+func init() {
+	// pattern to match email addresses in the form of `@redhat.com`, `@ibm.com` or subdomains (eg: `@fr.ibm.com`)
+	internalDomainPattern = regexp.MustCompile(`^.*@redhat.com$|^.*@((.+)\.)?ibm.com$`)
+}
```
MatousJobanek, line 17: why not:
```suggestion
var internalDomainPattern = regexp.MustCompile(`^.*@redhat.com$|^.*@((.+)\.)?ibm.com$`)
```
?

MatousJobanek (commented):
:+1:

pkg/metrics/email_domain.go
xcoulon: indeed, why not :)

pkg/counter/cache.go
xcoulon: > I'm curious - why should we remove it?
> I would expect to be part of the previous if statement with the fact that it should stay there forever.
> https://github.com/codeready-toolchain/host-operator/blob/f21acd30e5bc453ddacd144b1c4f24cc936806d3/pkg/counter/cache.go#L239-L242
> 

good point. I guess I was in the mindset of the migration when I wrote this comment. I've updated the logic in 3b967e8

xcoulon: > And we could do the same also for the other metric `toolchainv1alpha1.UsersPerActivationMetricKey`

done in 86045de

pkg/controller/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -61,8 +67,10 @@ func TestAddFinalizer(t *testing.T) {
 		murtest.AssertThatMasterUserRecord(t, "john", hostClient).
 			HasConditions(toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "")).
 			HasFinalizer()
-		AssertThatCounters(t).HaveMasterUserRecords(1).
-			HaveUserAccountsForCluster(test.MemberClusterName, 2)
+		AssertThatCounters(t).
+			HaveMasterUserRecords(1).                                                              // unchanged (not complete yet)
+			HaveMasterUserRecordsPerDomain(toolchainv1alpha1.Metric{string(metrics.Internal): 1}). // unchanged (not complete yet)
```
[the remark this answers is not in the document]
xcoulon: ok, fair point, I've update these comments in 37a6b26

pkg/controller/usersignup/masteruserrecord.go
```diff
@@ -37,7 +37,8 @@ func migrateOrFixMurIfNecessary(mur *toolchainv1alpha1.MasterUserRecord, nstempl
 			changed = true
 		}
 	}
-	// migration for CRT-1075: add an annotation with the email address (same as for associated UserSignup resource)
+	// Migration for CRT-1075: add an annotation with the email address (same as for associated UserSignup resource)
+	// TODO: remove as part of CRT-1074
```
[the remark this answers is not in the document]
xcoulon: no, indeed. I remove the comment on line 40 in 8e785d8

pkg/controller/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1153,11 +1164,14 @@ func TestSynchronizationWithCounter(t *testing.T) {
...
 			}),
+			WithMetric(toolchainv1alpha1.MasterUserRecordsPerDomainMetricKey, toolchainv1alpha1.Metric{
+				string(metrics.External): 8,
+			}),
 		)
 		reconciler, req, fakeClient := prepareReconcile(t, requestName, newResponseGood(), []string{"member-1", "member-2"}, hostOperatorDeployment, memberStatus, registrationServiceDeployment, registrationService, toolchainStatus)
 
 		// when
-		counter.IncrementMasterUserRecordCount()
+		counter.IncrementMasterUserRecordCount(metrics.Internal)
```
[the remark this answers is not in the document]
xcoulon: yes, indeed ;) done in b15a4b7

pkg/counter/cache.go
MatousJobanek: I'm really sorry, but I got confused by the changes in the commit you referenced - you changed a lot of the code plus also the `setStateLabel` which is not related at all.
Are the changes in the tests related to the fact that it initializes the counter for every test? But why changing the `setStateLabel`?

MatousJobanek: Now I see that the changes related to this specific metric were introduced in 3b967e8288765f71744bea25b94d392ce11c7bd3 - in the commit that you referenced, you added the `UsersPerActivationMetricKey` (which could have been done separately, but it probably doesn't matter now).

xcoulon: ok, let me try to explain (or confirm what you wrote above):
- the changes in the `setStateLabel` func are just about renaming some local variables and adding comments. There is no change in the logic itself. I "touched" this func because I was confused with some tests results and while debugging, I found that the var deserved to be renamed to be more adequate with the new "state" stuff. That being said, I can revert the changes here.
- there are indeed a lot of changes related to the tests. As you noted, that's related to the counter initialization, which will look at the resources (`UserSignup` and `MasterUserRecord`) if one if the metric is missing. So, to avoid switching to that kind if initialization, I had to set the `UsersPerActivationMetricKey` metric to all these tests.

I hope I answered all your questions, so please let me know if you have further concerns

pkg/controller/usersignup/masteruserrecord.go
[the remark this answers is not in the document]
xcoulon: ok, I see what you mean: the `if mur.Annotations == nil || mur.Annotations[toolchainv1alpha1.MasterUserRecordEmailAnnotationKey] == ""` statement should return `false` now, so this code block is not used anymore, but as the comment above says, I intend to clean up all the migration code in https://issues.redhat.com/browse/CRT-1074, which will be the last subtask of this story. 

I though it would be better to keep the PR focused on a single subject, but let me know if you prefer that I address this clean-up immediatly.

go.mod
```diff
@@ -46,4 +46,9 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace (
+	github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20210519064822-2ae896a20bf3
+	github.com/codeready-toolchain/toolchain-common => github.com/xcoulon/toolchain-common v0.0.0-20210519065518-4d26e9c0253f
```
MatousJobanek, line 51: just a reminder

MatousJobanek (approved):
Thanks for addressing my comments & improving the metrics :+1: :1st_place_medal:

rajivnathan (approved):

PR #438, opened 2021-05-18
ToolchainConfig CRD and controller

rajivnathan:
Adds generated files for ToolchainConfig CRD and implements a controller for it. The plan is to put the ToolchainConfig in place first before actually using it in the product code. Once it's in place we can begin migrating existing HostOperatorConfig configuration to it and then migrating other configuration after that.

Related PRs:
https://github.com/codeready-toolchain/api/pull/236
https://github.com/codeready-toolchain/toolchain-common/pull/175


pkg/controller/toolchainconfig/cache_test.go
```diff
@@ -0,0 +1,163 @@
...
+			assert.Equal(t, 666, actual.AutomaticApproval().MaxNumberOfUsersOverall())
+			assert.Empty(t, actual.AutomaticApproval().MaxNumberOfUsersSpecificPerMemberCluster())
+			assert.Equal(t, 0, actual.AutomaticApproval().ResourceCapacityThresholdDefault())
+			assert.Empty(t, actual.AutomaticApproval().ResourceCapacityThresholdSpecificPerMemberCluster())
+			assert.Equal(t, 5, actual.Deactivation().DeactivatingNotificationInDays())
+		})
+	})
+}
+
+func TestGetConfigErrored(t *testing.T) {
```
alexeykazakov, line 89: Do we have a test for the case where there is no ToolchainConfig resource with the expected name?

pkg/controller/toolchainconfig/configuration.go
```diff
@@ -0,0 +1,61 @@
+package toolchainconfig
+
+import "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+
+type ToolchainConfig struct {
+	cfg *v1alpha1.ToolchainConfigSpec
+}
+
+func (c *ToolchainConfig) AutomaticApproval() *autoApproval {
```
alexeykazakov, line 9: Maybe it's better to return the value here and all other places where we do not use pointers in the API?
I mean:
```suggestion
func (c *ToolchainConfig) AutomaticApproval() autoApproval {
```
Here and in all the places. Just to avoid confusion. We will never return a nil here (and we should not).

pkg/controller/toolchainconfig/configuration.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return defaultValue
+}
+
+func getInt(value *int, defaultValue int) int {
+	if value != nil {
+		return *value
+	}
+	return defaultValue
+}
```
alexeykazakov, line 61: Let's move all relevant tests from the cache test to configuration_test.go.
In cache_test.go we need to test the cache itself. And in configuration_test.go we need to test the Configuration (wrapper) API & default values.

pkg/controller/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -0,0 +1,71 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("toolchainconfig-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource ToolchainConfig
+	return c.Watch(&source.Kind{Type: &toolchainv1alpha1.ToolchainConfig{}}, &handler.EnqueueRequestForObject{})
```
alexeykazakov, line 27: While we don't use Status for ToolchainConfig yet I would still use the generation predicate here so we watch Spec changes only here.

pkg/controller/toolchainconfig/toolchainconfig_controller_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// when
+	_, err := controller.Reconcile(newRequest())
+
+	// then
+	require.NoError(t, err)
+	actual, err := GetConfig(NewFakeClient(t), HostOperatorNs)
+	require.NoError(t, err)
+	assert.False(t, actual.AutomaticApproval().IsEnabled())
+	assert.Equal(t, 123, actual.AutomaticApproval().MaxNumberOfUsersOverall())
```
alexeykazakov, line 34: Same here. Don't really need to test all these params.

alexeykazakov (commented):
Looks good overall! But a have a few comments to address.

pkg/controller/toolchainconfig/configuration.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/438/commits/90af2f9b7c30cd88fe31d833c812ade8e60a2315

pkg/controller/toolchainconfig/configuration.go
rajivnathan: Good point, done in https://github.com/codeready-toolchain/host-operator/pull/438/commits/90af2f9b7c30cd88fe31d833c812ade8e60a2315

pkg/controller/toolchainconfig/cache.go
```diff
@@ -0,0 +1,76 @@
...
+
+func updateConfig(config *v1alpha1.ToolchainConfig) {
+	configCache.set(config)
+}
+
+func loadLatest(cl client.Client, namespace string) error {
+	config := &v1alpha1.ToolchainConfig{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: "config"}, config); err != nil {
+		if apierrors.IsNotFound(err) {
+			cacheLog.Error(err, "ToolchainConfig resource with the name 'config' wasn't found", "namespace", namespace)
```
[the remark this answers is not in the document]
rajivnathan: You're right, we shouldn't treat it as an error. We should treat it as a default case and log it as an info message. I've updated it to `cacheLog.Info("ToolchainConfig resource with the name 'config' wasn't found, default configuration will be used", "namespace", namespace)` in https://github.com/codeready-toolchain/host-operator/pull/438/commits/c9cc996c362634d70059fe8b625c8fb3846b6623

pkg/controller/toolchainconfig/cache_test.go
```diff
@@ -0,0 +1,163 @@
...
+	defaultConfig, err := GetConfig(cl, HostOperatorNs)
+
+	// then
+	require.NoError(t, err)
+	assert.False(t, defaultConfig.AutomaticApproval().IsEnabled())
+	assert.Equal(t, 0, defaultConfig.AutomaticApproval().MaxNumberOfUsersOverall())
+	assert.Empty(t, defaultConfig.AutomaticApproval().MaxNumberOfUsersSpecificPerMemberCluster())
+	assert.Equal(t, 0, defaultConfig.AutomaticApproval().ResourceCapacityThresholdDefault())
+	assert.Empty(t, defaultConfig.AutomaticApproval().ResourceCapacityThresholdSpecificPerMemberCluster())
+	assert.Equal(t, 3, defaultConfig.Deactivation().DeactivatingNotificationInDays())
```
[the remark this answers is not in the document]
rajivnathan: Right, created a configuration_test.go for details configuration testing.

pkg/controller/toolchainconfig/cache_test.go
rajivnathan: Added it in https://github.com/codeready-toolchain/host-operator/pull/438/commits/c9cc996c362634d70059fe8b625c8fb3846b6623

pkg/controller/toolchainconfig/toolchainconfig_controller.go
rajivnathan: Sounds good, done in https://github.com/codeready-toolchain/host-operator/pull/438/commits/3a8428fec28d6b6619fd9c33a532e7ea001c9efc

pkg/controller/toolchainconfig/toolchainconfig_controller_test.go
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/438/commits/e5188e1c6702fa0801f91dea0520ec7e3a0da5de

pkg/controller/toolchainconfig/configuration_test.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+	t.Run("non-default", func(t *testing.T) {
+		cfg := newToolchainConfigWithReset(t, DeactivationCfg().DeactivatingNotificationDays(5))
+		toolchainCfg := ToolchainConfig{cfg: &cfg.Spec}
+
+		assert.Equal(t, 5, toolchainCfg.Deactivation().DeactivatingNotificationInDays())
+	})
+}
+
+func matchesDefaultConfig(t *testing.T, actual ToolchainConfig) {
```
alexeykazakov, line 48: Is it used anywhere?

rajivnathan: it's used in toolchainconfig_controller_test.go

rajivnathan: I thought it would be used in configuration_test.go but I didn't end up using it since we're testing by section so I moved it to toolchainconfig_controller_test.go https://github.com/codeready-toolchain/host-operator/pull/438/commits/22e0bc3aeb1791a2b0ebf4113b90b50b7551e24b

alexeykazakov (approved):
Nice!

pkg/controller/toolchainconfig/configuration.go
```diff
@@ -0,0 +1,61 @@
...
+func (a AutoApprovalConfig) ResourceCapacityThresholdDefault() int {
+	return getInt(a.approval.ResourceCapacityThreshold.DefaultThreshold, 0)
+}
+
+func (a AutoApprovalConfig) ResourceCapacityThresholdSpecificPerMemberCluster() map[string]int {
+	return a.approval.ResourceCapacityThreshold.SpecificPerMemberCluster
+}
+
+func (a AutoApprovalConfig) MaxNumberOfUsersOverall() int {
+	return getInt(a.approval.MaxNumberOfUsers.Overall, 0)
```
MatousJobanek, line 34: This was required too, I would set for example:
```suggestion
	return getInt(a.approval.MaxNumberOfUsers.Overall, 1000)
```

pkg/controller/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -0,0 +1,72 @@
...
+	return add(mgr, r)
+}
+
+// blank assignment to verify that Reconciler implements reconcile.Reconciler
+var _ reconcile.Reconciler = &Reconciler{}
+
+// Reconciler reconciles a ToolchainConfig object
+type Reconciler struct {
+	// This client, initialized using mgr.client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
```
MatousJobanek, line 42: It's just a detail, but you can drop the lines that are not used in the new format anymore:
```suggestion
// Reconciler reconciles a ToolchainConfig object
type Reconciler struct {
```

MatousJobanek (approved):
Looks good, just please set the default values for the fields that were originally required

PR #440, opened 2021-05-19
adding foreground propagation policy for userSignup deletion

ranakan19:
fix wrt: https://issues.redhat.com/browse/CRT-898

Unit tests not added, because it was getting too complicated to test it: and the user_signup_cleanup here in host-operator is only for scenarios which do not have an mur. 

Similar changes added in sandbox-sre: https://github.com/codeready-toolchain/sandbox-sre/pull/235


go.mod
```diff
@@ -41,6 +41,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common => github.com/ranakan19/toolchain-common v0.0.0-20210518183450-d7c417a42bd9
```
alexeykazakov, line 44: You don't need it, do you?

alexeykazakov (approved):

go.mod
ranakan19: no. I was wondering if it would be of any help to keep it nevertheless, like in sync everywhere, but i guess just setting blockOwnerDeletion without other details of owner reference doesn't help anyone. 
I'll go ahead and remove this.

rajivnathan (approved):

MatousJobanek (approved):

pkg/controller/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -134,8 +135,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 // DeleteUserSignup deletes the specified UserSignup
 func (r *Reconciler) DeleteUserSignup(userSignup *toolchainv1alpha1.UserSignup, logger logr.Logger) error {
-
-	err := r.Client.Delete(context.TODO(), userSignup)
+	propagationPolicy := metav1.DeletePropagationForeground
```
xcoulon, line 138: to test this option, you could use the mock client and set the `foregroundDeletion` finalizer on the UserSignup instead of deleting it, as the doc says:

> API sever will put the "foregroundDeletion" finalizer on the object, and sets its deletionTimestamp. This policy is cascading, i.e., the dependents will be deleted with Foreground.

it's not much as a test, but if someone in the future removes this option, the test will fail, so at least we have a (small) safety

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -143,6 +147,53 @@ func TestUserCleanup(t *testing.T) {
...
+		// now remove finalizer, reset MockDelete to nil and call reconcile
+		util.RemoveFinalizer(userSignup, "foregroundDeletion")
+		require.NoError(t, r.Client.Update(context.TODO(), userSignup))
+		fakeClient.MockDelete = nil
+
+		_, err = r.Reconcile(req)
+		require.NoError(t, err)
+
+		// now usersignup should be deleted
+		require.Error(t, r.Client.Get(context.Background(), key, userSignup))
```
alexeykazakov, line 194: With this finalizer you are actually testing the mock function. Not the controller. Controller doesn't add any finalizer, at least explicitly. It's added by your mock function. The only thing we care here is if the correct propagation option was passed to the delete client. So I would simplify the test:
```suggestion
		deleted := false
		fakeClient.MockDelete = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
			deleted = true
			require.Len(t, opts, 1)
			deleteOptions, ok := opts[0].(*client.DeleteOptions)
			require.True(t, ok)
			require.NotNil(t, deleteOptions)
			require.NotNil(t, deleteOptions.PropagationPolicy)
			assert.Equal(t, metav1.DeletePropagationForeground, *deleteOptions.PropagationPolicy)
			return nil
		}

		_, err := r.Reconcile(req)
		require.NoError(t, err)

		assert.True(t, deleted)
```

I actually mentioned it when saw that comment: https://github.com/codeready-toolchain/host-operator/pull/440/files#r636213278 days ago but forgot to press the submit button to submit the review :(

ranakan19: @alexeykazakov oh my! You're right, it's more of a test for the MockDelete. I guess i got too caught up on the "The object's metadata.finalizers contains the value "foregroundDeletion" condition check". I will certainly replace with your commit. I hope I didn't misunderstand @xcoulon's comment aimed at checking finalizers. I believe the suggested change should indeed be enough to get the security wrt the propagation policy.

alexeykazakov (approved):

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -143,6 +146,29 @@ func TestUserCleanup(t *testing.T) {
...
+
+		r, req, fakeClient := prepareReconcile(t, userSignup.Name, userSignup)
+		deleted := false
+		fakeClient.MockDelete = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
+			deleted = true
+			require.Len(t, opts, 1)
+			deleteOptions, ok := opts[0].(*client.DeleteOptions)
+			require.True(t, ok)
+			require.NotNil(t, deleteOptions)
+			require.NotNil(t, deleteOptions.PropagationPolicy)
```
xcoulon, line 163: I don't think you need this particular statement, but it's fine

pkg/controller/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -143,6 +146,29 @@ func TestUserCleanup(t *testing.T) {
...
 	})
 
+	t.Run("test propagation policy", func(t *testing.T) {
+		userSignup := test2.NewUserSignup(
+			test2.CreatedBefore(threeYears),
+			test2.VerificationRequired(),
+		)
+
+		r, req, fakeClient := prepareReconcile(t, userSignup.Name, userSignup)
+		deleted := false
```
xcoulon, line 156: this is not needed, I believe. What really matters in this test (as you did) is verifying that the object is deleted with the `DeletePropagationForeground`

PR #441, opened 2021-05-20
chore: update api with new format & kubebuilder version

MatousJobanek:
result of codeready-toolchain/api#237
* CRDs use a new name format: `toolchain.dev.openshift.com_<crd-plural-name>.yaml`
* sample CRs renamed to use the new name format as well
* slightly modified generated CRD as a result of using new kubebuilder (controller-tools)
* small changes in go.mod, so we can use the new controller-tools 0.4.1 together with client-go 0.18.3
* update the imports to the api repo/types


cmd/manager/main.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	api "github.com/codeready-toolchain/api/pkg/apis"
+	api "github.com/codeready-toolchain/api/api/v1alpha1"
```
xcoulon, line 13: same comment as in https://github.com/codeready-toolchain/toolchain-common/pull/177#discussion_r636104231 regarding the alias for the `github.com/codeready-toolchain/api/api/v1alpha1` package

deploy/crds/toolchain.dev.openshift.com_tiertemplates.yaml
```diff
@@ -0,0 +1,126 @@
+
```
xcoulon, line 1: is it a new manifest, or maybe git did not detect that it's a file that was renamed?

deploy/crds/toolchain.dev.openshift.com_bannedusers.yaml
```diff
@@ -1,6 +1,11 @@
+
+---
```
alexeykazakov, line 2: Previously when we tried to apply a CRD yaml with multiple manifests it failed. This is why we converted it to a single manifest yaml after generation. So, I assume it's not a problem anymore?

alexeykazakov (approved):

cmd/manager/main.go
alexeykazakov: That would require a massive refactoring and since this PR doesn't change any aliases let's maybe keep it as is in this PR and do this refactoring later?

deploy/crds/toolchain.dev.openshift.com_tiertemplates.yaml
alexeykazakov: I noticed it too but it seem there are some minor changes in the file and this is why it's not detected as renaming but as two changes: new file + deletion of the old one.

xcoulon (approved):

cmd/manager/main.go
MatousJobanek: I already did it in other PRs and since it's just a matter of applying a few `sed` commands, then it's fine to do it now: 6b734e8

deploy/crds/toolchain.dev.openshift.com_bannedusers.yaml
MatousJobanek: I tested it locally and it seems to be working fine, so kept it there.
The latest operator-sdk also works with this format.

deploy/crds/toolchain.dev.openshift.com_tiertemplates.yaml
MatousJobanek: I saw that too when I was making the commit - as you can see, there is also deleted `toolchain_v1alpha1_tiertemplate_crd.yaml` file so GH just probably didn't recognize the change. :shrug:

PR #458, opened 2021-06-09
added foreground propagation policy to deleteOptions of userAccount

ranakan19:
added foreground propagation policy to deleteOptions of userAccount

wrt story: https://issues.redhat.com/browse/CRT-897
related PRs for the fix:
- https://github.com/codeready-toolchain/member-operator/pull/262
- https://github.com/codeready-toolchain/toolchain-e2e/pull/322


controllers/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -312,8 +312,11 @@ func (r *Reconciler) deleteUserAccount(logger logr.Logger, targetCluster, name s
 		}
 		return requeueTime, nil
 	}
-
-	if err := memberCluster.Client.Delete(context.TODO(), userAcc); err != nil {
+	propagationPolicy := v1.DeletePropagationForeground
+	err = memberCluster.Client.Delete(context.TODO(), userAcc, &client.DeleteOptions{
+		PropagationPolicy: &propagationPolicy,
```
alexeykazakov, line 317: I don't think this will work. There is no owner reference between UA and MUR. They are in different clusters/namespaces.

ranakan19: @alexeykazakov But there is an owner reference between UA and NSTemplateSet, and they are in the same cluster.

alexeykazakov: Right.. so this will take care of foreground deletion between UA and dependent objects like NSTemplateSet.. Not between MUR and UA.
So, please disregard my comment :)

alexeykazakov (approved):
Looks good! But this doesn't really fully solve https://issues.redhat.com/browse/CRT-897, does it?
It's just one of multiple pieces we need to fix the issue, right?

xcoulon (approved):
Looks good! 🙌

PR #459, opened 2021-06-11
fix: don't delete reactivated UserSignups

MatousJobanek:
paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/319

### DIFF
--- a/controllers/usersignupcleanup/usersignup_cleanup_controller.go
+++ b/controllers/usersignupcleanup/usersignup_cleanup_controller.go
@@ -82,6 +82,8 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
 
+	// Check if the UserSignup is waiting for phone verification to be finished
+	// Check also if the status doesn't contain the approved condition set to true - it there is, then it means that it's reactivated UserSignup and we should skip it here
 	if states.VerificationRequired(instance) && !condition.IsTrue(instance.Status.Conditions, toolchainv1alpha1.UserSignupApproved) {
 
 		createdTime := instance.ObjectMeta.CreationTimestamp
@@ -104,6 +106,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		}, nil
 	}
 
+	// Check if the UserSignup is:
+	// * either deactivated
+	// * or if it is waiting for phone verification to be finished and at the same time the status contains the approved condition set to true
+	//   (in other words the UserSignup is reactivated but the phone verification hasn't been finished)
 	if states.Deactivated(instance) || (states.VerificationRequired(instance) && condition.IsTrue(instance.Status.Conditions, toolchainv1alpha1.UserSignupApproved)) {
 		// Find the UserSignupComplete condition
 		cond, found := condition.FindConditionByType(instance.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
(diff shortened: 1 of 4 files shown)


test/usersignup.go
```diff
@@ -62,9 +63,19 @@ func DeactivatedWithLastTransitionTime(before time.Duration) UserSignupModifier
 	}
 }
 
-func VerificationRequired() UserSignupModifier {
+func VerificationRequired(before time.Duration) UserSignupModifier {
 	return func(userSignup *toolchainv1alpha1.UserSignup) {
 		states.SetVerificationRequired(userSignup, true)
+
+		deactivatedCondition := toolchainv1alpha1.Condition{
```
xcoulon, line 70: so all users with `VerificationRequired` will get an extra `deactivatedCondition`? what about the tests for fresh/first-time users?

MatousJobanek: I copy-pasted the part and forgot to rename the condition var: bbb58c2

xcoulon: ok, thanks, I got confused with the condition name 🤦‍♂️

xcoulon (approved):

controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -83,7 +82,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
 
-	if states.VerificationRequired(instance) && !states.Approved(instance) {
+	if states.VerificationRequired(instance) && !condition.IsTrue(instance.Status.Conditions, toolchainv1alpha1.UserSignupApproved) {
```
alexeykazakov, line 85: While this will work I'm concert with the approach of relaying on leftover status conditions here...
Ideally we should introduce a new condition or maybe a new state? Something like `previously-provisioned` or something like this to mark signups which were at least once approved/provisioned in the past.
Just to separate the state/status of `approved-at-least-once-in-the-past` from `currently-approved`.
But it would require much bigger changes in the code. So, let's keep it simple for now and think about improving later when we have time.

But let's add a very detailed comment here in the code with explaining why we are checking the Approved status here.

alexeykazakov (approved):
Looks good! See my comment though. Let's add a detailed explanation of why we checking the approved state.

controllers/usersignupcleanup/usersignup_cleanup_controller.go
MatousJobanek: yeah, it's maybe not the best, but it was the quickest solution :stuck_out_tongue_closed_eyes: 

Another solution (without adding a new condition/status) would be checking the `toolchain.dev.openshift.com/activation-counter` annotations if it contains a non-zero value. But in this case I had a concern that we use the annotation only for the metrics, and it doesn't have anything in common with cleanup or any kind of state/status.

MatousJobanek: better? d95ef07

alexeykazakov: Actually.. now I remember. I thought about using this annotation in the first place. I think it's not that bad approach. This annotation is exactly for counting activations and this is what we need. We don't want to cleanup such resources because of the metrics. So, I think it's fine to use the annotation and we don't need any new status/state.
WDYT?

MatousJobanek: well, both approaches have pros and cons... Using the annotation was my first thought, but since we mainly rely on conditions in that controller, I decided to use the same approach. The second reason for using the condition was what I described in the previous comment.

But I'm fine with using the annotation too. I can change it, but not today - I can do it on Monday. Or - if you want - you can contribute to my PR :wink:

alexeykazakov: I'm on PTO today ;)

alexeykazakov: My main concern with using Approved condition is that in becomes `approved-at-some-point-in-the-past` instead of how we treated it up to now `currently-approved`. It's unclear and error prone IMO.
Annotation counter is IMO better. But I will leave it to you. Keep it as is or switch to annotation on Monday. Or merge it now and then switch to annotation later. I'm fine either way.

PR #461, opened 2021-06-14
Sync config to members

rajivnathan:
Updates the toolchainconfig controller to automatically sync member operator configuration to member clusters every 10 seconds.

Also contains updated package usage from the toolchain-common dependency.

After each reconcile the toolchainconfig status is updated. The status includes a`SyncComplete` condition that is either True or False depending on whether the sync was successful and if unsuccessful it will also contain a map of sync errors indexed by cluster name for debugging/investigation purposes.


controllers/toolchainconfig/sync_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+func TestSyncMemberConfigs(t *testing.T) {
+	defaultMemberConfig := config.NewMemberOperatorConfig(config.MemberStatus().RefreshPeriod("5s"))
+	specificMemberConfig := config.NewMemberOperatorConfig(config.MemberStatus().RefreshPeriod("10s"))
+
+	t.Run("sync success", func(t *testing.T) {
+
+		t.Run("no member clusters available - skip sync", func(t *testing.T) {
+			// given
+			toolchainConfig := config.NewToolchainConfig(config.AutomaticApproval().Enabled().MaxUsersNumber(123, config.PerMemberCluster("member1", 321)), config.Members().Default(defaultMemberConfig.Spec), config.Members().SpecificPerMemberCluster("member2", specificMemberConfig.Spec))
```
MatousJobanek, line 30: such a line is super long and hard to read - could you please break it for example to:
```suggestion
			toolchainConfig := config.NewToolchainConfig(
				config.AutomaticApproval().Enabled().MaxUsersNumber(123, 
					config.PerMemberCluster("member1", 321)), 
					config.Members().Default(defaultMemberConfig.Spec), 
					config.Members().SpecificPerMemberCluster("member2", specificMemberConfig.Spec))
```

controllers/toolchainconfig/sync.go
```diff
@@ -0,0 +1,83 @@
...
+			syncErrors[toolchainCluster.Name] = err.Error()
+		}
+	}
+
+	// add errors for any MemberOperatorConfigs that haven't been synced because there is no matching toolchaincluster
+	if len(membersWithSpecificConfig) > 0 {
+		for k := range membersWithSpecificConfig {
+			syncErrors[k] = "specific member configuration exists but no matching toolchaincluster was found"
+		}
+	}
```
MatousJobanek, line 57: this returns an error for clusters that are not returned from `getMembersFunc()` but are specified in the ToolchainConfig. However, this code is not executed when no cluster is returned from `getMembersFunc()`. Shouldn't we remove the lines 30-33 https://github.com/codeready-toolchain/host-operator/blob/f38406bd2fb10c21cb9eb57d20e3f26c2cf32965/controllers/toolchainconfig/sync.go#L30-L33 ?

controllers/toolchainconfig/sync_test.go
MatousJobanek: Just out of curiosity - why do we need to specify the max number of users per member cluster `config.PerMemberCluster("member1", 321)`? And why do we need to specify the AutomaticApproval configuration at all? I guess that this is not needed for the purpose of these tests.

controllers/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -60,12 +73,70 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	}
+
+	if syncErrs := sync.syncMemberConfigs(toolchainConfig); len(syncErrs) > 0 {
+		for cluster, errMsg := range syncErrs {
+			err := fmt.Errorf(errMsg)
+			reqLogger.Error(err, "error syncing to member cluster '%s'", cluster)
+		}
+		return defaultReconcile, r.setSyncFailureStatusCondition(reqLogger, toolchainConfig, syncErrs)
+	}
+	return defaultReconcile, r.setCompleteStatusCondition(reqLogger, toolchainConfig)
```
MatousJobanek, line 98: Just out of curiosity - what will happen when we will modify the `ToolchainConfig` for a few times (let say 3 times) between the deployments? How many reconciliations there will be? Always only one, or 4 of them? 
IIRC, there is only one reconciliation scheduled for one CR with the given name `config`, is that correct?

controllers/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -60,12 +73,70 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+}
+
+// setSyncFailureStatusCondition sets the ToolchainConfig status condition to `SyncComplete=false/reason=syncFailure`
+func (r *Reconciler) setSyncFailureStatusCondition(reqLogger logr.Logger, toolchainConfig *toolchainv1alpha1.ToolchainConfig, syncErrs map[string]string) error {
+	toolchainConfig.Status.SyncErrors = syncErrs
+	toolchainConfig.Status.Conditions = []toolchainv1alpha1.Condition{ToSyncFailure()}
+	err := r.Client.Status().Update(context.TODO(), toolchainConfig)
```
MatousJobanek, line 105: Why not using the function from toolchain-common?
```
AddOrUpdateStatusConditionsWithLastUpdatedTimestamp
```

controllers/toolchainconfig/toolchainconfig_controller_test.go
```diff
@@ -6,103 +6,173 @@ import (
...
-		actual, err := GetConfig(NewFakeClient(t), HostOperatorNs)
-		require.NoError(t, err)
-		assert.False(t, actual.AutomaticApproval().IsEnabled())
-		assert.Equal(t, 123, actual.AutomaticApproval().MaxNumberOfUsersOverall())
-		assert.Equal(t, config.Spec.Host.AutomaticApproval.MaxNumberOfUsers.SpecificPerMemberCluster, actual.AutomaticApproval().MaxNumberOfUsersSpecificPerMemberCluster())
-		assert.Equal(t, 100, actual.AutomaticApproval().ResourceCapacityThresholdDefault())
+	defaultMemberConfig := testconfig.NewMemberOperatorConfig(testconfig.MemberStatus().RefreshPeriod("5s"))
+	specificMemberConfig := testconfig.NewMemberOperatorConfig(testconfig.MemberStatus().RefreshPeriod("10s"))
+
+	t.Run("success", func(t *testing.T) {
```
MatousJobanek, line 30: Could you please verify in any of the tests that the member configs were really updated? In other words, that the synchronization logic was triggered?

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -42,6 +44,7 @@ func newNsTemplateTier(tierName string, nsTypes ...string) *toolchainv1alpha1.NS
 			TemplateRef: nstemplatetiers.NewTierTemplateName(tierName, nsType, revision),
 		}
 	}
+	fmt.Printf("Hello")
```
MatousJobanek, line 47: Hi :hand: 
```suggestion
```
:ghost:

test/toolchainconfig_assertions.go
```diff
@@ -0,0 +1,71 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ToolchainConfigAssertion struct {
+	config         *toolchainv1alpha1.ToolchainConfig
+	client         client.Client
+	namespacedName types.NamespacedName
+	t              test.T
+}
```
MatousJobanek, line 19: Wouldn't it make sense to move the methods to toolchain-common repo? I guess that they could be useful also in e2e tests. WDYT?

controllers/toolchainconfig/sync.go
```diff
@@ -0,0 +1,83 @@
...
+		s.logger.Info("No toolchainclusters were found, skipping MemberOperatorConfig syncing")
+		return syncErrors
+	}
+
+	membersWithSpecificConfig := toolchainConfig.Spec.Members.SpecificPerMemberCluster
+
+	for _, toolchainCluster := range memberToolchainClusters {
+		memberConfigSpec := toolchainConfig.Spec.Members.Default
+
+		if c, ok := membersWithSpecificConfig[toolchainCluster.Name]; toolchainCluster.Type == cluster.Member && ok {
```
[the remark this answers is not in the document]
rajivnathan: You're right, it's not needed. Removed it in 83a653b2b5db2f88a73a770838fac9536c6a3ffd

controllers/toolchainconfig/sync_test.go
rajivnathan: Sure, I can do that to make it easier to read. No, it's not needed but I figured the majority of the time the toolchainconfig resource would have those values set so I included it but I can remove it since it might be confusing. Done in 83a653b2b5db2f88a73a770838fac9536c6a3ffd

controllers/toolchainconfig/toolchainconfig_controller.go
rajivnathan: Updated the status updates to use a common function that makes use of the AddOrUpdateStatusConditions. See 83a653b2b5db2f88a73a770838fac9536c6a3ffd

controllers/toolchainconfig/toolchainconfig_controller_test.go
rajivnathan: Done with efe10d48c01cf0288d61cf0898f5983448f33ca3 and da1c26c3963382219fa07206e74cccd1949f4ac7

controllers/usersignup/usersignup_controller_test.go
rajivnathan: Oops!

test/toolchainconfig_assertions.go
rajivnathan: I thought we'd focus on checking only the status and actual config behaviour in e2e but it doesn't hurt to have it in toolchain-common so I'll add it there.

rajivnathan: Done https://github.com/codeready-toolchain/host-operator/pull/461/commits/709b826b8247c6b3232684d9627ea7f3e600723b

And here's the PR for toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/188

controllers/toolchainconfig/toolchainconfig_controller.go
rajivnathan: So it's interesting, I tried out this scenario (3 updates) and this is what I noticed from the logs:

 20:43:33
 20:43:43
 20:43:46 Update #1
 20:43:53 
 20:43:55 Update #2
 20:44:03 
 20:44:03 Update #3
 20:44:13
 20:44:23
 20:44:33
...

So it appears that the updates are just additional reconciles but the original cadence of scheduled reconciles is maintained throughout. So yes, only one reconciliation schedule for one CR is what it seems to be.

controllers/toolchainconfig/sync.go
MatousJobanek: Right, keep it as it is :+1:

controllers/toolchainconfig/toolchainconfig_controller.go
MatousJobanek: Great - so it confirms my expectation :-) - thanks for verifying it :+1:

controllers/toolchainconfig/toolchainconfig_controller.go
MatousJobanek: btw, I proposed to use `AddOrUpdateStatusConditionsWithLastUpdatedTimestamp` because that one also updates the `LastUpdatedTime`.
`AddOrUpdateStatusConditions` only compares the content of the condition but not timestamps, so if there is no change, then the condition is not updated.

MatousJobanek (approved):
Thanks for addressing my comments. :+1: 
I'm approving it but please modify the status update logic - use the `AddOrUpdateStatusConditionsWithLastUpdatedTimestamp` there so it updates the condition everytime with new `LastUpdatedTime` timestamp. You can also add a simple test for that.

controllers/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -60,12 +75,69 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+func (r *Reconciler) updateStatus(reqLogger logr.Logger, toolchainConfig *toolchainv1alpha1.ToolchainConfig, syncErrs map[string]string, newCondition toolchainv1alpha1.Condition) error {
+	var conditionUpdated bool
+	syncErrsUpdated := reflect.DeepEqual(syncErrs, toolchainConfig.Status.SyncErrors)
+
+	toolchainConfig.Status.SyncErrors = syncErrs
+	toolchainConfig.Status.Conditions, conditionUpdated = condition.AddOrUpdateStatusConditions(toolchainConfig.Status.Conditions, newCondition)
+	if !conditionUpdated && !syncErrsUpdated {
+		// Nothing changed
+		return nil
+	}
```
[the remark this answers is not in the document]
rajivnathan: And I guess it also serves as a way to see that sync is working.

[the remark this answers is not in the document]
rajivnathan: Updated to use AddOrUpdateStatusConditionsWithLastUpdatedTimestamp: d86a1be62e23bc188e198cfde0e211d9a6301da5

controllers/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -60,12 +75,69 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+
+// ToBeComplete condition when the update completed with success
+func ToBeComplete() toolchainv1alpha1.Condition {
+	currentTime := metav1.Now()
+	return toolchainv1alpha1.Condition{
+		Type:               toolchainv1alpha1.ToolchainConfigSyncComplete,
+		Status:             corev1.ConditionTrue,
+		Reason:             toolchainv1alpha1.ToolchainConfigSyncedReason,
+		LastTransitionTime: currentTime,
+		LastUpdatedTime:    &currentTime,
```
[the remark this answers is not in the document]
rajivnathan: Done in d90f96ce42b64622e7d0ab71c508be73cf62fbf4

PR #462, opened 2021-06-14
fix: use activations annotation in cleanup controller

MatousJobanek:
https://github.com/codeready-toolchain/host-operator/pull/459#discussion_r650143229


controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -82,9 +82,11 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	reqLogger = reqLogger.WithValues("username", instance.Spec.Username)
 
+	activations, activationsAnnotationPresent := instance.Annotations[toolchainv1alpha1.UserSignupActivationCounterAnnotationKey]
+
 	// Check if the UserSignup is waiting for phone verification to be finished
-	// Check also if the status doesn't contain the approved condition set to true - it there is, then it means that it's reactivated UserSignup and we should skip it here
-	if states.VerificationRequired(instance) && !condition.IsTrue(instance.Status.Conditions, toolchainv1alpha1.UserSignupApproved) {
+	// Check also if the UserSignup doesn't have at least one activation - if it has, then it means that it's reactivated UserSignup and we should skip it here
+	if states.VerificationRequired(instance) && (!activationsAnnotationPresent || activations == "0") {
```
xcoulon, line 89: IIRC, we never set the annotation value to `0` (but it's fine/safe to keep this assertion)

controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -108,9 +110,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	// Check if the UserSignup is:
 	// * either deactivated
-	// * or if it is waiting for phone verification to be finished and at the same time the status contains the approved condition set to true
+	// * or if it is waiting for phone verification to be finished and at the same it has at least one activation stored in the annotation
 	//   (in other words the UserSignup is reactivated but the phone verification hasn't been finished)
-	if states.Deactivated(instance) || (states.VerificationRequired(instance) && condition.IsTrue(instance.Status.Conditions, toolchainv1alpha1.UserSignupApproved)) {
+	if states.Deactivated(instance) || (states.VerificationRequired(instance) && activationsAnnotationPresent && activations != "0") {
```
xcoulon, line 115: if a user is already deactivated, her UserSignup resource should always have the `activation-counter` annotation (with a value >= `1`)

controllers/usersignupcleanup/usersignup_cleanup_controller.go
MatousJobanek: ah, right - I got confused with this code https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/usersignup_controller.go#L420-L428
at the first glance, it's not really clear that the update for `"0"` is skipped

controllers/usersignupcleanup/usersignup_cleanup_controller.go
MatousJobanek: However, at this point, the `UserSignup` CR is not in the deactivated state anymore - that's why we check if the state contains `verificationRequired` and then also the annotation.

xcoulon: still, if the user has been deactivated in the past, doesn't it imply that she already has the `activation-counter` annotation with value equal or greater than 1? 
(btw, I find `states.Deactivated(...)` to be a bit confusing: at first sight, I assume that this func checks the *current* state)

MatousJobanek: in this part, we want to clean:
* deactivated UserSignups that has been deactivated for more than one year
* UserSignups that were deactivated, the user tried to reactivated but hasn't verified the phone for more than one year

that being said - we need to check both:
* if the current state is `deactivated`
* or the current state is `verificationRequired` and the number of activations is > 0

xcoulon: ok, got it now. Please ignore my previous comments, they were wrong 🤦‍♂️

xcoulon (approved):

alexeykazakov (approved):

PR #463, opened 2021-06-14
feat: add option to force metric synchronization from resources

xcoulon:
trigger counter synchronization from existing resources at startup
when the `spec.host.metrics.forceSynchronization` field is set to `true`
in the `ToolchainConfig` resource  loaded by the host-operator

Related PRs:
- https://github.com/codeready-toolchain/toolchain-e2e/pull/321
- https://github.com/codeready-toolchain/api/pull/244
- https://github.com/codeready-toolchain/toolchain-common/pull/187


alexeykazakov (approved):
Looks good but I have the same question as in the API PR. For what cases do we need such a flag?

xcoulon:
> Looks good but I have the same question as in the API PR. For what cases do we need such a flag?

This flag will force the recount of the resources when it's set to `true`, which we will need to do to adjust the metrics now that we fixed the clean-up controller (this bug may explain why we had a drop of 350 users last time we triggered a recount).

it's implemented as a "feature flag" instead of hard-coded, so we can enable/disable it more easily.

rajivnathan (approved):

MatousJobanek (approved):
Nice and simple :+1:

xcoulon:
@rajivnathan I had to make some changes in the `controllers/toolchainconfig` package in order to avoid import cycles as that this PR introduces 😬 :

```
package github.com/codeready-toolchain/host-operator/controllers/toolchainconfig (test)
        imports github.com/codeready-toolchain/host-operator/test
        imports github.com/codeready-toolchain/host-operator/pkg/counter
        imports github.com/codeready-toolchain/host-operator/controllers/toolchainconfig
```
        
Most of the changes are related to tests which were moved into the `toolchainconfig_test` package and as a consequence, some types and funcs were also needed to be package-exported. See 7336c72

Please let me know if those changes are ok for you ;)

PR #468, opened 2021-06-18
feat: metric to track phone verification failures

xcoulon:
when a UserSignup resource is deleted because
the external user failed to pass the phone verification,
the cleanup controller checks if the user initiated an
attempt (and maybe some technical issue occurred)
or not (maybe the user did not want to input her phone number,
for privacy reasons).
This info is found by checking if the `toolchain.dev.openshift.com/verification-code`
annotation has been set or not.

Fixes https://issues.redhat.com/browse/CRT-1064

### DIFF
--- a/controllers/usersignupcleanup/usersignup_cleanup_controller.go
+++ b/controllers/usersignupcleanup/usersignup_cleanup_controller.go
@@ -8,6 +8,7 @@ import (
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	crtCfg "github.com/codeready-toolchain/host-operator/pkg/configuration"
+	"github.com/codeready-toolchain/host-operator/pkg/metrics"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	"github.com/codeready-toolchain/toolchain-common/pkg/states"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):
This looks good but the downside of this approach is that it shows verification failures for signups created 7 days ago. Let's try to reflect this somehow in grafana so it's clear what this counter represents.

alexeykazakov:
Can we add any e2e tests for that metric?

xcoulon:
> Can we add any e2e tests for that metric?

we can't change the `creationTime` on existing resources (it's an immutable field), so I don't see how we could trigger the UserSignupCleanupController at all. AFAIK, there is no E2E tests at all on this topic, sadly.

xcoulon:
> This looks good but the downside of this approach is that it shows verification failures for signups created 7 days ago. Let's try to reflect this somehow in grafana so it's clear what this counter represents.

yes, I agree with you, but the solution is so much simpler than having to watch the UserSignup resource in another controller, especially because some of the annotations related to phone verification are reset after 24hrs, etc.
See discussion on 

> > Can we add any e2e tests for that metric?
> 
> we can't change the `creationTime` on existing resources (it's an immutable field), so I don't see how we could trigger the UserSignupCleanupController at all. AFAIK, there is no E2E tests at all on this topic, sadly.

see discussion on https://coreos.slack.com/archives/CHK0J6HT6/p1624010602481500

xcoulon:
> This looks good but the downside of this approach is that it shows verification failures for signups created 7 days ago. Let's try to reflect this somehow in grafana so it's clear what this counter represents.

I agree that this shows failures with a 7 days delay, but I opted for this solution because it was much simpler and straightforward than having to watch the UserSignup resource and track relevant changes in the annotations related to phone verification (while taking into account multiple attempts per code, and subsequent retries after 3 failed attempts and 24h of pause, etc.)

See discussion about it on https://coreos.slack.com/archives/CHK0J6HT6/p1623935963440300

PR #469, opened 2021-06-21
Deactivation controller to watch UserSignup as secondary resource

sbryzak:
This PR modifies the Deactivation controller to watch UserSignup as a secondary resource, rather than forcing a requeue of set duration.

Fixes https://issues.redhat.com/browse/CRT-1070


controllers/deactivation/mapper.go
```diff
@@ -0,0 +1,53 @@
...
+		mur := &toolchainv1alpha1.MasterUserRecord{}
+
+		if err := b.client.Get(context.TODO(), client.ObjectKey{
+			Namespace: userSignup.Namespace,
+			Name:      userSignup.Status.CompliantUsername,
+		}, mur); err != nil {
+			mapperLog.Error(err, "Could not find MasterUserRecord resource with name", "name",
+				userSignup.Status.CompliantUsername)
+			return nil
+		}
```
MatousJobanek, line 34: May I know why do you get the MUR? As I see it - the mapper should only deal with mapping, which means that it should only return the `userSignup.Status.CompliantUsername`. If the MUR doesn't exist, then it's not the responsibility of the mapper, but of the controller.

MatousJobanek: In addition to that, the fact that the MUR doesn't exist is not an error - it's a pretty valid use-case, so we should not log an error here.
If you want to filter out those UserSignups whose MUR doesn't exist, then you should use predicated for that, not a mapper.

controllers/deactivation/mapper.go
```diff
@@ -0,0 +1,53 @@
...
+
+		req := []reconcile.Request{}
+
+		ns, err := k8sutil.GetWatchNamespace()
+		if err != nil {
+			mapperLog.Error(err, "Could not determine watched namespace")
+			return nil
+		}
```
MatousJobanek, line 42: I would rather take the namespace name from the UserSignup resource - they both should exist in the same namespace

controllers/deactivation/mapper.go
```diff
@@ -0,0 +1,53 @@
...
+
+		req = append(req, reconcile.Request{
+			NamespacedName: types.NamespacedName{Namespace: ns, Name: mur.Name},
+		})
+
+		return req
+	}
+
+	// the obj was not a UserSignup
+	return []reconcile.Request{}
```
MatousJobanek, line 52: In this case, I would log an error, because the mapper should be called only for UserSignups.

controllers/deactivation/deactivation_controller.go
```diff
@@ -178,19 +187,17 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		// 4. This  controller reconciles again, and if it doesn't find the notification created status condition as
-		//    expected, requeues again, otherwise:
-		// 5. This controller calculates the amount of time that has passed since the deactivating notification was sent,
+		// 3. This controller watches UserSignup as a secondary resource, so will be reconciled again once the
+		//    UserSignup is updated.
+		// 4. This controller calculates the amount of time that has passed since the deactivating notification was sent,
 		//  based on the LastTransitionTime of the condition. If enough time has now passed, it sets the UserSignup to deactivated.
-
-		return reconcile.Result{Requeue: true, RequeueAfter: time.Duration(10) * time.Second}, nil
+		return reconcile.Result{}, nil
```
MatousJobanek, line 200: I guess that this should be done also at line 210
 https://github.com/codeready-toolchain/host-operator/blob/91e62f4a75cdcbcd998a304ce7b66afa20cf2096/controllers/deactivation/deactivation_controller.go#L210

MatousJobanek (commented):
Thanks for taking care of it

controllers/deactivation/deactivation_controller.go
sbryzak: Fixed, thanks.

controllers/deactivation/mapper.go
sbryzak: Done

controllers/deactivation/mapper.go
sbryzak: Done

controllers/deactivation/mapper.go
sbryzak: Thanks for the explanation, I've removed the MUR lookup.

controllers/deactivation/mapper.go
```diff
@@ -0,0 +1,38 @@
...
+var _ handler.Mapper = UserSignupToMasterUserRecordMapper{}
+var mapperLog = ctrl.Log.WithName("UserSignupToMasterUserRecordMapper")
+
+func (b UserSignupToMasterUserRecordMapper) Map(obj handler.MapObject) []reconcile.Request {
+	if userSignup, ok := obj.Object.(*toolchainv1alpha1.UserSignup); ok {
+
+		req := []reconcile.Request{}
+
+		req = append(req, reconcile.Request{
+			NamespacedName: types.NamespacedName{Namespace: userSignup.Namespace, Name: userSignup.Status.CompliantUsername},
```
alexeykazakov, line 27: `CompliantUsername` can be empty. This should be fine but maybe we want to return an empty `[]Request{}` in this case?

controllers/deactivation/mapper.go
```diff
@@ -0,0 +1,38 @@
...
+	client client.Client
+}
+
+var _ handler.Mapper = UserSignupToMasterUserRecordMapper{}
+var mapperLog = ctrl.Log.WithName("UserSignupToMasterUserRecordMapper")
+
+func (b UserSignupToMasterUserRecordMapper) Map(obj handler.MapObject) []reconcile.Request {
+	if userSignup, ok := obj.Object.(*toolchainv1alpha1.UserSignup); ok {
+
+		req := []reconcile.Request{}
```
alexeykazakov, line 24: Or just simply:
```
return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: userSignup.Namespace, Name: userSignup.Status.CompliantUsername}}
```

controllers/deactivation/mapper_test.go
```diff
@@ -0,0 +1,66 @@
...
+		req := mapper.Map(handler.MapObject{
+			Object: userSignup,
+		})
+
+		require.Len(t, req, 1)
+		require.Equal(t, types.NamespacedName{
+			Namespace: mur.Namespace,
+			Name:      mur.Name,
+		}, req[0].NamespacedName)
+	})
```
alexeykazakov, line 65: Let's also add tests for:
- UserSignup doesn't map to any MUR
- UserSignup doesn't have ComplainUsername to map to any MUR
And add an extra MUR as a noise so we don't always return the only MUR available.

alexeykazakov: And also the test for the case when the object is not a Usersignup.

controllers/deactivation/mapper_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Status: toolchainv1alpha1.UserSignupStatus{
+				CompliantUsername: "",
+			},
+		}
+
+		mapper := &UserSignupToMasterUserRecordMapper{}
+
+		// This is required for the mapper to function
+		restore := test.SetEnvVarAndRestore(t, k8sutil.WatchNamespaceEnvVar, test.HostOperatorNs)
+		defer restore()
```
alexeykazakov, line 61: You could do it once before both tests.

alexeykazakov (approved):

controllers/deactivation/mapper_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestUserSignupToMasterUserRecordMapper(t *testing.T) {
+	// This is required for the mapper to function
+	restore := test.SetEnvVarAndRestore(t, k8sutil.WatchNamespaceEnvVar, test.HostOperatorNs)
+	defer restore()
```
MatousJobanek, line 22: @sbryzak Just curious - do we still need to set this env variable? If not, then feel free to keep it as it is - I'll remove it in my PR. I just wanted to check it with you

PR #470, opened 2021-06-21
build: use Codecov GitHub Action

xcoulon:
Paired with https://github.com/openshift/release/pull/19491

Fixes https://issues.redhat.com/browse/CRT-1033

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   test-with-coverage:
     name: Test with Coverage
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - name: Install Go
       uses: actions/setup-go@v2
@@ -40,7 +40,7 @@ jobs:
     
   golangci:
     name: GolangCI Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - name: Install Go
       uses: actions/setup-go@v2
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -10,6 +10,34 @@ on:
       - master
 
 jobs:
+  test-with-coverage:
+    name: Test with Coverage
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


MatousJobanek (approved):

.github/workflows/ci-build.yml
```diff
@@ -10,6 +10,34 @@ on:
       - master
 
 jobs:
+  test-with-coverage:
+    name: Test with Coverage
+    runs-on: ubuntu-latest
```
alexeykazakov, line 15: As far as I remember we had issues with some dependencies in the latest Ubuntu and had to switch to some specific version. So maybe it's safer to stick with some LTS version (*.04 versions)? cc: @MatousJobanek

alexeykazakov (approved):

.github/workflows/ci-build.yml
xcoulon: oh, I did not know about that 😮 
yes, indeed, we could use Ubuntu 18.04 or 20.04: https://github.com/actions/virtual-environments#available-environments

xcoulon: switched to Ubuntu 20.04 (https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) in b8109ce

alexeykazakov: Yeah. 20.04 would be preferable: https://ubuntu.com/about/release-cycle

MatousJobanek: it's not a big deal for these CI jobs - there was some problem with a specific version of Podman that was installed in the latest version of Ubuntu - that's why we use 18.04 in CD job

MatousJobanek: apart from that - I guess that 20.04 & latest is the same: https://github.com/actions/virtual-environments#available-environments

xcoulon: right, `20.04` and `latest` are equivalent. I'll go ahead and merge with `20.04` for now, but we can always change back to `latest` (or something else) in the future if needed

PR #475, opened 2021-06-24
fix: don't create notification when already exists

MatousJobanek:
when the update of the status (that adds the condition) fails, then the controller creates two (or even more) "provisioned" notifications for the same user. I faced this problem when running e2e tests locally.

So I added a call that lists the notification by their labels - if there is one already present, then it sets only the condition, but doesn't create the notification CR.


controllers/masteruserrecord/sync.go
```diff
@@ -183,32 +183,41 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
...
+			toolchainv1alpha1.NotificationUserNameLabelKey: s.record.Name,
+			// NotificationTypeLabelKey is only used for easy lookup for debugging and e2e tests
+			toolchainv1alpha1.NotificationTypeLabelKey: toolchainv1alpha1.NotificationTypeProvisioned,
+		}
+		opts := client.MatchingLabels(labels)
+		notificationList := &toolchainv1alpha1.NotificationList{}
+		if err := s.hostClient.List(context.TODO(), notificationList, opts); err != nil {
 			return false, err
 		}
+		// if there is no existing notification with that labels
```
alexeykazakov, line 197:
```suggestion
		// if there is no existing notification with these labels
```
```suggestion
		// if there is no existing notification with that labels
```

controllers/masteruserrecord/sync.go
```diff
@@ -183,32 +183,41 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
...
-				// (or removed) in a future PR
-				UserID:   s.record.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey],
-				Template: notificationtemplates.UserProvisioned.Name,
-			},
-		}
-
-		err := controllerutil.SetControllerReference(s.record, notification, s.scheme)
-		if err != nil {
+		labels := map[string]string{
+			// NotificationUserNameLabelKey is only used for easy lookup for debugging and e2e tests
```
alexeykazakov, line 187: It is not for tests or debugging anymore. You do rely on this label in the controller now.

controllers/masteruserrecord/sync.go
```diff
@@ -183,32 +183,41 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
...
+			// NotificationTypeLabelKey is only used for easy lookup for debugging and e2e tests
+			toolchainv1alpha1.NotificationTypeLabelKey: toolchainv1alpha1.NotificationTypeProvisioned,
+		}
+		opts := client.MatchingLabels(labels)
+		notificationList := &toolchainv1alpha1.NotificationList{}
+		if err := s.hostClient.List(context.TODO(), notificationList, opts); err != nil {
 			return false, err
 		}
+		// if there is no existing notification with that labels
+		if len(notificationList.Items) == 0 {
```
alexeykazakov, line 198: Let's log an info message saying that the notification was not created because it's already exists.

alexeykazakov (approved):
Looks good. Just a few minor comments.

controllers/masteruserrecord/sync.go
MatousJobanek: right :+1: 
I just blindly copy-pasted it and didn't pay attention on the comments - fixed in ea4f95c

controllers/masteruserrecord/sync.go
MatousJobanek: done 425e620

PR #479, opened 2021-06-30
fix: skip reconcile when UserSignup is being deleted

MatousJobanek:
When UserSignup is being deleted, then (in my opinion) we should not do anything in the reconcile. 
My guess is that the flakiness is caused by reconciliation that creates MUR just after their deletion - this is at least what it looks like from the logs. 
I can be wrong, but still - when the UserSignup is being deleted, we should just exit the reconcile.
I'm open to discussion ;-)

### DIFF
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -139,7 +139,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	logger = logger.WithValues("username", userSignup.Spec.Username)
 
 	if util.IsBeingDeleted(userSignup) {
-		logger.Error(err, "The UserSignup is being deleted")
+		logger.Info("The UserSignup is being deleted")
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/usersignup/usersignup_controller.go
```diff
@@ -137,6 +138,11 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	logger = logger.WithValues("username", userSignup.Spec.Username)
 
+	if util.IsBeingDeleted(userSignup) {
+		logger.Error(err, "The UserSignup is being deleted")
```
alexeykazakov, line 142:
```suggestion
		logger.Info("The UserSignup is being deleted")
```

alexeykazakov (approved):
Yeah.. this might be the reason... especially when we delete the UserSingup with Foreground deletion option which holds the UserSignup alive (with the deletion timestamp) until the MUR is actually deleted.
Good catch.

controllers/usersignup/usersignup_controller.go
MatousJobanek: copy-paste - thanks :-)

PR #480, opened 2021-07-02
Formatted email address to be more RFC2822 compliant

sbryzak:
After much digging through the RFC2822 spec, it seems the simplest fix for countering invalid characters that users insert into their names is to just surround their name with double quotes, making it a `quoted-string` (see https://datatracker.ietf.org/doc/html/rfc2822#section-3.2.5)

Fixes https://issues.redhat.com/browse/CRT-1166

### DIFF
--- a/controllers/notification/mailgun_notification_delivery_service.go
+++ b/controllers/notification/mailgun_notification_delivery_service.go
@@ -3,8 +3,9 @@ package notification
 import (
 	"context"
 	"fmt"
-	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"time"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 
 	"github.com/mailgun/mailgun-go/v4"
 )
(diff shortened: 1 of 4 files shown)


alexeykazakov (commented):
What if the name itself contains "?

sbryzak:
If the name itself contains a double quote, then it will still fail.  There's probably numerous other ways that the user could still break compliance with RFC2822, probably the only way around it is to strip out anything that's not an alpha character.

alexeykazakov:
Should we just use the email only without the name?

sbryzak:
We could.. it's nicer to include a full email address though.  In its current form this PR should fix all currently broken notifications, however I am about to push through some further changes which will strip out double quote `"` characters also.

MatousJobanek:
Maybe a dumb question - what is actually the benefit of adding the name? In which way it's nicer? 
Will it look different than when we just use the email?

sbryzak:
I guess it depends on the mail client used, in general I think gmail will kind of hide those details anyway (and just put `to me` in the header).  I've got nothing against removing the name altogether though.

MatousJobanek:
In some of the UserSignups I also saw names like `[PRIVATE]` or other crazy formats. What if the names are not filled at all - then it would be an empty string - would it be OK?
Personally, I would vote for keeping it simple and use only email. But that is my point of view.

sbryzak:
No problem, I've simplified it to just use the provided email address without the user's name.

MatousJobanek (approved):
I like this simpler format - thanks for dropping the name :pray:

PR #486, opened 2021-07-27
Create tier for idling extended to 24 hours

rajivnathan:
Creates a new tier that's the same as the base tier but extends the idling period to 24 hours.


alexeykazakov (approved):
Thanks!

PR #490, opened 2021-08-03
feat: reuse already existing tier templates in other tiers

MatousJobanek:
As you probably all know, the amount of duplication in the tier templates has been driving me crazy for a long time. I also noticed that we have some inconsistency there (for example advanced, team & test tier don't contain RHOAS and SBO CRQs).

## based_on_tier.yaml

So, this PR enables (with the help of a  new config file) reusing of already existing tier templates for defining a new tier. 
The idea is:
* everything that should be changed in the new tier is moved to parameters (to make it easily configurable) - you can see that in the changed `base` templates
* The new tier is defined via a `based-on-tier.yaml` file which says:
    * which tier should be reused
    * what the target tier is
    * the new values for the parameters
 for example:
```
from:
  name: base
to:
  name: baseextended
  parameters:
  - name: DEACTIVATION_TIMEOUT_DAYS
    value: 180 
```
which defines that for tier `baseextended` the logic should reuse the templates from `base` tier and replace the `DEACTIVATION_TIMEOUT_DAYS` parameter value with the new one: `180`

I believe that in this way, we can cover 90% of all use-cases and thus avoid code duplication as much as possible. 
I already configured `basedeactivationdisabled`, `basedeactivationdisabled`, `baseextended`, and `advanced`.

## Revisions

There is a small (I believe that cosmetic) change in the revisions that are used for the `TierTemplates`. Right now we take the revisions for each of the templates and then we use them in the name of the `TierTemplates`. However, most of the tiers won't contain the template files, but the `based_on_tier.yaml` file. And we need to catch the changes in both of them... 
So, this PR combines the revisions from both files - from the `based_on_tier.yaml` & from the particular template file that is being reused. If the tier doesn't contain `based_on_tier.yaml` file, then it just duplicates the same revision as is set for the template file.

let's consider this scenario:
```yaml
advanced/based_on_tier: "1234567"
base/cluster: "abc1234"
base/ns_dev: "bcd2345"
base/ns_stage: "cde3456"
```
as a result, there will be `TierTemplate` CRs with names:
* `advanced-clusterresources-1234567-abc1234`  
( `1234567` is revision of `based_on_tier.yaml` file, `abc1234` is revision of `base/cluster.yaml`)
* `advanced-dev-1234567-bcd2345`   
( `1234567` is revision of `based_on_tier.yaml` file, `bcd2345` is revision of `base/ns_dev.yaml`)
* `advanced-stage-1234567-cde3456`  
( `1234567` is revision of `based_on_tier.yaml` file, `cde3456` is revision of `base/ns_stage.yaml`)

for `base` it's simpler as it duplicates the same revision - it doesn't have the `based_on_tier.yaml` file
* `base-clusterresources-abc1234-abc1234`  
* `base-dev-bcd2345-bcd2345`   
* `base-stage-cde3456-cde3456`  

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/365


deploy/templates/nstemplatetiers/basedeactivationdisabled/tier-config.yaml
```diff
@@ -0,0 +1,7 @@
+from:
+  name: base
+to:
+  name: basedeactivationdisabled
```
alexeykazakov, line 4: Maybe we can call it `based_on_tier.yaml` instead of `tier-config.yaml`?

alexeykazakov (commented):
Awesome! I like the approach!
And yes, let's remove `team`.

deploy/templates/nstemplatetiers/basedeactivationdisabled/tier-config.yaml
MatousJobanek: done

alexeykazakov (approved):
Nice!

deploy/templates/nstemplatetiers/baseextendedidling/based_on_tier.yaml
```diff
@@ -0,0 +1,8 @@
+from:
```
xcoulon, line 1: so, having such a single `based_on_tier.yaml` file will trigger the generation of `cluster.yaml`, `ns_dev.yaml`, `ns_stage.yaml` and `tier.yaml` for this `baseextendingidling` tier?

xcoulon (commented):
what happens if in the future, a particular tier needs to have a totally different template for one of the namespaces? for example, could we have a custom `deploy/templates/nstemplatetiers/newadvanced/ns_dev.yaml` file in that would not be overridden by the generator when it processes `deploy/templates/nstemplatetiers/newadvanced/based_on_tier.yaml`? (I haven't seen a test that covers this specific usecase)

deploy/templates/nstemplatetiers/baseextendedidling/based_on_tier.yaml
MatousJobanek: not really - it doesn't generate any yaml files. 
This config will ensure creation/update of `NSTemplateTier` with the name `baseextendingidling` and related `TierTemplates` CRs

MatousJobanek:
> what happens if in the future, a particular tier needs to have a totally different template for one of the namespaces? 

for now it's all or nothing. You either define all templates that are needed for the tier or use the `based_on_tier.yaml` config which will reuse everything. The only change that is supported by this config is in the parameters of the templates, nothing else.

deploy/templates/nstemplatetiers/baseextendedidling/based_on_tier.yaml
xcoulon: ok, I see. So there is a risk that the NSTemplateTier generation fails during deployment (eg: missing template, etc.)? I assume this should be caught by our e2e tests?

xcoulon:
> > what happens if in the future, a particular tier needs to have a totally different template for one of the namespaces?
> 
> for now it's all or nothing. You either define all templates that are needed for the tier or use the `based_on_tier.yaml` config which will reuse everything. The only change that is supported by this config is in the parameters of the templates, nothing else.

in that case, should we have a meachnism that raises an error when there's a mix of custom and extension?

deploy/templates/nstemplatetiers/baseextendedidling/based_on_tier.yaml
MatousJobanek: I believe that it should be already caught by unit-tests

MatousJobanek:
> in that case, should we have a mechanism that raises an error when there's a mix of custom and extension?

did you mean something like this? b165b4b 
thanks for the proposal :+1: :wink:

xcoulon:
> > in that case, should we have a mechanism that raises an error when there's a mix of custom and extension?
> 
> did you mean something like this? [b165b4b](https://github.com/codeready-toolchain/host-operator/commit/b165b4bd90d59392e21f4fc43f2ed15df5c5a102)
> thanks for the proposal 👍 😉

yes 👍  thanks for taking my comment into account 🙌

deploy/templates/nstemplatetiers/baseextendedidling/based_on_tier.yaml
xcoulon: ok, we're good then

xcoulon (approved):

deploy/templates/nstemplatetiers/advanced/based_on_tier.yaml
```diff
@@ -0,0 +1,9 @@
+from:
+  name: base
+to:
+  name: advanced
+  parameters:
```
rajivnathan, line 5: Should we put the parameters field at the top level in case we want to expand on the composition in the future?
eg.
```
from:
   name: base
with:
   name: advanced-networkpolicies
to:
   name: advanced
parameters:
  - name: DEACTIVATION_TIMEOUT_DAYS
    value: 0
  - name: IDLER_TIMEOUT_SECONDS
    value: 0
  - name: NETWORK_PARAM
    value: 10
```

rajivnathan (approved):
This is a really great improvement! I really like the implementation too! 🥇

deploy/templates/nstemplatetiers/advanced/based_on_tier.yaml
rajivnathan: Or maybe it makes more sense for the parameters to go with the source template?
eg.
```
from:
   name: base
   parameters:
     - name: DEACTIVATION_TIMEOUT_DAYS
        value: 0
     - name: IDLER_TIMEOUT_SECONDS
        value: 0
with:
   name: advanced-networkpolicies
   parameters:
    - name: NETWORK_PARAM
       value: 10
to:
   name: advanced
```

rajivnathan: Or I'm fine with keeping it as is and dealing with it if/when we actually have a need. 🙂

MatousJobanek: I used this structure because it seemed logical to me to put all new parameters into `to:` section. In other words, it defines what is the new tier and which new parameters/values are used there.
But I'm open to discussion, let's discuss it in our sync call. I'll merge the PRs to unblock the other work on tiers. Change of the structure should be pretty simple.

PR #492, opened 2021-08-05
don't decrement when UserAccount is accidentally deleted

MatousJobanek:
paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/358

### DIFF
--- a/controllers/masteruserrecord/masteruserrecord_controller.go
+++ b/controllers/masteruserrecord/masteruserrecord_controller.go
@@ -87,12 +87,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		}
 		logger.Info("ensuring user accounts")
 		for _, account := range mur.Spec.UserAccounts {
-			requeueTime, err := r.ensureUserAccount(logger, account, mur)
+			err := r.ensureUserAccount(logger, account, mur)
 			if err != nil {
 				logger.Error(err, "unable to synchronize with member UserAccount")
 				return reconcile.Result{}, err
-			} else if requeueTime > 0 {
-				return reconcile.Result{Requeue: true, RequeueAfter: requeueTime}, err // waiting for a few seconds to give time to the member cluster to finish its deletions
 			}
 		}
 		// If the UserAccount is being deleted, delete the UserAccounts in members.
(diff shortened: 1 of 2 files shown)


controllers/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -144,32 +142,23 @@ func (r *Reconciler) ensureUserAccount(logger logr.Logger, murAccount toolchainv
 			// does not exist - should create
 			userAccount = newUserAccount(nsdName, murAccount.Spec, mur.Spec)
 			if err := memberCluster.Client.Create(context.TODO(), userAccount); err != nil {
-				return 0, r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToCreateUserAccountReason), err,
+				return r.wrapErrorWithStatusUpdate(logger, mur, r.setStatusFailed(toolchainv1alpha1.MasterUserRecordUnableToCreateUserAccountReason), err,
 					"failed to create UserAccount in the member cluster '%s'", murAccount.TargetCluster)
 			}
-			counter.IncrementUserAccountCount(logger, murAccount.TargetCluster)
-			return 0, updateStatusConditions(logger, r.Client, mur, toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, ""))
+			if murAccount.SyncIndex != "0" && murAccount.SyncIndex != "deleted" {
```
alexeykazakov, line 148: Should we really check if it's not "0"? It's "0" for the case when it's just created (but not synced yet), right? But we still need to count it, no?

alexeykazakov (commented):
I'm approving this PR but please take a look at my comment.

alexeykazakov (approved):

controllers/masteruserrecord/masteruserrecord_controller.go
MatousJobanek: if the MUR is created with the UserAccount def in its spec, then it contains the default value of the `syncIndex` - don't forget that it's a string, so the default value will be an empty string: `""`
We don't set anything specific to the syncIndex when creating MUR: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/masteruserrecord.go#L52-L62
I added the condition, because right now the useraccount_status controller sets the syncIndex to zero, when the UserAccount is gone. As soon as I modify it to `delete` then I can drop the zero from this condition. But still, it shouldn't have any negative effect at all

alexeykazakov: Ah right. Forgot that it's a string. So the default value is "". Not 0.

PR #493, opened 2021-08-06
Allow reg service to get toolchainconfigs

rajivnathan:
Since the reg service will get configuration directly from the ToolchainConfig resource, the reg service SA needs to be able to get toolchainconfigs


alexeykazakov (approved):
I guess we need to add it here too: https://github.com/codeready-toolchain/registration-service/blob/master/deploy/registration-service.yaml#L34
It used to be used when deployed in dev mode from the reg-service repo. Not sure if it's still used though. If not used then let's delete it.

PR #494, opened 2021-08-11
feat: trigger GH Actions workflow for retest comment

MatousJobanek:
The same change as: https://github.com/codeready-toolchain/member-operator/pull/289

### DIFF
--- a/.github/workflows/publish-operators-for-e2e-tests.yml
+++ b/.github/workflows/publish-operators-for-e2e-tests.yml
@@ -1,6 +1,9 @@
 name: publish-operators-for-e2e-tests
 on:
-  pull_request_target
+  issue_comment:
+    types: [created]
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

alexeykazakov:
Hm... `/retest` doesn't seem to be working?

MatousJobanek:
Yeah, it's most likely because the new config is only in PR, but not in master. Right now, the workflow has configuration only for `pull_request_target` which means that the workflow is taken only from master branch.

In addition, even if it was executed you wouldn't see it in the status checks - I haven't found a way to update the status check to `pending` and then either to `success` or `failed` based on the PR comment. The reason is that the PR (and mainly the check status) is tied to a specific ref, but comment is not.

PR #496, opened 2021-08-12
Use toolchain config for the remaining host configuration

rajivnathan:
The host operator was already using ToolchainConfig for the automatic approval stuff. This PR makes it use the rest of the ToolchainConfig.

Related issue: Related Issue: https://issues.redhat.com/browse/CRT-1122


controllers/toolchainconfig/configuration.go
```diff
@@ -1,15 +1,82 @@
...
+	if !ok {
+		// return default config
+		logger.Error(fmt.Errorf("cache does not contain toolchainconfig resource type"), "failed to get ToolchainConfig from resource, using default configuration")
+		return ToolchainConfig{cfg: &toolchainv1alpha1.ToolchainConfigSpec{}}
+	}
+	return ToolchainConfig{cfg: &toolchaincfg.Spec, secrets: secrets}
+}
+
+func (c *ToolchainConfig) Print() {
+	logger.Info("Toolchain configuration variables", "ToolchainConfigSpec", c.cfg)
```
MatousJobanek, line 75: wouldn't make sense to convert it into a YAML format so it's easily parsable in case of later troubleshooting?

controllers/toolchainconfig/deliveryserviceconfig.go
```diff
@@ -0,0 +1,25 @@
...
+	return d.Notifications().MailgunAPIKey()
+}
+
+func (d DeliveryServiceFactoryConfig) GetMailgunSenderEmail() string {
+	return d.Notifications().MailgunSenderEmail()
+}
+
+func (d DeliveryServiceFactoryConfig) GetMailgunReplyToEmail() string {
+	return d.Notifications().MailgunReplyToEmail()
+}
```
MatousJobanek, line 25: Do we still need this level of abstraction? It looks overcomplicated to me and setting up the tests should be pretty easy now.

In addition, in this way we won't be able to modify any of the values on the fly - we would need to restart the host-operator deployment after doing some change.

MatousJobanek (commented):
Looks good :1st_place_medal:  I tried to go through all changes, but it was pretty hard to pay attention to all lines :-) But I understand that there is no other way to do that :+1:

controllers/toolchainconfig/configuration.go
```diff
@@ -52,28 +142,151 @@ type DeactivationConfig struct {
...
+	return n.notificationSecret(key)
+}
+
+type RegistrationServiceConfig struct {
+	c toolchainv1alpha1.RegistrationServiceConfig
+}
+
+func (r RegistrationServiceConfig) Environment() string {
+	return commonconfig.GetString(r.c.Environment, "prod")
+}
```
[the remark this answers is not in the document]
rajivnathan: I figured the environment is not something that should be changeable without redeploying? It's not used in host operator yet but when I remove the registration service controller and CRD we can choose to get the value from the ToolchainConfig and then set it as an environment variable on the registration-service deployment.

controllers/toolchainconfig/configuration.go
```diff
@@ -52,28 +142,151 @@ type DeactivationConfig struct {
...
+
+func (r RegistrationServiceConfig) Namespace() string {
+	return commonconfig.GetString(r.c.Namespace, "toolchain-host-operator")
+}
```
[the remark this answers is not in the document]
rajivnathan: Sure, sounds good. 👍

deploy/registration-service/registration-service.yaml
```diff
@@ -112,62 +112,8 @@ objects:
...
-                - name: REGISTRATION_VERIFICATION_EXCLUDED_EMAIL_DOMAINS
-                  valueFrom:
-                    configMapKeyRef:
-                      name: registration-service
-                      key: verification.excluded_email_domains
-                - name: REGISTRATION_VERIFICATION_MESSAGE_TEMPLATE
-                  valueFrom:
-                    configMapKeyRef:
-                      name: registration-service
-                      key: verification.message_template
```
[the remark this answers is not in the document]
rajivnathan: Added it as part of https://github.com/codeready-toolchain/registration-service/pull/208

controllers/toolchainconfig/configuration.go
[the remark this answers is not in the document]
rajivnathan: Dropped in 85f68e73eb56012d30167f6e8f676d9653f0548a

controllers/toolchainconfig/configuration.go
rajivnathan: How about if it's JSON? Like this:

```
2021-08-16T23:00:21.402Z	INFO	toolchainconfig	Toolchain configuration variables
{
	"host": {
		"automaticApproval": {
			"enabled": true,
			"resourceCapacityThreshold": {},
			"maxNumberOfUsers": {}
		},
		"deactivation": {
			"deactivationDomainsExcluded": "@redhat.com"
		},
		"metrics": {},
		"notifications": {
			"secret": {}
		},
		"registrationService": {
			"analytics": {},
			"auth": {},
			"environment": "dev",
			"verification": {
				"secret": {
					"ref": "host-operator-secret",
					"twilioAccountSID": "twilio.account.sid",
					"twilioAuthToken": "twilio.auth.token",
					"twilioFromNumber": "twilio.from_number"
				}
			}
		},
		"tiers": {},
		"toolchainStatus": {},
		"users": {}
	},
	"members": {
		"default": {
			"auth": {},
			"autoscaler": {
				"deploy": false
			},
			"che": {
				"secret": {}
			},
			"console": {},
			"memberStatus": {
				"refreshPeriod": "1s"
			},
			"toolchainCluster": {
				"healthCheckPeriod": "5s"
			},
			"webhook": {}
		}
	}
```

controllers/toolchainconfig/deliveryserviceconfig.go
rajivnathan: I agree with you but I think it was done intentionally this way so that the notification service can swap out one mail provider with another without requiring many changes? Maybe we should discuss it in a sync call and do that work separately if we want to remove the abstraction.

rajivnathan:
Unrelated Note: Registration Service CRD cannot be removed until the registration service is deployed by the ToolchainConfig controller https://issues.redhat.com/browse/CRT-1205 is done. Have started working on that.

controllers/toolchainconfig/configuration.go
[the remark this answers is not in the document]
MatousJobanek: as was mentioned in the call - do whatever is easier (and more logical). In other words, feel free to keep it as it is :+1:

controllers/toolchainconfig/configuration.go
MatousJobanek: JSON looks good :+1:

controllers/toolchainconfig/deliveryserviceconfig.go
MatousJobanek: As was discussed - keep it as it is in this PR and let's try to remove it in separate one.

MatousJobanek (approved):
Thanks for addressing my comments. :+1: 
Just a reminder - please keep in mind to check the registration service that it's properly updated when you merge the PR. :pray:

controllers/deactivation/deactivation_controller.go
```diff
@@ -56,9 +54,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	logger := log.FromContext(ctx)
 	logger.Info("Reconciling Deactivation")
 
-	config, err := toolchainconfig.GetConfig(r.Client, request.Namespace)
+	config, err := toolchainconfig.GetToolchainConfig(r.Client)
```
MatousJobanek, line 57: I have one more question - why do you use `GetToolchainConfig` and not `GetCachedToolchainConfig`?

rajivnathan: I created `GetCachedToolchainConfig` for use in the registration service because there isn't always a client available in all the places we wanted to get config values. So the simple reason is I used `GetToolchainConfig` because I started the host operator migration before I did the reg service changes. 🙂 

Since the toolchainconfig controller updates the cache it's probably fine for us to switch to `GetCachedToolchainConfig` but I think it's okay to keep using `GetToolchainConfig` since it can handle the edge case where config is used before the toolchainconfig controller has initialized the cache.

MatousJobanek: As I see it - we should ideally use only the `GetToolchainConfig` function. 
Maybe we could remove the `GetCachedToolchainConfig` function from the code to avoid confusion. If I'm not mistaken, it's used only in tests now.

PR #501, opened 2021-08-20
Temporary fix to set notification.Spec.Recipient when it is blank

sbryzak:
This PR contains a temporary fix that will attempt to lookup the notification's corresponding UserSignup resource, and set the Notification.Spec.Recipient to the users email address if found.

Fixes https://issues.redhat.com/browse/CRT-1220


MatousJobanek (approved):
Would be good to have a unit-test for it, but it looks fine :+1:

PR #506, opened 2021-08-26
Enhance notification templates to support customisable replyTo address

sbryzak:
This PR modifies the notification templates to replace the hard-coded reply-to address with a customisable parameter value.  It also upgrades the mailgun client library to the latest release version.

Fixes https://issues.redhat.com/browse/CRT-1216

### DIFF
--- a/controllers/notification/mailgun_notification_delivery_service_test.go
+++ b/controllers/notification/mailgun_notification_delivery_service_test.go
@@ -1,7 +1,11 @@
 package notification
 
 import (
+	"net/http"
+	"net/url"
 	"testing"
+
+	"gopkg.in/h2non/gock.v1"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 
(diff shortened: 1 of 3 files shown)


controllers/notification/mailgun_notification_delivery_service.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/mailgun/mailgun-go/v4"
 )
 
+const (
+	ContextReplyTo = "ReplyTo"
```
MatousJobanek, line 14: @sbryzak so the resulting spec will look like:
```yaml
spec:
  context:
    ReplyTo: some-reply@dummy.com
```

shouldn't the `ReplyTo` start with lower-case `r`?

MatousJobanek: it's just a detail, of course - I mentioned it because all fields in our CRs start with lower-case, so it would be maybe good to follow that pattern :-).

sbryzak: That would be inconsistent with our other template variable naming.  Take a look through the notification templates [here](https://github.com/codeready-toolchain/host-operator/tree/master/deploy/templates/notificationtemplates) for other examples.

MatousJobanek: I forgot about that - thanks for letting me know. :+1:

PR #507, opened 2021-08-28
Move Registration Service Deployment To ToolchainConfig controller

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1205

Summary of changes:
- Moves the RegistrationService deploy code to the ToolchainConfig controller
- Removes the RegistrationService controller


controllers/toolchainconfig/toolchainconfig_controller.go
```diff
@@ -3,19 +3,27 @@ package toolchainconfig
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/go-logr/logr"
+	v1 "github.com/openshift/api/template/v1"
```
xcoulon, line 10: can we have `templatev1` as an alias? (like we have `corev1`, `metav1`, etc.)

xcoulon (approved):
looks good overall

controllers/toolchainconfig/toolchainconfig_controller.go
rajivnathan: done in 853e82cf304e083fe7a50390404297f8c81a1546

PR #510, opened 2021-09-10
feat: add Role & RoleBindings to grant crtadmins view & exec perms

MatousJobanek:
adds Role & RoleBindings which will grant crtadmins (that in `crtadmin-users-view` group) a view & exec permissions in all users namespaces.

This will still require some additional changes in sandbox-sre repo to create the Group and add specific crtadmin users there.

paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/381

### DIFF
--- a/deploy/templates/nstemplatetiers/base/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/base/ns_dev.yaml
@@ -57,6 +57,49 @@ objects:
   subjects:
   - kind: User
     name: ${USERNAME}
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: exec-pods
+    namespace: ${USERNAME}-dev
+  rules:
+  - apiGroups:
+    - ""
+    resources:
+    - pods/exec
+    verbs:
+    - get
+    - list
+    - watch
+    - create
+    - delete
+    - update
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: crtadmin-view
+    namespace: ${USERNAME}-dev
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: view
+  subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: crtadmin-users-view
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: crtadmin-pods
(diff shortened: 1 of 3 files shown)


deploy/templates/nstemplatetiers/base/ns_dev.yaml
```diff
@@ -57,6 +57,49 @@ objects:
   subjects:
   - kind: User
     name: ${USERNAME}
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
```
xcoulon, line 61: just to be sure: we can't create a ClusterRole on OSD, can we?

xcoulon (approved):

deploy/templates/nstemplatetiers/base/ns_dev.yaml
MatousJobanek: We can. 
Are you asking because we would create one ClusterRole that would be referenced by the RoleBindings? We could technically do that, but the creation of the ClusterRole would be done by sandbox-sre. I would rather keep Role & RoleBinding together as there is no way we could modify the reference to a ClusterRole in an existing RoleBinding - that field is immutable.

deploy/templates/nstemplatetiers/base/ns_stage.yaml
```diff
@@ -57,6 +57,49 @@ objects:
   subjects:
   - kind: User
     name: ${USERNAME}
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: exec-pods
```
alexeykazakov, line 63: What do we need this role for?
Having an view role makes sense. Plus edit role would make sense for some sre accounts too. But how would you use the exec-pod role?

MatousJobanek: My idea is that crtadmin users have the lowest needed permissions to be able to investigate a problem in a user's namespace and/or check if the user is mining bitcoins. To be able to exec into user's pod (and check what user is running there) you have to have permissions to `create` the `pods/exec` resource - and this is exactly what the role is about.

MatousJobanek: Just to be clear - by the "lowest needed permissions" I meant that I don't want to give `edit` permissions to crtadmins - view + exec in pods should be enough.

alexeykazakov: Ok. We can add more permissions later if needed.

alexeykazakov (approved):

PR #514, opened 2021-09-17
Don't create duplicate notifications when UserSignup status update fails.

sbryzak:
This PR prevents duplicate deactivating and deactivated notifications from being created if they have already been previously created in cases where the UserSignup status failed to update.

Fixes https://issues.redhat.com/browse/CRT-1243

### DIFF
--- a/controllers/usersignup/usersignup_controller_test.go
+++ b/controllers/usersignup/usersignup_controller_test.go
@@ -2396,7 +2396,7 @@ func TestUserSignupDeactivatingNotificationCreated(t *testing.T) {
 		ObjectMeta: NewUserSignupObjectMeta("", "edward.jones@redhat.com"),
 		Spec: toolchainv1alpha1.UserSignupSpec{
 			Userid:   "UserID089",
-			Username: "freja.johanssen@redhat.com",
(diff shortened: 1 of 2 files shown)


controllers/usersignup/usersignup_controller_test.go
```diff
@@ -1839,6 +1839,78 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
 		require.Equal(t, "https://registration.crt-placeholder.com", notification.Spec.Context["RegistrationURL"])
 		assert.Equal(t, "userdeactivated", notification.Spec.Template)
 	})
+
+	t.Run("second deactivated notification not created when it already exists", func(t *testing.T) {
```
alexeykazakov, line 1843: What about a similar test for the second deactivating notification case?

alexeykazakov (commented):
Looks good. I just want to make sure we are not missing tests here. See my comment.

controllers/usersignup/usersignup_controller_test.go
sbryzak: Added

PR #522, opened 2021-10-07
Propagated OriginalSub from MUR to UserAccount

sbryzak:
This PR propagates the OriginalSub property value from MasterUserRecord.Spec to UserAccount.Spec.

Fixes https://issues.redhat.com/browse/CRT-1272

### DIFF
--- a/controllers/masteruserrecord/masteruserrecord_controller_test.go
+++ b/controllers/masteruserrecord/masteruserrecord_controller_test.go
@@ -122,6 +122,8 @@ func TestCreateUserAccountSuccessful(t *testing.T) {
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
 	s := apiScheme(t)
 	mur := murtest.NewMasterUserRecord(t, "john")
+	mur.Spec.OriginalSub = "original-sub:12345"
+
 	require.NoError(t, murtest.Modify(mur, murtest.Finalizer("finalizer.toolchain.dev.openshift.com")))
 	memberClient := test.NewFakeClient(t)
 	hostClient := test.NewFakeClient(t, mur)
--- a/controllers/masteruserrecord/sync.go
+++ b/controllers/masteruserrecord/sync.go
@@ -43,7 +43,7 @@ type Synchronizer struct {
 	logger            logr.Logger
 }
 
-// synchronizeSpec synhronizes the useraccount in the MasterUserRecord with the corresponding UserAccount on the member cluster.
+// synchronizeSpec synchronizes the useraccount in the MasterUserRecord with the corresponding UserAccount on the member cluster.
 func (s *Synchronizer) synchronizeSpec() error {
 
 	if !s.isSynchronized() {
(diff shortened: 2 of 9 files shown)


controllers/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -122,6 +122,8 @@ func TestCreateUserAccountSuccessful(t *testing.T) {
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
 	s := apiScheme(t)
 	mur := murtest.NewMasterUserRecord(t, "john")
+	mur.Spec.OriginalSub = "original-sub:12345"
```
alexeykazakov, line 125: Shouldn't we also add an assertion for the UserAcount bellow to make sure it does has the OriginalSub set?

controllers/masteruserrecord/sync.go
```diff
@@ -54,6 +54,7 @@ func (s *Synchronizer) synchronizeSpec() error {
 		s.memberUserAcc.Spec.UserAccountSpecBase = s.recordSpecUserAcc.Spec.UserAccountSpecBase
 		s.memberUserAcc.Spec.Disabled = s.record.Spec.Disabled
 		s.memberUserAcc.Spec.UserID = s.record.Spec.UserID
+		s.memberUserAcc.Spec.OriginalSub = s.record.Spec.OriginalSub
```
alexeykazakov, line 57: Any tests for that?

pkg/templates/registrationservice/template_test.go
```diff
@@ -35,13 +37,15 @@ func TestDeploymentAssetContainsAllNecessaryInformation(t *testing.T) {
 	roleBindingFound := false
 	for _, toolchainObject := range toolchainObjects {
 		assert.Equal(t, "my-namespace", toolchainObject.GetNamespace())
-		fmt.Println(toolchainObject.GetGvk())
+		gvk, err := apiutil.GVKForObject(toolchainObject, s)
+		require.NoError(t, err)
+		fmt.Println(gvk)
 		fmt.Println(appsv1.SchemeGroupVersion.WithKind("Deployment"))
```
alexeykazakov, line 43: Is there any reason for keeping these Prtintln? I would remove them.

alexeykazakov (commented):
Looks good overall but it looks like it's missing some unit tests.

controllers/masteruserrecord/masteruserrecord_controller_test.go
sbryzak: I added this assertion in toolchain-common, see https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/useraccount/assertion.go#L69

alexeykazakov: Ah right..

controllers/masteruserrecord/sync.go
alexeykazakov: I assume it's covered by https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/useraccount/assertion.go#L69 too?

sbryzak: Like above, it's tested in `TestSynchronizeSpec` which also uses `MatchMasterUserRecord`.

PR #524, opened 2021-10-14
Reset deactivating state when promoting users

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1222

Related PRs:
https://github.com/codeready-toolchain/toolchain-common/pull/214
https://github.com/codeready-toolchain/toolchain-e2e/pull/393

Updates the ChangeTierRequest controller to look up the related UserSignup and reset its deactivating state

### DIFF
--- a/controllers/changetierrequest/changetierrequest_controller_test.go
+++ b/controllers/changetierrequest/changetierrequest_controller_test.go
@@ -14,6 +14,7 @@ import (
 	. "github.com/codeready-toolchain/host-operator/test"
 
 	commonconfig "github.com/codeready-toolchain/toolchain-common/pkg/configuration"
+	"github.com/codeready-toolchain/toolchain-common/pkg/states"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
 	testconfig "github.com/codeready-toolchain/toolchain-common/pkg/test/config"
 	murtest "github.com/codeready-toolchain/toolchain-common/pkg/test/masteruserrecord"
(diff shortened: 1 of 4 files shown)


controllers/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -224,6 +227,22 @@ func TestChangeTierFailure(t *testing.T) {
 			AllUserAccountsHaveTier(murtest.DefaultNSTemplateTier)
 		AssertThatChangeTierRequestHasCondition(t, cl, changeTierRequest.Name, toBeComplete(), toBeDeletionError("unable to delete ChangeTierRequest object 'request-name': error"))
 	})
+
+	t.Run("the change will fail because there is no corresponding UserSignup found", func(t *testing.T) {
```
alexeykazakov, line 231: This test actually asserts that the reconcile fails if there is no owner label in the MUR, right? We need this test, so it's good. But we also need a test for the case when there is no UserSignup found. Or I'm missing something?

controllers/changetierrequest/changetierrequest_controller_test.go
```diff
@@ -224,6 +227,22 @@ func TestChangeTierFailure(t *testing.T) {
...
+
+		// when
+		_, err := controller.Reconcile(context.TODO(), request)
+
+		// then
+		require.Error(t, err)
+		assert.Equal(t, err.Error(), `failed to get corresponding UserSignup for MasterUserRecord with name 'john': MasterUserRecord is missing label 'toolchain.dev.openshift.com/owner'`)
+		AssertThatChangeTierRequestHasCondition(t, cl, changeTierRequest.Name, toBeNotComplete(`MasterUserRecord is missing label 'toolchain.dev.openshift.com/owner'`))
+	})
 }
```
alexeykazakov, line 246: And what about the test which verifies that the deactivating state was dropped for the UserSignup?

rajivnathan: I'm not sure how I missed that. 🤦 Added in 5fda34e9a2ec5ebf433621ea20aaf7dd8ef0699c

PR #530, opened 2021-10-19
update notification templates to include slack channel invite

ranakan19:
fix wrt: https://issues.redhat.com/browse/CRT-1267

Agreed to have a common slack channel for sandbox and RHODS, thus only updating the template instead of adding it to config.

### DIFF
--- a/deploy/templates/nstemplatetiers/hackathon/based_on_tier.yaml
+++ b/deploy/templates/nstemplatetiers/hackathon/based_on_tier.yaml
@@ -1,4 +1,4 @@
 from: base
 parameters:
 - name: DEACTIVATION_TIMEOUT_DAYS
-  value: 50
+  value: 80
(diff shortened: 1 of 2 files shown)


deploy/templates/notificationtemplates/userdeactivated/notification.html
```diff
@@ -50,6 +50,11 @@
         You can request new access by signing up again at {{.RegistrationURL}}
     </p>
 
+    <p>
+        Share your feedback, request an extension for your Sandbox environment from the #dev-sandbox channel on DevNation Slack.
+        You can join using the following <a href="https://dn.dev/DevNationSlack">invitation</a>.
```
alexeykazakov, line 55: We previously had issues with rendering links in our emails. We need to test if it works now.

PR #534, opened 2021-11-04
Appstudio tier

alexeykazakov:
This PR introduces a new Appstudio tier required by the proxy. It's still user based tier for now.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/405

### DIFF
--- a/controllers/toolchainconfig/toolchainconfig_controller_test.go
+++ b/controllers/toolchainconfig/toolchainconfig_controller_test.go
@@ -296,7 +296,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncFailure(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
 				HasSyncErrors(map[string]string{"missing-member": "specific member configuration exists but no matching toolchaincluster was found"})
 		})
 	})
--- a/controllers/toolchainconfig/toolchainconfig_controller_test.go
+++ b/controllers/toolchainconfig/toolchainconfig_controller_test.go
@@ -92,7 +92,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncComplete(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
 				HasNoSyncErrors()
 
 			// check member1 config
--- a/controllers/toolchainconfig/toolchainconfig_controller_test.go
+++ b/controllers/toolchainconfig/toolchainconfig_controller_test.go
@@ -92,7 +92,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncComplete(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
 				HasNoSyncErrors()
 
 			// check member1 config
@@ -296,7 +296,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncFailure(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
 				HasSyncErrors(map[string]string{"missing-member": "specific member configuration exists but no matching toolchaincluster was found"})
 		})
 	})
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -18,39 +18,38 @@ objects:
       name: registration-service
       namespace: ${NAMESPACE}
     rules:
-    - apiGroups:
-      - toolchain.dev.openshift.com
-      resources:
-      - usersignups
-      verbs:
-      - create
-      - get
-      - update
-      - list
-    - apiGroups:
-      - toolchain.dev.openshift.com
-      resources:
-      - masteruserrecords
-      - toolchainconfigs
-      - toolchainstatuses
-      verbs:
-      - get
-    - apiGroups:
-      - toolchain.dev.openshift.com
-      resources:
-      - bannedusers
-      - toolchainclusters
-      verbs:
-      - get
-      - list
-    - apiGroups:
-      - ""
-      resources:
-      - secrets
-      - configmaps
-      verbs:
-      - get
-      - list
+      - apiGroups:
+          - toolchain.dev.openshift.com
+        resources:
+          - usersignups
+        verbs:
+          - create
+          - get
+          - update
+          - list
+      - apiGroups:
+          - toolchain.dev.openshift.com
+        resources:
+          - masteruserrecords
+          - toolchainconfigs
+          - toolchainstatuses
+        verbs:
+          - get
+      - apiGroups:
+          - toolchain.dev.openshift.com
+        resources:
+          - bannedusers
+        verbs:
+          - get
+          - list
+      - apiGroups:
+          - ""
+        resources:
+          - secrets
+          - configmaps
+        verbs:
+          - get
+          - list
   - kind: RoleBinding
     apiVersion: rbac.authorization.k8s.io/v1
     metadata:
@@ -89,7 +88,6 @@ objects:
               image: ${IMAGE}
               ports:
                 - containerPort: 8080
-                - containerPort: 8081
               command:
                 - registration-service
               imagePullPolicy: IfNotPresent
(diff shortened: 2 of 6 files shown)


deploy/registration-service/registration-service.yaml
```diff
@@ -155,6 +157,43 @@ objects:
...
+      host: ''
+      port:
+        targetPort: "8081"
+      to:
+        kind: Service
+        name: api
+        weight: 100
+      tls:
+        termination: edge
+      wildcardPolicy: None
```
MatousJobanek, line 196: I guess that this (and some other changes) is not needed for the tier

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
```diff
@@ -0,0 +1,214 @@
...
+  metadata:
+    name: user-view
+    namespace: ${USERNAME}
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: view
+  subjects:
+  - kind: User
+    name: ${USERNAME}
```
MatousJobanek, line 27: I thought that users should have no access to the namespace at all

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -572,18 +595,11 @@ func assertNamespaceTemplate(t *testing.T, decoder runtime.Decoder, actual templ
 		default:
 			t.Errorf("unexpected kind: '%s'", kind)
 		}
+	case "appstudio":
+		// Nothing
```
MatousJobanek, line 599: all the networkpolicies but `allow-from-other-user-namespaces` are still there - why shouldn't we check them as well?

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -543,8 +553,21 @@ func assertNamespaceTemplate(t *testing.T, decoder runtime.Decoder, actual templ
 	// Namespace
 	containsObj(t, actual, namespaceObj(kind))
 
-	// RoleBinding "user-edit"
-	containsObj(t, actual, userEditRoleBindingObj(kind))
+	// RBAC
+	switch tier {
+	case "appstudio":
+
```
MatousJobanek, line 559: is this intentional?

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -534,6 +542,8 @@ func assertNamespaceTemplate(t *testing.T, decoder runtime.Decoder, actual templ
 		} else {
 			require.Len(t, actual.Objects, 14)
 		}
+	case "appstudio":
+		require.Len(t, actual.Objects, 16)
 	case "test":
 		return // Don't care what objects are defined in the test tier
```
MatousJobanek, line 548: I also noticed that you don't verify the ServiceAccount anywhere, nor the related RoleBindings.

So, if you don't want to verify any of the objects from AppStudio, you could just return for the `appstudio` case here as well and skip the tier.

MatousJobanek (commented):
There are some redundant changes (probably related to the Proxy work) and the unit test needs more love.

controllers/toolchainconfig/toolchainconfig_controller_test.go
```diff
@@ -92,7 +92,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncComplete(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
```
alexeykazakov, line 95:
```suggestion
					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
```

controllers/toolchainconfig/toolchainconfig_controller_test.go
```diff
@@ -296,7 +296,7 @@ func TestReconcile(t *testing.T) {
 				Exists().
 				HasConditions(
 					toolchainconfig.ToSyncFailure(),
-					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
+					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service Service: api Route: api]")).
```
alexeykazakov, line 299:
```suggestion
					toolchainconfig.ToRegServiceDeploying("updated resources: [ServiceAccount: registration-service Role: registration-service RoleBinding: registration-service Deployment: registration-service Service: registration-service Route: registration-service]")).
```

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
alexeykazakov: Let's keep it here for now. It can be useful during development. We will need to adjust this tier anyway.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -460,130 +505,29 @@ func TestNewTierTemplate(t *testing.T) {
...
 func assertClusterResourcesTemplate(t *testing.T, decoder runtime.Decoder, actual templatev1.Template, tier string) {
-	switch tier {
-	case "base", "test":
+	if !basedOnOtherTier(tier) {
 		expected := templatev1.Template{}
 		content, err := Asset(fmt.Sprintf("%s/cluster.yaml", tier))
 		require.NoError(t, err)
 		_, _, err = decoder.Decode(content, nil, &expected)
 		require.NoError(t, err)
 		assert.Equal(t, expected, actual)
```
MatousJobanek, line 514: How about a small sanity check that the template contains some objects?
```suggestion
		assert.Equal(t, expected, actual)
		assert.NotEmpty(t, actual.Objects)
```

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -460,130 +505,29 @@ func TestNewTierTemplate(t *testing.T) {
...
 		templatePath = expectedTemplateFromBasedOnTierConfig(t, tier, fmt.Sprintf("ns_%s.yaml", kind))
+	} else {
+		templatePath = fmt.Sprintf("%s/ns_%s.yaml", tier, kind)
 	}
 	content, err := Asset(templatePath)
 	require.NoError(t, err)
 	expected := templatev1.Template{}
 	_, _, err = decoder.Decode(content, nil, &expected)
 	require.NoError(t, err)
 	assert.Equal(t, expected, actual)
```
MatousJobanek, line 530: the same here
```suggestion
	assert.Equal(t, expected, actual)
	assert.NotEmpty(t, actual.Objects)
```

MatousJobanek (approved):
Looks good :+1: I really like the simplification. 
Could you please change the names of the functions so they don't use the `kind` word?

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
```diff
@@ -0,0 +1,214 @@
...
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${USERNAME}
+      openshift.io/display-name: ${USERNAME}
+      openshift.io/requester: ${USERNAME}
+    labels:
+      name: ${USERNAME}
+    name: ${USERNAME}
```
MatousJobanek, line 15: BTW, what do you think about my comment: https://coreos.slack.com/archives/CHK0J6HT6/p1636028680487800?thread_ts=1635784649.338700&cid=CHK0J6HT6

just  in short what I have written there:

> member cluster will have not only the system (openshift-*, kube-*) namespaces whose names shouldn't be used by the workspace  - for example, related to SRE as well as those where operators will be installed in. The set of names that should be excluded can be large and mainly, we won't be able to define all of them from the very beginning.
That being said - having a suffix would be much simpler & safer.

alexeykazakov: so you saying we will need to add generate namespace names anyway? But we will need to keep the list of excluded names to generate a name anyway. Like we do now when generating the user namespaces.
Anyway, we don't have to decide it now. We can change the naming strategy later.

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -27,6 +27,67 @@ import (
...
+func namespaceKind(kind string) bool {
+	for _, k := range allNamespaceKinds() {
+		if kind == k {
+			return true
+		}
+	}
+	return false
+}
+
+func allNamespaceKinds() []string {
```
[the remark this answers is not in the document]
alexeykazakov: I just kept the same terms we already using in the tests. But I agree that kinds is confusing. Not sure why we use "kinds" here and not "types".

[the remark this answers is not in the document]
alexeykazakov: Done in https://github.com/codeready-toolchain/host-operator/pull/534/commits/eb3a120143be064b9b8519deab01300152bcdac4

xcoulon (approved):

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
MatousJobanek: No, I didn't mean that we would generate the namespace names - I was talking about adding the suffix. If we added the suffix to the namespace names, then I guess that we shouldn't hit a conflict with the already existing namespaces.
So, I would vote for adding one of those suffixes that were proposed in the slack channel:
* `-aps`
* `-app` (for me the most favorite one)
* `-poc`

alexeykazakov: My point is that we will need to detect invalid namespace names anyway. For example we can't start namespaces with openshift- or kube-
And namespace suffixes won't help with that.
So we either should catch it on the workspace creation stage or always transform the workspace name to a namespace name, so it's guaranteed that the namespace name is valid and unique but it means there is always can be cases when workspace name != namespace name. This is what we do with UserSignups and MUR name generation right now.

MatousJobanek: Catching the prefix is as simple as using the regex that is used in one of the admission webhooks that prevents from creating the projects with these prefixes.

> So we either should catch it on the workspace creation stage 

which is not possible for all namespace names - as I have written above 

>  always transform the workspace name to a namespace name, so it's guaranteed that the namespace name is valid and unique 

That would be harder to achieve because you would need to list all namespaces of the member clusters to check if there is no conflict with the already existing names.
This would be much simpler with a defined suffix (or prefix). 

> This is what we do with UserSignups and MUR name generation right now.

I know what you mean, but that's to achieve the uniqueness of the MURs/users in the whole platform, and not to avoid conflicts with the names of the SRE and other namespaces. 
In addition, all used MUR names are present in the host cluster, but that's not true for namespaces - they are present in member clusters.
The uniqueness of the workspace names will be achieved easily by listing the Space CRs. If we automatically added a defined suffix (or prefix) to every namespace, then we will avoid conflicts with SRE namespaces. We just need to follow the rule that when we create some SRE/operator namespace in the member cluster, then we won't use the prefix - which is not hard.

My main point is that we've used the defined suffixes of the users' namespaces in Dev Sandbox for some time already and we haven't faced any problem with it yet. It's also predictable, so UI/proxy could generate the namespace name from the workspace name easily.

PR #536, opened 2021-11-09
Added ToolchainEvent CRD to API

sbryzak:
Regenerated due to api changes.

https://issues.redhat.com/browse/CRT-1299

### DIFF
--- a/config/crd/bases/toolchain.dev.openshift.com_toolchainevents.yaml
+++ b/config/crd/bases/toolchain.dev.openshift.com_toolchainevents.yaml
@@ -1,0 +1,139 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.1
+  creationTimestamp: null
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

MatousJobanek:
@sbryzak This PR is missing a few important changes.
you need to
1. modify the kustomize config files (`config/crd/kustomization.yaml`)
2. define this new resource in PROJECT file
3. add the CRD to the kustomized CSV file `config/manifests/bases/host-operator.clusterserviceversion.yaml`

then, when you create a controller, don't forget to add kubebuilder annotations so the deployment has permissions to work with the CRD.

alexeykazakov:
Right. I forgot about those places too. @sbryzak let's update the API readme with instructions how to add a new API/CRD. And even maybe add some warning to the make generate target??? Something like: `!!!!! MAKE SURE YOU UPDATED x,y,z IF ADDING NEW CRD/API !!!!` kinda warning.

MatousJobanek:
Yeah, the host/member repos need a small fix to at least partially support generation via operator-sdk. Will work on it next week.

PR #541, opened 2021-11-19
Adds serviceaccounts so that proxy can list them when mapping users

rajivnathan:
Required so that the serviceaccount lookup will work here: https://github.com/codeready-toolchain/registration-service/blob/d9a98e513a7bb2170d8bbec14d30c2bc545081bb/pkg/proxy/service/cluster_service.go#L62-L67

I will add it to https://github.com/codeready-toolchain/registration-service/blob/d9a98e513a7bb2170d8bbec14d30c2bc545081bb/deploy/registration-service.yaml as well

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -48,6 +48,7 @@ objects:
         resources:
           - secrets
           - configmaps
+          - serviceaccounts
         verbs:
           - get
           - list


deploy/registration-service/registration-service.yaml
```diff
@@ -48,6 +48,7 @@ objects:
         resources:
           - secrets
           - configmaps
+          - serviceaccounts
```
alexeykazakov, line 51: What do you need it for?
We don't use the reg-service SA to get service accounts.
The API call you are referring to (https://github.com/codeready-toolchain/registration-service/blob/d9a98e513a7bb2170d8bbec14d30c2bc545081bb/pkg/proxy/service/cluster_service.go#L62-L67) doesn't use the reg-service SA. It uses the member client created by ToolchainCluster Service.
See https://github.com/codeready-toolchain/registration-service/blob/d9a98e513a7bb2170d8bbec14d30c2bc545081bb/pkg/proxy/proxy.go#L50
And https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/cluster/service.go#L90

rajivnathan: You're right. I originally saw an error related to serviceaccounts when running the test and made this change. But the actual change I needed was the namespace fix in https://github.com/codeready-toolchain/registration-service/pull/228 so this change isn't needed.

PR #539, opened 2021-11-18
refactor: add MasterUserRecord.Spec.TierName field

xcoulon:
migrate existing MasterUserRecords by setting their `spec.TierName`
with `spec.UserAccounts[0].Spec.NSTemplateSet.TierName`
and set the field in new MasterUserRecords as well

Fixes https://issues.redhat.com/browse/CRT-1321

See also:
https://github.com/codeready-toolchain/api/pull/266
https://github.com/codeready-toolchain/member-operator/pull/313
https://github.com/codeready-toolchain/toolchain-e2e/pull/411

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

PR #543, opened 2021-11-23
refactor: add MasterUserRecord.Spec.TierName field (only)

xcoulon:
Just a module update with the new field in `MasterUserRecord.Spec`,
but no change in the business logic.
Applying the same module update on both host-operator and member-operator
so that when the business logic is applied, both operators can use the
new field.

This is a workaround the limitations of e2e test pairing.
See [discussion](https://github.com/codeready-toolchain/host-operator/pull/539#issuecomment-975630203)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/codeready-toolchain/host-operator
 
 require (
 	cloud.google.com/go v0.60.0 // indirect
-	github.com/codeready-toolchain/api v0.0.0-20211116140337-1aaf7ef57cc2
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -31,4 +31,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.8.3
 )
 
+replace github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20211123080725-72e8d12f0de1
```
MatousJobanek, line 34: reminder

MatousJobanek (approved):

go.mod
xcoulon: done

PR #548, opened 2021-11-29
feat Space Controller

xcoulon:
Add SpaceController with capability to watch `NSTemplateSet` resources
on member clusters

See also: https://github.com/codeready-toolchain/toolchain-e2e/pull/415

Fixes https://issues.redhat.com/browse/CRT-1310


controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+// deleteNSTemplateSet triggers the deletion of the NSTemplateSet on the target member cluster.
+// Returns `true/nil` if the NSTemplateSet was deleted during this call, or if it is already being deleted
+// (ie, the deletion was triggered during a previous call/reconcile loop, and it's not fully done yet)
+// Returns `false/nil` if the NSTemplateSet was already deleted (ie, resource doesn't exist anymore),
+//   or if there is no target cluster specified in the given space, of if the target cluster is unknown.
+// Returns `false/error` if an error occurred
+func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		logger.Info("cannot deleted NSTemplateSet: no target cluster specified")
+		return false, nil // abort deletion
```
MatousJobanek, line 223: We should clearly define what happens in the cases when the `TargetCluster` is removed from the already provisioned user. My guess would be that it would remove the NSTemplateSet from the member cluster where it was provisioned. In that case, we would need to add a `targetCluster` to the `Space.status` section.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		return reconcile.Result{}, errs.Wrap(err, "unable to get the current Space")
+	}
+	if !util.IsBeingDeleted(space) {
+		// Add the finalizer if it is not present
+		if err := r.addFinalizer(logger, space); err != nil {
+			return reconcile.Result{}, err
+		}
+	} else {
+		requeue, err := r.ensureSpaceDeletion(logger, space)
+		return reconcile.Result{Requeue: requeue}, err
```
MatousJobanek, line 83: we don't need to force to requeue immediately - if there was something deleted, then we should just wait for the event when the NSTemplateSet is finally gone.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	// ensure that there's a NSTemplateSet on the Target Cluster
+	// will trigger a requeue until the NSTemplateSet exists and is ready,
+	// so the Space can be in `ready` status as well
+	requeue, err := r.ensureNSTemplateSet(logger, space)
+	return ctrl.Result{Requeue: requeue}, err
```
MatousJobanek, line 89: it's the same as for deletion - if there was something created/updated in the member cluster, then we should just wait until the next reconcile is triggered by k8s.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	return nil
+}
+
+// ensureNSTemplateSet creates the NSTemplateSet on the target member cluster if it does not exist,
+// and updates the space's status accordingly.
+// returns `true` after creating the NSTemplateSet and until it is `ready`
+func (r *Reconciler) ensureNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		return false, fmt.Errorf("missing target member cluster for Space '%s'", space.Name)
+	}
```
MatousJobanek, line 111: This is a valid state - there is gonna be a completion controller that will set the `targetCluster` based on the capacity of the member clusters: https://issues.redhat.com/browse/CRT-1333
so instead of returning an error, we should just set the state that it's not ready because there is no targetCluster set (for example because of the capacity issues)

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+				TemplateRef: tmplTier.Spec.ClusterResources.TemplateRef,
+			},
+			Namespaces: make([]toolchainv1alpha1.NSTemplateSetNamespace, len(tmplTier.Spec.Namespaces)),
+		},
+	}
+	for i, tmpl := range tmplTier.Spec.Namespaces {
+		nsTmplSet.Spec.Namespaces[i] = toolchainv1alpha1.NSTemplateSetNamespace{
+			TemplateRef: tmpl.TemplateRef,
+		}
+	}
```
MatousJobanek, line 187: Could you please use the already existing function https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/masteruserrecord.go#L109-L127
We should be sure that it will create the very same content of the NSTempalteSet  CRs

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+
+	readyCond, found := condition.FindConditionByType(nsTmplSet.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	// if UserAccount ready condition is already set to false, then don't update if no message is provided
+	// also, don't update when NSTemplateSet ready condition is in an invalid state
+	if !found || readyCond.Status == corev1.ConditionUnknown || readyCond.Status == corev1.ConditionFalse {
+		logger.Info("NSTemplateSet is not ready", "ready-condition", readyCond)
+		if err := r.setStatusFromNSTemplateSet(space, readyCond.Reason, readyCond.Message); err != nil {
+			return false, err
+		}
+		return true, nil
```
MatousJobanek, line 152: > // if UserAccount ready

It's apparent that it was partially copied from UserAccount controller, but not entirely. 
Anyway, if the condition is either not found or unknown, then we should not replicate it to Space CR.
In UserAccount controller, we don't replicate the status if the UserAccount ready condition is already set to false to just not overload the synchronization mechanism. We could technically do the same also here, but it's not super needed https://github.com/codeready-toolchain/member-operator/blob/master/controllers/useraccount/useraccount_controller.go#L404-L416

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		if !errors.IsNotFound(err) {
+			return false, err // something wrong happened
+		}
+		logger.Info("the NSTemplateSet resource is already deleted")
+		return false, nil // NSTemplateSet was already deleted
+	}
+	if util.IsBeingDeleted(nstmplSet) {
+		logger.Info("the NSTemplateSet resource is already being deleted")
+		return true, nil // requeue until fully deleted
+	}
```
MatousJobanek, line 247: we should also check for how long the NStempalteSet is being deleted - to catch the cases when the deletion is blocked

test/nstemplateset/nstemplateset_assertion.go
```diff
@@ -0,0 +1,237 @@
...
+
+// TODO: remove in favor of github.com/codeready-toolchain/toolchain-common/test/nstemplateset
+// (because there's already a `go mod replace` directive for github.com/codeready-toolchain/toolchain-common)
+
+type Assertion struct {
+	nsTmplSet      *toolchainv1alpha1.NSTemplateSet
+	client         client.Client
+	namespacedName types.NamespacedName
+	t              test.T
+}
```
MatousJobanek, line 28: I think that this whole assertion file is already in toolchain-common repo, isn't it?

MatousJobanek (commented):
Thanks for the PR and for improving the usage of the member Clusters :+1: 
To be honest, I haven't reviewed it completely yet - for example I haven't taken a look at the tests at all. I'll do that tomorrow, I hope that I will feel better.
Anyway, I noticed that there were quite a lot of Codecov warnings - mainly for the error cases. Could you please take a look at them?

controllers/space/space_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	// no need to update the status of the Space once the finalizer has been removed, since
+	// the resource will be deleted
+	return false, nil // no need to requeue
+}
+
+// deleteNSTemplateSet triggers the deletion of the NSTemplateSet on the target member cluster.
+// Returns `true/nil` if the NSTemplateSet was deleted during this call, or if it is already being deleted
+// (ie, the deletion was triggered during a previous call/reconcile loop, and it's not fully done yet)
+// Returns `false/nil` if the NSTemplateSet was already deleted (ie, resource doesn't exist anymore),
+//   or if there is no target cluster specified in the given space, of if the target cluster is unknown.
```
rajivnathan, line 220:
```suggestion
//   or if there is no target cluster specified in the given space, or if the target cluster is unknown.
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,345 @@
...
+// Returns `false/error` if an error occurred
+func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		logger.Info("cannot deleted NSTemplateSet: no target cluster specified")
```
rajivnathan, line 224:
```suggestion
		logger.Info("cannot delete NSTemplateSet: no target cluster specified")
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// deleteNSTemplateSet triggers the deletion of the NSTemplateSet on the target member cluster.
+// Returns `true/nil` if the NSTemplateSet was deleted during this call, or if it is already being deleted
+// (ie, the deletion was triggered during a previous call/reconcile loop, and it's not fully done yet)
+// Returns `false/nil` if the NSTemplateSet was already deleted (ie, resource doesn't exist anymore),
+//   or if there is no target cluster specified in the given space, of if the target cluster is unknown.
+// Returns `false/error` if an error occurred
+func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
```
rajivnathan, line 223: Isn't the targetcluster field optional? If it's empty does it mean the NSTemplateSet could still be in one of the member clusters?

controllers/space/space_controller.go
```diff
@@ -0,0 +1,345 @@
...
+		logger.Info("cannot deleted NSTemplateSet: no target cluster specified")
+		return false, nil // skip NSTemplateSet deletion
+	}
+	memberCluster, found := r.MemberClusters[space.Spec.TargetCluster]
+	if !found {
+		logger.WithValues("target_cluster", space.Spec.TargetCluster).Info("Cannot deleted NSTemplateSet: unknown target member cluster")
```
rajivnathan, line 229:
```suggestion
		logger.WithValues("target_cluster", space.Spec.TargetCluster).Info("Cannot delete NSTemplateSet: unknown target member cluster")
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,345 @@
...
+		logger.Error(err, "failed to remove finalizer")
+		return false, r.setStatusTerminatingFailed(space, "failed to remove finalizer")
+	}
+	logger.Info("removed finalizer")
+	// no need to update the status of the Space once the finalizer has been removed, since
+	// the resource will be deleted
+	return false, nil // no need to requeue
+}
+
+// deleteNSTemplateSet triggers the deletion of the NSTemplateSet on the target member cluster.
```
rajivnathan, line 216: May I request to add a clarification comment about what the bool return value represents? Looking at the code it's not clear to me whether it is indicating whether the NSTemplateSet is deleted or if it's supposed to trigger a requeue. To me the function looks to treat it as a requeue but where the function is called in the `ensureSpaceDeletion` func the variable is named `deleted` so I'm confused

rajivnathan (commented):
Nice, we have launched into Space! 🚀  😂 

Have a question regarding the NSTemplateSet deletion and left a few minor comments as well.

main.go
```diff
@@ -309,12 +316,12 @@ func main() {
 	}
 }
 
-func addMemberClusters(mgr ctrl.Manager, cl client.Client, namespace string) ([]runtimecluster.Cluster, error) {
-	memberConfigs, err := cluster.ListToolchainClusterConfigs(cl, namespace, cluster.Member, memberClientTimeout)
+func addMemberClusters(mgr ctrl.Manager, cl client.Client, namespace string) (map[string]cluster.Cluster, error) {
```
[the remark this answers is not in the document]
xcoulon: yes, this new `cluster.Cluster` struct gives access to the client, cache and operator namespace (which is not exposed in the `runtimecluster.Cluster`)

test/nstemplateset/nstemplateset_assertion.go
xcoulon: yes, I used it here while the PR was still in review (and I was already using another version of toolchain-common)

main.go
```diff
@@ -261,6 +263,11 @@ func main() {
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "UserSignupCleanup")
 	}
+
+	if err = space.SetupWithManager(mgr, memberClusters); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Space")
+		os.Exit(1)
```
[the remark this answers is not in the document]
xcoulon: it's probably a left-over from the generated code

main.go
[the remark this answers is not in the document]
MatousJobanek: Wait a sec - you can use `runtimecluster.Cluster` to get the client as well as the cache, by calling:
```go
memberCluster.GetClient(),
memberCluster.GetCache(),
```
The only information that is not exposed is the namespace so we could change the `cluster.Cluster` abstraction from
```go
type Cluster struct {
	OperatorNamespace string
	Client            client.Client
	Cache             cache.Cache
}
```
to:
```go
type Cluster struct {
	runtimecluster.Cluster
	OperatorNamespace string
}
```
or am I missing something?

controllers/space/space_controller.go
MatousJobanek: no really. The `targetCluster` is an optional field so the UI (or who will create Spaces for AppStudio workspaces) doesn't have to care about capacity of member clusters. There is a plan to create a `SpaceCompletion` controller that will check the capacity of member clusters and will assign the `targetCluster` to those `Spaces` that don't have it set yet: https://issues.redhat.com/browse/CRT-1333
This means that the NSTemplateSet won't be created until the `targetCluster` is set.

In DevSandbox case, the `targetCluster` will be always set by `UserSignup` controller, so we don't have to deal with it right now. This is only for the case of AppStudio so UI can just simply create `Space` without specifying `targetCluster` and let the controllers deal with it.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+type Reconciler struct {
+	Client         client.Client
+	MemberClusters map[string]cluster.Cluster
+}
+
+// SetupWithManager sets up the controller reconciler with the Manager and the given member clusters.
+// Watches the Space resources in the current (host) cluster as its primary resources.
+// Watches NSTemplateSets on the member clusters as its secondary resources.
+func SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
+	builder := ctrl.NewControllerManagedBy(mgr).
```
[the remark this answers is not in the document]
xcoulon: renamed to `b` in b558b85

controllers/space/space_controller.go
xcoulon: good point, let me add this field in the API and use it here

test/nstemplateset/nstemplateset_assertion.go
xcoulon: changed in 301bc73

controllers/space/space_controller.go
xcoulon: I believe this was addressed in 1285823 (even though I missed your comment here)

controllers/space/space_controller.go
MatousJobanek: this can be done in separate PR ;-)

controllers/space/space_controller.go
xcoulon: I removed call to update status of Space resource in faedeeb since it's not necessary

controllers/space/space_controller.go
xcoulon: Tracking actual target cluster on which the NSTemplate is provisioned in faedeeb (even in case of failure)

MatousJobanek: by saying 

>We should clearly define what happens in the cases when the TargetCluster is removed from the already provisioned user. 

I meant that we should rather create a Jira story and spend some time with it the next sprint.
For example - removing or changing the target cluster could trigger move to another member cluster - the "retarget" operation.

xcoulon: well, that wasn't clear in your initial comment :/

xcoulon: but ok, I will add a subtask in CRT-1308 for this particular case (which is "retargeting to another cluster", anyways)

xcoulon: https://issues.redhat.com/browse/CRT-1392

controllers/space/space_controller.go
xcoulon: done in d34a855

controllers/space/space_controller.go
xcoulon: done in d34a855, too

controllers/space/space_controller.go
xcoulon: I suggest we address this in https://issues.redhat.com/browse/CRT-1393

controllers/space/space_controller.go
xcoulon: sure! done in 36a81e1

controllers/space/space_controller.go
MatousJobanek: ok, but the `Space.status.targetCluster` needs an update in api repo, right?

xcoulon: yes, I forgot to open the PR while dealing with the changes here: https://github.com/codeready-toolchain/api/pull/272

config/samples/toolchain.dev.openshift.com_v1alpha1_space.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: Space
+metadata:
+  name: space-sample
+spec:
+  tierName: base
```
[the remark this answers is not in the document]
xcoulon: yes, good point!

main.go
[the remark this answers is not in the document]
xcoulon: removed in 05334cc

controllers/space/space_controller.go
xcoulon: I clarified a bit the comment in 35a2d2f, but note that after d34a855, the controller does not requeue explicitly while the NSTemplateSet is being deleted (since the controller watches the NSTemplateSets on the member clusters, a new reconcile loop will automatically be triggered when the NSTemplateSet is deleted). 
Please let me know if it's better like this now ;)

controllers/space/space_controller.go
rajivnathan: Cool, thanks for the detailed explanation!

controllers/space/space_controller.go
xcoulon: this was addressed in https://github.com/codeready-toolchain/host-operator/pull/549

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Cluster struct {
+	OperatorNamespace string
+	Client            client.Client
+	Cache             cache.Cache
+}
```
MatousJobanek, line 12: I already mentioned it in another comment: https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r759920112
this could be simplified
```suggestion
type Cluster struct {
	runtimecluster.Cluster
	OperatorNamespace string
}
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+//   or if there is no target cluster specified in the given space, or if the target cluster is unknown.
+// Returns `false/error` if an error occurred
+func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		logger.Info("cannot delete NSTemplateSet: no target cluster specified")
+		return false, nil // skip NSTemplateSet deletion
+	}
+	memberCluster, found := r.MemberClusters[space.Spec.TargetCluster]
+	if !found {
+		logger.WithValues("target_cluster", space.Spec.TargetCluster).Info("Cannot delete NSTemplateSet: unknown target member cluster")
```
MatousJobanek, line 218: so you introduced the `Space.Status.TargetCluster` but you didn't use it here
it should have been:
```suggestion
	targetCluster := space.Spec.TargetCluster
	if targetCluster == "" {
		targetCluster = space.Status.TargetCluster
	}
	if targetCluster == "" {
		logger.Info("cannot delete NSTemplateSet: no target cluster specified")
		return false, nil // skip NSTemplateSet deletion
	}
	memberCluster, found := r.MemberClusters[targetCluster]
	if !found {
		logger.WithValues("target_cluster", targetCluster).Info("Cannot delete NSTemplateSet: unknown target member cluster")
```
plus also a change in the unit test.

controllers/space/space_controller.go
MatousJobanek: I see that you force requeue again - why? And why immediately and not with some delay?

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	return nil
+}
+
+// ensureNSTemplateSet creates the NSTemplateSet on the target member cluster if it does not exist,
+// and updates the space's status accordingly.
+// returns `true` after creating the NSTemplateSet and until it is `ready`
+func (r *Reconciler) ensureNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		return false, r.setStatusProvisioningFailed(logger, space, fmt.Errorf("unspecified target member cluster"))
```
MatousJobanek, line 116: as I already mentioned in the comment in the last review https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r759324703
this is a valid state and we should not return an error here - only set condition and wait until the targetCluster is set.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	if err := r.updateStatus(
+		space,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.SpaceTerminatingFailedReason,
+			Message: cause.Error(),
+		}); err != nil {
+		logger.Error(cause, "unable to terminate Space")
+		return err
```
MatousJobanek, line 306: just out of curiosity - why did you introduce a new way of setting the failure conditions instead of using the same functions that are used in other controllers - eg.: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/masteruserrecord/masteruserrecord_controller.go#L218-L226
it would be nice to keep the usage consistent.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+func (r *Reconciler) ensureNSTemplateSet(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
+	if space.Spec.TargetCluster == "" {
+		return false, r.setStatusProvisioningFailed(logger, space, fmt.Errorf("unspecified target member cluster"))
+	}
+	memberCluster, found := r.MemberClusters[space.Spec.TargetCluster]
+	if !found {
+		return false, r.setStatusProvisioningFailed(logger, space, fmt.Errorf("unknown target member cluster '%s'", space.Spec.TargetCluster))
+	}
+	name := space.Name // NSTemplateSet will have the same name as the Space
+	logger = logger.WithValues("target_member_cluster", space.Spec.TargetCluster, "name", name)
```
MatousJobanek, line 123: correct me if I'm wrong, but the name should have been included in the logger as part of the logger initialization at the very beginning of the reconcile.

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	// create the NSTemplateSet from the NSTemplateTier
+
+	nsTmplSet := &toolchainv1alpha1.NSTemplateSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: memberOperatorNS,
+			Name:      space.Name,
+			Finalizers: []string{
+				toolchainv1alpha1.FinalizerName,
+			},
```
MatousJobanek, line 174: why do you add the finalizer here - this should have been done by the NSTemplateSet controller, right?
```suggestion
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		},
+		Spec: *usersignup.NewNSTemplateSetSpec(tmplTier),
+	}
+	for i, tmpl := range tmplTier.Spec.Namespaces {
+		nsTmplSet.Spec.Namespaces[i] = toolchainv1alpha1.NSTemplateSetNamespace{
+			TemplateRef: tmpl.TemplateRef,
+		}
+	}
```
MatousJobanek: This is not needed, correct? everything should be set by `usersignup.NewNSTemplateSetSpec`
```suggestion
```

controllers/space/space_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+		return false, r.setStatusNSTemplateSetCreationFailed(logger, space, err)
+	}
+	logger.Info("NSTemplateSet already exists")
+
+	readyCond, found := condition.FindConditionByType(nsTmplSet.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	if !found || readyCond.Status != corev1.ConditionTrue {
+		logger.Info("NSTemplateSet is not ready", "ready-condition", readyCond)
+		return true, nil // here we need to explicitly requeue since the controller doesn't watch the NSTemplateSetStatus
+	}
```
MatousJobanek, line 154: We need to replicate the unready conditions. I mentioned it in the last review https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r759338529
 
in addition - you wrote 

> // here we need to explicitly requeue since the controller doesn't watch the NSTemplateSetStatus

but the controller watches the changes in NSTemplatSet.status - we don't use any predicate for it. Or am I missing something?

main.go
```diff
@@ -261,6 +263,10 @@ func main() {
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "UserSignupCleanup")
 	}
+
+	if err = space.SetupWithManager(mgr, namespace, memberClusters); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Space")
+	}
```
MatousJobanek, line 269: out of curiosity - why do you use a different way of creating the controller compared to the other controllers? The previous format is taken from the code that was generated by operator-sdk - I thought that we wanted to be as close to it as possible.

MatousJobanek (commented):
added a few comments
still need to review the unit tests though.

controllers/space/space_controller.go
xcoulon: because each time I read the code you linked, I have to figure out what are the arguments to pass. I don't find it convenient at all.

controllers/space/space_controller.go
xcoulon: already removed in https://github.com/codeready-toolchain/host-operator/pull/548/commits/d44ab408cb72a5dd0375480e5e40aed42a4ece1d

pkg/cluster/cluster.go
xcoulon: no, I prefer keeping it like this so it's easier to configure in the tests.

controllers/space/space_controller.go
xcoulon: removed in 6415808

controllers/space/space_controller.go
xcoulon: fixed in 47eab1f

main.go
xcoulon: refactored in 6b4027f

controllers/space/space_controller.go
xcoulon: I'm confused because your comment above ended with 

> Anyway, if the condition is either not found or unknown, then we should not replicate it to Space CR.
In UserAccount controller, we don't replicate the status if the UserAccount ready condition is already set to false to just not overload the synchronization mechanism. **We could technically do the same also here, but it's not super needed**

so, am I supposed to do it or not? Besides, we don't have the same kind of sync mechanism as in the UserAccount, so what's the point here?

xcoulon: > but the controller watches the changes in NSTemplatSet.status - we don't use any predicate for it. Or am I missing something?

this was addressed in https://github.com/codeready-toolchain/host-operator/pull/548/commits/47eab1f38e323375dd1afa43d1b39fa6de741d3e

controllers/space/space_controller.go
MatousJobanek: ok, but wouldn't it be good to keep the consistency? 
Feel free to create a story for simplification of these functions.

MatousJobanek: in addition - in your version, you don't wrap the error if the update of the status fails.

controllers/space/space_controller.go
MatousJobanek: @xcoulon thanks for applying my change, but please don't forget to change also the unit tests.

controllers/space/space_controller.go
xcoulon: good idea! See https://issues.redhat.com/browse/CRT-1406

xcoulon: > in addition - in your version, you don't wrap the error if the update of the status fails.

if the update fails, the error is returned and it will be logged by the controller (or manager) itself (ie, by the caller to `Reconcile`). In other words, the `cause` is logged here, and the status update error is logged afterwards. Or am I missing something?

MatousJobanek: Yeah, you are probably right. 
It's just that I'm used to the current way of updating the status, so it's harder to compare it with your version. That's why I proposed using the same way to keep the consistency - mainly when more people will work on this code.

xcoulon: I agree with you about the need to consistency but for now, can we keep this code as-is in this controller and see how we address the consistency in https://issues.redhat.com/browse/CRT-1406?

controllers/space/space_controller.go
MatousJobanek: @xcoulon We need to replicate the status from NSTempalteSet CR to Space CR so we know what's going on with the NSTemplateSet. 
By saying 

> Anyway, if the condition is either not found or unknown, then we should not replicate it to Space CR.
In UserAccount controller, we don't replicate the status if the UserAccount ready condition is already set to false to just not overload the synchronization mechanism. **We could technically do the same also here, but it's not super needed**

I was referring to this particular case https://github.com/codeready-toolchain/member-operator/blob/master/controllers/useraccount/useraccount_controller.go#L404-L408 - in UserAccount we don't replicate the status for the case when the ready condition is already set to false and NSTemplateSet doesn't provide any message. 
Since we don't have to deal with the synchronization mechanism here, we can skip that part and replicate the status for all cases when the condition is not ready https://github.com/codeready-toolchain/member-operator/blob/master/controllers/useraccount/useraccount_controller.go#L410-L416

xcoulon: ok, thanks for the clarification. I will fix that

xcoulon: done in ac77948, hopefully!

controllers/space/space_controller.go
MatousJobanek: We don't know when we will have time to work on CRT-1406 so we can stay with the mix of approaches for a few months. Just saying. :wink: :peace_symbol: 

Let's wait also for @rajivnathan's opinion on it as he will work with the code of this controller as well. I'm fine with keeping it as it is.

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+		})
+	})
+
+	t.Run("delete", func(t *testing.T) {
+
+		t.Run("success", func(t *testing.T) {
+
+			t.Run("after space was successfully provisioned", func(t *testing.T) {
+
+				t.Run("when space is deleted", func(t *testing.T) {
```
MatousJobanek, line 423: Do we really need this amount of subtests when there is no shared common setup? 
In addition. It would be nice to start a brand new test for the new use-case - deletion. It's hard to follow what setup is reused.

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+							nstmplSet.DeletionTimestamp = &now
+							nstmplSet.Status.Conditions = []toolchainv1alpha1.Condition{
+								nstemplatetsettest.Terminating(),
+							}
+							// instead of deleting the resource in the FakeClient,
+							// we update it with a `DeletionTimestamp`
+							return member1Client.Client.Update(ctx, obj)
+						}
+						return member1Client.Client.Delete(ctx, obj, opts...)
+					}
```
MatousJobanek, line 444: Why do we need this? The purpose of this test is:

1. There is a Space that is being deleted
2. There is a NSTemplateSet that is ready
3. We call `Reconcile`
4. Then we should expect that the `delete` was called
5. which means that NSTemplateSet is gone
6. and Space's condition is set to terminating 
7. plus still has the Finalizer set.

If we want to verify the case when the deletion was already executed before - which means that the NSTemplateSet has DeletionTimestamp set, then we could configure the test with:
1. There is a Space that is being deleted
2. There is a NSTemplateSet that is being deleted
3. We call `Reconcile`
4. Then we should expect that the `delete` was NOT called
5. which means that NSTemplateSet is still there with the DeletionTimestap
6. and Space's condition is set to terminating 
7. plus still has still the Finalizer set.

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+						res, err := ctrl.Reconcile(context.TODO(), reconcile.Request{
+							NamespacedName: types.NamespacedName{
+								Namespace: space.Namespace,
+								Name:      space.Name,
+							},
+						})
+
+						// then
+						require.NoError(t, err)
+						assert.Equal(t, reconcile.Result{Requeue: false}, res) // no need to explicitly requeue while the NSTemplate is terminating
```
MatousJobanek, line 494:
```suggestion
						assert.False(t, res.Requeue)
```

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+						spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
+							Exists().
+							HasStatusTargetCluster("member-1").
+							HasConditions(spacetest.Terminating()).
+							HasFinalizer(toolchainv1alpha1.FinalizerName)
+
+						t.Run("done when NSTemplateSet is deleted", func(t *testing.T) {
+							// given another round of requeue without with NSTemplateSet now *fully deleted*
+							member1Client.MockDelete = nil
+							err := member1.Client.Delete(context.TODO(), nsTmplSet)
```
MatousJobanek, line 510: by doing this, you more or less replicate the K8s logic. Wouldn't be better to configure the test properly and test the particular reconciliations in an isolated way?

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+							res, err := ctrl.Reconcile(context.TODO(), reconcile.Request{
+								NamespacedName: types.NamespacedName{
+									Namespace: space.Namespace,
+									Name:      space.Name,
+								},
+							})
+							// then
+							require.NoError(t, err)
+							assert.Equal(t, reconcile.Result{Requeue: false}, res) // no more requeue.
+							spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
```
MatousJobanek, line 523: you should also verify that NSTemplateSet is not there

controllers/space/space_controller.go
```diff
@@ -0,0 +1,340 @@
...
+		targetCluster = space.Status.TargetCluster
+	}
+	if targetCluster == "" {
+		logger.Info("cannot delete NSTemplateSet: no target cluster specified")
+		return false, nil // skip NSTemplateSet deletion
+	}
+	memberCluster, found := r.MemberClusters[targetCluster]
+	if !found {
+		logger.WithValues("target_cluster", targetCluster).Info("Cannot delete NSTemplateSet: unknown target member cluster")
+		return false, nil // skip NSTemplateSet deletion
```
MatousJobanek, line 218: If the targetCluster is set but is not found, then this is a problem - we should return an error here!

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+						},
+					})
+
+					// then
+					require.NoError(t, err)
+					assert.Equal(t, reconcile.Result{Requeue: false}, res) // no requeue neede
+					spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
+						Exists().
+						HasNoStatusTargetCluster().
+						HasNoFinalizers() // will allow for deletion of the Space CR
```
MatousJobanek, line 602: just commented in the reconcile code - this is an error case

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+			})
+
+			t.Run("error while adding finalizer", func(t *testing.T) {
+				// given
+				space := spacetest.NewSpace("oddity", basicTier.Name)
+				hostClient := test.NewFakeClient(t, space)
+				hostClient.MockUpdate = func(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+					fmt.Printf("updating object of type %T\n", obj)
+					if _, ok := obj.(*toolchainv1alpha1.Space); ok {
+						return fmt.Errorf("mock error!")
```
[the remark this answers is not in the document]
xcoulon: yes, it's just a detail. Honestly, it's a mock error, so let's not bother about it, especially after reaching 100 comments on this PR!

controllers/space/space_controller.go
xcoulon: you do realize that you suggested this code in https://github.com/codeready-toolchain/host-operator/pull/548/commits/2fcdba0b8ab1599fa4e03b052b39a8d59ffd771d?

controllers/space/space_controller_test.go
xcoulon: it's isolated per the use of chained subtest. I thought it would make sense to replicate the k8s logic in the unit tests, since it's the expected behaviour after all

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+			member2 := NewMemberCluster(t, "member-2", corev1.ConditionTrue)
+			ctrl := newReconciler(hostClient, member1, member2)
+
+			// when
+			res, err := ctrl.Reconcile(context.TODO(), reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: space.Namespace,
+					Name:      space.Name,
+				},
+			})
```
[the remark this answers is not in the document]
xcoulon: done in 7261943

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+
+			// then
+			require.NoError(t, err)
+			assert.False(t, res.Requeue)
+			nsTmplSet := nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member1.Client).
+				Exists().
+				Get()
```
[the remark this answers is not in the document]
xcoulon: done in d51991c

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+				nsTmplSet.Status.Conditions = []toolchainv1alpha1.Condition{
+					{
+						Type:   toolchainv1alpha1.ConditionReady,
+						Status: corev1.ConditionFalse,
+						Reason: toolchainv1alpha1.NSTemplateSetProvisioningReason,
+					},
+				}
+				err := member1.Client.Update(context.TODO(), nsTmplSet)
+				require.NoError(t, err)
+				ctrl := newReconciler(hostClient, member1, member2)
```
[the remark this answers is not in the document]
xcoulon: I thought that we had issues in the past where a unit test reused a controller, which resulted in errors? or am I missing something?

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+	s := scheme.Scheme
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+	basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates)
+
+	t.Run("create", func(t *testing.T) {
+
+		t.Run("success", func(t *testing.T) {
+			// given
+			space := spacetest.NewSpace("oddity", basicTier.Name, spacetest.WithTargetCluster("member-1"))
```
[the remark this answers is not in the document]
xcoulon: renamed in ac73a08

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+				require.EqualError(t, err, "unspecified target member cluster")
+				assert.False(t, res.Requeue)
+				spacetest.AssertThatSpace(t, space.Namespace, space.Name, hostClient).
+					HasNoStatusTargetCluster().
+					HasConditions(toolchainv1alpha1.Condition{
+						Type:    toolchainv1alpha1.ConditionReady,
+						Status:  corev1.ConditionFalse,
+						Reason:  toolchainv1alpha1.SpaceProvisioningPendingReason,
+						Message: "unspecified target member cluster",
+					})
```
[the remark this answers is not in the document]
xcoulon: done in 6b8445e

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+				require.EqualError(t, err, "unknown target member cluster 'unknown'")
+				assert.False(t, res.Requeue)
+				spacetest.AssertThatSpace(t, space.Namespace, space.Name, hostClient).
+					HasStatusTargetCluster("unknown").
+					HasConditions(toolchainv1alpha1.Condition{
+						Type:    toolchainv1alpha1.ConditionReady,
+						Status:  corev1.ConditionFalse,
+						Reason:  toolchainv1alpha1.SpaceProvisioningFailedReason,
+						Message: "unknown target member cluster 'unknown'",
+					})
```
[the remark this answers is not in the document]
xcoulon: done in 6b8445e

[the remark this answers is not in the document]
xcoulon: see also 94e8079

controllers/space/space_controller_test.go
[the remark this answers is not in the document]
xcoulon: see also 94e8079

controllers/space/space_controller_test.go
xcoulon: I like it that way, but if it's really an issue, then I'm fine with splitting the test.

controllers/space/space_controller_test.go
xcoulon: as I wrote before, the test mimicks the logic of k8s and the sequence of events that lead to the deletion of the Space resource, so we can cover all the cases implemented in the controller (trigger deletion of NSTemplateSet, then wait until NSTemplateSet is deleted to remove the finalizer on the Space)

controllers/space/space_controller_test.go
xcoulon: huh? it's already explicitly deleted in line 510:
```
member1Client.MockDelete = nil
err := member1.Client.Delete(context.TODO(), nsTmplSet)
```

controllers/space/space_controller_test.go
xcoulon: see comment above

controllers/space/space_controller.go
xcoulon: I'm fine with returning an error, but then what are we supposed to do with the Space resource? Leave it in a `TerminatingFailed` state? Let the reconciler return the error until the Space resource has a `targetCluster` set by an admin?

MatousJobanek: > you do realize that you suggested this code in 2fcdba0?

no, my suggestion didn't contain the "return" line - this is what I suggested
```go 
	targetCluster := space.Spec.TargetCluster
	if targetCluster == "" {
		targetCluster = space.Status.TargetCluster
	}
	if targetCluster == "" {
		logger.Info("cannot delete NSTemplateSet: no target cluster specified")
		return false, nil // skip NSTemplateSet deletion
	}
	memberCluster, found := r.MemberClusters[targetCluster]
	if !found {
		logger.WithValues("target_cluster", targetCluster).Info("Cannot delete NSTemplateSet: unknown target member cluster")
```

> I'm fine with returning an error, but then what are we supposed to do with the Space resource? Leave it in a TerminatingFailed state? Let the reconciler return the error until the Space resource has a targetCluster set by an admin?

yes, exactly - keep it in the failed state. The reason can be either a wrong targetCluster name or missing ToolchainCluster CR when the operator deployment was started.

controllers/space/space_controller_test.go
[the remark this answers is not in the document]
MatousJobanek: I'm not aware of anything like that. In the subtests you use the same objects, the same clients, so I would expect that you should use also the same Controller object.
if it wasn't a subtests, then it would be a different story - we shouldn't reuse the same controller between other tests that are at the same level.

controllers/space/space_controller_test.go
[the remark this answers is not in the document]
MatousJobanek: Yeah, it just screams in my IDE

controllers/space/space_controller_test.go
MatousJobanek: Please split it at least into two tests:
1. provisioning
2. deleting

and please avoid so many subtests without using any common setup - it's really hard to review it. 

Just out of curiosity - what is the benefit of having such 4 subtests.

controllers/space/space_controller_test.go
MatousJobanek: ok, but why not having one isolated unit tests per one event? As we have in all other controller unit tests?

1. one unit test that checks that the controller calls deletion (which means that it's really gone without any mocks)
2. one unit test that checks what happens when the NSTemplateSet is already being deleted
3. one unit test that checks what happens when the NSTemplateSet is not there and Space is being deleted
4. one or more unit tests for some error cases

controllers/space/space_controller_test.go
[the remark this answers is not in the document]
xcoulon: I probably got confused with this [discussion on our Slack channel](https://coreos.slack.com/archives/CHK0J6HT6/p1633509468123100?thread_ts=1633503180.103700&cid=CHK0J6HT6)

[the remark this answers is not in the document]
xcoulon: removed in 44a3b31

controllers/space/space_controller_test.go
```diff
@@ -0,0 +1,796 @@
...
+					nsTmplSet.Status.Conditions = []toolchainv1alpha1.Condition{
+						{
+							Type:   toolchainv1alpha1.ConditionReady,
+							Status: corev1.ConditionTrue,
+							Reason: toolchainv1alpha1.NSTemplateSetProvisionedReason,
+						},
+					}
+					err := member1.Client.Update(context.TODO(), nsTmplSet)
+					require.NoError(t, err)
+					ctrl := newReconciler(hostClient, member1, member2)
```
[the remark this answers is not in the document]
xcoulon: removed in 44a3b31

controllers/space/space_controller_test.go
xcoulon: just a matter of organizing the tests, but if you prefer otherwise, then fine

xcoulon: done in 87b2b42

controllers/space/space_controller_test.go
xcoulon: I don't understand: the way it is organized with subtests is exactly for this purpose!

```
t.Run("after space was successfully provisioned", func(t *testing.T) {
	t.Run("when space is deleted", func(t *testing.T) {
		t.Run("wait while NSTemplateSet is terminating", func(t *testing.T) {
			t.Run("done when NSTemplateSet is deleted", func(t *testing.T) {
				...
			})
		})
	})
})
```

controllers/space/space_controller_test.go
MatousJobanek: I meant that you should call the assertion to verify that the NSTemplateSet is still not there and the Space controller didn't create it.

controllers/space/space_controller.go
xcoulon: > no, my suggestion didn't contain the "return" line - this is what I suggested

ok, I'm confused because what you said you suggested contains exactly the same code:
```
return false, nil // skip NSTemplateSet deletion
```

that being said, ok, let me return an error, and we'll see how we can deal with this case when/if it happens

xcoulon: ok, I see. We have 2 identical lines like this
```
return false, nil // skip NSTemplateSet deletion
``` 
and I was looking at the wrong case :(

controllers/space/space_controller_test.go
xcoulon: done in ba33697

controllers/space/space_controller_test.go
alexeykazakov: @xcoulon let me start with some general principles we all I believe follow when writing unit tests. Just for sake of making sure we are the same page :)

What we usually do in unit testes is testing a single reconcile loop:

Create a resource in the state you need it (including setting status if you expect it in the certain state)
Call the reconcile
Check that the resource is now in the state set by the controller (not by kube!). Including checking the status. And, if applicable, check that other resources created or updated by the controller.
If we need to test multiple reconcile loops (controller modifies some resources which triggers another reconcile loop) then we usually do it in separate tests (they can be encapsulated like in your example) when you re-create the state you need before the reconcile and check it after the reconcile.

And this is what you are mostly doing in this test as far as I can see. But it looks like you are skipping one step... and saving one additional test :)
If I'm not mistaken you need to test a step when the NSTemplateSet is only marked as deleted, and not actually deleted yet to make sure that the Space is still waiting for the actually deletion, correct?
If we wanted to follow the principles I described above then we would create another test for that when you manually create a NSTemplateSet with a deletion timestamp in the fake client and then reconcile the Space controller to make sure it behaves properly in this case instead of trying to emulate kube.
But... I personally think that we could cut corners sometime and do little tricks (like this) if it helps us to have tests which easier to read and maintain and covering the functionality in the same time. This is why we use some generic principles in the first place. Not because we follow the rules just because they are the rules.
If it's the case here? I don't know. I'm hesitant... I probably didn't take a good enough look at the tests to have a strong opinion :)

@xcoulon, @MatousJobanek et al WDYT?

controllers/space/space_controller_test.go
xcoulon: done in 5b4b684

controllers/space/space_controller_test.go
xcoulon: > @xcoulon let me start with some general principles we all I believe follow when writing unit tests. Just for sake of making sure we are the same page :)
> 
> What we usually do in unit testes is testing a single reconcile loop:
> 
> Create a resource in the state you need it (including setting status if you expect it in the certain state) Call the reconcile Check that the resource is now in the state set by the controller (not by kube!). Including checking the status. And, if applicable, check that other resources created or updated by the controller. If we need to test multiple reconcile loops (controller modifies some resources which triggers another reconcile loop) then we usually do it in separate tests (they can be encapsulated like in your example) when you re-create the state you need before the reconcile and check it after the reconcile.
> 
> And this is what you are mostly doing in this test as far as I can see. But it looks like you are skipping one step... and saving one additional test :) If I'm not mistaken you need to test a step when the NSTemplateSet is only marked as deleted, and not actually deleted yet to make sure that the Space is still waiting for the actually deletion, correct? If we wanted to follow the principles I described above then we would create another test for that when you manually create a NSTemplateSet with a deletion timestamp in the fake client and then reconcile the Space controller to make sure it behaves properly in this case instead of trying to emulate kube. But... I personally think that we could cut corners sometime and do little tricks (like this) if it helps us to have tests which easier to read and maintain and covering the functionality in the same time. This is why we use some generic principles in the first place. Not because we follow the rules just because they are the rules. If it's the case here? I don't know. I'm hesitant... I probably didn't take a good enough look at the tests to have a strong opinion :)
> 
> @xcoulon, @MatousJobanek et al WDYT?

I don't see where I "cut corners" in the series of tests here:

> If I'm not mistaken you need to test a step when the NSTemplateSet is only marked as deleted, and not actually deleted yet to make sure that the Space is still waiting for the actually deletion, correct?

this is the purpose of subtest <2> below, AFAICS

```
<1> t.Run("when space is deleted", func(t *testing.T) {                                          
<2>  t.Run("wait while NSTemplateSet is still terminating", func(t *testing.T) {
<3>    t.Run("done when NSTemplateSet is deleted", func(t *testing.T) {
```

I'm happy to discuss about the structure of the tests but frankly, if the logic is covered, should we spend more time on this topic?

controllers/space/space_controller_test.go
MatousJobanek: Thanks, but it looks like you added it to the other tests - not this particular one.

MatousJobanek (approved):
Thanks a lot for addressing my comments :+1: :1st_place_medal: 

I left one minor comment https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r766653632

controllers/space/space_controller_test.go
xcoulon: yes, I did add it: https://github.com/codeready-toolchain/host-operator/pull/548/files#diff-92b9cf612f3442ae1d6b7359f59a2c5683ab7edb129c96fdc050bd4c083e5671R528-R529

rajivnathan (approved):

PR #559, opened 2021-12-16
Template update requests for Spaces

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1315

When TemplateUpdateRequest with Spec.CurrentHash is created, then it will only monitor the associated Space until the “toolchain.dev.openshift.com/<tiername>-tier-hash” label at Space CR is different from the one stored in TemplateUpdateRequest.Spec.CurrentHash. In other words, it will wait until the “toolchain.dev.openshift.com/<tiername>-tier-hash” label has changed. When the label is changed, then the TemplateUpdateRequest controller will wait until the Ready condition in Space CR is true.

### DIFF
--- a/controllers/templateupdaterequest/templateupdaterequest_controller_test.go
+++ b/controllers/templateupdaterequest/templateupdaterequest_controller_test.go
@@ -819,6 +819,30 @@ func TestReconcile(t *testing.T) {
 				assert.EqualError(t, err, "unable to get the Space associated with the TemplateUpdateRequest: mock error")
 				assert.Equal(t, reconcile.Result{}, res) // no need to requeue, the Space is watched
 			})
+
+			t.Run("error updating status", func(t *testing.T) {
+				// given
+				previousBasicTier := tiertest.BasicTier(t, tiertest.PreviousBasicTemplates)
+				basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates, tiertest.WithCurrentUpdateInProgress())
+				space := spacetest.NewSpace("user-1", spacetest.WithTierNameAndHashLabelFor(previousBasicTier)) // space's hash matches TUR hash
+				initObjs := []runtime.Object{basicTier, space}
+				previousHash, err := tierutil.ComputeHashForNSTemplateTier(previousBasicTier)
+				require.NoError(t, err)
+				initObjs = append(initObjs, turtest.NewTemplateUpdateRequest("user-1", *basicTier, turtest.CurrentTierHash(previousHash)))
+				r, req, cl := prepareReconcile(t, initObjs...)
+				cl.MockStatusUpdate = func(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+					if _, ok := obj.(*toolchainv1alpha1.TemplateUpdateRequest); ok {
+						return fmt.Errorf("mock error")
+					}
+					return cl.Client.Status().Update(ctx, obj, opts...)
+				}
+				// when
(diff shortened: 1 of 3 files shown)


controllers/templateupdaterequest/templateupdaterequest_controller_test.go
```diff
@@ -690,6 +692,159 @@ func TestReconcile(t *testing.T) {
...
+		t.Run("when space tier hash label still the same", func(t *testing.T) {
+			// given
+			previousBasicTier := tiertest.BasicTier(t, tiertest.PreviousBasicTemplates)
+			basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates, tiertest.WithCurrentUpdateInProgress())
+			space := spacetest.NewSpace("user-1", spacetest.WithTierNameAndHashLabelFor(previousBasicTier)) // space's hash matches TUR hash
+			initObjs := []runtime.Object{basicTier, space}
+			previousHash, err := tierutil.ComputeHashForNSTemplateTier(previousBasicTier)
+			require.NoError(t, err)
+			initObjs = append(initObjs, turtest.NewTemplateUpdateRequest("user-1", *basicTier, turtest.CurrentTierHash(previousHash)))
+			r, req, cl := prepareReconcile(t, initObjs...)
```
MatousJobanek, line 705: I guess that we could pass the init objects directly to the `prepareReconcile` function?
```suggestion
			previousHash, err := tierutil.ComputeHashForNSTemplateTier(previousBasicTier)
			require.NoError(t, err)
			r, req, cl := prepareReconcile(t, basicTier, space, turtest.NewTemplateUpdateRequest("user-1", *basicTier, turtest.CurrentTierHash(previousHash)))
```
the same applies for the following cases as well.

controllers/templateupdaterequest/templateupdaterequest_controller_test.go
```diff
@@ -690,6 +692,159 @@ func TestReconcile(t *testing.T) {
...
+					Reason: toolchainv1alpha1.TemplateUpdateRequestUpdatingReason,
+				})
+		})
+
+		t.Run("when space tier hash label has changed", func(t *testing.T) {
+			t.Run("space condition not ready yet", func(t *testing.T) {
+				// given
+				previousBasicTier := tiertest.BasicTier(t, tiertest.PreviousBasicTemplates)
+				basicTier := tiertest.BasicTier(t, tiertest.CurrentBasicTemplates, tiertest.WithCurrentUpdateInProgress())
+				space := spacetest.NewSpace("user-1", spacetest.WithTierNameAndHashLabelFor(basicTier)) // space's hash updated to new tier
```
MatousJobanek, line 724: the condition of the space is actually not set at all - which doesn't have to be clear from the code. It would be nice to set the condition to "updating" as part of the test setup.

MatousJobanek (approved):
Looks good - simple and clean :+1: 
I have just two comments regarding the tests.

controllers/templateupdaterequest/templateupdaterequest_controller_test.go
rajivnathan: Done in 634cc5e4be3e3398ad3a3c36e418b20c92b1cee1

controllers/templateupdaterequest/templateupdaterequest_controller_test.go
rajivnathan: Done in a569ce2497328bebf5cf79199df106595cc108a1

xcoulon (approved):

PR #560, opened 2021-12-16
fix: update common to not watch changes of the ToolchainCluster.Status

MatousJobanek:
https://github.com/codeready-toolchain/toolchain-common/pull/223


MatousJobanek:
My assumption was wrong - we need to watch also the changes of the status, because we need to store the updated status in the cache

PR #568, opened 2022-01-10
Update Space controller to watch TemplateUpdateRequests

rajivnathan:
Updates the space controller to watch for TemplateUpdateRequest CRs so that it will reconcile and proceed with the NSTemplateSet update flow. The flow is exactly the same as for tier promotion except instead of only comparing the tier name it now does a comparison of the tier hash using the new `tierutil.TierHashMatches()` func

Related Issue: https://issues.redhat.com/browse/CRT-1314

Related PRs:
- https://github.com/codeready-toolchain/toolchain-common/pull/228
- https://github.com/codeready-toolchain/toolchain-e2e/pull/435


controllers/space/space_controller_test.go
```diff
@@ -737,16 +737,126 @@ func TestPromoteSpace(t *testing.T) {
...
+					assert.Equal(t, reconcile.Result{Requeue: false}, res) // no more requeue.
+					spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
+						Exists().
+						HasStatusTargetCluster("member-1").
+						HasConditions(spacetest.Ready()).
+						HasLabel(tierutil.TemplateTierHashLabelKey(basicTier.Name)).
+						HasFinalizer()
+				})
+			})
+		})
```
MatousJobanek, line 846: I guess that we don't need this sub-tests as all of the cases are already covered in the test-case above (in promotion).
I know that it's not clear for the first glance as it's done as a sub-test of the promotion, but there is nothing new covered here, right?

This is one of the reasons why not nesting such test-cases into sub-tests (and keeping them at the same level) would be much better/clearer. @xcoulon

controllers/space/space_controller_test.go
```diff
@@ -737,16 +737,126 @@ func TestPromoteSpace(t *testing.T) {
...
 
+	t.Run("tier update (same tier but updated references)", func(t *testing.T) {
+		// given that Space is set to the same tier that has been updated and the corresponding NSTemplateSet is not up-to-date
+		s := spacetest.NewSpace("oddity",
+			spacetest.WithTierNameAndHashLabelFor(basicTier), // assume that at this point, the `TemplateTierHash` label was already removed by the ChangeTierRequestController
+			spacetest.WithSpecTargetCluster("member-1"),
+			spacetest.WithStatusTargetCluster("member-1"), // already provisioned on a target cluster
+			spacetest.WithFinalizer())
+		hostClient := test.NewFakeClient(t, s, basicTier, otherTier)
+		nstmplSet := nstemplatetsettest.NewNSTemplateSet("oddity", nstemplatetsettest.WithReadyCondition()) // NSTemplateSet has references to old basic tier
```
MatousJobanek, line 748: if you used references from the "older" version of the basic tier:
```
nstemplatetsettest.WithReferencesFor(olderBasicTier)
```
it would make the code more clear why the content is not matching

controllers/space/space_controller_test.go
```diff
@@ -737,16 +737,126 @@ func TestPromoteSpace(t *testing.T) {
...
+			Requeue:      true,
+			RequeueAfter: 1 * time.Second,
+		}, res) // explicitly requeue while the NSTemplate update is triggered by its controller
+		spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
+			Exists().
+			HasTier(basicTier.Name).
+			HasSpecTargetCluster("member-1").
+			HasStatusTargetCluster("member-1").
+			HasConditions(spacetest.Updating()).
+			HasLabel(tierutil.TemplateTierHashLabelKey(basicTier.Name))
```
MatousJobanek, line 769: could we also verify the value of the label? at this point, it should be still the older version.
That also means that we need to add a test-case that will verify a situation when:
1. The Space has the hash label value outdated 
2. The NSTemplateSet is up-to-date
3. The controller update the hash label value to the latest one

but there is maybe already such a test-case - not sure :man_shrugging:

MatousJobanek: btw, it would be maybe good to modify
```
HasLabel(tierutil.TemplateTierHashLabelKey(
```
so it automatically checks the value of the label - that it matches the content of the given tier

controllers/space/space_controller_test.go
```diff
@@ -737,16 +737,126 @@ func TestPromoteSpace(t *testing.T) {
...
+		nsTmplSet := nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member1.Client).
+			Exists().
+			HasTierName(basicTier.Name).
+			Get()
```
MatousJobanek, line 773: but you actually don't verify the content of the NSTemplateSet anywhere, right?

This is missing also in the promotion test-case.

controllers/space/space_controller_test.go
```diff
@@ -737,16 +737,126 @@ func TestPromoteSpace(t *testing.T) {
...
+		require.False(t, tierutil.TierHashMatches(basicTier, nstmplSet.Spec))
```
MatousJobanek, line 774: This should match! The NSTemplateSet.spec should be updated. 
```suggestion
		require.True(t, tierutil.TierHashMatches(basicTier, nsTmplSet.Spec))
```

controllers/space/space_controller_test.go
rajivnathan: Ok, I think this should do it? e84e7132a0bad6f885cfb3e11cbf3c576716225c

controllers/space/space_controller_test.go
rajivnathan: That's a good suggestion. Thanks! 3a63b3ca6927a4f2ccc3943606adae44b7121c33

controllers/space/space_controller_test.go
rajivnathan: Added a func `HasMatchingTierLabelForTier` that verifies both the label exists and its hash matches: ea242bebc5906482ce2f79135bc2f30ee311d08b

There is already a test named `update not needed` that doesn't set the hash label on the space and the NSTemplateSet is up-to-date which is basically the same thing: https://github.com/codeready-toolchain/host-operator/pull/568/files#diff-92b9cf612f3442ae1d6b7359f59a2c5683ab7edb129c96fdc050bd4c083e5671R830

controllers/space/space_controller_test.go
rajivnathan: Right, good catch. I've added them for this and promotion as well: 224c8d716aa953e1143560bcdf10044a67351575

controllers/space/space_controller_test.go
rajivnathan: 😱 Thank you for noticing this! Fixed in 12fab16b5184a9025ff9901c49ff62d0fd4004d1 I didn't notice that I had `nstmplSet` and `nsTmplSet` vars so I fixed that too.

rajivnathan:
@MatousJobanek all comments have been addressed. Can you please take another look when you have some time. Thanks!

controllers/space/space_controller_test.go
```diff
@@ -729,24 +732,117 @@ func TestPromoteSpace(t *testing.T) {
...
+				require.NoError(t, err)
+				assert.Equal(t, reconcile.Result{Requeue: false}, res) // no more requeue.
+				spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
+					Exists().
+					HasStatusTargetCluster("member-1").
+					HasConditions(spacetest.Ready()).
+					HasMatchingTierLabelForTier(basicTier). // label updated
+					HasFinalizer()
+			})
+		})
```
MatousJobanek, line 832: If you slightly modify the tests then you can easily remove also the `"requeue while NSTemplateSet is not ready"` one - it doesn't verify anything new which wouldn't be already verified earlier, am I correct?
```suggestion
		s = spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
			Exists().
			HasTier(basicTier.Name).
			HasSpecTargetCluster("member-1").
			HasStatusTargetCluster("member-1").
			HasConditions(spacetest.Updating()).
			HasMatchingTierLabelForTier(olderBasicTier).
			Get()
		nsTmplSet = nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member1.Client).
			Exists().
			HasTierName(basicTier.Name).
			HasClusterResourcesTemplateRef("basic-clusterresources-123456new").
			HasNamespaceTemplateRefs("basic-code-123456new", "basic-dev-123456new", "basic-stage-123456new").
			Get()
		require.True(t, tierutil.TierHashMatches(basicTier, nsTmplSet.Spec))

		t.Run("NSTemplateSet is provisioned", func(t *testing.T) {
			// given another round of requeue with NSTemplateSet now *ready*
			nsTmplSet.Status.Conditions = []toolchainv1alpha1.Condition{
				nstemplatetsettest.Provisioned(),
			}
			err := member1.Client.Update(context.TODO(), nsTmplSet)
			require.NoError(t, err)
			// hack: change Space's updating condition timestamp so that it appears the status is ready for >1 second since it's required before can go to Ready condition
			s.Status.Conditions[0].LastTransitionTime = metav1.NewTime(s.Status.Conditions[0].LastTransitionTime.Time.Add(-1 * time.Second))
			err = hostClient.Status().Update(context.TODO(), s)
			require.NoError(t, err)

			// when
			res, err = ctrl.Reconcile(context.TODO(), requestFor(s))

			// then
			require.NoError(t, err)
			assert.Equal(t, reconcile.Result{Requeue: false}, res) // no more requeue.
			spacetest.AssertThatSpace(t, test.HostOperatorNs, "oddity", hostClient).
				Exists().
				HasStatusTargetCluster("member-1").
				HasConditions(spacetest.Ready()).
				HasMatchingTierLabelForTier(basicTier). // label updated
				HasFinalizer()
		})
```

MatousJobanek (approved):
Thanks a lot for addressing my comments. I have one more proposal for simplifying the tests, but it's just a minor thing.

controllers/space/space_controller_test.go
rajivnathan: Well actually, I thought of it as verifying the Space status when the NSTemplateSet status changes to Updating. Maybe it's worth renaming the test to `NSTemplateSet status changes to Updating` and its subtest to `NSTemplateSet status changes to Provisioned` for clarity instead?

MatousJobanek: But what is the new functionality that is tested here? It's already covered in the previous "promotion" test, right?

rajivnathan: Ok, that's a fair point. Since it's covered in the promotion test then I guess we can remove it here.

PR #570, opened 2022-01-10
refactor: remove UserSignupController.wrapErrorWithStatusUpdate

xcoulon:


controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -150,7 +150,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 	err = r.aggregateAndUpdateStatus(reqLogger, toolchainStatus)
 	if err != nil {
-		reqLogger.Error(err, "Failed to update status")
+		reqLogger.Error(err, "failed to update status")
```
alexeykazakov, line 153: What's wrong with upper case?

alexeykazakov (commented):
What's the purpose of this refactoring?

xcoulon:
> What's the purpose of this refactoring?

to improve readability of the code. IMO, it's hard it's understand how the arguments passed to the `wrapErrorWithStatusUpdate` func are used. In the new `SpaceController`, I decided to avoid using such a func, and IMO the code is easier to read

controllers/toolchainstatus/toolchainstatus_controller.go
xcoulon: when errors are wrapped, the cause and error message are joined with a `:` character, but there should not be an uppercase. Also, I thought that uppercases in error messages were discouraged?

xcoulon: > > What's the purpose of this refactoring?
> 
> to improve readability of the code. IMO, it's hard it's understand how the arguments passed to the `wrapErrorWithStatusUpdate` func are used. In the new `SpaceController`, I decided to avoid using such a func, and IMO the code is easier to read

in other words, the extra level of indirection introduced with the use of a `statusUpdater` parameter in makes the `wrapErrorWithStatusUpdate` func unnecessarily complicated (IMHO)

MatousJobanek:
I remember our discussion regarding this topic https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r763859438 but I would prefer postponing this refactoring effort to the time when the migration to Spaces is done - it has much lower priority and addressing it in all controllers is quite a big task. 

I know that you proposed one solution in the Jira issue https://issues.redhat.com/browse/CRT-1406 but I believe that there are more ways of making it better. For example, introducing something more generic that could be used in all controllers. I have one idea. We can also take a look at the new functions that have been introduced in apimachinery lib https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go (but I don't think that we could use it though)

So please, put it on hold for now, and let's focus on the migration. In the meantime, we can create some doc where we could write a few proposals and discuss them with the entire team. As soon as we agree on one approach in the whole team, then we can plan the effort accordingly.

xcoulon:
> I remember our discussion regarding this topic [#548 (comment)](https://github.com/codeready-toolchain/host-operator/pull/548#discussion_r763859438) but I would prefer postponing this refactoring effort to the time when the migration to Spaces is done - it has much lower priority and addressing it in all controllers is quite a big task.
> 
> I know that you proposed one solution in the Jira issue https://issues.redhat.com/browse/CRT-1406 but I believe that there are more ways of making it better. For example, introducing something more generic that could be used in all controllers. I have one idea. We can also take a look at the new functions that have been introduced in apimachinery lib https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go (but I don't think that we could use it though)
> 
> So please, put it on hold for now, and let's focus on the migration. In the meantime, we can create some doc where we could write a few proposals and discuss them with the entire team. As soon as we agree on one approach in the whole team, then we can plan the effort accordingly.

ok, I'm not sure what idea you have in mind, but what concerns me now is that if we wait for too long, the PR will be hard to merge with the latest changes :/

MatousJobanek:
My idea is to introduce either an interface or some additional type for all `*Status` types in the `api` repo which would help us with managing the conditions. Then we could use this abstraction in a generic function across all controllers. I've had that idea for more than one year already, but I haven't got to give it a try yet as it has always been the lowest priority in the list of my tasks.

I understand your concern, but this change affects the entire project as it should be applied in all controllers (to keep the consistency). I raised the fact that it could take some time to implement it also in the discussion I linked above.

Anyway, since it affects the entire team, we should transparently discuss it and agree on a final solution that would be suitable for all of us. Otherwise, we could end up refactoring it in a few months again.

xcoulon:
> My idea is to introduce either an interface or some additional type for all `*Status` types in the `api` repo which would help us with managing the conditions. Then we could use this abstraction in a generic function across all controllers. I've had that idea for more than one year already, but I haven't got to give it a try yet as it has always been the lowest priority in the list of my tasks.
> 
> I understand your concern, but this change affects the entire project as it should be applied in all controllers (to keep the consistency). I raised the fact that it could take some time to implement it also in the discussion I linked above.
> 
> Anyway, since it affects the entire team, we should transparently discuss it and agree on a final solution that would be suitable for all of us. Otherwise, we could end up refactoring it in a few months again.

well, I'm not sure that an addition interface will simplify the readability, but let's discuss about this altogether, indeed. In the mean time, I'm putting this PR in "draft" mode so we know it's not "ready" yet.

PR #572, opened 2022-01-12
fix: remove deactivating state after promoting to a new tier

MatousJobanek:
When promoting a user to a new tier I saw that there was a race-condition:
1. `ChangeTierRequest` was created
2. The `ChangeTierRequest` controller removed `Deactivating` state from `UserSignup`
3. The `Deactivation` controller was triggered based on the change in the spec
4. The `Deactivation` controller checked the tier of the MUR - which was still the old one - and added the `deactivating` state back to `UserSignup`
5. The `ChangeTierRequest` controller updated the `TierName` in MUR but it was too late

So this PR moves the logic that updates `UserSignup` resources after changing TierName in MUR 

There won't be any e2e test since it's impossible to write a reliable one - it's race-condition issue. But I added unit-test for it.


alexeykazakov (approved):
Good catch!

PR #577, opened 2022-01-24
chore: update linter config

xcoulon:
disable 'golint', use 'revive' instead

### DIFF
--- a/controllers/toolchainstatus/toolchainstatus_controller.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller.go
@@ -64,7 +64,7 @@ const (
 	hostOperatorTag        statusComponentTag = "hostOperator"
 	memberConnectionsTag   statusComponentTag = "members"
 	counterTag             statusComponentTag = "MasterUserRecord and UserAccount counter"
-	minutesAfterUnready    time.Duration      = 10 * time.Minute
+	durationAfterUnready   time.Duration      = 10 * time.Minute
 )
 
 const (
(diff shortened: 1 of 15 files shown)


MatousJobanek (approved):

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -64,7 +64,7 @@ const (
 	hostOperatorTag        statusComponentTag = "hostOperator"
 	memberConnectionsTag   statusComponentTag = "members"
 	counterTag             statusComponentTag = "MasterUserRecord and UserAccount counter"
-	minutesAfterUnready    time.Duration      = 10
+	minutesAfterUnready    time.Duration      = 10 * time.Minute
```
alexeykazakov, line 67: Let's rename it. It's not the number of minutes actually. It's a duration.
durationAfterUnready?

controllers/usersignup/usersignup_controller.go
```diff
@@ -548,7 +548,7 @@ func (r *Reconciler) DeleteMasterUserRecord(mur *toolchainv1alpha1.MasterUserRec
 
 	err := r.updateStatus(logger, userSignup, inProgressStatusUpdater)
 	if err != nil {
-		return nil
+		return err
```
alexeykazakov, line 551: I guess we are missing a test for this case? Otherwise we would catch it earlier.

alexeykazakov (approved):
Looks good. Just a minor comment.

controllers/usersignup/usersignup_controller.go
xcoulon: ah right, I intended to add a test for that, as I did in https://github.com/codeready-toolchain/member-operator/pull/328/files#diff-f57f1b01b263fd00b64ab1ec2d007285e03675feb420c7732360f639b38ce736R1262 when making a similar change, but I forgot about it 🤦‍♂️ ! Thanks for the reminder ;)

xcoulon: done in 1af071a by adding a test inspired from `TestUserSignupDeactivatedButMURDeleteFails` just above

controllers/toolchainstatus/toolchainstatus_controller.go
xcoulon: done in fd6f00a

PR #579, opened 2022-01-25
feat: limit permissions granted to AppStudio users

MatousJobanek:
This PR changes the `appstudio` tier template:

cleanup of :ghost::
* drops `Role` & `RoleBindings` that granted view & exec permissions to crtadmin users - we don't use crtadmin users in AppStudio cluster
* drops `RoleBinding` that granted view permissions to a `User` resource - there is no user provisoned for `appstudio` tier - we should use only SAs

limit the user's permissions:
* changes the `ClusterRole` that is granted to the user's SA in the workspace namespace from `edit` to `view`
* adds new `Role` & `RoleBinding` that grants these permissions to the user's namespace:
  * CRUD on HAS Application & Component CRDs
  * CRUD on PipelineRuns
  * only create & delete permissions for secrets
  
I also added some basic comments describing the purpose of the Roles/RoleBindings

for more info see: https://coreos.slack.com/archives/C02CTEB3MMF/p1643044182095500

paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/456


deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
```diff
@@ -14,93 +14,98 @@ objects:
...
+    - applications
+    - components
     verbs:
     - get
     - list
     - watch
     - create
     - delete
     - update
+    - patch
```
alexeykazakov, line 44: Could be just * I guess.

alexeykazakov (approved):

rajivnathan (approved):
Looks good 👍

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
MatousJobanek: good point - updated in bfe899bd

PR #580, opened 2022-01-25
Set tier label on useraccount when created or updated

rajivnathan:
As part of the migration to use Spaces, the UserAccountSpec in MUR will not contain NSTemplateSet going forward so the UserAccount will not have tier information either. That impacts this part of the code in the member-operator: https://github.com/codeready-toolchain/member-operator/blob/61c6ce1a07519d894e6bff4b58e4b6b6e82df517/controllers/useraccount/useraccount_controller.go#L163-L184

That code checks whether it's the appstudio tier so as a workaround we'll set a tier label on the useraccount and the check can then be changed to 
```
if userAcc.Labels != nil && userAcc.Labels[toolchainv1alpha1.TierLabelKey] != "appstudio"
```

Related PRs:
- https://github.com/codeready-toolchain/toolchain-common/pull/232
- https://github.com/codeready-toolchain/toolchain-e2e/pull/458


go.mod
```diff
@@ -31,4 +29,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.8.3
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/rajivnathan/toolchain-common v0.0.0-20220125233151-b258a6cb1a08
```
MatousJobanek, line 32: reminder

MatousJobanek (approved):

go.mod
rajivnathan: updated

PR #588, opened 2022-02-01
feat: introduce SpaceCompletion controller

MatousJobanek:
Adds SpaceCompletion controller which will set TierName and TargetCluster if they are not set already.
To choose the correct TargetCluter it uses the same capacity manager as is used for UserSignup

paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/469

### DIFF
--- a/controllers/spacecompletion/space_completion_controller.go
+++ b/controllers/spacecompletion/space_completion_controller.go
@@ -35,6 +35,7 @@ type Reconciler struct {
 // Watches the Space resources and the ToolchainStatus CRD
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		Named("spacecompletion").
 		// watch Spaces in the host cluster
 		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(
@@ -47,7 +48,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Reconcile ensures that Space has set all missing fields that are needed for proper provisoning
 func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx, "namespace", r.Namespace).WithName("completion")
+	logger := log.FromContext(ctx, "namespace", r.Namespace)
 	logger.Info("reconciling Space")
 
 	// Fetch the Space
--- a/controllers/spacecompletion/space_completion_controller.go
+++ b/controllers/spacecompletion/space_completion_controller.go
@@ -1,0 +1,110 @@
+package spacecompletion
+
+import (
+	"context"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/host-operator/controllers/toolchainconfig"
+	"github.com/codeready-toolchain/host-operator/pkg/capacity"
+	"github.com/codeready-toolchain/host-operator/pkg/pending"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+	"github.com/go-logr/logr"
+	"github.com/redhat-cop/operator-utils/pkg/util"
+
+	errs "github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


controllers/spacecompletion/space_completion_controller.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Fetch the Space
+	space := &toolchainv1alpha1.Space{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: r.Namespace,
+		Name:      request.Name,
+	}, space)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Space not found")
```
xcoulon, line 62: the `IsNotFound` case could be tested

controllers/spacecompletion/space_completion_controller.go
```diff
@@ -0,0 +1,111 @@
...
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, errs.Wrap(err, "unable to get the current Space")
```
xcoulon, line 69: this case could be tested, too

controllers/spacecompletion/space_completion_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	if space.Spec.TargetCluster == "" {
+		targetCluster, err := capacity.GetOptimalTargetCluster("", space.Namespace, r.GetMemberClusters, r.Client)
+		if err != nil {
+			return false, errs.Wrapf(err, "unable to get the optimal target cluster")
+		}
+		if targetCluster == "" {
+			logger.Info("no cluster available")
+			return false, nil
```
xcoulon, line 103: so in that case, the `Space` will remain in "pending" state until there is a new reconcile loop trigger by a chanfge on the resource. 
Is it what we want to do, or should be requeue with a delay until there is room on a cluster? 
Also, in such a case, should we have an alert? 🚨

controllers/spacecompletion/space_completion_controller_test.go
```diff
@@ -0,0 +1,203 @@
...
+		space := spacetest.NewSpace("without-fields",
+			spacetest.WithTierName(""))
+		r, req, cl := prepareReconcile(t, space, getMemberClusters)
+
+		// when
+		_, err := r.Reconcile(context.TODO(), req)
+
+		// then
+		require.NoError(t, err)
+		spacetest.AssertThatSpace(t, test.HostOperatorNs, space.Name, cl).
```
xcoulon, line 41: could you add an assertion on the response, in particular, on the `requeue` field? (assuming we expect `false` at this point?)

xcoulon: (same comment applies on all tests below)

controllers/spacecompletion/space_completion_controller.go
MatousJobanek: We don't test these two cases in 99% of all controllers, so I didn't see it important doing that here. 
But I can add test case for that.

controllers/spacecompletion/space_completion_controller.go
MatousJobanek: The controller also watches ToolchainStatus, so as soon as there is any change in the clusters, then the controller will be triggered for the oldest Space that is in pending state.
In addition, ToolchainStatus is updated every 5 seconds or so, which will trigger the reconcile as well.

controllers/spacecompletion/space_completion_controller_test.go
MatousJobanek: it's supposed to be false literally in all cases, so I followed the same approach as we use in all other controllers like that - I don't check that.

controllers/spacecompletion/space_completion_controller.go
xcoulon: ah ok, I missed that :/ It explains why the e2e tests passed, then!

controllers/spacecompletion/space_completion_controller.go
MatousJobanek: added c7a09ec8

controllers/spacecompletion/space_completion_controller.go
MatousJobanek: added c7a09ec8

xcoulon (approved):

PR #592, opened 2022-02-07
feat: choose the target cluster based on capacity in a balanced way

MatousJobanek:
By default, choose the target cluster based on capacity in a balanced way, because I got bored by trying to achieve the same thing via modifying the thresholds over and over again.

The capacity manager returns the cluster with the most capacity available there.

### DIFF
--- a/pkg/capacity/manager_test.go
+++ b/pkg/capacity/manager_test.go
@@ -31,7 +31,8 @@ func TestGetOptimalTargetCluster(t *testing.T) {
 			"1,external": 700,
 		}),
 		WithMember("member1", WithUserAccountCount(700), WithNodeRoleUsage("worker", 68), WithNodeRoleUsage("master", 65)),
-		WithMember("member2", WithUserAccountCount(200), WithNodeRoleUsage("worker", 55), WithNodeRoleUsage("master", 60)))
+		WithMember("member2", WithUserAccountCount(200), WithNodeRoleUsage("worker", 55), WithNodeRoleUsage("master", 60)),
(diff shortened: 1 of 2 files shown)


pkg/capacity/manager_test.go
```diff
@@ -70,6 +71,24 @@ func TestGetOptimalTargetCluster(t *testing.T) {
 		assert.Equal(t, "member1", clusterName)
 	})
 
+	t.Run("with two clusters and enough capacity in both of them so it returns the with more capacity (the third one)", func(t *testing.T) {
```
alexeykazakov, line 74:
```suggestion
	t.Run("with three clusters and enough capacity in both of them so it returns the with more capacity (the third one)", func(t *testing.T) {
```

alexeykazakov (approved):
Yeah... A good engineer is a lazy engineer. Instead of doing something multiple times manually just automate it once and forget about it ;)

Regarding this specific way of balancing the load. The downside of this option is that if there are two users registering in the same time then it's most likely that they will end up in different clusters. Which might be not very convenient if a few ppl from the same team want to explore namespace sharing.
To mitigate that we could provision users by chunks. Let's say 50 users goes to one member, then we re-calculate the preferable cluster and the next 50 users go to that cluster. Then re-calculate again.
But, the suggested approach is still better than we are currently using anyway. So, let's merge it and then maybe improve it later.

MatousJobanek:
> To mitigate that we could provision users by chunks. Let's say 50 users goes to one member, then we re-calculate the preferable cluster and the next 50 users go to that cluster. Then re-calculate again.
But, the suggested approach is still better than we are currently using anyway. So, let's merge it and then maybe improve it later.

@alexeykazakov here you go https://github.com/codeready-toolchain/host-operator/pull/593

PR #593, opened 2022-02-09
feat: privision in batches of 50

MatousJobanek:
https://github.com/codeready-toolchain/host-operator/pull/592#pullrequestreview-875208283

### DIFF
--- a/pkg/capacity/manager.go
+++ b/pkg/capacity/manager.go
@@ -2,6 +2,7 @@ package capacity
 
 import (
 	"context"
+	"math"
 	"sort"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/capacity/manager.go
```diff
@@ -82,7 +83,7 @@ func GetOptimalTargetCluster(preferredCluster, namespace string, getMemberCluste
 		provisioned2 := counts.UserAccountsPerClusterCounts[optimalTargetClusters[j]]
 		threshold2 := config.AutomaticApproval().MaxNumberOfUsersSpecificPerMemberCluster()[optimalTargetClusters[j]]
 
-		return float32(provisioned1)/float32(threshold1) < float32(provisioned2)/float32(threshold2)
+		return math.Round(float64(provisioned1)/float64(threshold1)*10) < math.Round(float64(provisioned2)/float64(threshold2)*10)
```
alexeykazakov, line 86: If I'm not mistaken, this will give you batches of 50 only if the max capacity is 1000.
But for example if it's 2000 then the batches will be 100.
And for 10000 it will be 500.
Which is fine IMO. But let's add a comment to `GetOptimalTargetCluster()` function explaining the batches. And some comments in the tests would be helpful too.

alexeykazakov (approved):

alexeykazakov (approved):
Looks good!

PR #602, opened 2022-02-28
reduce log level when no spacebinding exists

xcoulon:
no need to dump a stacktrace in the logs when there is no spacebinding


MatousJobanek (approved):
Thanks for fixing this copy-paste mistake :man_facepalming:

PR #603, opened 2022-02-28
feat: update k8s to v0.22.7 and OpenShift to 4.9

MatousJobanek:
* Update k8s and OpenShift dependencies so they match the version of OpenShift clusters we run. It also corresponds to the version that is generated by operator-sdk 1.16.0.
* updates the version of controller-gen binary
* modifies (or also deletes) some of the tests (related to finalizers & deletions) because FakeClient newly automatically deletes all resources that are marked for deletion (have DeletionTimestamp set) and don't have any finalizer set. 
* one test in `space_controller_test.go` file was significantly refactored because it was hard to maintain/modify it in the original format.
* updates auth-proxy and manager deployment manifests so they match what is generated by operator-sdk 1.16.0

related PRs:
https://github.com/codeready-toolchain/api/pull/295
https://github.com/codeready-toolchain/toolchain-common/pull/243


controllers/space/space_controller_test.go
```diff
@@ -1035,47 +976,47 @@ func TestRetargetSpace(t *testing.T) {
 			HasConditions(spacetest.Retargeting()).
 			HasStatusTargetCluster("member-1") // not reset yet
 
-		t.Run("status target cluster is not reset while NSTemplateSet is still being deleted on member-1", func(t *testing.T) {
-			// once NSTemplateSet resource is fully deleted, the SpaceController is triggered again
-			// and it can create the NSTemplateSet on member-2 cluster
-
+		//t.Run("status target cluster is not reset while NSTemplateSet is still being deleted on member-1", func(t *testing.T) {
```
xcoulon, line 979: do we need to keep this?

controllers/space/space_controller_test.go
```diff
@@ -1035,47 +976,47 @@ func TestRetargetSpace(t *testing.T) {
...
+		//		HasStatusTargetCluster("member-1") // not updated yet
+		//	nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member1.Client).
+		//		HasDeletionTimestamp()
+		//	nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member2.Client).
+		//		DoesNotExist()
+
+		t.Run("status target cluster is reset when NSTemplateSet is deleted on member-1", func(t *testing.T) {
+			//// once NSTemplateSet resource is fully deleted, the SpaceController is triggered again
+			//// and it can create the NSTemplateSet on member-2 cluster
+			//err := member1Client.Client.Delete(context.TODO(), nstmplSet) // bypass the mock func
```
xcoulon, line 1001: same here: do we need to keep this?

controllers/space/space_controller_test.go
```diff
@@ -1035,47 +976,47 @@ func TestRetargetSpace(t *testing.T) {
...
-					DoesNotExist()
-				nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member2.Client).
-					Exists().
-					HasTierName(basicTier.Name)
-			})
+			nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member2.Client).
+				Exists().
+				HasTierName(basicTier.Name)
 		})
+		//})
```
xcoulon, line 1019: same here

controllers/space/space_controller_test.go
```diff
@@ -356,225 +356,182 @@ func TestDeleteSpace(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, reconcile.Result{Requeue: false}, res) // no need to explicitly requeue while the NSTemplate is terminating
+			spacetest.AssertThatSpace(t, s.Namespace, s.Name, hostClient).
+				Exists().
+				HasStatusTargetCluster("member-1").
+				HasConditions(spacetest.Terminating()).
+				HasFinalizer() // finalizer is still present while the NSTemplateSet is not fully deleted
+			nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member1.Client).
+				DoesNotExist()
+			// stop the test here: it verified that the NSTemlateSet deletion was triggered (the rest is already covered above)
```
xcoulon, line 453:
```suggestion
			// stop the test here: it verified that the NSTemplateSet deletion was triggered (the rest is already covered above)
```

controllers/space/space_controller_test.go
```diff
@@ -356,225 +356,182 @@ func TestDeleteSpace(t *testing.T) {
...
-					HasConditions(nstemplatetsettest.Terminating()).
-					Get()
+		t.Run("Space controller deletes NSTemplateSet", func(t *testing.T) {
+			// given
+			hostClient := test.NewFakeClient(t, s, basicTier)
+			memberClient := test.NewFakeClient(t, nsTmplSet)
+			member := NewMemberClusterWithClient(memberClient, "member-1", corev1.ConditionTrue)
+			ctrl := newReconciler(hostClient, member)
 
-				t.Run("wait while NSTemplateSet is still terminating", func(t *testing.T) {
```
xcoulon, line 412: sorry but I disagree with the way subtests are reorganized in this PR: they are intended to follow flow/steps that happen for real, ie, ieach subtest is a step towards the deletion of the space and its associated resources. 
IMO it's harder to figure out why there's a "when NSTemplateSet is in terminating state" subtest (L395) at the same level as the previous one now.

xcoulon (changes requested):

controllers/space/space_controller_test.go
MatousJobanek: My understanding of a unit test is that it should verify one single functionality ideally in an isolated way. The unit test should have a defined set of inputs (initial resources in some state) and the assertions that verify the expected values after running either the method or the function that is being verified. Is your understanding the same?

This is exactly what I did:
1. I defined a set of inputs like the NSTemplateSet in the expected state (no matter why it is in this state)
2. Ran the `Reconcile` function
3. Verified the result

in this way, it's easy to understand what is being tested in this particular test and thus it is also very easy to modify and maintain.

When we combine multiple reconcile loops in a set of sub-tests to simulate the k8s flow and to mock the member-operator functionality, then this looks like more an integration test. For this, we could use the EnvTest framework. 
I know that it's in sub-tests that verify the particular reconcile loops, but in order to modify it, you need to understand and follow the entire flow (and not one single functionality). Mainly when there is some additional mocking that depends on the previous tests.
There are some cases when we use sub-tests to verify the second or the third reconcile loop of the same controller because it is just much easier to reuse the already existing state, but in these cases, we don't mock nor simulate the k8s  nor other operator functionality as part of the setup.

controllers/space/space_controller_test.go
MatousJobanek: I forgot to push a commit that removes it 7dcc5402

controllers/space/space_controller_test.go
MatousJobanek: the same - just forgot to push it 7dcc5402

controllers/space/space_controller_test.go
MatousJobanek: the same 7dcc5402

controllers/space/space_controller_test.go
xcoulon: > My understanding of a unit test is that it should verify one single functionality ideally in an isolated way. The unit test should have a defined set of inputs (initial resources in some state) and the assertions that verify the expected values after running either the method or the function that is being verified. Is your understanding the same?
> 

yes, but you did not mention the case of subtests.


> This is exactly what I did:
> 
> 1. I defined a set of inputs like the NSTemplateSet in the expected state (no matter why it is in this state)
> 2. Ran the `Reconcile` function
> 3. Verified the result
> 
> in this way, it's easy to understand what is being tested in this particular test and thus it is also very easy to modify and maintain.
> 

I do not believe it makes it easier to understand than the subtests, especially since you need to "deepcopy" some resources in some cases.


> When we combine multiple reconcile loops in a set of sub-tests to simulate the k8s flow and to mock the member-operator functionality, then this looks like more an integration test. For this, we could use the EnvTest framework. 

yes, we could use EnvTest framework, but it's not a requirement

> I know that it's in sub-tests that verify the particular reconcile loops, but in order to modify it, you need to understand and follow the entire flow (and not one single functionality). 

IMO it's easier to understand the flow with nested subtests rather than sibling subtest which are de-factor isolated. Also, assuming we don't understand the entire flow is wrong.


> Mainly when there is some additional mocking that depends on the previous tests. 

this is the benefit of nested subtests: they reuse the results from the previous (parent) tests as input.

> There are some cases when we use sub-tests to verify the second or the third reconcile loop of the same controller because it is just much easier to reuse the already existing state, but in these cases, we don't mock nor simulate the k8s nor other operator functionality as part of the setup.

Also, it's more complicated to understand the purpose of some tests when they are isolated. 
For example, can you explain the purpose of this test when it's not a subtest (ie, part of the flow):

```
t.Run("when NSTemplateSet is in terminating state", func(t *testing.T) {
			// given another reconcile loop while the NSTemplateSet is *terminating* (ie, user namespaces are being deleted)
			nsTmplSet := nsTmplSet.DeepCopy()
			nsTmplSet.Status.Conditions = []toolchainv1alpha1.Condition{
				nstemplatetsettest.Terminating(),
			}
			nsTmplSet.Finalizers = []string{"kubernetes"}
			hostClient := test.NewFakeClient(t, s, basicTier)
			memberClient := test.NewFakeClient(t, nsTmplSet)
			member := NewMemberClusterWithClient(memberClient, "member-1", corev1.ConditionTrue)
			ctrl := newReconciler(hostClient, member)

			// when
			res, err := ctrl.Reconcile(context.TODO(), requestFor(s))

			// then
			require.NoError(t, err)
			assert.Equal(t, reconcile.Result{Requeue: false}, res) // no need to explicitly requeue while the NSTemplate is terminating
			// no changes
			spacetest.AssertThatSpace(t, s.Namespace, s.Name, hostClient).
				Exists().
				HasFinalizer().
				HasStatusTargetCluster("member-1").
				HasConditions(spacetest.Terminating())
			nstemplatetsettest.AssertThatNSTemplateSet(t, test.MemberOperatorNs, "oddity", member.Client).
				Exists().
				HasDeletionTimestamp().
				HasConditions(nstemplatetsettest.Terminating())
		})
```

MatousJobanek: > I do not believe it makes it easier to understand than the subtests, especially since you need to "deepcopy" some resources in some cases.

The `deepCopy` is used in several other cases - it's just for the sake of reusing the same setup object + doing some small modification there. The main point is that it's reusing the same object with the initial state, not reusing the FakeClient and modifying the resource that is in some state there.

> yes, we could use EnvTest framework, but it's not a requirement

my main point was that we want to write unit tests, not integration tests!

> IMO it's easier to understand the flow with nested subtests rather than sibling subtest which are de-factor isolated. 

I tried that - I started changing the original format, but I failed several times. Rewriting it was much easier, otherwise, I was afraid that I could introduce some bugs/flakiness. 

> Also, assuming we don't understand the entire flow is wrong.

When I'm looking at one unit test I'm thinking about the single functionality that is being tested - what happens when there is this input and what is the expected result. I don't want to think about the whole flow what happens when one sub-test contains some mocking mechanism for k8s/member-operator and what happens then it is removed or modified in the following subtest, etc... 

> this is the benefit of nested subtests: they reuse the results from the previous (parent) tests as input.

in most of the cases, we reuse the setup, not the result. As I mentioned before - there are some cases when we do that, but that are usually special cases when it makes the code easier/shorter.

> For example, can you explain the purpose of this test when it's not a subtest (ie, part of the flow):

that's easy - what happens when the NSTemplateSet is:
* being deleted
* and has a condition set to `Terminating`

xcoulon: > my main point was that we want to write unit tests, not integration tests!

I would rather say that we want to test without running/deploying on a cluster, but frankly I don't see why it's wrong to write test that match with the real flow?

anyways, let's stop arguing on these changes, we just have different views.

xcoulon (approved):

PR #604, opened 2022-03-02
update watch handler to map host namespace in reconcile request 

ranakan19:
currently while watching NSTemplateSet in all member clusters from Space Controller, the event handler used was 'EnqueueRequestForObject' which would set the Namespace as the member-operator's namespace in the reconcile request. This PR updates the watch handler to use a mapper to map the host-operator's namespace in the reconcile request for the watcher.

This was also causing a delay in cleanup - space would take longer than timeout at times to delete and fail. Made minor changes in e2e test to identify the failing object in logs here - https://github.com/codeready-toolchain/toolchain-e2e/pull/489

### DIFF
--- a/controllers/space/mapper.go
+++ b/controllers/space/mapper.go
@@ -8,6 +8,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
+
 var mapperLog = ctrl.Log.WithName("NSTemplateSetToSpaceMapper")
 
 func MapNSTemplateSetToSpace(hostNamespace string) func(object client.Object) []reconcile.Request {
--- a/controllers/space/mapper.go
+++ b/controllers/space/mapper.go
@@ -1,0 +1,15 @@
+package space
+
+import (
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


controllers/space/mapper.go
```diff
@@ -0,0 +1,28 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var mapperLog = ctrl.Log.WithName("NSTemplateSetToSpaceMapper")
+
+func MapNSTemplateSetToSpace(hostNamespace string) func(object client.Object) []reconcile.Request {
+	return func(obj client.Object) []reconcile.Request {
+		if nsTemplateSet, ok := obj.(*toolchainv1alpha1.NSTemplateSet); ok {
```
MatousJobanek, line 16: We don't have to cast the object and check if it's really `NSTemplateSet` - the watcher ensures that the mapper is triggered only for NSTemplateSets, but that's probably just a detail

MatousJobanek (approved):
Good catch @ranakan19!!!

controllers/space/mapper.go
MatousJobanek: fair enough :+1:

PR #607, opened 2022-03-08
feat: add optional DBaaSTenant resource

MatousJobanek:
Adds two DBaaSTenant cluster-scoped resources for RHODA operator. Both of them contain the `toolchain.dev.openshift.com/optional-resource` annotation to mark them as optional. The content of the annotation doesn't matter as we check only the presence of the annotation.
There won't be any e2e tests because the resources won't be created in e2e environment - RHODA operator is not installed there so it's missing the needed API group.

Depends on: https://github.com/codeready-toolchain/member-operator/pull/345

### DIFF
--- a/deploy/templates/nstemplatetiers/base/cluster.yaml
+++ b/deploy/templates/nstemplatetiers/base/cluster.yaml
@@ -164,6 +164,22 @@ objects:
     name: ${USERNAME}-stage
   spec:
     timeoutSeconds: ${{IDLER_TIMEOUT_SECONDS}}
+- apiVersion: dbaas.redhat.com/v1alpha1
+  kind: DBaaSTenant
+  metadata:
+    annotations:
+      toolchain.dev.openshift.com/optional-resource: "yes"
+    name: "tenant-${USERNAME}-stage"
+  spec:
+    inventoryNamespace: ${USERNAME}-stage
+- apiVersion: dbaas.redhat.com/v1alpha1
+  kind: DBaaSTenant
+  metadata:
+    annotations:
+      toolchain.dev.openshift.com/optional-resource: "yes"
+    name: "tenant-${USERNAME}-dev"
+  spec:
+    inventoryNamespace: ${USERNAME}-dev
 parameters:
 - name: USERNAME
   required: true


deploy/templates/nstemplatetiers/base/cluster.yaml
```diff
@@ -164,6 +164,22 @@ objects:
     name: ${USERNAME}-stage
   spec:
     timeoutSeconds: ${{IDLER_TIMEOUT_SECONDS}}
+- apiVersion: dbaas.redhat.com/v1alpha1
+  kind: DBaaSTenant
+  metadata:
+    annotations:
+      toolchain.dev.openshift.com/optional-resource: "yes"
```
xcoulon, line 171:
```suggestion
      toolchain.dev.openshift.com/optional-resource: true
```

deploy/templates/nstemplatetiers/base/cluster.yaml
```diff
@@ -164,6 +164,22 @@ objects:
...
+    name: "tenant-${USERNAME}-stage"
+  spec:
+    inventoryNamespace: ${USERNAME}-stage
+- apiVersion: dbaas.redhat.com/v1alpha1
+  kind: DBaaSTenant
+  metadata:
+    annotations:
+      toolchain.dev.openshift.com/optional-resource: "yes"
```
xcoulon, line 179:
```suggestion
      toolchain.dev.openshift.com/optional-resource: true
```

deploy/templates/nstemplatetiers/base/cluster.yaml
```diff
@@ -164,6 +164,22 @@ objects:
...
   spec:
     timeoutSeconds: ${{IDLER_TIMEOUT_SECONDS}}
+- apiVersion: dbaas.redhat.com/v1alpha1
+  kind: DBaaSTenant
+  metadata:
+    annotations:
+      toolchain.dev.openshift.com/optional-resource: "yes"
+    name: "tenant-${USERNAME}-stage"
+  spec:
+    inventoryNamespace: ${USERNAME}-stage
```
xcoulon, line 174: might not be needed once if we drop the `-stage` namespaces before merging this PR

xcoulon (commented):
should we have changes in e2e to make sure that these new resources are created?

btw, I'm surprised that e2e tests still pass: do we already install the DBaaS operator before running the e2e tests?

deploy/templates/nstemplatetiers/base/cluster.yaml
MatousJobanek: This PR has to be merged today and dropping the namespace will be done in the following days

deploy/templates/nstemplatetiers/base/cluster.yaml
MatousJobanek: I had exactly this at the very beginning, but the annotation wasn't properly processed, because it didn't contain a string but a boolean. 
So, I could have used:
```
toolchain.dev.openshift.com/optional-resource: "true"
```
or even
```
toolchain.dev.openshift.com/optional-resource: 'true'
```
but I wasn't sure if the apostrophes wouldn't be removed, so I chose using `"yes"`

deploy/templates/nstemplatetiers/base/cluster.yaml
xcoulon: ok, that was just a remark in case the ns drop would happen earlier

xcoulon (approved):

PR #613, opened 2022-03-11
feat: upgrade GH Actions env to Ubuntu 20.04

MatousJobanek:


.github/workflows/publish-operators-for-e2e-tests.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         python-version: '3.x'
 
     - name: Prepare tools
-      uses: codeready-toolchain/toolchain-cicd/prepare-tools-action@master
+      uses: codeready-toolchain/matousjobanek/prepare-tools-action@opm-v1.19.5
```
alexeykazakov, line 67: You don't need it, right?

alexeykazakov (approved):

.github/workflows/publish-operators-for-e2e-tests.yml
MatousJobanek: this is for testing this https://github.com/codeready-toolchain/toolchain-cicd/pull/55 it wasn't compatible with the older version of ubuntu

MatousJobanek: The GH actions worklfow passed, so the compatibility with the latest OPM is solved

PR #614, opened 2022-03-14
Drop counts for objects defined by CRDs from user quotas

alexeykazakov:
It looks like the CR object counts are not supported in 4.10. It was undocumented in 4.9 (see table 7.3 in https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/building_applications/quotas#quotas-resources-managed_quotas-setting-per-project) but worked. It doesn't work in 4.10 anymore.

See https://issues.redhat.com/browse/OHSS-10763 for more details.

E2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/506

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/cluster.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/cluster.yaml
@@ -125,33 +125,6 @@ objects:
       annotations:
         openshift.io/requester: ${USERNAME}
       labels: null
-- apiVersion: quota.openshift.io/v1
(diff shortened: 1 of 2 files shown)


ranakan19 (approved):
LGTM 
If i understand correctly, as of now we are only dropping the count, and look for an alternative that works for 4.10 later

alexeykazakov:
> If i understand correctly, as of now we are only dropping the count, and look for an alternative that works for 4.10 later

Yes.

PR #618, opened 2022-03-31
Remove TemplateUpdateRequests

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1441


controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -112,197 +83,21 @@ func (r *Reconciler) ensureStatusUpdateRecord(logger logr.Logger, tier *toolchai
...
-	// reset the `FailedAccounts` in the previous update as we don't want to retain the usernames
-	// for whom the update failed previously (no need to carry such data anymore)
-	tier.Status.Updates[len(tier.Status.Updates)-1].FailedAccounts = nil
 	logger.Info("Adding a new entry in tier.status.updates")
+	return true, r.addNewTierUpdate(tier, hash)
+}
+
+func (r *Reconciler) addNewTierUpdate(tier *toolchainv1alpha1.NSTemplateTier, hash string) error {
 	tier.Status.Updates = append(tier.Status.Updates, toolchainv1alpha1.NSTemplateTierHistory{
 		StartTime: metav1.Now(),
```
MatousJobanek, line 99: long-term, it would be fine to rename for example to "UpdateTime", but it's fine as it is for now

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -112,197 +83,21 @@ func (r *Reconciler) ensureStatusUpdateRecord(logger logr.Logger, tier *toolchai
...
 		Hash:      hash,
 	})
```
MatousJobanek, line 101: I would love to change the order of the history so the latest is the first:
```suggestion
	tier.Status.Updates = append([]toolchainv1alpha1.NSTemplateTierHistory{{
		StartTime: metav1.Now(),
		Hash:      hash,
	}}, tier.Status.Updates...)
```
but it's not urgent - it can be easily done in the next pr as a separate task.

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -56,28 +52,20 @@ func TestReconcile(t *testing.T) {
...
 				toolchainv1alpha1.NSTemplateTierHistory{
-					StartTime:      metav1.Now(),
-					CompletionTime: &now,
-					Failures:       2,
-					FailedAccounts: []string{"failed1", "failed2"},
+					StartTime: metav1.Now(),
+					Hash:      "def456",
 				}))
 			initObjs := []runtime.Object{basicTier}
 			r, req, cl := prepareReconcile(t, basicTier.Name, initObjs...)
```
MatousJobanek, line 71: take a look at the content of the `prepareReconcile` function - this can be simplified (shortened) as well

MatousJobanek (commented):
Nice cleanup :+1:

MatousJobanek (approved):
Wanted to approve it :-)

controllers/nstemplatetier/nstemplatetier_controller_test.go
rajivnathan: Good catch, done in 0ce848e2b13fe954c116bdf3f4945890a3a6803a

controllers/nstemplatetier/nstemplatetier_controller.go
xcoulon: if we reverse the order from now on, should we also reorder the already-existing status updates?

xcoulon (approved):

controllers/nstemplatetier/nstemplatetier_controller.go
MatousJobanek: I would rather wipe them out as they are not relevant anymore anyway :-) 
But as I mentioned in the comment - it can be done as a separate task. I'm afraid that it would maybe require some changes also in e2e tests

xcoulon: sure, addressing this in a separate PR (or set of PRs) makes sense. And if we can wipe out older entries, it's even simpler ;) 👍

PR #619, opened 2022-04-04
fix: prefix 'sa-read' rolebinding with 'appstudio-'

xcoulon:
This particular rolebinding applies to `system:serviceaccounts:${MEMBER_OPERATOR_NAMESPACE`, so it doesn't need to contain any `$USERNAME`, but instead, it should be prefixed with `appstudio-`, as all sibling rolebindings in this template.

see https://github.com/codeready-toolchain/toolchain-e2e/pull/519


deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
```diff
@@ -131,7 +131,7 @@ objects:
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: RoleBinding
   metadata:
-    name: ${USERNAME}-sa-read
+    name: appstudio-sa-read
```
alexeykazakov, line 134:
```suggestion
    name: member-operator-sa-read
```

alexeykazakov (commented):
I would rather call it `member-operator-sa-read` since this one binds all SAs in the member-operator namespace to the `toolchain-sa-read` role.

deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
xcoulon: I thought we would use `appstudio-` for all role bindings related to AppStudio, but I don't have any strong opinion on the names or prefixes.

xcoulon: ok, got it. As I said, I'm fine with both options ;)

alexeykazakov (approved):

MatousJobanek (approved):

PR #624, opened 2022-04-06
feat: SocialEvent controller

xcoulon:
controller checks that the TierName matches an NSTemplateTier

also: 
- skip linter on cyclomatic complexity
- rename `controllers/usersignup/status.go` -> `controllers/usersignup/status_updater.go` (and same for `_test.go` companion tests)

api: https://github.com/codeready-toolchain/api/pull/313
e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/535

Fixes https://issues.redhat.com/browse/CRT-1304


config/samples/kustomization.yaml
```diff
@@ -2,5 +2,6 @@
 resources:
 - toolchain.dev.openshift.com_v1alpha1_usersignup.yaml
 - toolchain.dev.openshift.com_v1alpha1_space.yaml
+- toolchain.dev.openshift.com_v1alpha1_socialevent.yaml
```
xcoulon, line 5: that came from operator-sdk, but I'm not sure we need this change

config/samples/toolchain.dev.openshift.com_v1alpha1_socialevent.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: toolchain.dev.openshift.com/v1alpha1
```
xcoulon, line 1: this also came from operator-sdk, but I'm not sure we need it

MatousJobanek: You can drop it. 
Technically speaking - we should have the sample for each of our CRD, but let's be practical here - the samples are not supposed to be directly used by anyone, really. As soon as we have to add them (because of any reason) then we can add them, but for now, let's make our lives and the maintenance easier :-)

xcoulon: yeah, that was my feeling, too ;)

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,92 @@
...
+func (r *Reconciler) checkTier(logger logr.Logger, event *toolchainv1alpha1.SocialEvent) error {
+	// fetch the NSTemplateTier tier specified in the SocialEvent
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: event.Namespace,
+		Name:      event.Spec.Tier,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("NSTemplateTier not found", "nstemplatetier_name", event.Spec.Tier)
+			event.Status.Conditions = []toolchainv1alpha1.Condition{
```
alexeykazakov, line 71: Let's not override all the conditions. Use https://github.com/codeready-toolchain/toolchain-common/blob/b7d36013dc0224b65ce8c020dd847ab9d9351206/pkg/condition/condition.go#L14 instead.
It's much safer to update the specific condition then resetting all the conditions which might be set in a different place in the controller.

And here is an example of how to use this function to update the status only if it changed: https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/status.go#L373:25

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,92 @@
...
+					Message: fmt.Sprintf("NSTemplateTier '%s' not found", event.Spec.Tier),
+				},
+			}
+			return r.Client.Status().Update(context.TODO(), event)
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the NSTemplateTier", "nstemplatetier_name", event.Spec.Tier)
+		return errs.Wrapf(err, "unable to get the '%s' NSTemplateTier", event.Spec.Tier)
+	}
+	event.Status.Conditions = []toolchainv1alpha1.Condition{
```
alexeykazakov, line 85: Same here.

config/samples/toolchain.dev.openshift.com_v1alpha1_socialevent.yaml
xcoulon: removed in 8a32d92

controllers/socialevent/socialevent_controller.go
xcoulon: thanks for the pointers! 
done in 1c8032c

controllers/socialevent/socialevent_controller.go
xcoulon: done in done in 1c8032c too ;)

config/samples/toolchain.dev.openshift.com_v1alpha1_socialevent.yaml
xcoulon: removed in https://github.com/codeready-toolchain/host-operator/pull/624/commits/8a32d924810cca8098cff6c76cf003c9bf2153db

config/samples/kustomization.yaml
xcoulon: removed in https://github.com/codeready-toolchain/host-operator/commit/8a32d924810cca8098cff6c76cf003c9bf2153db

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,87 @@
...
+			})
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the NSTemplateTier", "nstemplatetier_name", event.Spec.Tier)
+		return errs.Wrapf(err, "unable to get the '%s' NSTemplateTier", event.Spec.Tier)
+	}
+	return r.StatusUpdater.updateStatusConditions(event, toolchainv1alpha1.Condition{
+		Type:   toolchainv1alpha1.ConditionReady,
+		Status: corev1.ConditionTrue,
+	})
```
MatousJobanek, line 86: shouldn't we also:
* watch the UserSignups with the associated activation code
* count them
* report the number of UserSignups that used the code in `SocialEvent.Status.ActivationCount`

?

controllers/socialevent/socialevent_controller_test.go
```diff
@@ -0,0 +1,120 @@
...
+		se := socialeventtest.NewSocialEvent("lab", "basic")
+		hostClient := test.NewFakeClient(t, se, basicTier)
+		ctrl := newReconciler(hostClient)
+
+		// when
+		res, err := ctrl.Reconcile(context.TODO(), requestFor(se))
+
+		// then
+		require.NoError(t, err)
+		assert.False(t, res.Requeue)
```
MatousJobanek, line 44: the `Requeue` should be false in all cases if I'm not mistaken. We don't verify that in such cases, so you could skip it here as well.

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,87 @@
...
+			return r.StatusUpdater.updateStatusConditions(event, toolchainv1alpha1.Condition{
+				Type:    toolchainv1alpha1.ConditionReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  toolchainv1alpha1.SocialEventInvalidTierReason,
+				Message: fmt.Sprintf("NSTemplateTier '%s' not found", event.Spec.Tier),
+			})
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the NSTemplateTier", "nstemplatetier_name", event.Spec.Tier)
+		return errs.Wrapf(err, "unable to get the '%s' NSTemplateTier", event.Spec.Tier)
```
MatousJobanek, line 81: Why don't we set the "failed" condition for this error as well?

xcoulon: because we're going to requeue the request, like we do for other resources in the same case

controllers/socialevent/socialevent_controller_test.go
xcoulon: it's not a big deal, is it?

xcoulon: removed in c0aefb2 and 026e91c

controllers/socialevent/socialevent_controller.go
xcoulon: done in https://github.com/codeready-toolchain/host-operator/pull/624/commits/7b901ca9244cef0deb97c853d2ea1b9d8ec6756f: the status will be updated with the new `UnableToGetTier` reason introduced in [api#311](https://github.com/codeready-toolchain/api/pull/311/files#diff-21b047e72efa2ef881ad05b5b46f4fa54d832b1d43040f5282632b99527bf400R11), and the error will be returned by the controller, which should trigger another reconcile loop (as it was the case before)

controllers/socialevent/socialevent_controller.go
xcoulon: Just to clarify: this is the first version of the SocialEvent controller, and it covers the case of the creation of a SocialEvent. Of course, all the cases you mentioned above will be implemented, but in subsequent PRs.

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,75 @@
...
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) checkTier(logger logr.Logger, event *toolchainv1alpha1.SocialEvent) error {
+	// fetch the NSTemplateTier tier specified in the SocialEvent
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{
```
MatousJobanek, line 64: we will need to update this part (as well as the SocialEvent CRD) to include also the UserTier information

MatousJobanek (approved):

controllers/socialevent/socialevent_controller.go
xcoulon: yes, indeed!

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,75 @@
...
+		Namespace: event.Namespace,
+		Name:      event.Spec.Tier,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			return r.StatusUpdater.tierNotFound(logger, event)
```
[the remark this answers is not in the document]
xcoulon: good point, let me check that

controllers/socialevent/status_updater.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+}
+
+func (u *StatusUpdater) userTierNotFound(logger logr.Logger, event *toolchainv1alpha1.SocialEvent) error {
+	logger.Info("NSTemplateTier not found", "nstemplatetier_name", event.Spec.UserTier)
+	return u.updateStatusConditions(event, toolchainv1alpha1.Condition{
+		Type:    toolchainv1alpha1.ConditionReady,
+		Status:  corev1.ConditionFalse,
+		Reason:  toolchainv1alpha1.SocialEventInvalidUserTierReason,
+		Message: fmt.Sprintf("NSTemplateTier '%s' not found", event.Spec.UserTier),
```
rajivnathan, line 33:
```suggestion
		Message: fmt.Sprintf("UserTier '%s' not found", event.Spec.UserTier),
```

controllers/socialevent/status_updater.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (u *StatusUpdater) ready(event *toolchainv1alpha1.SocialEvent) error {
+	return u.updateStatusConditions(event, toolchainv1alpha1.Condition{
+		Type:   toolchainv1alpha1.ConditionReady,
+		Status: corev1.ConditionTrue,
+	})
+}
+
+func (u *StatusUpdater) userTierNotFound(logger logr.Logger, event *toolchainv1alpha1.SocialEvent) error {
+	logger.Info("NSTemplateTier not found", "nstemplatetier_name", event.Spec.UserTier)
```
rajivnathan, line 28:
```suggestion
	logger.Info("UserTier not found", "usertier_name", event.Spec.UserTier)
```

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,86 @@
...
+	// fetch the SocialEvent resource
+	event := &toolchainv1alpha1.SocialEvent{}
+	if err := r.Client.Get(context.TODO(), request.NamespacedName, event); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("SocialEvent not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current SocialEvent")
+		return reconcile.Result{}, errs.Wrap(err, "unable to get the current NSTeSocialEventplateTier")
```
rajivnathan, line 51:
```suggestion
		return reconcile.Result{}, errs.Wrap(err, "unable to get the current SocialEvent")
```

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,86 @@
...
+		For(&toolchainv1alpha1.SocialEvent{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=socialevents,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=socialevents/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=socialevents/finalizers,verbs=update
+
+// Reconcile takes care of:
+// - checking that the target tier specifiec in the SocialEvent is valid and set the status condition accordingly
```
rajivnathan, line 38:
```suggestion
// - checking that the target User and Space tiers specified in the SocialEvent are valid and set the status condition accordingly
```

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (r *Reconciler) checkTier(logger logr.Logger, event *toolchainv1alpha1.SocialEvent) error {
+	// fetch the NSTemplateTier tier specified in the SocialEvent
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: event.Namespace,
+		Name:      event.Spec.UserTier,
+	}, tier); err != nil {
```
rajivnathan, line 68:
```suggestion
	// fetch the UserTier tier specified in the SocialEvent
	userTier := &toolchainv1alpha1.UserTier{}

	if err := r.Client.Get(context.TODO(), types.NamespacedName{
		Namespace: event.Namespace,
		Name:      event.Spec.UserTier,
	}, userTier); err != nil {
```

controllers/socialevent/socialevent_controller.go
```diff
@@ -0,0 +1,86 @@
...
+		if errors.IsNotFound(err) {
+			return r.StatusUpdater.userTierNotFound(logger, event)
+		}
+		// Error reading the object - requeue the request.
+		return r.StatusUpdater.unableToGetUserTier(logger, event, err)
+	}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: event.Namespace,
+		Name:      event.Spec.SpaceTier,
+	}, tier); err != nil {
```
rajivnathan, line 78:
```suggestion
	// fetch the NSTemplateTier tier specified in the SocialEvent
	nsTmplTier := &toolchainv1alpha1.NSTemplateTier{}
	if err := r.Client.Get(context.TODO(), types.NamespacedName{
		Namespace: event.Namespace,
		Name:      event.Spec.SpaceTier,
	}, nsTmplTier); err != nil {
```

test/nstemplatetier/nstemplatetier.go
```diff
@@ -98,6 +98,26 @@ var CurrentBasicTemplates = toolchainv1alpha1.NSTemplateTierSpec{
 	},
 }
 
+// BaseTier returns a "base" NSTemplateTier with template refs in the given spec
+func BaseTier(t *testing.T, spec toolchainv1alpha1.NSTemplateTierSpec, options ...TierOption) *toolchainv1alpha1.NSTemplateTier {
```
rajivnathan, line 102: Is there any difference between this `BaseTier` and the `BasicTier` from below? If not, then maybe we can use `BasicTier` instead? If your intent was to update our code to use `BaseTier` then I guess we can rename the func in a separate PR?

rajivnathan (commented):
Looks clean and straightforward 👍  Just found a few instances where I think we're meant to update them to use UserTier instead of NSTemplateTier

controllers/socialevent/socialevent_controller.go
xcoulon: ah my bad, I did not realize it was a different type! 🤦‍♂️

xcoulon: fixed in 42926ad

controllers/socialevent/status_updater.go
xcoulon: fixed in 42926ad

test/nstemplatetier/nstemplatetier.go
xcoulon: we don't have such `basic` tier so I thought that `base` would be more appropriate. But ok, let me use `BasicTier ` for now and I'll change to `BaseTier` in a separate PR

xcoulon: done in 85579a0

MatousJobanek (approved):

PR #628, opened 2022-04-14
Migrate deactivated UserSignups to use new username-based naming

sbryzak:
This PR will migrate deactivated UserSignups to use the new username-based naming convention, rather than using the User ID which comes from the `sub` claim of the user's bearer token.

Fixes: https://issues.redhat.com/browse/CRT-1402

Currently a work in progress until comprehensive tests have been added.


controllers/usersignup/usersignup_controller.go
```diff
@@ -112,6 +118,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			// If the annotation exists however the original UserSignup isn't found, then remove the annotation
+			if errors.IsNotFound(err) {
+				// Remove the annotation
+				delete(userSignup.Annotations, migrationAnnotationName)
+				// Update the UserSignup
+				err = r.Client.Update(ctx, userSignup)
+				if err != nil {
+					return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup,
+						r.setStatusMigrationFailedCleanup, err, fmt.Sprintf("Failed to remove migration annotation"))
+				}
```
alexeykazakov, line 147: We should either return here after successfully deleting the annotation (but you would need to update the predicate to watch the migration annotation too: https://github.com/codeready-toolchain/host-operator/blob/95d7140dd246549775d077cfd1e1b039f6678187/controllers/usersignup/predicate.go#L24-L41) or skip the rest of the migration code (do not try to delete the original usersignup below) and proceed with the reconcile.

controllers/usersignup/usersignup_controller.go
```diff
@@ -112,6 +118,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+		// Delete the original UserSignup
+		err = r.Client.Delete(context.TODO(), userSignupToDelete)
+		if err != nil {
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup, r.setStatusMigrationFailedCleanup,
+				err, fmt.Sprintf("Failed to remove original UserSignup [%s]",
+					userSignup.Annotations[migrationAnnotationName]))
+		}
+		// Requeue so that the annotation will now be removed also
+		return reconcile.Result{Requeue: true}, nil
```
alexeykazakov, line 163: I wonder if we want to add some delay (less than a second?) here...

controllers/usersignup/usersignup_controller.go
```diff
@@ -112,6 +118,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+				}
+			} else {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup,
+					r.setStatusMigrationFailedLookup, err, fmt.Sprintf("Failed to lookup original UserSignup [%s]",
+						userSignup.Annotations[migrationAnnotationName]))
+			}
+		}
+
+		// Delete the original UserSignup
+		err = r.Client.Delete(context.TODO(), userSignupToDelete)
```
alexeykazakov, line 156: Before deleting any usersignup let's check that it's deactivated.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +251,97 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
 		return reconcile.Result{}, r.updateStatus(logger, userSignup, r.setStatusDeactivated)
 	}
 
 	return reconcile.Result{}, r.ensureNewMurIfApproved(logger, config, userSignup)
 }
 
+// EncodeUserIdentifier is a temporary function used by the migration procedure
+// REMOVE AFTER MIGRATION IS COMPLETE
+func EncodeUserIdentifier(subject string) string {
```
alexeykazakov, line 316: We already have this function somewhere, right? We use it when creating a usersignup I suppose... Is it just a copy of that function (in case you don't want to move it from reg-service to common for a temporal migration step or something like this)?

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +251,97 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+					// We need to:
+					//
+					// 1) Copy the existing UserSignup
+					// 2) Override the new UserSignup resource name with the encoded username
+					// 3) Set the starting status to Deactivated (technically we could let the reconciler function do this
+					//    when it reconciles the migrated UserSignup, but it shouldn't hurt to set this status up front)
+					migratedUserSignup = userSignup.DeepCopy()
+					migratedUserSignup.Name = encodedUsername
+					r.setStatusDeactivated(migratedUserSignup, "")
+					err = r.Client.Create(ctx, migratedUserSignup)
```
alexeykazakov, line 295: Do we also want to cleanup the rest of the meta which we do not set for new resources, like the resource ID, version, generation, etc? Or kube ignores them when creating the resource?

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +251,97 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+					if err != nil {
+						// If there was an error creating the migrated UserSignup, then set the status and requeue
+						return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup,
+							r.setStatusMigrationFailedCreate, err, "Failed to migrate UserSignup")
+					}
```
alexeykazakov, line 300: If no error, then:
- set the status to Deactivated.
- set the migration annotation and return. But you need to update the predicate to watch the migration annotation: https://github.com/codeready-toolchain/host-operator/blob/95d7140dd246549775d077cfd1e1b039f6678187/controllers/usersignup/predicate.go#L24-L41

alexeykazakov (commented):
Overall looks good but I have a few comments to address.

controllers/usersignup/usersignup_controller.go
sbryzak: Restructured this so that the delete occurs up front

controllers/usersignup/usersignup_controller.go
sbryzak: I think a 1 second delay is acceptable, added.

controllers/usersignup/usersignup_controller.go
sbryzak: We do have it in registration service, I was tempted to move it to toolchain-common however after the migration is complete the only place it will still be used will only be registration service again which is why I simply made a temporary copy of it.

controllers/usersignup/usersignup_controller.go
sbryzak: Fixed this with a slight restructure

controllers/usersignup/usersignup_controller.go
sbryzak: Done, I wasn't certain what to do here exactly in the case that it wasn't deactivated, but I think that setting the migration failed status on the new UserSignup and returning an error is probably reasonable.

controllers/usersignup/usersignup_controller.go
sbryzak: I'll investigate deeper into whether this is required when I start to debug the unit tests.

sbryzak: Looks like I need to at least reset some of these values.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}, userSignupToDelete)
+	if err == nil {
+		// Delete the original UserSignup, if it has finished deactivating
+		cond, found := condition.FindConditionByType(userSignupToDelete.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
+		if !found || cond.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason {
+			// The UserSignup to delete isn't finished deactivating yet, return an error
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup, r.setStatusMigrationFailedCleanup,
+				errs.New("Original UserSignup not deactivated"),
+				fmt.Sprintf("Original UserSignup [%s] not yet deactivated", userSignupToDelete.Name))
+		}
```
MatousJobanek, line 329: wouldn't make more sense to check the migration condition - that it's true because it successfully created the new UserSignup resource?

controllers/usersignup/status.go
```diff
@@ -182,6 +182,43 @@ func (u *StatusUpdater) setStatusDeactivated(userSignup *toolchainv1alpha1.UserS
 		})
 }
 
+const (
+	UserMigrationFailed toolchainv1alpha1.ConditionType = "UserMigrationFailed"
```
MatousJobanek, line 186: I should have commented that as the first thing - the condition should have more states than just "failed" ones, so we can fully monitor the process.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+		err = r.Client.Delete(context.TODO(), userSignupToDelete)
+		if err != nil {
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup, r.setStatusMigrationFailedCleanup,
+				err, fmt.Sprintf("Failed to remove original UserSignup [%s]",
+					userSignup.Annotations[migrationAnnotationName]))
+		}
+		// Requeue so that the annotation will now be removed also
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second}, nil
```
MatousJobanek, line 338: could you also set the "migration" condition to true to make it clear that the old UserSignup has been successfully deleted?

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+				}
+			}
+			userSignup.Status.Conditions = conditions
+
+			// Update the UserSignup
+			err = r.Client.Update(context.TODO(), userSignup)
+			if err != nil {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup,
+					r.setStatusMigrationFailedCleanup, err, "Failed to remove migration annotation")
+			}
```
MatousJobanek, line 362: btw, do we need to remove the annotation and the condition at all? would it be a big deal if it stays there?
I would keep it there - can be useful for monitoring purposes - to see what has been migrated and what has not.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -3952,3 +3952,213 @@ func TestUserSignupLastTargetClusterAnnotation(t *testing.T) {
...
+
+	var migrated *toolchainv1alpha1.UserSignup
+
+	for _, us := range userSignups.Items {
+		if us.Name == "1cf93821-fooredhatcom" {
+			value := us
+			migrated = &value
+			break
+		}
+	}
```
MatousJobanek, line 3987:
```suggestion
	migrated := AssertThatUserSignup(t, test.HostOperatorNs, "1cf93821-fooredhatcom", cl).Get()
```

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -3952,3 +3952,213 @@ func TestUserSignupLastTargetClusterAnnotation(t *testing.T) {
...
+
+	require.NotNil(t, migrated)
+	require.Equal(t, userSignup.Spec.Username, migrated.Spec.Username)
+	require.Equal(t, userSignup.Spec.Userid, migrated.Spec.Userid)
+	require.Equal(t, userSignup.Spec.OriginalSub, migrated.Spec.OriginalSub)
+	require.Equal(t, userSignup.Spec.Company, migrated.Spec.Company)
+	require.Equal(t, userSignup.Spec.FamilyName, migrated.Spec.FamilyName)
+	require.Equal(t, userSignup.Spec.GivenName, migrated.Spec.GivenName)
```
MatousJobanek, line 3995: why not:
```suggestion
	require.Equal(t, userSignup.Spec, migrated.Spec)
```

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -3952,3 +3952,213 @@ func TestUserSignupLastTargetClusterAnnotation(t *testing.T) {
...
+		Namespace: userSignup.Namespace,
+		Name:      userSignup.Name}, userSignup)
+	require.NoError(t, err)
+
+	// Confirm that the original has had its Deactivated condition set
+	require.True(t, condition.HasConditionReason(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete,
+		toolchainv1alpha1.UserSignupUserDeactivatedReason))
+
+	// Confirm that the migrated still has its migration annotation
+	require.Contains(t, migrated.Annotations, "toolchain.dev.openshift.com/migration-replaces")
```
MatousJobanek, line 4009: please verify that also the already existing annotations plus the value of this "migration-replaces" annotation.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			migratedUserSignup.ObjectMeta.Generation = 0
+			migratedUserSignup.ObjectMeta.CreationTimestamp = v1.Time{}
+			migratedUserSignup.Name = encodedUsername
+			migratedUserSignup.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(migratedUserSignup.Status.Conditions,
+				toolchainv1alpha1.Condition{
+					Type:    toolchainv1alpha1.UserSignupComplete,
+					Status:  corev1.ConditionTrue,
+					Reason:  toolchainv1alpha1.UserSignupUserDeactivatedReason,
+					Message: "",
+				})
```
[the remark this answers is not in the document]
sbryzak: No problem, modified to create a new object instead.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			migratedUserSignup.Annotations[migrationAnnotationName] = userSignup.Name
```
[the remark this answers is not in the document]
sbryzak: Done

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			err = r.Client.Create(context.TODO(), migratedUserSignup)
```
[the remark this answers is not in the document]
sbryzak: Good point, I've set this status with a reason of "UserSignupMigrated" so that the notification won't be created again.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			if err != nil {
+				// If there was an error creating the migrated UserSignup, then set the status and requeue
+				return r.wrapErrorWithStatusUpdate(logger, userSignup,
+					r.setStatusMigrationFailedCreate, err, "Failed to migrate UserSignup")
```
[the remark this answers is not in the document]
sbryzak: Done

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 			}
 		}
+
+		// Migrate the user if necessary
+		err = r.migrateUserIfNecessary(userSignup, request, logger)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
 		return reconcile.Result{}, r.updateStatus(logger, userSignup, r.setStatusDeactivated)
```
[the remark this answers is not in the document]
sbryzak: Sure, we can do it that way.

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+// 2) Remove the migration-replaces annotation
+func (r *Reconciler) cleanupMigration(userSignup *toolchainv1alpha1.UserSignup, originalUserSignupName string,
+	request ctrl.Request, logger logr.Logger) (ctrl.Result, error) {
+
+	userSignupToDelete := &toolchainv1alpha1.UserSignup{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: request.Namespace,
+		Name:      originalUserSignupName,
+	}, userSignupToDelete)
+	if err == nil {
```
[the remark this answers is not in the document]
sbryzak: Done.

controllers/usersignup/usersignup_controller.go
sbryzak: I've got no big problem leaving it there.

controllers/usersignup/usersignup_controller_test.go
sbryzak: That's much nicer, done.

controllers/usersignup/usersignup_controller_test.go
sbryzak: Nice improvement, thanks.

controllers/usersignup/usersignup_controller_test.go
sbryzak: Done

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +217,191 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			migratedUserSignup.Annotations[migrationAnnotationName] = userSignup.Name
+			err = r.Client.Create(context.TODO(), migratedUserSignup)
+			if err != nil {
+				// If there was an error creating the migrated UserSignup, then set the status and requeue
+				return r.wrapErrorWithStatusUpdate(logger, userSignup,
+					r.setStatusMigrationFailedCreate, err, "Failed to migrate UserSignup")
+			}
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
sbryzak: I feel that this kind of status update is redundant.  If we wish to know whether the newly migrated UserSignup has been created, then why can't we just check whether it exists or not?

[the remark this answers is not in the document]
MatousJobanek: there are two reasons:

1. what if the creation of the UserSignup fails for whatever reason - in that case the condition would be set to:
```
"Failed to migrate UserSignup"
```
in the next reconcile the controller would successfully create the UserSignup, but the condition would still keep the old "failed" value, which is confusing.

2. you could use that condition in the next step of the migration as I mentioned in this comment https://github.com/codeready-toolchain/host-operator/pull/628#discussion_r853037579

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +218,229 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+				err, fmt.Sprintf("Failed to remove original UserSignup [%s]",
+					userSignup.Annotations[migrationAnnotationName]))
+		}
+		// Requeue so that the annotation will now be removed also
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Second}, nil
+	}
+
+	// If everything was cleaned up successfully, requeue one more time to invoke the standard reconciliation workflow
+	return reconcile.Result{Requeue: true}, nil
+}
```
MatousJobanek, line 411: you could make the code easier to read by moving the `return reconcile.Result{Requeue: true}, nil` to the line after deletion and the dropping the `if err == nil {` statement
```suggestion
			}
			// If everything was cleaned up successfully, requeue one more time to invoke the standard reconciliation workflow
			return reconcile.Result{Requeue: true}, nil
		} else {
			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup,
				r.setStatusMigrationFailedLookup, err, fmt.Sprintf("Failed to lookup original UserSignup [%s]",
					userSignup.Annotations[migrationAnnotationName]))
		}
	}

	// Delete the original UserSignup if it has finished migration
	if !condition.IsTrueWithReason(userSignupToDelete.Status.Conditions, toolchainv1alpha1.UserSignupComplete,
		toolchainv1alpha1.UserSignupUserDeactivatedReason) {
		// The UserSignup to delete isn't finished migrating yet, return an error
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup, r.setStatusMigrationFailedCleanup,
			errs.New("Original UserSignup not finished migration"),
			fmt.Sprintf("Original UserSignup [%s] not yet finished migration", userSignupToDelete.Name))

	}

	err = r.Client.Delete(context.TODO(), userSignupToDelete)
	if err != nil {
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, userSignup, r.setStatusMigrationFailedCleanup,
			err, fmt.Sprintf("Failed to remove original UserSignup [%s]",
				userSignup.Annotations[migrationAnnotationName]))
	}
	// Requeue so that the annotation will now be removed also
	return reconcile.Result{Requeue: true, RequeueAfter: time.Second}, nil
}
```

sbryzak: Done

controllers/usersignup/usersignup_controller.go
```diff
@@ -200,12 +218,229 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+					Status:  corev1.ConditionTrue,
+					Reason:  toolchainv1alpha1.UserSignupUserDeactivatedReason,
+					Message: "",
+				},
+				toolchainv1alpha1.Condition{
+					Type:    toolchainv1alpha1.UserSignupUserDeactivatedNotificationCreated,
+					Status:  corev1.ConditionTrue,
+					Reason:  "UserSignupMigrated",
+					Message: "",
+				})
```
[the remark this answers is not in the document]
sbryzak: I've refactored this somewhat to deal with setting the status separately.  While we do copy the existing conditions, we also need to ensure that certain condition values are set also.

controllers/usersignup/usersignup_controller.go
```diff
@@ -43,6 +50,12 @@ import (
 
 type StatusUpdaterFunc func(userAcc *toolchainv1alpha1.UserSignup, message string) error
 
+const (
+	migrationAnnotationName           = toolchainv1alpha1.LabelKeyPrefix + "migration-replaces"
+	migratedAnnotationName            = toolchainv1alpha1.LabelKeyPrefix + "migrated"
```
xcoulon, line 55: what is the purpose of this annotation?

controllers/usersignup/usersignup_controller.go
```diff
@@ -107,11 +120,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 	logger = logger.WithValues("username", userSignup.Spec.Username)
 
+	// Don't reconcile while migration is in progress
+	// TODO delete this after migration is complete
+	_, ok := userSignup.Annotations[migrationInProgressAnnotationName]
```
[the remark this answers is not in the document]
MatousJobanek: It shouldn't be blocked. 
There are two phases:
1. creation of a new UserSignup
2. deletion of the old one

the deletion can happen only when the `migrationInProgressAnnotationName` is not present.
The first phase consists of two steps
1. actual creation of the new UserSignup
2. update of the `UserSignup.status` and removal of the annotation `migrationInProgressAnnotationName`

if the second step fails (for example because of a conflict) then the controller returns an error and got reconciled again. in the next run it updates the status & removes the annotation. Since the UserSignup predicates checks also for the changes in annotation, then the controller should be triggered for the "new" UserSignup which doesn't have the annotation anymore.

This annotation and the check was introduced so the controller is not triggered before the "new" UserSignup has all needed status conditions set.

Correct me if I missed some detail

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
sbryzak: Refactored further to just set the migrated status.

controllers/usersignup/usersignup_controller.go
```diff
@@ -180,30 +206,276 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+			err = r.Client.Create(context.TODO(), migratedUserSignup)
+			if err != nil {
+				// If there was an error creating the migrated UserSignup, then set the status and requeue
+				return r.wrapErrorWithStatusUpdate(logger, userSignup,
+					r.setStatusMigrationFailedCreate, err, "Failed to create migrated UserSignup")
+			}
+
+			// Reload the UserSignup in order to populate the status (which happens outside of this condition block)
+			err = r.Client.Get(context.TODO(), types.NamespacedName{
```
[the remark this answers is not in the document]
sbryzak: This is done as the status is required to be updated separately.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -3953,3 +3954,222 @@ func TestUserSignupLastTargetClusterAnnotation(t *testing.T) {
...
+		// Given
+		userSignup := commonsignup.NewUserSignup()
+		userSignup.Labels[toolchainv1alpha1.UserSignupStateLabelKey] = "deactivated"
+		userSignup.Annotations["toolchain.dev.openshift.com/migration-replaces"] = "foo"
+		states.SetDeactivated(userSignup, true)
+		members := NewGetMemberClusters(NewMemberCluster(t, "member1", v1.ConditionTrue))
+
+		userSignupToDelete := commonsignup.NewUserSignup()
+		userSignupToDelete.Name = "foo"
+		userSignupToDelete.Annotations[migratedAnnotationName] = "true"
```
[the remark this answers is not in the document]
sbryzak: The usage of this annotation is in its presence, not in its value (which can be anything).

controllers/usersignup/usersignup_controller.go
sbryzak: This annotation is essentially informational, indicating to anyone looking at the UserSignup that it has been migrated.  After migration is complete for all users we can remove them.

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
sbryzak: This is all verified now.

MatousJobanek (approved):
Looks good :+1: 
Thanks a lot for addressing all the comments. I really appreciate your patience with the PR :1st_place_medal: 

Just please, don't merge the PR yet - I will merge it tomorrow while watching the cluster and the flow of the migration.

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
MatousJobanek: but the error is returned during the reconciliation of the "old" UserSignup:
* "Old UserSignup" is reconciled
   1. it creates "new UserSignup" with the annotation
   2. it tries to update the status
   3. it fails and returns an error
* "New UserSignup" is reconciled based on its creation
    1. it exits because the annotation is still present
* "Old UserSignup" is reconciled because the previous was returned with an error
   1. it doesn't create anything because the new is already present
   2. it tries to update the status of the "new UserSignup"
   3. it removes the annotation
* "New UserSignup" is reconciled based on the removal of the annotation
   1. deletes the "old UserSignup"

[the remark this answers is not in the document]
rajivnathan: Ah, you're right. I was in the habit of thinking of reconciliations on the same resource but in this case we're migrating to a new resource altogether to please disregard what I said. Sorry about that.

rajivnathan (approved):

PR #632, opened 2022-04-21
Define and Create User Tiers

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1357

Summary of changes:
- Defined a UserTier for each of the unique deactivation days counts we had from NSTemplateTiers eg. 0, 30, ...
- Update go generator make target to generate UserTier assets
- Added UserTier generator similar to NSTemplateTier generator but more simple because it's just the main tier.yaml we have for UserTiers
- Unit tests
- e2e tests coming soon...

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -7,15 +7,16 @@ resources:
 - bases/toolchain.dev.openshift.com_masteruserrecords.yaml
 - bases/toolchain.dev.openshift.com_notifications.yaml
 - bases/toolchain.dev.openshift.com_nstemplatetiers.yaml
+- bases/toolchain.dev.openshift.com_socialevents.yaml
+- bases/toolchain.dev.openshift.com_spacebindings.yaml
 - bases/toolchain.dev.openshift.com_spaces.yaml
 - bases/toolchain.dev.openshift.com_templateupdaterequests.yaml
 - bases/toolchain.dev.openshift.com_tiertemplates.yaml
 - bases/toolchain.dev.openshift.com_toolchainclusters.yaml
 - bases/toolchain.dev.openshift.com_toolchainconfigs.yaml
-- bases/toolchain.dev.openshift.com_socialevents.yaml
 - bases/toolchain.dev.openshift.com_toolchainstatuses.yaml
 - bases/toolchain.dev.openshift.com_usersignups.yaml
-- bases/toolchain.dev.openshift.com_spacebindings.yaml
+- bases/toolchain.dev.openshift.com_usertiers.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
 # the following config is for teaching kustomize how to do kustomization for CRDs.
--- a/make/go.mk
+++ b/make/go.mk
@@ -26,7 +26,11 @@ vendor:
 NSTEMPLATES_BASEDIR = deploy/templates/nstemplatetiers
 NSTEMPLATES_FILES = $(wildcard $(NSTEMPLATES_BASEDIR)/**/*.yaml)
 NSTEMPLATES_TEST_BASEDIR = test/templates/nstemplatetiers
-NSTEMPLATES_TEST_SUBDIRS = $(wildcard $(NSTEMPLATES_TEST_BASEDIR)/**/*)
+
+USERTEMPLATES_BASEDIR = deploy/templates/usertiers
+USERTEMPLATES_FILES = $(wildcard $(USERTEMPLATES_BASEDIR)/**/*.yaml)
+USERTEMPLATES_TEST_BASEDIR = test/templates/usertiers
+
 NOTIFICATION_BASEDIR = deploy/templates/notificationtemplates
 REGISTRATION_SERVICE_DIR=deploy/registration-service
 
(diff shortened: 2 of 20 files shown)


config/crd/kustomization.yaml
```diff
@@ -7,15 +7,16 @@ resources:
...
 - bases/toolchain.dev.openshift.com_spaces.yaml
 - bases/toolchain.dev.openshift.com_templateupdaterequests.yaml
 - bases/toolchain.dev.openshift.com_tiertemplates.yaml
 - bases/toolchain.dev.openshift.com_toolchainclusters.yaml
 - bases/toolchain.dev.openshift.com_toolchainconfigs.yaml
-- bases/toolchain.dev.openshift.com_socialevents.yaml
 - bases/toolchain.dev.openshift.com_toolchainstatuses.yaml
 - bases/toolchain.dev.openshift.com_usersignups.yaml
-- bases/toolchain.dev.openshift.com_spacebindings.yaml
+- bases/toolchain.dev.openshift.com_usertiers.yaml
```
MatousJobanek, line 19: and don't forget to modify the PROJECT file https://github.com/codeready-toolchain/host-operator/blob/master/PROJECT and also the CSV: https://github.com/codeready-toolchain/host-operator/blob/master/config/manifests/bases/host-operator.clusterserviceversion.yaml
forgot about that in the last review yesterday

make/go.mk
```diff
@@ -43,6 +47,9 @@ clean-metadata:
 generate-metadata: clean-metadata
 	@echo "generating templates metadata for manifests in $(NSTEMPLATES_BASEDIR)" 
 	@$(foreach tmpl,$(wildcard $(NSTEMPLATES_FILES)),$(call git_commit,$(tmpl),$(NSTEMPLATES_BASEDIR),metadata.yaml))
+	@echo "generating templates metadata for manifests in $(USERTEMPLATES_BASEDIR)" 
+	@rm ./$(USERTEMPLATES_BASEDIR)/metadata.yaml 2>/dev/null || true
+	@$(foreach tmpl,$(wildcard $(USERTEMPLATES_FILES)),$(call git_commit,$(tmpl),$(USERTEMPLATES_BASEDIR),metadata.yaml))
```
MatousJobanek, line 52: we don't need to version the UserTiers, right? or do we need that?

pkg/templates/usertiers/usertier_generator.go
```diff
@@ -0,0 +1,256 @@
...
+func loadTemplatesByTiers(assets assets.Assets) (map[string]*tierData, error) {
+	metadataContent, err := assets.Asset("metadata.yaml")
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to load templates")
+	}
+	metadata := make(map[string]string)
+	err = yaml.Unmarshal([]byte(metadataContent), &metadata)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to load templates")
+	}
```
MatousJobanek, line 114: similar question - what do we need this metadata for?
also the `revision` field in the `template` struct is probably not read/used anywhere.

pkg/templates/usertiers/usertier_generator.go
```diff
@@ -0,0 +1,256 @@
...
+
+// loadTemplatesByTiers loads the assets and dispatches them by tiers, assuming the given `assets` has the following structure:
+//
+// metadata.yaml
+// advanced/
+//   tier.yaml
+// basic/
+//   tier.yaml
+// team/
+//   tier.yaml
```
MatousJobanek, line 100: maybe update the names to match what we have in the directory

config/crd/kustomization.yaml
rajivnathan: Done in https://github.com/codeready-toolchain/host-operator/pull/632/commits/2a861ab3aabe7a28555b76daa2d40a1df5d6c572

Should we also add them to https://github.com/codeready-toolchain/api/blob/master/.github/pull_request_template.md so it's less likely we forget in the future?

rajivnathan: It looks like SpaceBinding is missing from the CSV? Should it be added there as well?

make/go.mk
rajivnathan: ah that's right, we don't have any references so it's not necessary... 🤦

pkg/templates/usertiers/usertier_generator.go
rajivnathan: Done in 85d8b484e5d0b8a28a92e116d1508f16babcf3f6

config/crd/kustomization.yaml
MatousJobanek: > Should we also add them to https://github.com/codeready-toolchain/api/blob/master/.github/pull_request_template.md so it's less likely we forget in the future?

good idea :+1: 

> It looks like SpaceBinding is missing from the CSV? Should it be added there as well?


whoops :grimacing: yeah, please add it there as well. Good catch :+1:

make/go.mk
rajivnathan: Removed: https://github.com/codeready-toolchain/host-operator/pull/632/commits/ec8bb521c62159dee5097715788a22cf7ce4282d

pkg/templates/usertiers/usertier_generator.go
rajivnathan: Removed: https://github.com/codeready-toolchain/host-operator/pull/632/commits/ec8bb521c62159dee5097715788a22cf7ce4282d

MatousJobanek (approved):
Thanks for addressing my comments :+1:

PR #633, opened 2022-04-23
Idler active

ranakan19:


MatousJobanek (commented):
the suffix "activated" in the new directory name can bring a small confusion with all the other names like "deactivated" and "deactivating" etc...
how about idlertriggered?

ranakan19:
> the suffix "activated" in the new directory name can bring a small confusion with all the other names like "deactivated" and "deactivating" etc... how about `idlertriggered`?
> 
> btw, have you checked the content of the email with Parag?



> 

Sure, idlertriggered sounds much better. Done in 39f9fdc. And yes, the email content was given by parag

deploy/templates/notificationtemplates/idlertriggered/notification.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>
+        Notice: Idler is Running on your Developer Sandbox.
```
rajivnathan, line 8:
```suggestion
        Notice: Your applications are being idled
```

deploy/templates/notificationtemplates/idlertriggered/subject.txt
```diff
@@ -0,0 +1 @@
+Notice: Your running application has been idled
```
rajivnathan, line 1:
```suggestion
Notice: Your application is being idled
```

deploy/templates/notificationtemplates/idlertriggered/notification.html
```diff
@@ -0,0 +1,64 @@
...
+
+    <p>
+        You are receiving this email because one or more of your applications in the Developer Sandbox for
+        Red Hat OpenShift has been running for 12 hours.
+    </p>
+
+    <p>
+        In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
+        application to zero (0). You can restart your application(s) by increasing the number of instances from the
+        Developer Sandbox User Interface.
```
rajivnathan, line 51: If the user just deployed a pod then I think the idler just deletes the pod so I guess the UI wouldn't have a way to bring it back. The user would have to recreate it so maybe we need to make this message more generic?

rajivnathan (commented):
Looks good 👍 Just had a couple minor comments/suggestions

MatousJobanek (approved):
Looks nice :+1:

xcoulon (approved):

PR #637, opened 2022-04-29
Update deactivation controller to work with NSTemplateTiers and UserTiers

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1361

Related PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/550

- Adds a function `getDeactivationTimeoutDays` that first checks if the MasterUserRecord `TierName` has a corresponding UserTier, if one is not found then it will fall back to searching for a NSTemplateTier with the same name. This function can be removed once the migration to UserTiers is completed.
- Unit tests are split by UserTiers and NSTemplateTiers, the NSTemplateTiers unit tests will be removed after the UserTier migration is complete
- Space controller will remove the old tier hash label as part of the space update
- Deactivation controller now resets the Deactivating state so that when a user is promoted they will still receive the Deactivating notification at the correct time
- This also means we don't need to use ChangeTierRequest for tier updates anymore, can promote MURs and Spaces individually


MatousJobanek (approved):
Looks fine :+1:

rajivnathan:
FYI, this PR was just updated today so that the space controller will remove the old tier hash label as part of the space update. Also, the deactivation controller now resets the Deactivating state so that when a user is promoted they will still receive the Deactivating notification at the correct time so this fixes the e2e test failure for this PR. This also means we don't need to use ChangeTierRequest for tier updates anymore, can promote MURs and Spaces individually. Description of the PR has been updated

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -243,47 +456,159 @@ func TestReconcile(t *testing.T) {
...
+			// Set the notification status condition as sent
+			userSignupFoobar.Status.Conditions = []toolchainv1alpha1.Condition{
+				{
+					Type:               toolchainv1alpha1.UserSignupUserDeactivatingNotificationCreated,
+					Status:             corev1.ConditionTrue,
+					LastTransitionTime: metav1.Time{Time: time.Now()},
+					Reason:             toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason,
+				},
+			}
+
```
MatousJobanek, line 548: it's just a detail, but we don't need to set the condition, right? The same applies for the following test-case

MatousJobanek (approved):
Thanks for improving the deactivation & space controllers

controllers/deactivation/deactivation_controller_test.go
rajivnathan: Yeah, that's right. They're not needed. I had them set because it would be set in the actual scenario but I can remove them to make the test more concise.

rajivnathan: Removed in 82cc799afa2d193006a3f8611d26992908d90131

PR #644, opened 2022-05-31
refactor UserSignupController: do not watch SpaceBindings

xcoulon:
Create SpaceBindings with a name based on the Space name and the MasterUserRecord name, so that if the UserSignupController is called just after a SpaceBinding was created but before it was indexed by Etcd (hence, `client.List` with a label matcher would return no result), then creating another SpaceBinding will fail because it would have the same name as the existing one.

see also https://github.com/codeready-toolchain/toolchain-e2e/pull/561


rajivnathan (approved):
The changes look good to me but two error messages were changed and I didn't see any unit test changes. Are we missing some coverage?

ranakan19 (approved):
LGTM

PR #648, opened 2022-06-02
Separate quotas for builds in appstudio

alexeykazakov:


xcoulon (approved):

xcoulon:
just curious: why moving from `ClusterResourceQuota` to `ResourceQuota`? I know there's a single namespace, so `ClusterResourceQuota` may be "overkill", but it there any other reason?

alexeykazakov:
Because ClusterResourceQuota doesn't support scopes. So I had to switch to ResourceQuota.

xcoulon:
> Because ClusterResourceQuota doesn't support scopes. So I had to switch to ResourceQuota.

ok, I missed that. Thanks for the explanation!

sbose78:
Thanks!

PR #655, opened 2022-06-10
Introduce more logging for user migration

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-1530


MatousJobanek (commented):
TBH, I don't think that these logs will help. We should add more complex logs - like for every significant step you do in the migration flow. 
Or, for example, every time when the code returns saying why and what it is the current status of the migration.

PR #669, opened 2022-07-26
feat: push event/action to Segment on user activations

xcoulon:
When a user is activated, an event is sent to Segment (if properly configured) with a hash of the `UserSignup.Spec.Username` as a key for the user.


controllers/usersignup/usersignup_controller.go
```diff
@@ -655,7 +657,14 @@ func (r *Reconciler) ensureNewMurIfApproved(reqLogger logr.Logger, config toolch
 	}
 
 	// Provision the MasterUserRecord
-	return r.provisionMasterUserRecord(reqLogger, config, userSignup, targetCluster, userTier)
+	if err := r.provisionMasterUserRecord(reqLogger, config, userSignup, targetCluster, userTier); err != nil {
+		return err
+	}
+	// track activation in Segment
+	if r.SegmentClient != nil {
+		return r.SegmentClient.TrackAccountActivation(userSignup.Spec.Username)
```
xcoulon, line 665: @ibuziuk @alexeykazakov I would need to get confirmation that `userSignup.Spec.Username` is the field to use here, so it matches with what CRW and Web Console send.

alexeykazakov: `userSignup.Spec.Username` is the user's sso username. My understanding it's what CRW and WebConsole use as well.

controllers/usersignup/usersignup_controller.go
```diff
@@ -655,7 +657,14 @@ func (r *Reconciler) ensureNewMurIfApproved(reqLogger logr.Logger, config toolch
 	}
 
 	// Provision the MasterUserRecord
-	return r.provisionMasterUserRecord(reqLogger, config, userSignup, targetCluster, userTier)
+	if err := r.provisionMasterUserRecord(reqLogger, config, userSignup, targetCluster, userTier); err != nil {
+		return err
+	}
+	// track activation in Segment
+	if r.SegmentClient != nil {
```
alexeykazakov, line 664: I wonder if we can add some tests for that. I guess segment provides some mock libs for testing.

alexeykazakov, line 664: Instead of reporting it here I think we should report it in provisionMasterUserRecord(). Right after or just before incrementing our user count metrics. Otherwise if there is a problem with provisioning a user you will report it multiple times.

pkg/segment/segment.go
```diff
@@ -0,0 +1,37 @@
...
+type Client struct {
+	client analytics.Client
+}
+
+var logger = logf.Log.WithName("segment_client")
+
+func NewClient(key string) (*Client, error) {
+	cl, err := analytics.NewWithConfig(key,
+		analytics.Config{
+			// Logger: nil,
```
alexeykazakov, line 17: ?

controllers/usersignup/usersignup_controller.go
xcoulon: good point, let me check that!

xcoulon: good point, let me refactor that!

pkg/segment/segment.go
xcoulon: ahem... forgot to implement an adapter to log when an event is sent to Segment 😂 
I will take care of it

controllers/usersignup/usersignup_controller.go
xcoulon: addressed in e701dec: done in the `updateUserSignupMetricsByState` func when metrics are updated

controllers/usersignup/usersignup_controller.go
xcoulon: ok, thanks for confirming!

controllers/usersignup/usersignup_controller.go
xcoulon: I couldn't find any mock to test the Segment client, but I refactored the code in addressed in e701dec to optionally replace it in the unit tests with a mock instance that we can use to assert that the message was queued

pkg/segment/segment.go
xcoulon: done in e701dec, there is a logger adapter which logs all info/errors. We can disable it later if it's too verbose.

alexeykazakov (approved):

PR #670, opened 2022-08-04
ensure status propagation - only when resources ready

ranakan19:
fix wrt : https://issues.redhat.com/browse/CRT-1471
This PR ensures UserSignup is complete only when space is ready.


controllers/usersignup/usersignup_controller.go
```diff
@@ -840,6 +856,18 @@ func (r *Reconciler) ensureSpace(logger logr.Logger, userSignup *toolchainv1alph
...
+	logger.Info("Ensuring Space is Ready", "Space", space.Name)
+	readyCond, ok := condition.FindConditionByType(space.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	if ok {
+		if readyCond.Status == corev1.ConditionTrue && readyCond.Reason == toolchainv1alpha1.SpaceProvisionedReason {
+			return true, nil
+		}
+		return false, nil
+	}
+	return false, fmt.Errorf("ready condition not found on space %s", space.Name)
+}
```
rajivnathan, line 869: Since we're only checking the status of the space and not actually changing anything I think renaming this func to something like `isSpaceReady` is more appropriate and it can probably be simplified to just return a `bool` because even if the Space isn't ready it's not necessarily a problem since we expect it to take some time. Otherwise we'd be printing errors in the logs under normal operating conditions.

```suggestion
func (r *Reconciler) isSpaceReady(logger logr.Logger, space *toolchainv1alpha1.Space) (bool, error) {
	logger.Info("Checking whether Space is Ready", "Space", space.Name)
	readyCond, ok := condition.FindConditionByType(space.Status.Conditions, toolchainv1alpha1.ConditionReady)
	return ok && readyCond.Status == corev1.ConditionTrue && readyCond.Reason == toolchainv1alpha1.SpaceProvisionedReason
}
```

controllers/usersignup/usersignup_controller.go
```diff
@@ -798,6 +804,16 @@ func (r *Reconciler) provisionMasterUserRecord(logger logr.Logger, config toolch
...
+
+	//check MUR was actually created
+	murRetrieved := &toolchainv1alpha1.MasterUserRecord{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: mur.Name, Namespace: mur.Namespace}, murRetrieved)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return fmt.Errorf(`MUR %s created isn't available yet`, mur.Name)
+		}
+		return err
+	}
```
rajivnathan, line 816: I think we can remove this check because we're creating the MUR just above here so I believe checking if it exists here is redundant because we should get an error if the creation failed.

controllers/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -167,6 +167,15 @@ func (r *Reconciler) ensureUserAccount(logger logr.Logger, murAccount toolchainv
...
 			}
+			// check the userAccount has actually been created
+			userAccountRetrieved := &toolchainv1alpha1.UserAccount{}
+			if err := memberCluster.Client.Get(context.TODO(), types.NamespacedName{
+				Namespace: userAccount.Namespace,
+				Name:      userAccount.Name}, userAccountRetrieved); err != nil {
+				if errors.IsNotFound(err) {
+					return updateStatusConditions(logger, r.Client, mur, toBeNotReady(toolchainv1alpha1.MasterUserRecordProvisioningReason, "UserAccount created isn't ready yet"))
+				}
+			}
```
rajivnathan, line 178: Do we need to check if the UserAccount exists here since we're creating it just a few lines above? Are we supposed to check if the UserAccount is Ready instead? That would need to be done outside of this block though, maybe just after the Space status check?

rajivnathan (commented):
Nice, this should fix the problems. Just added some questions/suggestions about which checks we need

controllers/usersignup/usersignup_controller.go
ranakan19: I like the idea of just returning the bool. Lemme make that change! Thanks!

ranakan19: done in [bc13ccb](https://github.com/codeready-toolchain/host-operator/pull/670/commits/bc13ccbc3af4a5f2e63d48f1cd695476e0e516aa)

controllers/usersignup/usersignup_controller.go
ranakan19: done in [d6bfc32](https://github.com/codeready-toolchain/host-operator/pull/670/commits/d6bfc3296a49c51e700d78063c0e676e719079a5)

controllers/usersignup/usersignup_controller.go
```diff
@@ -591,6 +591,11 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 			if err = r.ensureSpaceBinding(reqLogger, userSignup, mur, space); err != nil {
 				return true, err
 			}
+
+			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
+			if !r.isSpaceReady(reqLogger, space) {
+				return true, fmt.Errorf("space %s is not Ready", space.Name)
```
MatousJobanek, line 597: We don't need to return an error - this is not an "error case". The controller should get automatically reconciled when the Space's status is modified.

controllers/usersignup/usersignup_controller.go
```diff
@@ -840,6 +846,12 @@ func (r *Reconciler) ensureSpace(logger logr.Logger, userSignup *toolchainv1alph
 	return space, true, nil
 }
 
+func (r *Reconciler) isSpaceReady(logger logr.Logger, space *toolchainv1alpha1.Space) bool {
+	logger.Info("Ensuring Space is Ready", "Space", space.Name)
+	readyCond, ok := condition.FindConditionByType(space.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	return ok && readyCond.Status == corev1.ConditionTrue && readyCond.Reason == toolchainv1alpha1.SpaceProvisionedReason
```
MatousJobanek, line 852: you can use the `IsTrue` function https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go#L63

MatousJobanek (commented):
Looks good - just two small comments.

controllers/usersignup/usersignup_controller.go
MatousJobanek: I mean - just keep the status of the UserSignup in an "incomplete" state (you can also introduce a new reason "Provisioning"), and just return here without an error. As soon as the Space.status is updated because it gets "ready" then the UserSignup will be triggered and the status will be updated to "complete".

controllers/masteruserrecord/masteruserrecord_controller.go
```diff
@@ -167,6 +167,7 @@ func (r *Reconciler) ensureUserAccount(logger logr.Logger, murAccount toolchainv
 			if !isStatusPresent {
 				counter.IncrementUserAccountCount(logger, murAccount.TargetCluster)
 			}
+
```
rajivnathan, line 170: Is there some UserAccount Status we should be checking at all?

controllers/usersignup/usersignup_controller.go
```diff
@@ -840,6 +846,12 @@ func (r *Reconciler) ensureSpace(logger logr.Logger, userSignup *toolchainv1alph
 	return space, true, nil
 }
 
+func (r *Reconciler) isSpaceReady(logger logr.Logger, space *toolchainv1alpha1.Space) bool {
+	logger.Info("Ensuring Space is Ready", "Space", space.Name)
```
rajivnathan, line 850:
```suggestion
	logger.Info("Checking whether Space is Ready", "Space", space.Name)
```
Just a suggested rewording since we're not really "ensuring" it, we're just "checking" it

controllers/usersignup/usersignup_controller.go
ranakan19: I want to check the reason as well, ended up using `IsTrueWithReason` https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go#L91
Thanks for pointing it out

controllers/usersignup/usersignup_controller.go
ranakan19: done in [5ea85c0](https://github.com/codeready-toolchain/host-operator/pull/670/commits/5ea85c0b2ec4c72293ee59beea0f8b8d3c8af09d)

controllers/usersignup/usersignup_controller.go
ranakan19: done in [5ea85c0](https://github.com/codeready-toolchain/host-operator/pull/670/commits/5ea85c0b2ec4c72293ee59beea0f8b8d3c8af09d)

controllers/usersignup/usersignup_controller.go
ranakan19: done in [5ea85c0](https://github.com/codeready-toolchain/host-operator/pull/670/commits/5ea85c0b2ec4c72293ee59beea0f8b8d3c8af09d)

controllers/masteruserrecord/masteruserrecord_controller.go
ranakan19: I beleive it is not needed, as the SynchStatus and SyncSpec take care of it.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -183,22 +182,23 @@ func TestUserSignupCreateSpaceAndSpaceBindingOk(t *testing.T) {
...
 				res, err := r.Reconcile(context.TODO(), req)
-
-				// then
 				require.NoError(t, err)
+				// then
 				require.Equal(t, reconcile.Result{}, res)
 				switch testname {
 				case "without skip space creation annotation", "with skip space creation annotation set to false":
+					//require.Error(t, err)
+					//require.Equal(t, "space foo is not Ready", err.Error())
```
rajivnathan, line 193:
```suggestion
```

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -183,22 +182,23 @@ func TestUserSignupCreateSpaceAndSpaceBindingOk(t *testing.T) {
...
 					spacebindingtest.AssertThatSpaceBinding(t, test.HostOperatorNs, "foo", "foo", r.Client).
 						Exists().
 						HasLabelWithValue(toolchainv1alpha1.SpaceCreatorLabelKey, userSignup.Name).
 						HasLabelWithValue(toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey, "foo").
 						HasLabelWithValue(toolchainv1alpha1.SpaceBindingSpaceLabelKey, "foo").
 						HasSpec("foo", "foo", "admin")
 				case "with skip space creation annotation set to true":
+					//require.NoError(t, err)
```
rajivnathan, line 201:
```suggestion
```

go.mod
```diff
@@ -29,5 +29,7 @@ require (
 	k8s.io/klog/v2 v2.9.0
 	sigs.k8s.io/controller-runtime v0.10.3
 )
-
+replace (
+	github.com/codeready-toolchain/api v0.0.0-20220712190140-fca2282ae885 => github.com/ranakan19/api  v0.0.0-20220907163128-ec82f43e4482
+)
```
rajivnathan, line 34: Friendly reminder

PR #676, opened 2022-08-18
update toolchain-common to support insecure connection

MatousJobanek:
https://github.com/codeready-toolchain/toolchain-common/pull/265

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/codeready-toolchain/host-operator
 
 require (
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
 	github.com/codeready-toolchain/api v0.0.0-20220712190140-fca2282ae885
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20220628122057-9108fb0609bd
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/codeready-toolchain/host-operator
 
 require (
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
```
xcoulon, line 4: why do we suddenly need that `assert` pkg? 🤔

MatousJobanek: don't ask me - this is automatically pulled when I run `go mod tidy` on master without any changes.
btw, it's indirect dependency.

MatousJobanek: most likely transitive dependency of 
https://github.com/codeready-toolchain/host-operator/commit/d72e68a0c78ab9a1cb14c80d488f5628f2a61d51#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R17
as this is the commit that introduced it

alexeykazakov (approved):

PR #677, opened 2022-08-18
refactor: no need for runtime.Scheme in ApplyClient

xcoulon:

### DIFF
--- a/deploy/resources.go
+++ b/deploy/resources.go
@@ -1,8 +1,0 @@
-package resources
-
-import (
-	"embed"
-)
-
-//go:embed *
-var Resources embed.FS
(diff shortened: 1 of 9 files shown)


deploy/resources.go
```diff
@@ -0,0 +1,8 @@
+package resources
```
alexeykazakov, line 1: This is not related to this PR, is it?

alexeykazakov (approved):

deploy/resources.go
xcoulon: right, my bad, I created this branch on top of an existing one, for something not completed yet 🤦‍♂️

xcoulon: removed in 28ec1eb

PR #682, opened 2022-09-16
Create tier without idling

rajivnathan:
Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/600


alexeykazakov (approved):

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
```diff
@@ -30,6 +30,7 @@ var expectedProdTiers = map[string]bool{
 	"advanced":           true, // tier_name: true/false (if based on the other tier)
 	"base":               false,
 	"base1ns":            false,
+	"basenoidling":       true,
```
MatousJobanek, line 33: it's just a detail and I don't insist on it, but shouldn't we rather name it `base1nsnoiddling`?

MatousJobanek (approved):

pkg/templates/nstemplatetiers/nstemplatetier_generator_whitebox_test.go
rajivnathan: I did think about that but thought it's a bit long and probably not necessary. But I can change it if we want.

rajivnathan: Renamed it to `base1nsnoidling` just in case https://github.com/codeready-toolchain/host-operator/pull/682/commits/b40bb93bc2dcf5e5cd5e03acd6a64fd1ba0fd7ed

PR #702, opened 2022-11-07
test: usersignup approved funcs

xcoulon:
see https://github.com/codeready-toolchain/toolchain-common/pull/277


sbryzak (approved):
Looking at the API again I think we probably should have gone with `ApprovedManually` / `ApprovedAutomatically` / `SetApprovedManually` which I think makes more sense from a developer perspective however this is a very minor nitpick and I'm happy to merge as is.

PR #703, opened 2022-11-07
enable cross-platform docker build

mfrancisc:
While building docker images on Mac M1 chipset and deploying to AWS dev environment (running on linux/amd64) I need to change the build platform from my default one.

This PR enables configuring the docker build platform option, example command to run from within the toolchain-e2e dir:

`make dev-deploy-e2e HOST_REPO_PATH=$(PWD)/../host-operator goarch=amd64 IMAGE_BUILDER=docker`

**NOTE:** doesn't work with podman on Apple M1 apparently, the images starting with `FROM scratch` are not being built for the architecture specified with `--platform` option but with the default one (in my case `arm64` ). This could be an issue with podman on Mac only, since docker works fine.


make/run-cicd-script.mk
```diff
@@ -1,4 +1,4 @@
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=mfrancisc/toolchain-cicd/build_platform_config
```
MatousJobanek, line 1: reminder

MatousJobanek (approved):

make/run-cicd-script.mk
mfrancisc: yes, I'll update this as soon as https://github.com/codeready-toolchain/toolchain-cicd/pull/74 is merged.

Thanks!

mfrancisc:
I see codecov is complaining about less coverage, but there are no Go changes here. should I merge this anyway?

MatousJobanek:
Yes, sure. As for Codecov, we are kind of practical here. We use Codecov as a tool that helps us to keep the code coverage high, but the report is not a blocker for us anyhow - we don't have any limits that we should meet. So, if you see that the coverage is somehow low, then take a look at the report if you didn't miss anything. But for example in such cases like this one, you don't have to do even that :-).

PR #711, opened 2022-11-24
ASC-115 update capacity manager to use spaces

mfrancisc:
Switch Capacity Manager to count spaces and not user accounts.

Paired with: [toolchain-e2e#626](https://github.com/codeready-toolchain/toolchain-e2e/pull/626)

Jira ticket: [ASC-115](https://issues.redhat.com/browse/ASC-115)


pkg/capacity/manager_test.go
```diff
@@ -145,11 +145,11 @@ func TestGetOptimalTargetCluster(t *testing.T) {
...
+	t.Run("with two clusters, none of them is returned since it reaches max number of Spaces, no matter what is defined as preferred", func(t *testing.T) {
 		// given
 		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
-			testconfig.AutomaticApproval().
-				MaxNumberOfUsers(800, testconfig.PerMemberCluster("member1", 6000), testconfig.PerMemberCluster("member2", 1000)).
+			testconfig.CapacityThresholds().
+				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 6), testconfig.PerMemberCluster("member2", 10)).
 				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 60), testconfig.PerMemberCluster("member2", 75)))
 		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
 		InitializeCounters(t, toolchainStatus)
```
MatousJobanek, line 155: If I'm not mistaken, this one was testing the "overall" limit which was dropped, so I guess that we can drop also this test.

pkg/capacity/manager_test.go
```diff
@@ -166,8 +166,8 @@ func TestGetOptimalTargetCluster(t *testing.T) {
 	t.Run("with two clusters but only the second one has enough capacity - using the default values", func(t *testing.T) {
 		// given
 		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
-			testconfig.AutomaticApproval().
-				MaxNumberOfUsers(2000).
+			testconfig.CapacityThresholds().
+				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 700), testconfig.PerMemberCluster("member2", 1000)).
```
MatousJobanek, line 170: we should keep it empty because it's testing the memory limit
```suggestion
				MaxNumberOfSpaces().
```

pkg/capacity/manager_test.go
```diff
@@ -184,8 +184,8 @@ func TestGetOptimalTargetCluster(t *testing.T) {
 	t.Run("with two clusters but none of them has enough capacity - using the default memory values", func(t *testing.T) {
 		// given
 		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
-			testconfig.AutomaticApproval().
-				MaxNumberOfUsers(5000).
+			testconfig.CapacityThresholds().
+				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 2500), testconfig.PerMemberCluster("member2", 2500)).
```
MatousJobanek, line 188: the same
```suggestion
				MaxNumberOfSpaces().
```

MatousJobanek (approved):
Looks good, just a few minor comments

pkg/capacity/manager_test.go
mfrancisc: right, but now is covering the case in which there is no more room for spaces in both members (even if the limit is per member cluster and not overall anymore).

Do you think we don't need to test that?

MatousJobanek (approved):
Nice & clean - thanks for the additional changes :+1:

PR #719, opened 2022-12-14
Remove appstudio user SA

rajivnathan:
The proxy now uses impersonation so the user SA and its RoleBindings are no longer needed. See https://github.com/codeready-toolchain/registration-service/pull/289

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/642

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -27,6 +27,7 @@ objects:
           - get
           - update
           - list
+          - watch
       - apiGroups:
           - toolchain.dev.openshift.com
         resources:
--- a/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -4,13 +4,6 @@ metadata:
   name: appstudio-spacerole-admin # name is used in e2e tests
 objects:
 
-# ServiceAccounts that represents the AppStudio user - the token of this SA is used by the proxy for forwarding the requests from UI (or any other client)
-- apiVersion: v1
-  kind: ServiceAccount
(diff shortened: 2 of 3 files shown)


deploy/registration-service/registration-service.yaml
```diff
@@ -27,6 +27,7 @@ objects:
           - get
           - update
           - list
+          - watch
```
rajivnathan, line 30: This change is not actually related to this issue. It is related to proxy caching work but just leaving it here since it's a small change.

alexeykazakov: Just keep in mind that we need to keep it consistent with https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml#L25-L29

alexeykazakov (approved):

alexeykazakov:
@rajivnathan you need to update the e2e tests too.

deploy/registration-service/registration-service.yaml
rajivnathan: I guess you meant https://github.com/codeready-toolchain/registration-service/blob/master/deploy/registration-service.yaml#L25-L29  

Good point, I totally forgot about that! Will create a PR for that.

rajivnathan: https://github.com/codeready-toolchain/registration-service/pull/296

rajivnathan:
@alexeykazakov right, thanks. e2e tests here: https://github.com/codeready-toolchain/toolchain-e2e/pull/642

deploy/registration-service/registration-service.yaml
alexeykazakov: Yes! :)

PR #721, opened 2023-01-03
fix: provision Spaces in batches by 50

MatousJobanek:
The original implementation didn't work fine because it blindly divided by 50 without reminder, but the result was in most cases the same for both clusters. 

The new approach stores the last member cluster that was used.
* If the reminder is not 0 (there is still some capacity until a multiply of 50 is reached), then it uses the last used member cluster
* otherwise, it compares the clusters and chooses the freest one for the next batch of 50 Spaces

### DIFF
--- a/pkg/capacity/manager.go
+++ b/pkg/capacity/manager.go
@@ -51,6 +51,19 @@ func hasMemberREnoughResources(memberStatus toolchainv1alpha1.Member, threshold 
 	return false
 }
 
+func NewClusterManager(getMemberClusters cluster.GetMemberClustersFunc, cl client.Client) *ClusterManager {
+	return &ClusterManager{
+		getMemberClusters: getMemberClusters,
+		client:            cl,
+	}
+}
+
+type ClusterManager struct {
+	getMemberClusters cluster.GetMemberClustersFunc
+	client            client.Client
+	lastUsed          string
+}
+
 // GetOptimalTargetCluster returns the name of the cluster with the most available capacity where a Space could be provisioned.
 //
 // If two clusters have the same limit and they both have the same usage, then the logic distributes spaces in a batches of 50.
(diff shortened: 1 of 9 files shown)


pkg/capacity/manager.go
```diff
@@ -83,15 +105,16 @@ func GetOptimalTargetCluster(preferredCluster, namespace string, getMemberCluste
 		provisioned2 := counts.SpacesPerClusterCounts[optimalTargetClusters[j]]
 		threshold2 := config.CapacityThresholds().MaxNumberOfSpacesSpecificPerMemberCluster()[optimalTargetClusters[j]]
 
-		// Let's round the number of provisioned users down to closest multiple of 50
+		// Let's round the number of provisioned users down to the closest multiple of 50
 		// This is a trick we need to do before comparing the capacity, so we can distribute the users in batches by 50 (if the clusters have the same limit)
-		provisioned1 = (provisioned1 / 50) * 50
-		provisioned2 = (provisioned2 / 50) * 50
+		provisioned1By50 := (provisioned1 / 50) * 50
```
sonyccd, line 110: (y/x)*x=y because (y/x)*(x/1)=y/1=y therefore provisioned1By50=provisioned1

MatousJobanek: in theory (on paper) yes. But not in Golang when using integers (whole numbers).
eg. 
```
provisoned1 = 108
(provisioned1 / 50) is evaluated as 2
(provisioned1 / 50) * 50 is 100
```
I could change it to 
```
provisioned1 - (provisioned1 % 50)
```
if it would make more sense ( `%` is modulo)

alexeykazakov (approved):

MatousJobanek:
/retest
infra

PR #724, opened 2023-01-10
Reduce log verbosity

xcoulon:
see also https://github.com/codeready-toolchain/toolchain-common/pull/285

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/codeready-toolchain/host-operator
 
 require (
 	github.com/codeready-toolchain/api v0.0.0-20221220133801-e9fb0b2352db
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20221220133846-3415e8af3654
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20230112091129-d06f31ddd2f6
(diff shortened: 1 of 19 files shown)


mfrancisc:
Even here it would be nice if the logging library could provide a debug level (`logger.Debug(string)`) so that we can keep those logs turned on only when the operator is deployed in debugging mode for examlpe.

MatousJobanek (changes requested):
TBH, I would love to keep these logs there because they are super helpful when analyzing logs for some troubleshooting.

xcoulon:
> TBH, I would love to keep these logs there because they are super helpful when analyzing logs for some troubleshooting.

all of them?

go.mod
```diff
@@ -31,6 +31,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.12.2
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/xcoulon/toolchain-common v0.0.0-20230111085908-1dc1cb70f1e9
```
mfrancisc, line 34: friendly reminder

mfrancisc (approved):

MatousJobanek (approved):
Approving since it doesn't remove any logs now.
Thanks for changing it

PR #727, opened 2023-01-12
Update parentSpace label 

mfrancisc:
Set in Space objects the parentSpace label value to .spec.ParentSpace value when they differ.

This is PR covers the acceptance criteria from [Jira](https://issues.redhat.com/browse/ASC-247):

- The Space controller makes sure that the value of the label corresponds to the value stored in Spec.ParentSpace parameter.

Following PRs will cover the other acceptance criteria.


controllers/space/space_controller.go
```diff
@@ -312,6 +314,12 @@ func (r *Reconciler) ensureNSTemplateSet(logger logr.Logger, space *toolchainv1a
 	}
 }
 
+func setParentSpaceLabel(space *toolchainv1alpha1.Space) {
+	if parentSpace, found := space.Labels[toolchainv1alpha1.ParentSpaceLabelKey]; !found || parentSpace != space.Spec.ParentSpace {
+		space.Labels[toolchainv1alpha1.ParentSpaceLabelKey] = space.Spec.ParentSpace
```
MatousJobanek, line 319: if the space is not a sub-space, which means that it doesn't have the `Space.Spec.ParentSpace` set, then the label shouldn't be present. However, this would set the label to empty string `""`

MatousJobanek (commented):
Please add some tests as well

controllers/space/space_controller.go
mfrancisc: Thanks! 

I've pushed a change to handle it and unit tests, please see: https://github.com/codeready-toolchain/host-operator/pull/727/commits/5ec86f07f3db70397e2dc2d60e356c5ccbc21415

mfrancisc:
/test e2e

https://issues.redhat.com/browse/CRT-1643

xcoulon (approved):

MatousJobanek (approved):

PR #732, opened 2023-01-20
Add reg service permissions to watch spaces and add startup probe

rajivnathan:
Related to https://issues.redhat.com/browse/CRT-1637

Reg service PR with matching deployment yaml: https://github.com/codeready-toolchain/registration-service/pull/300

*Update*
Since populating the informers cache only happens on startup I reverted the changes to the liveness and readiness probes and introduced a startup probe with a higher failureThreshold.

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -32,9 +32,10 @@ objects:
           - toolchain.dev.openshift.com
         resources:
           - masteruserrecords
+          - socialevents
+          - spaces
           - toolchainconfigs
           - toolchainstatuses
-          - socialevents
         verbs:
           - get
           - list
@@ -98,7 +99,7 @@ objects:
                 - registration-service
               imagePullPolicy: IfNotPresent
               livenessProbe:
-                failureThreshold: 3
+                failureThreshold: 5
                 httpGet:
                   path: /api/v1/health
                   port: 8080
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -119,13 +119,13 @@ objects:
                 successThreshold: 1
                 timeoutSeconds: 1
               startupProbe:
-                failureThreshold: 18
+                failureThreshold: 180
                 httpGet:
                   path: /api/v1/health
                   port: 8080
                   scheme: HTTP
                 initialDelaySeconds: 1
-                periodSeconds: 10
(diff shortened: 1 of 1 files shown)


MatousJobanek (commented):
don't forget to update also the template in reg-service repo.
So, do we want to keep the `initialDelaySeconds`? I thought that we wanted to make it smaller https://redhat-internal.slack.com/archives/CHK0J6HT6/p1674055211798019?thread_ts=1673965463.977119&cid=CHK0J6HT6
Wouldn't make sense to decrease it for example to 3 seconds and the attempts increase eg. to 15?

rajivnathan:
@MatousJobanek good point, I thought the decision was to not increase it *more* than we already did but reading it again I think your interpretation is correct. While I was considering what failureThreshold would be good I stumbled upon the startup probes documentation and that would actually be a perfect fit for our problem since populating the informers cache only happens on startup so reverted the changes to the liveness and readiness probes and introduced a startup probe with a higher failureThreshold. See https://github.com/codeready-toolchain/host-operator/pull/732/commits/1f2b09fd50011385b165e008b49a9b244a34cb8b

deploy/registration-service/registration-service.yaml
```diff
@@ -113,10 +114,20 @@ objects:
...
+              startupProbe:
+                failureThreshold: 18
+                httpGet:
+                  path: /api/v1/health
+                  port: 8080
+                  scheme: HTTP
+                initialDelaySeconds: 1
+                periodSeconds: 10
+                successThreshold: 1
+                timeoutSeconds: 1
```
alexeykazakov, line 130: It means the prob will be called 18 times (after one second initial delay) every 10 seconds before failing. So it's 1, 10, 11,  21, ... 181 seconds before failing. Correct?

If so, wouldn't it be better to have smaller intervals, something like:
```suggestion
              startupProbe:
                failureThreshold: 60
                httpGet:
                  path: /api/v1/health
                  port: 8080
                  scheme: HTTP
                initialDelaySeconds: 1
                periodSeconds: 3
                successThreshold: 1
                timeoutSeconds: 1
```

rajivnathan: Good point, since this is only for startup we can have a shorter intervals. 👍

alexeykazakov (approved):

MatousJobanek (approved):

PR #736, opened 2023-01-24
feat: Support cluster-roles when specifying the needed target cluster in Spaces

mfrancisc:
- update capacity manager to select targetCluster based on `cluster-roles` if provided
- update UserSignup Controller to set 'tenant' cluster-role label to all Spaces it provisions
- update tests,packages and controllers to use new `Space.Spec.TargetCluster` and `Space.Status.TargetCluster` structs

Jira ticket: https://issues.redhat.com/browse/ASC-251

Related PRs:
- api: https://github.com/codeready-toolchain/api/pull/336
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/289

Paired with:
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/656

### DIFF
--- a/controllers/usersignup/approval_test.go
+++ b/controllers/usersignup/approval_test.go
@@ -70,7 +70,10 @@ func TestGetClusterIfApproved(t *testing.T) {
 				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 70), testconfig.PerMemberCluster("member2", 75)))
 		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
 		InitializeCounters(t, toolchainStatus)
-		clusters := NewGetMemberClusters(NewMemberCluster(t, "member1", corev1.ConditionTrue), NewMemberCluster(t, "member2", corev1.ConditionTrue))
+		clusters := NewGetMemberClusters(
+			NewMemberClusterWithClient(NewFakeClient(t), "member1", corev1.ConditionTrue), // no cluster-role label on this member
+			NewMemberCluster(t, "member2", corev1.ConditionTrue),
+		)
 
 		// when
 		approved, clusterName, err := getClusterIfApproved(fakeClient, signup, capacity.NewClusterManager(clusters, fakeClient))
@@ -85,7 +88,7 @@ func TestGetClusterIfApproved(t *testing.T) {
 		// given
 		signup := commonsignup.NewUserSignup()
 		signup.Annotations = map[string]string{
-			toolchainv1alpha1.UserSignupLastTargetClusterAnnotationKey: "member1",
+			toolchainv1alpha1.UserSignupLastTargetClusterAnnotationKey: "member1", // set preferred member cluster name
 		}
 		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
 			testconfig.AutomaticApproval().
--- a/controllers/usersignup/approval_test.go
+++ b/controllers/usersignup/approval_test.go
@@ -71,8 +71,8 @@ func TestGetClusterIfApproved(t *testing.T) {
 		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
 		InitializeCounters(t, toolchainStatus)
 		clusters := NewGetMemberClusters(
-			NewMemberClusterWithClient(NewFakeClient(t), "member1", corev1.ConditionTrue), // no cluster-role label on this member
-			NewMemberCluster(t, "member2", corev1.ConditionTrue),
+			NewMemberClusterWithoutClusterRoles(t, "member1", corev1.ConditionTrue), // no cluster-role label on this member
+			NewMemberCluster(t, "member2", corev1.ConditionTrue),                    // by default all member clusters will have the 'tenant' cluster-role
 		)
 
 		// when
@@ -100,8 +100,8 @@ func TestGetClusterIfApproved(t *testing.T) {
 		InitializeCounters(t, toolchainStatus)
 		clusters := NewGetMemberClusters(
 			// member1 doesn't have the cluster-role tenant but it's preferred one
-			NewMemberClusterWithClient(NewFakeClient(t), "member1", corev1.ConditionTrue),
-			NewMemberCluster(t, "member1", corev1.ConditionTrue),
+			NewMemberClusterWithoutClusterRoles(t, "member1", corev1.ConditionTrue),
+			NewMemberCluster(t, "member1", corev1.ConditionTrue), // by default all member clusters will have the 'tenant' cluster-role
 		)
 
 		// when
--- a/controllers/usersignup/approval_test.go
+++ b/controllers/usersignup/approval_test.go
@@ -48,7 +48,7 @@ func TestGetClusterIfApproved(t *testing.T) {
 				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 70), testconfig.PerMemberCluster("member2", 75)))
 		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
 		InitializeCounters(t, toolchainStatus)
-		clusters := NewGetMemberClusters(NewMemberCluster(t, "member1", corev1.ConditionTrue), NewMemberCluster(t, "member2", corev1.ConditionTrue))
+		clusters := NewGetMemberClusters(NewMemberClusterWithTenantRole(t, "member1", corev1.ConditionTrue), NewMemberClusterWithTenantRole(t, "member2", corev1.ConditionTrue))
 
 		// when
 		approved, clusterName, err := getClusterIfApproved(fakeClient, signup, capacity.NewClusterManager(clusters, fakeClient))
@@ -65,14 +65,12 @@ func TestGetClusterIfApproved(t *testing.T) {
 		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
 			testconfig.AutomaticApproval().
 				Enabled(true),
-			testconfig.CapacityThresholds().
-				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 1000), testconfig.PerMemberCluster("member2", 1000)).
-				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 70), testconfig.PerMemberCluster("member2", 75)))
+		)
 		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
 		InitializeCounters(t, toolchainStatus)
 		clusters := NewGetMemberClusters(
 			NewMemberClusterWithoutClusterRoles(t, "member1", corev1.ConditionTrue), // no cluster-role label on this member
-			NewMemberCluster(t, "member2", corev1.ConditionTrue),                    // by default all member clusters will have the 'tenant' cluster-role
+			NewMemberClusterWithTenantRole(t, "member2", corev1.ConditionTrue),      // by default all member clusters will have the 'tenant' cluster-role
 		)
 
 		// when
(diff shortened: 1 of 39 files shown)


mfrancisc:
/retest

pushed small time.sleep in e2e: [206ea79](https://github.com/codeready-toolchain/toolchain-e2e/pull/656/commits/206ea7948813c268aff0221f5deaa0cbf9a5bdc9)

controllers/usersignup/approval_test.go
```diff
@@ -60,6 +59,60 @@ func TestGetClusterIfApproved(t *testing.T) {
...
 
+	t.Run("with two clusters available, the second one has required cluster-role label", func(t *testing.T) {
+		// given
+		signup := commonsignup.NewUserSignup()
+		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
+			testconfig.AutomaticApproval().
+				Enabled(true),
+			testconfig.CapacityThresholds().
+				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 1000), testconfig.PerMemberCluster("member2", 1000)).
+				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 70), testconfig.PerMemberCluster("member2", 75)))
```
MatousJobanek, line 70: for the simplicity of the tests, you can drop the capacity thresholds. 
```suggestion
				Enabled(true))
```

controllers/usersignup/approval_test.go
```diff
@@ -60,6 +59,60 @@ func TestGetClusterIfApproved(t *testing.T) {
...
+		signup := commonsignup.NewUserSignup()
+		signup.Annotations = map[string]string{
+			toolchainv1alpha1.UserSignupLastTargetClusterAnnotationKey: "member1", // set preferred member cluster name
+		}
+		toolchainConfig := commonconfig.NewToolchainConfigObjWithReset(t,
+			testconfig.AutomaticApproval().
+				Enabled(true),
+			testconfig.CapacityThresholds().
+				MaxNumberOfSpaces(testconfig.PerMemberCluster("member1", 1000), testconfig.PerMemberCluster("member2", 1000)).
+				ResourceCapacityThreshold(80, testconfig.PerMemberCluster("member1", 70), testconfig.PerMemberCluster("member2", 75)))
```
MatousJobanek, line 98: the same
```suggestion
				Enabled(true))
```

controllers/usersignup/approval_test.go
```diff
@@ -60,6 +59,60 @@ func TestGetClusterIfApproved(t *testing.T) {
...
+		fakeClient := NewFakeClient(t, toolchainStatus, toolchainConfig)
+		InitializeCounters(t, toolchainStatus)
+		clusters := NewGetMemberClusters(
+			// member1 doesn't have the cluster-role tenant but it's preferred one
+			NewMemberClusterWithoutClusterRoles(t, "member1", corev1.ConditionTrue),
+			NewMemberCluster(t, "member1", corev1.ConditionTrue), // by default all member clusters will have the 'tenant' cluster-role
```
MatousJobanek, line 104: you already created member1 in the previous line
```suggestion
			NewMemberCluster(t, "member2", corev1.ConditionTrue), // by default all member clusters will have the 'tenant' cluster-role
```

pkg/capacity/manager.go
```diff
@@ -72,7 +86,8 @@ type ClusterManager struct {
 // Let's say that the limit for member1 is 1000 and for member2 is 2000, then the batch of spaces would be 50 for member1 and 100 for member2.
 //
 // If the preferredCluster is provided and it is also one of the available clusters, then the same name is returned.
-func (b *ClusterManager) GetOptimalTargetCluster(preferredCluster, namespace string) (string, error) {
+// In case the preferredCluster was not provided or not found/available and the clusterRoles are provided then the candidates optimal cluster pool will be made out by only those matching the labels, if any available.
+func (b *ClusterManager) GetOptimalTargetCluster(optimalClusterFilter *OptimalTargetClusterFilter) (string, error) {
```
MatousJobanek, line 90: it's just a detail, but it doesn't have to be passed as a pointer - the value should be totally fine
```suggestion
func (b *ClusterManager) GetOptimalTargetCluster(optimalClusterFilter OptimalTargetClusterFilter) (string, error) {
```

MatousJobanek: out of curiosity - what is the benefit of introducing the new type `OptimalTargetClusterFilter` instead of taking 3 parameters?
```suggestion
func (b *ClusterManager) GetOptimalTargetCluster(preferredCluster, toolchainStatusNamespace string, clusterRoles []string) (string, error) {
```
I didn't notice that there would be some particular place where the new type really simplifies the code.

pkg/capacity/manager.go
```diff
@@ -134,9 +154,31 @@ func getOptimalTargetClusters(preferredCluster string, getMemberClusters cluster
...
 	}
 
-	memberNames := make([]string, len(members))
-	for i := range members {
-		memberNames[i] = members[i].Name
+	var memberNames []string
+	for _, member := range members {
+		// if cluster-role labels were provided, it will check for matching on the member labels
+		// if no clusterRoles labels are required, then the function will simple return true
+		if hasClusterRoles(clusterRoles, member) {
```
MatousJobanek, line 161: if no clusterRole is provided, then we should automatically pick the `tenant` cluster. I also mentioned it in the Jira:
> Capacity manager is updated so it reflects the roles (if specified and targetCluster is empty) and selects the right member clusters. If no role nor name is provided then it uses the standard "tenant" cluster defined by the label cluster-role.toolchain.dev.openshift.com/tenant: "".

test/cluster.go
```diff
@@ -33,16 +33,30 @@ func NewGetMemberClusters(memberClusters ...*cluster.CachedToolchainCluster) clu
...
 
+type Modifier func(toolchainCluster *cluster.CachedToolchainCluster)
+
+func WithClusterRoleLabel(labelKey string) Modifier {
+	return func(cluster *cluster.CachedToolchainCluster) {
+		cluster.Config.Labels[labelKey] = "" // we don't care about the label value, only the key is used
+	}
+}
+
 func NewMemberCluster(t *testing.T, name string, status corev1.ConditionStatus) *cluster.CachedToolchainCluster {
```
MatousJobanek, line 44: I would rename it to 
```suggestion
func NewMemberClusterWithTenantRole(t *testing.T, name string, status corev1.ConditionStatus) *cluster.CachedToolchainCluster {
```
to make it explicit

pkg/capacity/manager.go
```diff
@@ -118,11 +133,16 @@ func (b *ClusterManager) GetOptimalTargetCluster(preferredCluster, namespace str
...
+func getOptimalTargetClusters(preferredCluster string, getMemberClusters cluster.GetMemberClustersFunc, clusterRoles []string, conditions ...cluster.Condition) []string {
+	emptyTargetCluster := []string{""}
 	// Automatic cluster selection based on cluster readiness
 	members := getMemberClusters(append(conditions, cluster.Ready)...)
 	if len(members) == 0 {
-		return []string{""}
+		return emptyTargetCluster
 	}
 
 	// if the preferred cluster is provided and it is also one of the available clusters, then the same name is returned, otherwise, it returns the first available one
```
MatousJobanek, line 148: personally, I think that the roles should have precedence over the preferred cluster. The preferred target cluster is the cluster where the user was provisioned last time. It's not the cluster that is specified in the Spec. Imagine:
1. you are provisioned to member1
2. you get deactivated
3. when you reactivate your account, then we try our best to provision you to member1 again. 
4. but, if the cluster is full or is gone then we provision you somewhere else

So if:
* either the purpose of the cluster has changed (we removed a label from the cluster)
* or the required set of labels has changed for your Space

Then we should not provision the Space to the same member cluster but somewhere else that matches the requirements

controllers/usersignup/approval_test.go
mfrancisc: removed in [6b01926](https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642)

controllers/usersignup/approval_test.go
mfrancisc: removed in [6b01926](https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642)

controllers/usersignup/approval_test.go
mfrancisc: thanks for spotting this , fixed in [6b01926](https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642)

pkg/capacity/manager.go
mfrancisc: I have removed the pointer and now the struct is passed by value, see [6b01926](https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642)

The advantages IMO for introducing the struct are mainly those 2:

1. you can avoid setting blank/empty values for optional parameters, for example when there is no `preferredCluster` and no `clusterRoles` the code will look like:
```
manager.GetOptimalTargetCluster("", toolchainStatusNamespace, []string{})
```
which was a bit weird from my point of view.

2. you can add more optional parameters without changing the caller code if not required.

I agree that it's more benefitting for the test code, but those were the main reasons for introducing this new struct.

pkg/capacity/manager.go
mfrancisc: I was assuming all member cluster have the `tenant` role, but I agree we can force it in case this will change in the future.

I've changed the `hasClusterRoles` to add the `tenant` role in case `clusterRoles` is empty, see https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642

test/cluster.go
mfrancisc: makes sense, change in https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642

pkg/capacity/manager.go
mfrancisc: Sorry I got confused between `targetCluster` and `preferredCluster` , changed the logic to first filter on cluster roles. see https://github.com/codeready-toolchain/host-operator/pull/736/commits/6b01926137a7f5c783ed0b886e61eaba8cc13642

mfrancisc:
/retest

infra

mfrancisc:
/retest

infra

pkg/capacity/manager.go
```diff
@@ -144,34 +144,46 @@ func getOptimalTargetClusters(preferredCluster string, getMemberClusters cluster
...
-		if hasClusterRoles(clusterRoles, member) {
-			memberNames = append(memberNames, member.Name)
-		}
-	}
-	// return empty string if no members available with roles
-	if len(memberNames) == 0 {
-		return emptyTargetCluster
+	for _, m := range filteredMembers {
+		memberNames = append(memberNames, m.Name)
 	}
```
MatousJobanek, line 174: just a side note: you could have extracted just the names while filtering them based on the roles - you don't need the whole `CachedToolchainCluster` types, only the names. But it's just a detail.

MatousJobanek (approved):
Great :+1: nice and clean work :1st_place_medal:

pkg/capacity/manager.go
mfrancisc: Good point! Not sure why I've created two lists in the first place 🤦 

Fixed in [f16977a](https://github.com/codeready-toolchain/host-operator/pull/736/commits/f16977a51010e145b15d0052602eccbe8879a9fc) 

Thanks!

mfrancisc:
/retest

infra

PR #737, opened 2023-01-24
Increase PVC quota for AppStudio to 12

Roming22:
Pipeline Service keeps the latest 10 PipelineRuns, so the quota is increased to 12 to allow for 2 PipelineRuns to be created before the pruner makes more space for additional PipelineRuns.

e2e-tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/657

Signed-off-by: Romain Arnaud <rarnaud@redhat.com>

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/ns_appstudio.yaml
@@ -87,10 +87,10 @@ objects:
     namespace: ${USERNAME}
   spec:
     hard:
-      limits.ephemeral-storage: 15Gi
-      requests.storage: 15Gi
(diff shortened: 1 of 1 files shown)


alexeykazakov (commented):
Could you please also create a PR for the e2e test repo?
You need to update this https://github.com/codeready-toolchain/toolchain-e2e/blob/master/testsupport/tiers/checks.go#L420
So the e2e tests can pass.

Please note that the e2e test PR has to be from a branch with the same name as this PR branch name so our CI can pair those PRs.

Here is an example of a paired e2e-test PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/654

alexeykazakov (approved):

PR #741, opened 2023-02-01
Add RHODS DSP labels to all user namespaces

VedantMahabaleshwarkar:
Adding the `modelmesh-enabled: "true"` and `opendatahub.io/dashboard: "true"` labels to all user namespaces across all tiers so that user's namespace is treated as a Data Science Project from a RHODS perspective. This allows users to use the Model Serving feature within RHODS

### DIFF
--- a/deploy/templates/nstemplatetiers/base/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/base/ns_dev.yaml
@@ -12,6 +12,9 @@ objects:
       openshift.io/requester: ${USERNAME}
     labels:
       name: ${USERNAME}-dev
+      # For RHODS: Allow user namespace to be treated as a DSP to enable Model Serving on this NS
+      modelmesh-enabled: "true" 
+      opendatahub.io/dashboard: "true"
     name: ${USERNAME}-dev
 
 # Role and RoleBindings for CRT administration (not associated with users)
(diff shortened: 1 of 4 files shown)


deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
```diff
@@ -14,6 +14,8 @@ objects:
       name: ${USERNAME}-tenant
       toolchain.dev.openshift.com/workspace: ${USERNAME}
       argocd.argoproj.io/managed-by: gitops-service-argocd
+      modelmesh-enabled: "true"
+      opendatahub.io/dashboard: "true"
```
alexeykazakov, line 18: Don't need it for the appstudio tier.
```suggestion
```

deploy/templates/nstemplatetiers/base/ns_dev.yaml
```diff
@@ -12,6 +12,8 @@ objects:
       openshift.io/requester: ${USERNAME}
     labels:
       name: ${USERNAME}-dev
+      modelmesh-enabled: "true"
```
alexeykazakov, line 15: Could you please add some comment explaining:
- that it's for RHODS
- what it does

deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
VedantMahabaleshwarkar: Removed it from the app studio tier

deploy/templates/nstemplatetiers/base/ns_dev.yaml
VedantMahabaleshwarkar: Added a comment with explanation

alexeykazakov (approved):

PR #747, opened 2023-02-16
Propagate SSO properties to UserAccount

sbryzak:
This PR modifies both UserSignupController and MasterUserRecordController to propagate the UserID and AccountID values through to the UserAccount resource.


alexeykazakov (commented):
Looks good but it seems that you missed the mur controller tests. I see only usersignup controller tests.

tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
 // +build tools
```
mfrancisc, line 2: probably we don't need this anymore, apparently was used for go version <= 1.16

mfrancisc (approved):

mfrancisc:
/retest

infra

PR #748, opened 2023-02-17
feat: add SpaceRequestConfig object in ToolchainConfig CRD

mfrancisc:
Add a new SpaceRequestConfig object that will store all configuration values for the Space Request controller (which is still under development).

Specifically add a flag to enable/disable the SpaceRequest controller, so that it can be enabled only for StoneSoup environments and disabled by default in DevSandbox envs.

This PR just adds the new type api version that has `SpaceRequestConfig` , the functionality to use this new object in host operator will be implemented once we start working at the basic skeleton for the SpaceRequest Controller.

Jira story: https://issues.redhat.com/browse/ASC-253

Related PRs:
- https://github.com/codeready-toolchain/api/pull/344


mfrancisc:
/retest

infra, cannot see logs

alexeykazakov (approved):

PR #754, opened 2023-02-27
enable selection of emails as per environment

ranakan19:
changes wrt https://issues.redhat.com/browse/CRT-1706

This PR makes the following changes:
- choose email templates based on notificationEnvironment
- notificationEnvironment is introduced in https://github.com/codeready-toolchain/api/pull/347
- replace bindata with go:embed to load static templates
- to test NotificationTemplateSetName for test added in https://github.com/codeready-toolchain/toolchain-common/pull/300
- add appstudio value in sandbox-sre to start sending out appstudio specific emails - https://github.com/codeready-toolchain/sandbox-sre/pull/958


controllers/toolchainconfig/deliveryserviceconfig.go
```diff
@@ -23,3 +23,7 @@ func (d DeliveryServiceFactoryConfig) GetMailgunSenderEmail() string {
 func (d DeliveryServiceFactoryConfig) GetMailgunReplyToEmail() string {
 	return d.Notifications().MailgunReplyToEmail()
 }
+
+//func (d DeliveryServiceFactoryConfig) GetNotificationEnvironment() string {
```
mfrancisc, line 27: probably we can delete this one if not needed

deploy/resources.go
```diff
@@ -0,0 +1,6 @@
+package deploy
+
+import "embed"
+
+//go:embed templates/notificationtemplates/*
+var Files embed.FS
```
mfrancisc, line 6: minor thing , but maybe we can name this var something more specific `notificationtemplateFS` or something similar ?

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -22,30 +30,40 @@
 }
 
 // GetNotificationTemplate returns a notification subject, body and a boolean
-// indicating whether or not a template was found. Otherwise, an error will be returned
-func GetNotificationTemplate(name string) (*NotificationTemplate, bool, error) {
-	templates, err := loadTemplates()
+// indicating whether a template was found. Otherwise, an error will be returned
+func GetNotificationTemplate(name string, notificationEnvironment string) (*NotificationTemplate, bool, error) {
+	templates, err := loadTemplates(notificationEnvironment)
```
mfrancisc, line 35: I was wondering if we could simplify this and just try and read the "subject" and "notification" templates directly:

```
    subject, err := deploy.Files.ReadFile("templates/notificationtemplates/" + notificationEnvironment + "/" + name + "/subject.txt")
	if err != nil {
		return nil, false, err
	}
	content, err := deploy.Files.ReadFile("templates/notificationtemplates/" + notificationEnvironment + "/" + name + "/notification.html")
	if err != nil {
		return nil, false, err
	}

	notificationTemplate := make(map[string]NotificationTemplate)
	notificationTemplate[name] = NotificationTemplate{
		Name:    name,
		Subject: string(subject),
		Content: string(content),
	}
	
	return &notificationTemplate, true, nil
```
and we can maybe drop all the logic that loops over the FS, checks for the existance of those 2 files and uses the directory name as name of the template. If I'm not missing anything , this may replace the usage of `templatesForAssets`, `getAllFilenames`, `getTemplatesFroNotificationEnvironment`, `loadTemplates`.

NOTE: this is just sample code, it may contain errors, plus the concatenation logic can be done better, and the constants can be defined somewhere else to avoid duplication.

deploy/resources.go
ranakan19: yes absolutely! that's a great name, Thanks!

pkg/templates/notificationtemplates/notification_generator.go
ranakan19: I like the idea, its more straight-forward for sure! I would still keep loading the templates in a map, so that we don't have to read from the file every time to get the template. Reading from the file only once and reading from the map thereafter should be faster. 
About the way templates are read and loaded, I agree that the current implementation is not the best and it depends on the length of the segmented file path. While I think reading the html and subject directly as you have suggested is a lot easier to understand, it is still dependent on the directory structure. In short, it is not completely flexible - and if there is any change in the folder structure the flow would break and the code would need to be updated. I'll update the dependency on structure it in the comments above the function.

If there is any other way anybody can think of that makes it completely flexible, I'd be happy to incorporate it.

deploy/resources.go
ranakan19: done in [fb7de08](https://github.com/codeready-toolchain/host-operator/pull/754/commits/fb7de08aadf0e35b6ffd34b707f24526645a3feb)

controllers/toolchainconfig/deliveryserviceconfig.go
ranakan19: done in [fb7de08](https://github.com/codeready-toolchain/host-operator/pull/754/commits/fb7de08aadf0e35b6ffd34b707f24526645a3feb)

deploy/templates/notificationtemplates/stonesoup/userprovisioned/notification.html
```diff
@@ -0,0 +1,67 @@
...
+  </p>
+
+  <p>
+    Join the Dev Sandbox community and engage with the Red Hat team from the #dev-sandbox channel on DevNation slack workspace.
+    You can join using the following invite - https://dn.dev/DevNationSlack. You can also reach us via email at {{.ReplyTo}} with any questions.
+  </p>
+
+  <p>
+    Thanks,<br />
+    The Developer Sandbox for Red Hat OpenShift team
```
alexeykazakov, line 63: Have you seen https://docs.google.com/document/d/11P8Gj8Q9gindbB9wmbFuIKv53SQL3b99vHUy0jemPfw/edit ?
Let's make the template as close to the proposed draft as possible. We will have to introduce more parameters like exact console link (instead of the reg-service URL) to make this template exactly as in the doc but we can do it in a different PR. For now let's just make it look very similar to what in the doc.

deploy/templates/notificationtemplates/stonesoup/userprovisioned/subject.txt
```diff
@@ -0,0 +1 @@
+Notice: Your Stone soup account is provisioned.
```
alexeykazakov, line 1: See https://docs.google.com/document/d/11P8Gj8Q9gindbB9wmbFuIKv53SQL3b99vHUy0jemPfw/edit

controllers/masteruserrecord/sync.go
```diff
@@ -221,7 +221,7 @@ func (s *Synchronizer) alignReadiness() (bool, error) {
 			_, err = notify.NewNotificationBuilder(s.hostClient, s.record.Namespace).
 				WithNotificationType(toolchainv1alpha1.NotificationTypeProvisioned).
 				WithControllerReference(s.record, s.scheme).
-				WithTemplate(notificationtemplates.UserProvisioned.Name).
+				WithTemplate(notificationtemplates.SandboxUserProvisioned.Name).
```
alexeykazakov, line 224: Why it's hardcoded to Sandbox instead of relying on the notification env from the config?

controllers/usersignup/usersignup_controller.go
```diff
@@ -736,7 +736,7 @@ func (r *Reconciler) sendDeactivatingNotification(logger logr.Logger, config too
 		}
 
 		notification, err := notify.NewNotificationBuilder(r.Client, userSignup.Namespace).
-			WithTemplate(notificationtemplates.UserDeactivating.Name).
+			WithTemplate(notificationtemplates.SandboxUserDeactivating.Name).
```
alexeykazakov, line 739: Same here. Why hardcoded to Sandbox?

controllers/usersignup/usersignup_controller.go
```diff
@@ -771,7 +771,7 @@ func (r *Reconciler) sendDeactivatedNotification(logger logr.Logger, config tool
 		}
 
 		notification, err := notify.NewNotificationBuilder(r.Client, userSignup.Namespace).
-			WithTemplate(notificationtemplates.UserDeactivated.Name).
+			WithTemplate(notificationtemplates.SandboxUserDeactivated.Name).
```
alexeykazakov, line 774: And here.

controllers/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -186,7 +186,7 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
 				Template:  "replyto",
 				Context:   notCtx,
 			},
-		})
+		}, "")
```
alexeykazakov, line 189: Maybe I'm missing something but I don't see any tests for the stone soup notification environment.

alexeykazakov (commented):
Is there only one notification template for the stonesoup env? Let's have the full set of all templates for every environment. Even if they are not currently used.

Also, let's not use "stonesoup". It's a code name which will change 100%.
While we don't really know the final name yet, let's stick with "appstudio" because it's kinda an internal name we already use in many places like the tier name or in APIs. Just for the sake of consistency.

controllers/masteruserrecord/sync.go
ranakan19: That is just the name of the template to be stored while creating the Notification. It does actually rely on the notification env when getting the template to fetch - https://github.com/ranakan19/host-operator/blob/emailConfig/controllers/notification/mailgun_notification_delivery_service.go#L77.

deploy/templates/notificationtemplates/stonesoup/userprovisioned/notification.html
ranakan19: added all four folders for appstudio here - [3df1eab](https://github.com/codeready-toolchain/host-operator/pull/754/commits/3df1eab38ec6e59824567e8d6c9d39bd6c43a3ff)

deploy/templates/notificationtemplates/stonesoup/userprovisioned/subject.txt
ranakan19: done in [3df1eab](https://github.com/codeready-toolchain/host-operator/pull/754/commits/3df1eab38ec6e59824567e8d6c9d39bd6c43a3ff)

controllers/masteruserrecord/sync.go
ranakan19: however the variable `SandboxUserProvisioned` was not needed, so i replaced the variables with constants in [3df1eab](https://github.com/codeready-toolchain/host-operator/pull/754/commits/3df1eab38ec6e59824567e8d6c9d39bd6c43a3ff). This is assuming that the template names remain the same for sandbox and appstudio. If we do however plan on having different templates in future, we can have separate constants for them.

controllers/usersignup/usersignup_controller.go
ranakan19: replaced with constant [3df1eab](https://github.com/codeready-toolchain/host-operator/pull/754/commits/3df1eab38ec6e59824567e8d6c9d39bd6c43a3ff)
Again, this is just template name, actual fetching based on notification env is done while sending.

controllers/usersignup/usersignup_controller.go
ranakan19: replaced with constant [3df1eab](https://github.com/codeready-toolchain/host-operator/pull/754/commits/3df1eab38ec6e59824567e8d6c9d39bd6c43a3ff)
Again, this is just template name, actual fetching based on notification env is done while sending.

controllers/notification/mailgun_notification_delivery_service_test.go
ranakan19: nope it is indeed missing, let me add that

deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
```diff
@@ -0,0 +1,73 @@
...
+  <p>
+    Dear xyz,
+  </p>
+
+  <p>
+    Welcome to Red Hat CI/CD! Red Hat CI/CD saves you weeks of time by automating the onboarding process for containerized apps with CI/CD. It also offers granular customization for pipelines, once they are running. And it provides unparalleled security for your software supply chain.
+  </p>
+
+  <p>
+    We have enabled your access, so to get started, you can now visit https://console.dev.redhat.com/hac/stonesoup and select + Create an application.
```
alexeykazakov, line 52: We will have to parameterize the console URL. Different environments (stage, prod, etc) will have different console URLs. Let's skip the URL for now.
For now we can use something like:
```suggestion
    We have enabled your access, so to get started, you can now visit Red Hat CI/CD console and select + Create an application.
```

alexeykazakov (commented):
Looks better!
There is one zombie: pkg/templates/notificationtemplates/testTemplates/fakeNotificationTemplates/appstudio/userprovisioned/test.txt

controllers/masteruserrecord/sync.go
alexeykazakov: The `Sandbox` in the template name confused me. It looks better now.

pkg/templates/notificationtemplates/notification_generator.go
mfrancisc: Keeping the map should not be a problem , even if I think (not 100% sure), that `embed.FS.ReadFile` is quite efficient since it may have the data already loaded in memory inside the global variable you define for the filesystem.

Regarding the flexibility, as we said, the `templatesForAssets` function will still need to be updated in case we change folder structure, since it uses indexes of the splitted path to get `directoryName` and `filename`.

Anyway, as discussed, I don't mind keeping the current implementation in case I'm the only one that has a different proposal for that area of the code  :)

deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
ranakan19: done in [10a70c1](https://github.com/codeready-toolchain/host-operator/pull/754/commits/10a70c1831306e3c08c37a3fa27740c5a8d66d0b)

pkg/templates/notificationtemplates/testTemplates/fakeNotificationTemplates/appstudio/userprovisioned/test.txt
```diff
@@ -0,0 +1 @@
+There should have been notification.html and subject.txt files in this repository instead.
```
alexeykazakov, line 1: What is this?

controllers/notification/mailgun_notification_delivery_service.go
```diff
@@ -107,7 +103,7 @@ func (s *MailgunNotificationDeliveryService) Send(notification *toolchainv1alpha
 		body = notification.Spec.Content
 	}
 
-	if subject == "" && body == "" {
+	if subject == "" || body == "" {
```
ranakan19, line 106: Changed this, because it made sense to return if either body or subject is empty.

controllers/notification/notification_controller.go
```diff
@@ -88,8 +88,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 	// if the environment is set to e2e do not attempt sending via mailgun
 	if config.Environment() != "e2e-tests" {
+		// get the notification environment
+		NotificationTemplateSetName := config.Notifications().NotificationTemplateSetName()
```
alexeykazakov, line 92:
```suggestion
		notificationTemplateSetName := config.Notifications().NotificationTemplateSetName()
```

controllers/notification/notification_controller.go
```diff
@@ -88,8 +88,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 		// Send the notification via the configured delivery service
-		err = r.deliveryService.Send(notification)
+		err = r.deliveryService.Send(notification, NotificationTemplateSetName)
```
alexeykazakov, line 94:
```suggestion
		err = r.deliveryService.Send(notification, notificationTemplateSetName)
```

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -21,31 +29,41 @@ type NotificationTemplate struct {
...
-	return &template, found, nil
+	if !found {
+		return &template, fmt.Errorf("notification template %v not found in %v", name, notificationTemplateSetName)
+	}
+	return &template, nil
 }
 
-func templatesForAssets(assets assets.Assets) (map[string]NotificationTemplate, error) {
-	paths := assets.Names()
+func templatesForAssets(efs embed.FS, root string, env string) (map[string]NotificationTemplate, error) {
```
alexeykazakov, line 47: I see you still use `env` as the template set name, right? Can we get it renamed to `setName` maybe?

alexeykazakov (approved):
Looks good overall! Just a few minor comments.

pkg/templates/notificationtemplates/notification_generator.go
ranakan19: done in aace73a

controllers/notification/notification_controller.go
ranakan19: done in aace73a

controllers/notification/notification_controller.go
ranakan19: done in aace73a

controllers/notification/mailgun_notification_delivery_service_test.go
```diff
@@ -186,11 +196,57 @@ func TestMailgunNotificationDeliveryService(t *testing.T) {
...
+		// when
+		mgds := NewMailgunNotificationDeliveryService(config, templateLoader, mockServerOption)
+		err := mgds.Send(&toolchainv1alpha1.Notification{
+			Spec: toolchainv1alpha1.NotificationSpec{
+				Recipient: "foo@bar.com",
+				Subject:   "test",
+				Content:   "abc",
+				Context:   notCtx,
+			},
+		}, notificationtemplates.AppstudioNotificationTemplateSetName)
```
MatousJobanek, line 247: this whole `TestMailgunNotificationDeliveryService` test can be changed to iterate over the set names:
https://gist.github.com/MatousJobanek/1496f2538a4ea71280f363594dfc1123

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -1,18 +1,26 @@
...
 	"github.com/pkg/errors"
 )
 
+const (
+	SandboxNotificationTemplateSetName   = "sandbox"
+	AppstudioNotificationTemplateSetName = "appstudio"
+	UserProvisionedTemplateName          = "userprovisioned"
+	UserDeactivatedTemplateName          = "userdeactivated"
+	UserDeactivatingTemplateName         = "userdeactivating"
+	IdlerTriggeredTemplateName           = "idlertriggered"
```
MatousJobanek, line 19: would it make sense to move it either to api repo or toolchain-common repo next to the `NotificationBuilder` so we use the constant also in member operator? https://github.com/codeready-toolchain/member-operator/blob/master/controllers/idler/idler_controller.go#L208
Can be done in separate PR :wink:

deploy/templates/notificationtemplates/appstudio/userdeactivating/notification.html
```diff
@@ -0,0 +1,58 @@
...
+       box-shadow: 0 2px 4px #d7d7d7;"
+>
+
+    <p>
+        You are receiving this email because your email account {{.UserEmail}} was provisioned.
+    </p>
+
+    <p>
+        Your account will expire in 3 days.  We recommend you save your work as all data will be
+        deleted upon expiry.  After deactivation, you can sign up again at any time for new access at {{.RegistrationURL}}.
```
MatousJobanek, line 49: out of curiosity - is it the plan to set the registration URL in ToolchainConfig to HAC? https://github.com/codeready-toolchain/sandbox-sre/blob/master/resources/stone-prd-host1.wdlc.p1.openshiftapps.com/configure/host/sandbox/toolchainconfig.yaml#L49

MatousJobanek (approved):
Looks great. Awesome job :1st_place_medal:

controllers/notification/notification_controller.go
```diff
@@ -88,8 +88,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 	// if the environment is set to e2e do not attempt sending via mailgun
 	if config.Environment() != "e2e-tests" {
+		// get the notification environment
+		notificationTemplateSetName := config.Notifications().NotificationTemplateSetName()
 		// Send the notification via the configured delivery service
-		err = r.deliveryService.Send(notification)
+		err = r.deliveryService.Send(notification, notificationTemplateSetName)
```
MatousJobanek, line 94: just a side note - it would be cool to separate the creation of templates & processing them from the actual sending over mailgun, so the first part 
https://github.com/codeready-toolchain/host-operator/blob/0ec0d89abee88a7527bd5a4648b6d3a979bcbf04/controllers/notification/mailgun_notification_delivery_service.go#L72-L108
could have been executed also as part of e2e tests and only the last part in the production
https://github.com/codeready-toolchain/host-operator/blob/0ec0d89abee88a7527bd5a4648b6d3a979bcbf04/controllers/notification/mailgun_notification_delivery_service.go#L110-L128
But let's do it as a separate story

deploy/templates/notificationtemplates/appstudio/userdeactivating/notification.html
alexeykazakov: Right. We need to set it to HAC URL instead. Similar to what we did for sandbox prod: https://github.com/codeready-toolchain/sandbox-sre/blob/master/resources/sandbox.x8i5.p1.openshiftapps.com/configure/host/sandbox/toolchainconfig.yaml#L50
In sandbox prod we set it to RHD.
In stonesoup we need to set it to HAC.

@ranakan19 let's add this `{{.RegistrationURL}}.` to the welcome email too! And we will update it in the configuration for the environments in sandbox-sre.

alexeykazakov: https://github.com/codeready-toolchain/sandbox-sre/pull/956

deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
```diff
@@ -0,0 +1,73 @@
...
+    Welcome to Red Hat CI/CD! Red Hat CI/CD saves you weeks of time by automating the onboarding process for containerized apps with CI/CD. It also offers granular customization for pipelines, once they are running. And it provides unparalleled security for your software supply chain.
+  </p>
+
+  <p>
+    We have enabled your access, so to get started, you can now visit Red Hat CI/CD console and select + Create an application.
+  </p>
+
+  <p>
+    If you’d rather access Red Hat CI/CD through your CLI, or if you have any questions that are not answered in the web UI, you can also visit our documentation at "dummy-link".
+  </p>
```
alexeykazakov, line 57: Let's modify the template to make it closer to https://docs.google.com/document/d/11P8Gj8Q9gindbB9wmbFuIKv53SQL3b99vHUy0jemPfw/edit
```suggestion
  <p>
    Welcome to Red Hat CI/CD! Red Hat CI/CD saves you weeks of time by automating the onboarding process for containerized apps with CI/CD. It also offers granular customization for pipelines, once they are running. And it provides unparalleled security for your software supply chain.
  </p>

  <p>
    We have enabled your access, so to get started, you can now visit {{.RegistrationURL}} and select + <b>Create an application</b>.
  </p>

  <p>
    If you’d rather access Red Hat CI/CD through your CLI, or if you have any questions that are not answered in the web UI, you can also visit our documentation at https://redhat-appstudio.github.io/docs.stonesoup.io/Documentation/main/index.html.
  </p>
```

deploy/templates/notificationtemplates/appstudio/userdeactivating/notification.html
alexeykazakov: And see my suggestion for the welcome email below.

deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
ranakan19: done in [29be8b4](https://github.com/codeready-toolchain/host-operator/pull/754/commits/29be8b41347898468009c261876303dde8b37305)

controllers/notification/notification_controller.go
ranakan19: wonderful idea, created a new story - https://issues.redhat.com/browse/CRT-1736

pkg/templates/notificationtemplates/notification_generator.go
ranakan19: yeah, it makes absolute sense

controllers/notification/mailgun_notification_delivery_service_test.go
ranakan19: done in [29be8b4](https://github.com/codeready-toolchain/host-operator/pull/754/commits/29be8b41347898468009c261876303dde8b37305)

pkg/templates/notificationtemplates/notification_generator.go
ranakan19: oops, i thought i fixed that already. done in [29be8b4](https://github.com/codeready-toolchain/host-operator/pull/754/commits/29be8b41347898468009c261876303dde8b37305) now

alexeykazakov (approved):

deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
ranakan19: @alexeykazakov I can also add the First name and Last Name as specified in the [doc](https://docs.google.com/document/d/11P8Gj8Q9gindbB9wmbFuIKv53SQL3b99vHUy0jemPfw/edit), since those values are already available [here](https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/notification/notification_builder.go#L145-L146) by passing [userContext](https://github.com/codeready-toolchain/host-operator/blob/master/controllers/masteruserrecord/sync.go#L225).

Also, just noticed that the content of the email is not exactly same. 

> _**Red Hat CI/CD**_ saves you weeks of time by automating the onboarding process for containerized apps with CI/CD. It also offers granular customization for _**pipelines, once they are running**_. And it provides unparalleled security for your software supply chain.

vs (provided in the doc):
> _**This new platform**_ saves you weeks of time by automating the onboarding process for containerized apps with CI/CD. It also offers granular customization for _**build pipelines**_. And it provides unparalleled security for your software supply chain.

Should I make the change as expected in the doc or keep the change? (Personally i don't mind the difference, and can follow up with the HAC team if we keep the current changes.)

xcoulon (commented):
you should also delete the https://github.com/codeready-toolchain/host-operator/blob/master/tools.go file and update `go.mod` to get rid of the dependency to `github.com/go-bindata/go-bindata`

MatousJobanek:
> you should also delete the https://github.com/codeready-toolchain/host-operator/blob/master/tools.go file and update go.mod to get rid of the dependency to github.com/go-bindata/go-bindata

@xcoulon why? the go-bindata is still used for the other templates. Kanika changed it only for email-related files.

xcoulon:
> > you should also delete the https://github.com/codeready-toolchain/host-operator/blob/master/tools.go file and update go.mod to get rid of the dependency to github.com/go-bindata/go-bindata
> 
> @xcoulon why? the go-bindata is still used for the other templates. Kanika changed it only for email-related files.

my bad, I thought it would be replaced everywhere, but I forgot about the other places where it's used 🤦‍♂️
maybe we could/should have replaced `go-bindata` with `go:embed` everywhere in the repo in a dedicated PR before dealing with new templates?

alexeykazakov:
> maybe we could/should have replaced go-bindata with go:embed everywhere in the repo in a dedicated PR before dealing with new templates?

I think it's fine to start using go:embed for new stuff and migrate the existing stuff in separate PRs later.

PR #755, opened 2023-03-01
refactor: ignore NSTemplateSet creation failure if already exists

xcoulon:
there may be a delay between the creation on the member cluster and the client update on the host cluster, so failing at the second attempt should not raise an error

Fixes https://issues.redhat.com/browse/CRT-1729 

### DIFF
--- a/controllers/space/space_controller.go
+++ b/controllers/space/space_controller.go
@@ -342,7 +342,7 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
 				if errors.IsAlreadyExists(err) {
 					// requeue immediately, there's probably a race condition between the host client cache and the member cluster state
-					return nsTmplSet, norequeue, nil
+					return nsTmplSet, requeueDelay, nil
 				}
 				return nsTmplSet, norequeue, r.setStatusNSTemplateSetCreationFailed(logger, space, err)
 			}
--- a/controllers/space/space_controller.go
+++ b/controllers/space/space_controller.go
@@ -340,7 +340,10 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			}
 			nsTmplSet = NewNSTemplateSet(memberCluster.OperatorNamespace, space, spaceBindings.Items, tmplTier)
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
-				logger.Error(err, "failed to create NSTemplateSet on target member cluster")
+				if errors.IsAlreadyExists(err) {
+					// move to next part but don't incremenent space count again (assuming it was already incremented before)
+					goto nstmpl_exists
+				}
 				return nsTmplSet, norequeue, r.setStatusNSTemplateSetCreationFailed(logger, space, err)
 			}
 			logger.Info("NSTemplateSet created on target member cluster")
@@ -350,6 +353,7 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 		}
 		return nsTmplSet, norequeue, r.setStatusNSTemplateSetCreationFailed(logger, space, err)
 	}
+nstmpl_exists:
 	logger.Info("NSTemplateSet already exists")
 
 	_, found := condition.FindConditionByType(nsTmplSet.Status.Conditions, toolchainv1alpha1.ConditionReady)
--- a/controllers/space/space_controller.go
+++ b/controllers/space/space_controller.go
@@ -341,8 +341,8 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			nsTmplSet = NewNSTemplateSet(memberCluster.OperatorNamespace, space, spaceBindings.Items, tmplTier)
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
 				if errors.IsAlreadyExists(err) {
-					// move to next part but don't incremenent space count again (assuming it was already incremented before)
-					goto nstmpl_exists
+					// requeue immediately, there's probably a race condition between the host client cache and the member cluster state
+					return nsTmplSet, norequeue, nil
 				}
 				return nsTmplSet, norequeue, r.setStatusNSTemplateSetCreationFailed(logger, space, err)
 			}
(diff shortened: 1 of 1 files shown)


controllers/space/space_controller.go
```diff
@@ -340,7 +340,10 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			}
 			nsTmplSet = NewNSTemplateSet(memberCluster.OperatorNamespace, space, spaceBindings.Items, tmplTier)
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
-				logger.Error(err, "failed to create NSTemplateSet on target member cluster")
+				if errors.IsAlreadyExists(err) {
+					// move to next part but don't incremenent space count again (assuming it was already incremented before)
+					goto nstmpl_exists
```
xcoulon, line 345: or if the usage of `goto` is not a good idea, we could do something like this:
```
return nsTmplSet, norequeue, nil
```
to immediately trigger another reconcile loop without changing the current status (ie, no call to `r.setStatusProvisioning(space)` since it should already be the in this status).

controllers/space/space_controller.go
```diff
@@ -340,7 +340,10 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			}
 			nsTmplSet = NewNSTemplateSet(memberCluster.OperatorNamespace, space, spaceBindings.Items, tmplTier)
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
-				logger.Error(err, "failed to create NSTemplateSet on target member cluster")
+				if errors.IsAlreadyExists(err) {
```
mfrancisc, line 343: dumb question: how can the `nstemplateset` exist immediately after the `client.Get` failed to find it? Is there any other controller that creates `nstemplateset` resources or concurrent process of this controller ?

xcoulon: not a dumb question! the NSTemplateSet is directly created on the target member cluster whereas this code runs on the host cluster, so I suspect it's a race condition

controllers/space/space_controller.go
mfrancisc: yeah it doesn't seem a good practice to use `goto` . I remember it was a bad practice since I was learning COBOL 😄 . 

But I think 1 single `goto` is not bad, better then reconciling IMO! 👍

controllers/space/space_controller.go
mfrancisc: Ha I was rereading your description of the PR, so you think it get's created but the cache for the member client in the host operator it's not yet updated with this new object , if I well understood ?

controllers/space/space_controller.go
xcoulon: yes indeed, we don't use `goto` anywhere else in our codebase, which is why I'm a bit reluctant, but I thought it could solve the problem without too much refactoring. As I said, the alternative would be to return and trigger a new reconcile loop, as we do in other cases

controllers/space/space_controller.go
xcoulon: yes, that's my assumption, indeed 🤷‍♂️

controllers/space/space_controller.go
mfrancisc: Checking what happens after `nstmpl_exists` I think it will still be reconciled since the `ConditionReady` will not be present in the current object right?

```
_, found := condition.FindConditionByType(nsTmplSet.Status.Conditions, toolchainv1alpha1.ConditionReady)
// skip until there's a `Ready` condition
if !found {
     // just created, but there is no `Ready` condition yet
     return nsTmplSet, requeueDelay, nil
}
```

xcoulon: ah right, good point! So I think it would be safer to use `return nsTmplSet, norequeue, nil` instead of `goto` and make sure we ignore the creation in the subsequent loops and move on properly.

xcoulon: refactored in 5aecb03

controllers/space/space_controller.go
```diff
@@ -340,7 +340,10 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
...
+					// requeue immediately, there's probably a race condition between the host client cache and the member cluster state
+					return nsTmplSet, norequeue, nil
```
mfrancisc, line 345:
```suggestion
					return nsTmplSet, requeueDelay, nil
```
should it return requeueDelay instead of norequeue ?

xcoulon: why? I'm not sure we need to wait any longer. Unless we need to give time to the client to update its cache?

mfrancisc: Sorry, my suggestion was incorrect as well!

Don't you need to return an error at least, so that the reconcile is triggered ?

xcoulon: you're right, I did not realize that `norequeue` was the wrong value to return here. I read that its value was zero but I missed the check upstream 🤦‍♂️ 
I applied your suggestion in 1af8c57

xcoulon: thanks for your patience @mfrancisc !

mfrancisc (approved):
Looks good! 👍 

controllers/space/space_controller.go
mfrancisc: No worries, it was unclear also for me what happened with the upstream functions !

MatousJobanek:
> there may be a delay between the creation on the member cluster and the client update on the host cluster, so failing at the second attempt should not raise an error

out of curiosity - is it really a problem? if the error is returned, then it just reconciles again, right?

controllers/space/space_controller.go
```diff
@@ -340,7 +340,10 @@ func (r *Reconciler) manageNSTemplateSet(logger logr.Logger, space *toolchainv1a
 			}
 			nsTmplSet = NewNSTemplateSet(memberCluster.OperatorNamespace, space, spaceBindings.Items, tmplTier)
 			if err := memberCluster.Client.Create(context.TODO(), nsTmplSet); err != nil {
-				logger.Error(err, "failed to create NSTemplateSet on target member cluster")
+				if errors.IsAlreadyExists(err) {
+					// requeue immediately, there's probably a race condition between the host client cache and the member cluster state
+					return nsTmplSet, requeueDelay, nil
```
MatousJobanek, line 345: you wrote 
> // requeue immediately

but 
```
requeueDelay = 1 * time.Second
```

xcoulon: granted, the wording was confusing so I removed the "immediately" word in the sentence in 26423d6

xcoulon:
> > there may be a delay between the creation on the member cluster and the client update on the host cluster, so failing at the second attempt should not raise an error
> 
> out of curiosity - is it really a problem? if the error is returned, then it just reconciles again, right?

no really a problem, except it pollutes the logs

xcoulon:
> Thanks for the answer. Does it happen often? For example compared to UserAccounts? it's the very same use-case, right?

not sure to be honest, but I noticed some occurrences last week

MatousJobanek (approved):

PR #759, opened 2023-03-06
fix: preferred cluster disrupts balanced provisioning of users

MatousJobanek:
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1677703479890809


pkg/capacity/manager.go
```diff
@@ -104,6 +104,10 @@ func (b *ClusterManager) GetOptimalTargetCluster(optimalClusterFilter OptimalTar
 	}
 	optimalTargetClusters := getOptimalTargetClusters(optimalClusterFilter.PreferredCluster, b.getMemberClusters, optimalClusterFilter.ClusterRoles, hasNotReachedMaxNumberOfSpacesThreshold(config, counts), hasEnoughResources(config, status))
 
+	if len(optimalTargetClusters) == 1 {
+		return optimalTargetClusters[0], nil
+	}
+
```
mfrancisc, line 110: sorry , but I was trying to understand better. If there is only 1 cluster it will still used that right? Is this change needed so that it doesn't go trough the logic of checking if usage of cluster is a multiple of 50 or sorting the clusters (which is not useful in this case since only 1 choice) ?

MatousJobanek: two things:
* there is no point in going through the logic when there is only one cluster selected
* second thing, which is more important, is that the users with the preferred clusters disrupted the balanced provisioning - consider this situation:
1. based on the load the logic decides that the following 50 users should go to m3
2. m3 cluster is stored as `lastUsed`
3. the next user has m2 cluster set as the preferred one
4. it goes through the logic and sets m2 as `lastUsed`, because it looks like there is nothing else we could choose from
5. the next 50 users go to m2 instead of m3.
6. this may happen several times in a row, so most of the users go to m2 instead of m3 as we see in Grafana right now

mfrancisc: Ok that makes sense, Thanks! 

So this avoids updating the `lastUsed` field when there is only one cluster returned (which most probably is because the user has already a preferred cluster, it's basically a reprovisioning). While "new" users will end up in the cluster that has more room at the time they signup up.

Just one more question at this point I'm not sure I totally understood why do we need this part here https://github.com/codeready-toolchain/host-operator/blob/69fcf1ec7688a2974b7cdd14185a2076407f47b2/pkg/capacity/manager.go#L108-L113

I mean if there is a preferred cluster the user will end up there , otherwise it will end up on the cluster that has more room according to the logic of batching 50 users at the time ? But probably I'm missing something .

rajivnathan: I believe we need that because if we don't then we could get into a situation where we always provision to `b.lastUsed`. That logic is essentially saying if the provisioned count is a multiple of 50 then lets recalculate to see if `b.lastUsed` needs to be changed.

rajivnathan (commented):
I don't think the unit tests are covering the ratio aspect of the balancing logic:
`return float64(provisioned1By50)/float64(threshold1) < float64(provisioned2By50)/float64(threshold2)`

The tests have
```
member2CurrentCount := numberOfSpaces
member3CurrentCount := numberOfSpaces
```
so they always start off with the same number of workspace right?

Do you think it would be beneficial to include a test where there's a large disparity ie. mimicking the situation when we have one new empty cluster and another that is nearly full?

rajivnathan (approved):
Just the 1 question about coverage of the ratio aspect of the balancing logic but I believe the PR would solve the problem 👍

alexeykazakov (approved):

pkg/capacity/manager.go
MatousJobanek: yes, what it technically does is that it checks if any of the available clusters is the `lastUsed` - if it is, then it checks what is the current number of users modulo 50. 
* If it's not 0 (the current number of users is not a multiply of 50), then it uses the `lastUsed`. 
* if it is 0, then it continues with the balancing logic to pick the freest cluster for the next batch of 50

In other words, it should make sure that the users are provisioned (more or less) in batches of 50. Why "more or less"? Because it may happen that a user gets deactivated in the cluster,so the batch can be more or less than 50. I decided to not care about such situations.

mfrancisc: > That logic is essentially saying if the provisioned count is a multiple of 50 then lets recalculate to see if b.lastUsed needs to be changed

I think I got it now , but the way I see it (if I'm not wrong) it's actually the opposite, in other words: we avoid recalculating and potentially picking selecting a "new" `lastUsed` if the current one is not multiple of 50, even if there may be a cluster that has less users at the moment. But we want to be sure we get to a number of users which is a multiple of 50 before we switch cluster.

Thanks both for the extra details.

mfrancisc (approved):

PR #762, opened 2023-03-09
PLNSRVCE-1166: add proxyplugins crd to registration-service role

gabemontero:
see test results analysis in

https://github.com/codeready-toolchain/registration-service/pull/311#issuecomment-1462382013
https://github.com/codeready-toolchain/registration-service/pull/311#issuecomment-1462185209

/assign @alexeykazakov 
/assign @rajivnathan

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -37,6 +37,7 @@ objects:
           - spaces
           - toolchainconfigs
           - toolchainstatuses
+          - proxyplugins
         verbs:
           - get
           - list


alexeykazakov (approved):
The PR itself looks good. However you have to use another branch for it.
You can't pair multiple PRs (== use the same branch name) with the same e2e-test PR. Our tools will block it. The reason is that we do not want to allow deploying different components in the same time when they depends on each other.

Move this PR to a different branch (don't forget to delete the branch from your fork!). If e2e-tests pass we will merge it.
Then we will be able to re-ran the other PRs tests and merge them when green.

gabemontero:
> The PR itself looks good. However you have to use another branch for it. You can't pair multiple PRs (== use the same branch name) with the same e2e-test PR. Our tools will block it. The reason is that we do not want to allow deploying different components in the same time when they depends on each other.
> 
> Move this PR to a different branch (don't forget to delete the branch from your fork!). If e2e-tests pass we will merge it. Then we will be able to re-ran the other PRs tests and merge them when green.
 

my guess is https://github.com/codeready-toolchain/toolchain-common/pull/302 will fall under the same situation

alexeykazakov:
> my guess is https://github.com/codeready-toolchain/toolchain-common/pull/302 will fall under the same situation

Not really. Common or api are used as dependencies / go modules only. They are not deployed to our clusters on their own. We have only host, member operators and reg-service which are deployble and can/have to be mapped to e2e-tests.

PR #765, opened 2023-03-13
Edit text and add image to app studio  welcome email

Chr1st1anSears:


deploy/templates/notificationtemplates/appstudio/userprovisioned/notification.html
```diff
@@ -68,10 +64,15 @@
     Thank you for your interest. We look forward to having your feedback about Red Hat CI/CD.
   </p>
 
+  <p>
+    <img src="hand_wave_big.png" alt="Hand wave">
```
alexeykazakov, line 68: I'm not sure if it will work as is in mailgun. Something we need to try.
@ranakan19 could you please take a look when you have time.

Chr1st1anSears: @alexeykazakov @ranakan19 I wanted to follow up on this PR and see if we could implement these changes. Ran, would you please let us know if we can include this image that Masha made as I've coded it in the message?

ranakan19: @Chr1st1anSears Sorry for the late reply. I can confirm that the current change is not enough to render the wave sign in the emails as expected.
We will need to use the AddInline [function](https://pkg.go.dev/github.com/BoltApp/mailgun-go#Message.AddInline) to send the image via mailgun. However to do that, we'd need to make some changes in the backend, either to add a special case for app-studio provisioning email to send images or a general way to send images whenever present. 
cc: @alexeykazakov

Chr1st1anSears: Understood--thanks for letting me know. I'll close this PR for now.

PR #766, opened 2023-03-14
Allow create access for SelfSubjectAccessReview

rottencandy:
Add create permission for `SelfSubjectAccessReview` to allow for RBAC checks in UI.

Test repo PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/681

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -158,6 +158,12 @@ objects:
     - watch
     - update
     - patch
+  - apiGroups:
+    - authorization.k8s.io
+    resources:
+    - selfsubjectaccessreview
+    verbs:
+    - create
 
 # RoleBinding that grants limited CRUD permissions to the User
 - apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
```diff
@@ -158,6 +158,12 @@ objects:
     - watch
     - update
     - patch
+  - apiGroups:
+    - authorization.k8s.io
+    resources:
+    - selfsubjectaccessreview
```
MatousJobanek, line 164: correct me if I'm wrong, but it should be in the plural form, right?:
```suggestion
    - selfsubjectaccessreviews
```

MatousJobanek: btw, is `selfsubjectaccessreview`  the only one that is used by UI, or does it also use SubjectAccessReview?

alexeykazakov: Yes, it should be in the plural form indeed! I keep missing it when reviewing those PRs :(

rottencandy: Updated!

The UI doesn't need to check access for users other than the current user, so SubjectAccessReview is not needed.

MatousJobanek (approved):

PR #783, opened 2023-03-31
ResourceQuota for Appstudion CRDs

MartinBasti:
Resource quota to prevent making unlimited number of CRs in namespace and causing degradation of service.

This is very initial change with high number of entries, this will be adjusted (lowered) in future.

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
@@ -103,6 +103,57 @@ objects:
       requests.ephemeral-storage: 50Gi
       count/persistentvolumeclaims: "12"
 - apiVersion: v1
+  kind: ResourceQuota
+  metadata:
+    name: appstudio-crds
+    namespace: ${USERNAME}-tenant
+  spec:
+    hard:
+      count/applications.appstudio.redhat.com: "512"
+      count/buildpipelineselectors.appstudio.redhat.com: "512"
+      count/componentdetectionqueries.appstudio.redhat.com: "512"
+      count/components.appstudio.redhat.com: "512"
+      count/deploymenttargetclaims.appstudio.redhat.com: "512"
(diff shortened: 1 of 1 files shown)


deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
```diff
@@ -102,6 +102,36 @@ objects:
       requests.storage: 50Gi
       requests.ephemeral-storage: 50Gi
       count/persistentvolumeclaims: "12"
+- apiVersion: v1
```
alexeykazakov, line 105: I would split this one big quota to multiple quotas. The RQ.status has a filed which is updated when the actual quota usage is changed: `used`. So every time some of the resources in the quota is created the corresponding RQ is updated. If we have too many items in the quota it will lead to too many conflicts. Basically there might be a high rate of errors when a CR from this quota is created.
It can be mitigated by creating multiple quotas. Let's say one quota for SPI resources, one for HAS, one for release, etc.

deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
```diff
@@ -102,6 +102,57 @@ objects:
...
+  kind: ResourceQuota
+  metadata:
+    name: appstudio-crds
```
MartinBasti, line 108: I still have to split this group, but trying to figure out how

deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
dirgim: Thanks @alexeykazakov we created a follow-up PR #786. Please let us know your thoughts on it

dirgim:
Created a follow-up PR #786

alexeykazakov:
Replaced by https://github.com/codeready-toolchain/host-operator/pull/786

PR #784, opened 2023-04-03
Generated captcha configuration files

rajivnathan:
Paired with https://github.com/codeready-toolchain/api/pull/355


deploy/registration-service/registration-service.yaml
```diff
@@ -130,13 +130,21 @@ objects:
...
                 - name: WATCH_NAMESPACE
                   value: ${NAMESPACE}
               resources:
                 requests:
                   cpu: "50m"
                   memory: "100M"
+          volumes:
+            - name: captcha-secret
+              secret:
+                secretName: captcha
```
alexeykazakov, line 147: If we introduce this dependency to a secret here then it will cause a registration service deployment failure OOTB if there is no such secret, correct?

Ideally we would want to load the secret only if the captcha is enabled. Can we add the captcha secret name to the toolchainconfig and load that secret via our [ToolchainConfig ](https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go)?

rajivnathan: Hmm, the secret change was only supposed to be part of my prototype changes. I'm not sure how it got here. Thanks for catching it! Will remove it, it should be added in a later PR.

rajivnathan: Removed

alexeykazakov (approved):

mfrancisc (approved):

PR #787, opened 2023-04-07
Use compliant username in segment event

alexeykazakov:
Currently there is mismatch between `userID` field in events sent to segment from web-console and Sandbox backend (host operator).
In our activation event we use the sso username while web-console uses the `User.metadata.name` which is the complaint username.
As a temporal solution we would like to start using the compliant username in the host operator too but unfortunately we don't have it available in the UserSignup when it's approved.

This PR does two things:
- Move the activation event report to the place where we create a MUR CR for the approved UserSignup instead of reporting it when it's approved.
- Start using the compliant username instead of the sso username.

See more details in slack: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1680830236435579

We still need to get confirmation from Brad and Jake that it's worth doing though.

### DIFF
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -502,12 +502,6 @@ func (r *Reconciler) updateUserSignupMetricsByState(logger logr.Logger, userSign
 	switch newState {
 	case toolchainv1alpha1.UserSignupStateLabelValueApproved:
 		metrics.UserSignupApprovedTotal.Inc()
-		// track activation in Segment
-		if r.SegmentClient != nil {
-			r.SegmentClient.TrackAccountActivation(userSignup.Spec.Username, userSignup.Spec.Userid, userSignup.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey])
-		} else {
-			logger.Info("segment client not configure to track account activations")
-		}
 	case toolchainv1alpha1.UserSignupStateLabelValueDeactivated:
 		if oldState == toolchainv1alpha1.UserSignupStateLabelValueApproved {
 			metrics.UserSignupDeactivatedTotal.Inc()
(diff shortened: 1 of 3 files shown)


alexeykazakov:
infra issue

/retest

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -353,7 +353,7 @@ func TestUserSignupWithAutoApprovalWithoutTargetCluster(t *testing.T) {
 	AssertMetricsCounterEquals(t, 0, metrics.UserSignupDeactivatedTotal)
 	AssertMetricsCounterEquals(t, 1, metrics.UserSignupApprovedTotal)
 	AssertMetricsCounterEquals(t, 1, metrics.UserSignupUniqueTotal)
-	segmenttest.AssertMessageQueued(t, r.SegmentClient, userSignup, segment.AccountActivated)
+	segmenttest.AssertMessageQueuedForProvisionedMur(t, r.SegmentClient, userSignup, murtest.AssertThatMasterUserRecord(t, userSignup.Name, r.Client).Get())
```
MatousJobanek, line 356: in this case, it's enough to move it a few lines below because at line 358 there is the assertion of the MUR

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -1598,7 +1597,6 @@ func TestUserSignupMUROrSpaceOrSpaceBindingCreateFails(t *testing.T) {
 			assert.Equal(t, "approved", userSignup.Labels[toolchainv1alpha1.UserSignupStateLabelKey])
 			AssertMetricsCounterEquals(t, 1, metrics.UserSignupApprovedTotal)
 			AssertMetricsCounterEquals(t, 1, metrics.UserSignupUniqueTotal)
-			segmenttest.AssertMessageQueued(t, r.SegmentClient, userSignup, segment.AccountActivated)
 		})
```
MatousJobanek, line 1600: also here
```suggestion
			segmenttest.AssertNoMessageQueued(t, r.SegmentClient)
		})
```
and in all other cases below

controllers/usersignup/usersignup_controller_test.go
alexeykazakov: Done in https://github.com/codeready-toolchain/host-operator/pull/787/commits/a60a2d65c75591ff5fa145824a2bb1e3911e06c7

controllers/usersignup/usersignup_controller.go
```diff
@@ -600,6 +594,13 @@ func (r *Reconciler) provisionMasterUserRecord(logger logr.Logger, config toolch
 	domain := metrics.GetEmailDomain(mur)
 	counter.IncrementMasterUserRecordCount(logger, domain)
 
+	// track the MUR creation as an account activation event in Segment
+	if r.SegmentClient != nil {
+		r.SegmentClient.TrackAccountActivation(compliantUsername, userSignup.Spec.Userid, userSignup.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey])
+	} else {
+		logger.Info("segment client not configure to track account activations")
```
xcoulon, line 601:
```suggestion
		logger.Info("segment client not configured to track account activations")
```

xcoulon: just a minor typo I made in the initial version, but not critical

xcoulon (approved):

MatousJobanek (approved):

PR #788, opened 2023-04-13
Generated files for additional captcha configuration

rajivnathan:
Paired with https://github.com/codeready-toolchain/api/pull/356


alexeykazakov (approved):

mfrancisc (approved):
👍 

I was looking for the `go.mod` change but I think it's 100% safe to avoid adding that temporary replace statement since it's all new configuration.

PR #791, opened 2023-04-18
fix: decrement counter when NSTemplateSet is deleted

MatousJobanek:
decrement counter when NSTemplateSet is deleted and not when finalizer is removed, otherwise, we can have wrong values in the counter when space is moved from one member cluster to another

paired PR and also the case when I faced the issue with wrong numbers: https://github.com/codeready-toolchain/toolchain-e2e/pull/707

### DIFF
--- a/controllers/space/space_controller_test.go
+++ b/controllers/space/space_controller_test.go
@@ -558,19 +558,10 @@ func TestDeleteSpace(t *testing.T) {
 			spacetest.WithSpecTargetCluster("member-1"),
 			spacetest.WithStatusTargetCluster("member-1"))
 		nsTmplSet := nstemplatetsettest.NewNSTemplateSet("oddity", nstemplatetsettest.WithReadyCondition())
-		InitializeCounters(t,
-			NewToolchainStatus(
-				WithMetric(toolchainv1alpha1.UserSignupsPerActivationAndDomainMetricKey, toolchainv1alpha1.Metric{
-					"1,internal": 1,
-				}),
-				WithMetric(toolchainv1alpha1.MasterUserRecordsPerDomainMetricKey, toolchainv1alpha1.Metric{
-					string(metrics.Internal): 1,
-				}),
-				WithMember("member-1", WithSpaceCount(1)),
-			))
 
 		t.Run("Space controller deletes NSTemplateSet", func(t *testing.T) {
 			// given
+			InitializeCounters(t, NewToolchainStatus(WithEmptyMetrics(), WithMember("member-1", WithSpaceCount(1))))
 			hostClient := test.NewFakeClient(t, s, basicTier)
 			memberClient := test.NewFakeClient(t, nsTmplSet)
 			member := NewMemberClusterWithClient(memberClient, "member-1", corev1.ConditionTrue)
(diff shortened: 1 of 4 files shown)


controllers/space/space_controller_test.go
```diff
@@ -687,20 +678,19 @@ func TestDeleteSpace(t *testing.T) {
...
 			hostClient := test.NewFakeClient(t, s, basicTier)
-			nstmplSet := nstemplatetsettest.NewNSTemplateSet("delete-space", nstemplatetsettest.WithReadyCondition())
-			member1Client := test.NewFakeClient(t, nstmplSet)
-			member1Client.MockDelete = mockDeleteNSTemplateSet(member1Client.Client)
+			member1Client := test.NewFakeClient(t)
 			member1 := NewMemberClusterWithClient(member1Client, "member-1", corev1.ConditionTrue)
 			member2 := NewMemberClusterWithTenantRole(t, "member-2", corev1.ConditionTrue)
 			ctrl := newReconciler(hostClient, member1, member2)
-			InitializeCounters(t, NewToolchainStatus(WithMember("member-1", WithSpaceCount(1))))
+			InitializeCounters(t, NewToolchainStatus(WithEmptyMetrics(), WithMember("member-1", WithSpaceCount(1))))
```
mfrancisc, line 693: just curious - was `NewToolchainStatus` failing with some error when called without this new function `WithEmptyMetrics`. 

Basically I'm trying to understand why this new function is now needed ?

controllers/space/space_controller_test.go
```diff
@@ -716,7 +706,7 @@ func TestDeleteSpace(t *testing.T) {
 			spacetest.AssertThatSpace(t, s.Namespace, s.Name, hostClient).
 				DoesNotExist()
 			AssertThatCountersAndMetrics(t).
-				HaveSpacesForCluster("member-1", 0) // space counter is decremented when there are no finalizers on the space resource that is being deleted
+				HaveSpacesForCluster("member-1", 1) // space counter is not decremented
```
mfrancisc, line 709: probably I'm missing something - but how can the counter not be decremented when Space is deleted ?

MatousJobanek: because this is the last step when the NSTemplateSet is gone and we just need to remove the finalizer.
We count the representation of the Spaces in the member clusters - the representation is NSTemplateSet, so we need to count when we create or delete it

controllers/space/space_controller_test.go
MatousJobanek: There was an issue with the initialization of the counters - in these tests, we want to initialize the counter based on the values provided in the ToolchainStatus, but if the ToolchainStatus doesn't contain some metrics, then it doesn't initialize from the ToolchainStatus, but based on all present resources of the given types in the cluster https://github.com/codeready-toolchain/host-operator/blob/34b494234cf66da80e289113fde7d5e1c0b169a5/pkg/counter/cache.go#L236-L242

So, I would need to create it with:
```
	InitializeCounters(t,
			NewToolchainStatus(
				WithMetric(toolchainv1alpha1.UserSignupsPerActivationAndDomainMetricKey, toolchainv1alpha1.Metric{
					"1,internal": 1,
				}),
				WithMetric(toolchainv1alpha1.MasterUserRecordsPerDomainMetricKey, toolchainv1alpha1.Metric{
					string(metrics.Internal): 1,
				}),
				WithMember("member-1", WithSpaceCount(1)),
			))

```  
but I was tool lazy to copy-paste the whole code, so I thought that the simple function will be easier to use.

mfrancisc: Ok, I see!  So previously this specific case at line 703 was initializing from resources and not from ToolchainStatus, which is the desired behavior.

controllers/space/space_controller_test.go
mfrancisc: ok , so in this case the counter was already decremented , the test now mainly ensures that the finalizer is correctly removed from the Space. 
I was just a bit confused by seeing the sapce counter initialized with 1 and not decremented.

mfrancisc (approved):
Thanks for fixing this 🏅  and for checking my comments!

xcoulon (approved):

PR #792, opened 2023-04-21
fix: Don't block Space deletion if it hasn't been provisiond to member cluster yet

mfrancisc:
Delete spaces with invalid `space.Spec.TargetCluster` if it wasn't provisioned before (meaning `space.Status.TargetCluster` is empty)

Jira bug: https://issues.redhat.com/browse/CRT-1775

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/711


controllers/space/space_controller.go
```diff
@@ -477,8 +477,12 @@ func extractUsernames(role string, bindings []toolchainv1alpha1.SpaceBinding) []
...
 	logger.Info("terminating Space")
 	if isBeingDeleted, err := r.deleteNSTemplateSet(logger, space); err != nil {
-		logger.Error(err, "failed to delete the NSTemplateSet")
-		return r.setStatusTerminatingFailed(logger, space, err)
+		// space was already provisioned to a cluster
+		// let's not proceed with deletion
+		if space.Status.TargetCluster != "" {
+			logger.Error(err, "failed to delete the NSTemplateSet")
+			return r.setStatusTerminatingFailed(logger, space, err)
+		}
```
MatousJobanek, line 485: what if it fails for any other reason even when the targetCluster is not set. Could be a temp glitch at the server side, or anything else. In other words - we should always return the error. 
Shouldn't this check (of the emptiness of the targetCluster) be done rather as part of the `deleteNSTemplateSet` function?

mfrancisc: Probably I've missed something, but following is my analysis :

the potential errors that are currently forwarded by `deleteNSTemplateSet` from `deleteNSTemplateSetFromCluster` are the following:

1. _cannot delete NSTemplateSet: unknown target member cluster: bla bla_
2. _something went wrong while trying to GET the NSTemplateSet resource_
3. _NSTemplateSet deletion has not completed in over 1 minute_ 
4. _something wrong happened while trying to DELETE the NSTemplateSet resource_

Keeping in mind that we are deleting a Space which has `space.Status.TargetCluster` field empty , which means that there should be no NSTemplateSet resource associated at that moment. 

I see that the `space.Status.TargetCluster` field gets populated here https://github.com/codeready-toolchain/host-operator/blob/03d197de76aeb04fda509c65bb5a539e66a65fcc/controllers/space/space_controller.go#L183
while the NSTemplateSet is created some lines after at https://github.com/codeready-toolchain/host-operator/blob/03d197de76aeb04fda509c65bb5a539e66a65fcc/controllers/space/space_controller.go#L200

That said , if I'm not mistaken cases 3 and 4 should never happen since there is no NSTemplateSet in this case!
 
If we return the error for case 1 we could end up in an endless reconcile loop since the member cluster is potentially "unknown" forever 
If we return the error for case 2 it would trigger another reconcile, and maybe second reconcile attempt will succeed discovering that there is no NSTemplateSet and proceed with Space deletion (but not sure how useful this second reconcile loop is).

However I agree that we could at least log the error even if we don't return it. This is how I see it but maybe I'm missing something. 

Again, If I'm not mistaken, we could even avoid calling `deleteNSTemplateSet` when `space.Status.TargetCluster` is empty, since there should be no NSTemplateSet to delete in this case.

MatousJobanek: ok, I see what you mean, thanks for the detailed description... it's not really visible from that part of the code. 

I see that you moved setting of the `space.Status.TargetCluster` down, which is good, otherwise, the target cluster would be always set in the status part -  even for invalid clusters :+1: 

> Again, If I'm not mistaken, we could even avoid calling deleteNSTemplateSet when space.Status.TargetCluster is empty, since there should be no NSTemplateSet to delete in this case.

Yeah, you are right. We should also probably change the logic so it takes the member cluster only based on the information stored in the Status, right? https://github.com/codeready-toolchain/host-operator/blob/03d197de76aeb04fda509c65bb5a539e66a65fcc/controllers/space/space_controller.go#L511-L514

controllers/space/space_controller.go
```diff
@@ -477,8 +479,13 @@ func extractUsernames(role string, bindings []toolchainv1alpha1.SpaceBinding) []
...
+		logger.Error(err, "error while deleting NSTemplateSet")
```
MatousJobanek, line 488:
```suggestion
		logger.Error(err, "error while deleting NSTemplateSet - ignored since the target cluster in the Status is empty")
```

MatousJobanek (approved):

controllers/space/space_controller.go
mfrancisc: Yeah, right, probably we should switch to use only `space.Status.TargetCluster` , but let's discuss this for a future improvement maybe. WDYT?

Thanks for review and approval!

MatousJobanek: > but let's discuss this for a future improvement maybe. WDYT?

yup yup :+1:

PR #793, opened 2023-04-21
Remove write access to all GitOpsDeployment* resources, from all roles

jgwest:
For security purposes, only the GitOps Service controllers should be able to create/modify/delete the `GitOpsDeployment*` resources, at this time.

* Note this does not effect the Environment API (Snapshots, Environments, SnapshotEnvironmentBindings, etc).

#### Corresponding PRs:
- *Book*: https://github.com/redhat-appstudio/book/pull/97
- *Toolchain-e2e*: https://github.com/codeready-toolchain/toolchain-e2e/pull/712

### DIFF
--- a/test/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/test/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -61,7 +61,9 @@ objects:
       - gitopsdeploymentrepositorycredentials
       - gitopsdeploymentsyncruns
     verbs:
-    - '*'
+      - get
+      - list
+      - watch
   - apiGroups:
     - tekton.dev
     resources:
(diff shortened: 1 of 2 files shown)


test/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
```diff
@@ -61,7 +61,9 @@ objects:
       - gitopsdeploymentrepositorycredentials
       - gitopsdeploymentsyncruns
     verbs:
-    - '*'
+      - get
```
alexeykazakov, line 64: You don't really have to update the test yaml. It's used to test that the templates are loaded. The content doesn't matter.
It doesn't hurt to upload it to but it's not necessary. Just FYI for future PRs :)

alexeykazakov (approved):

PR #795, opened 2023-04-28
fix: rename MapToSubSpacesByParentObjectName and improve comment

mfrancisc:


controllers/space/mapper.go
```diff
@@ -10,11 +10,17 @@ import (
...
-func MapToSubSpacesByParentObjectName(cl client.Client) func(object client.Object) []reconcile.Request {
+// MapSpaceBindingToSubSpaces maps the SpaceBinding of a given Space to it's subSpaces (if any).
+// It enables SpaceBinding inheritance from the parentSpace to the eventual subSpaces.
+//
+// The logic works as following:
+// - an event is triggered on a SpaceBinding object (something has changed)
+// - from the SpaceBinding labels we get the name of the Space object (a.k.a parentSpace)
+// - with the name of the Space (parentSpace) we search for eventual subSpaces
+// - in order to reflect the changes on SpaceBinding we trigger a `reconcile.Request` for the Space (parentSpace) and all it's subSpaces (if any)
+func MapSpaceBindingToSubSpaces(cl client.Client) func(object client.Object) []reconcile.Request {
```
rajivnathan, line 23: What about `MapSpaceBindingToParentAndSubSpaces`?

rajivnathan (commented):
Thanks for the clarifications! Just 1 small suggestion related to the name

controllers/space/mapper.go
mfrancisc: @rajivnathan nice , maybe it reflects better what it actually does indeed.

mfrancisc: @rajivnathan rename done , please see [bd25074](https://github.com/codeready-toolchain/host-operator/pull/795/commits/bd25074d43e0d6859d82584734c322c4dadca639)

Thanks for the suggestion!

alexeykazakov (approved):

controllers/space/mapper.go
ranakan19: Sorry for the late review, how does `MapSpaceBindingFromParentToSubSpaces` sound?
I don't mind the MapSpaceBindingToParentAndSubSpaces, I can get used to it. However, in the first look it looked like a one-to-many mapping because of 'and', maybe it was just me?

ranakan19 (approved):

controllers/space/mapper.go
mfrancisc: @ranakan19 your proposal makes sense, however the mapping of the SpaceBinding to the parent space is happening at the same time (same function) with the subSpaces. 
To me `MapSpaceBindingFromParentToSubSpaces` gives the idea that the mapping of the SpaceBinding to the parentSpace is already done (which is actually not true, there is only a label that links them, but events are not triggered on changes yet) , and only the one to the SubSpaces is performed here. 

I suggest to keep it as is because of this reason, hopefully makes sense. WDYT?

PR #804, opened 2023-05-11
Add approved metric that includes method of approval

rajivnathan:
https://issues.redhat.com/browse/CRT-1790

Introduces a new metric `user_signups_approved_with_method_total` that has 2 labels `automatic` and `manual`. It can be used to monitor the ratio of automatic to manual approvals.

```
# HELP sandbox_user_signups_approved_with_method_total Total number of UserSignups approved, includes either 'automatic' or 'manual' labels for the approval method
# TYPE sandbox_user_signups_approved_with_method_total counter
sandbox_user_signups_approved_with_method_total{method="manual"} 2
sandbox_user_signups_approved_with_method_total{method="automatic"} 2
```

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/727


pkg/metrics/metrics.go
```diff
@@ -16,6 +16,9 @@ var (
 	// UserSignupApprovedTotal is incremented each time a user signup is approved, can be multiple times per user if they reactivate multiple times
 	UserSignupApprovedTotal prometheus.Counter
 
+	// UserSignupApprovedWithMethodTotal is incremented each time a user signup is approved and includes either 'automatic' or 'manual' labels, can be multiple times per user if they reactivate multiple times
+	UserSignupApprovedWithMethodTotal *prometheus.CounterVec
```
alexeykazakov, line 20: Just curios why you are using a pointer here while we don't use pointers in other counters? Just curios if there is any specific reason for this inconsistency.

alexeykazakov (approved):

pkg/metrics/metrics.go
rajivnathan: `prometheus.NewCounter()` returns `prometheus.Counter` but `prometheus.NewCounterVec()` returns `*prometheus.CounterVec` and I was trying to be consistent with *prometheus.GaugeVec that we have in the same file.

mfrancisc (approved):
Nice 👍 ! 

The usage of `prometheus.CounterVec` makes sense for this use case!

PR #817, opened 2023-06-01
track event sent to Pendo

ranakan19:
fix wrt: https://issues.redhat.com/browse/SANDBOX-55

track user activation in pendo as well.


pkg/pendo/pendo.go
```diff
@@ -0,0 +1,70 @@
...
+	req, err := http.NewRequest("POST", TrackApiUrl, bytes.NewBuffer(payloadJSON))
+	if err != nil {
+		logger.Error(err, "Error creating post request")
+		return fmt.Errorf("Error creating post request: %s", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("x-pendo-integration-key", c.trackEventKey)
+
+	res, err := c.client.Do(req)
```
alexeykazakov, line 56: Use the default client instead:
```suggestion
	res, err := http.DefaultClient.Do(req)
```
You can easily test it using something like `gock` in unit tests.

pkg/pendo/pendo.go
```diff
@@ -0,0 +1,70 @@
...
+
+	//convert payload to json
+	payloadJSON, err := json.Marshal(payload)
+	if err != nil {
+		logger.Error(err, "Unable to marshall payload")
+	}
+
+	req, err := http.NewRequest("POST", TrackApiUrl, bytes.NewBuffer(payloadJSON))
+	if err != nil {
+		logger.Error(err, "Error creating post request")
```
alexeykazakov, line 49: Instead of logging the error and then returning it here and in other places just return the error. Check the error in the controller when calling `TrackAccountActivation()` and log it there if it's not nil.

pkg/pendo/pendo.go
```diff
@@ -0,0 +1,70 @@
...
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("x-pendo-integration-key", c.trackEventKey)
+
+	res, err := c.client.Do(req)
+	if err != nil {
+		fmt.Println(err)
+		return fmt.Errorf("Error sending request: %s", err)
+	}
+	defer res.Body.Close()
```
alexeykazakov, line 61: It's not enough just to close the body. You have to read the body before closing it to avoid FS leaks. It's golang specifics.

alexeykazakov:
I guess it can be closed now?

ranakan19:
not needed anymore, going the other route with https://github.com/codeready-toolchain/host-operator/pull/821

ranakan19:
> I guess it can be closed now?

Yep, closed all related ones as well

PR #820, opened 2023-06-02
Enable admins to work with the remote secrets.

metlos:
We're planning to enable the remote secrets in rhtap staging in the near future so we want the admin users to be able to play with them. I hope this is enough to do so.

e2e-tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/746

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -87,6 +87,11 @@ objects:
     - spiaccesstokens
     - spifilecontentrequests
     - spiaccesstokendataupdates
+    verbs:
+    - "*"
+  - apiGroups:
+    - appstudio.redhat.com
+    resources:
     - remotesecrets
     verbs:
     - "*"
--- a/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -87,6 +87,7 @@ objects:
     - spiaccesstokens
     - spifilecontentrequests
     - spiaccesstokendataupdates
+    - remotesecrets
     verbs:
     - "*"
   - apiGroups:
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):
The PR looks good.
But please:
1. Create a PR to update https://github.com/redhat-appstudio/book/blob/main/ADR/0011-roles-and-permissions.md to reflect this change.
2. Create a paired e2e-tests PR. That PR has to be from your fork and the branch has to have the same name as for this PR. So our CI can pair the PRs. See an example of such paired PR here: https://github.com/codeready-toolchain/toolchain-e2e/pull/716

metlos:
Thanks for pointing out the lack of tests and docs. This is only the second time I'm updating the host operator so I don't have the workflow under my skin yet :)

The documentation update PR is here https://github.com/redhat-appstudio/book/pull/106 and the PR with the test is here https://github.com/codeready-toolchain/toolchain-e2e/pull/746

alexeykazakov:
/retest

I fixed the test: https://github.com/codeready-toolchain/toolchain-e2e/pull/746/commits/139f4a6a4bd1c49d1fad9fb380d9b834458e5394

It should work now.

deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
```diff
@@ -89,6 +89,12 @@ objects:
     - spiaccesstokendataupdates
     verbs:
     - "*"
+  - apiGroups:
+    - appstudio.redhat.com
+    resources:
+    - remotesecrets
+    verbs:
+    - "*"
```
MatousJobanek, line 97: I would love to stop using the `*` sign in RBAC. I understand that it's used for other rules, feel free to keep it here. But please keep in mind that we should ideally avoid using it in the future and also fix it in the ADR.

MatousJobanek (approved):

deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
metlos: Agreed, but as you correctly guessed, I was just going with the flow here..

PR #822, opened 2023-06-06
Send epoch time in segment properties

ranakan19:
sending data from Amplitude -> pendo is failing due to incompatible timestamp formats. Thus sending the epoch time as part of properties to segment to propagate the field forward to pendo.

fix wrt: https://issues.redhat.com/browse/SANDBOX-55

### DIFF
--- a/pkg/segment/segment.go
+++ b/pkg/segment/segment.go
@@ -2,6 +2,7 @@ package segment
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/segmentio/analytics-go/v3"
@@ -48,7 +49,8 @@ func (c *Client) TrackAccountActivation(username, userID, accountID string) {
 		UserId: Hash(username),
 		Properties: analytics.NewProperties().
 			Set("user_id", userID).
-			Set("account_id", accountID),
+			Set("account_id", accountID).
+			Set("epoch_time", time.Now()),
 		Context: &analytics.Context{
 			Extra: map[string]interface{}{
 				"groupId": accountID,
--- a/pkg/segment/segment.go
+++ b/pkg/segment/segment.go
@@ -50,7 +50,7 @@ func (c *Client) TrackAccountActivation(username, userID, accountID string) {
 		Properties: analytics.NewProperties().
 			Set("user_id", userID).
 			Set("account_id", accountID).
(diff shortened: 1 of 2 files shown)


test/segment/assertions.go
```diff
@@ -30,7 +30,16 @@ func assertMessageQueued(t *testing.T, cl *segment.Client, username, userID, acc
...
+	}
+	actualMessage := cl.Client().(*MockClient).Queue[0]
+	actualTrackEvent, ok := actualMessage.(analytics.Track)
+	require.True(t, ok)
+	assert.Equal(t, expectedTrackEvent.UserId, actualTrackEvent.UserId)
+	assert.Equal(t, expectedTrackEvent.Event, actualTrackEvent.Event)
+	assert.Equal(t, expectedTrackEvent.Context, actualTrackEvent.Context)
+	assert.Equal(t, expectedTrackEvent.Properties["user_id"], actualTrackEvent.Properties["user_id"])
+	assert.Equal(t, expectedTrackEvent.Properties["account_id"], actualTrackEvent.Properties["account_id"])
+	assert.NotEmpty(t, actualTrackEvent.Properties["epoch_time"]) // not comparing the timestamp in test
```
drpaneas, line 42: If I understand the issue here is `incompatible timestamp format` and what this PR does is making sure that the timestamp is not empty, but not checking the compatible format. I might be wrong, though, in that case please disregard my comment.

pkg/segment/segment.go
```diff
@@ -48,7 +49,8 @@ func (c *Client) TrackAccountActivation(username, userID, accountID string) {
 		UserId: Hash(username),
 		Properties: analytics.NewProperties().
 			Set("user_id", userID).
-			Set("account_id", accountID),
+			Set("account_id", accountID).
+			Set("epoch_time", time.Now()),
```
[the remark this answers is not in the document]
ranakan19: Thanks Xavier! I definitely need the epoch time, that's the entire reason for change. Added the UnixMilli() because need it in milliseconds.

test/segment/assertions.go
ranakan19: @drpaneas The incompatible timestamp is between Amplitude and pendo. This change is to bypass the incompatibility by sending the compatibile timestamp in properties to segment. Hope it helps answer your question.

PR #830, opened 2023-06-27
Add appstudio-redhat nstemplatetier

jinqi7:
Add a new nstemplatetier for redhat internal tenant

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio-redhat/ns_tenant.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio-redhat/ns_tenant.yaml
@@ -1,0 +1,325 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: appstudio-redhat-tenant
+objects:
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    annotations:
+      openshift.io/description: ${SPACE_NAME}
+      openshift.io/display-name: ${SPACE_NAME}
+      openshift.io/requester: ${SPACE_NAME}
+    labels:
+      name: ${SPACE_NAME}-tenant
+      argocd.argoproj.io/managed-by: gitops-service-argocd
+      #billing labels ref: https://github.com/redhat-appstudio/book/blob/main/ADR/0010-namespace-metadata.md
+      appstudio.redhat.com/workspace_name: ${SPACE_NAME}
+      redhat-internal-tenant/label: "true"
+    name: ${SPACE_NAME}-tenant
+
+# Built-n developer environment is available for all users OOTB
+- apiVersion: appstudio.redhat.com/v1alpha1
+  kind: Environment
+  metadata:
+    name: development
+    namespace: ${SPACE_NAME}-tenant
+  spec:
+    deploymentStrategy: AppStudioAutomated
+    displayName: Development
+    type: Non-POC
+
+# Role & RoleBinding that grant limited read permissions to all SAs in member-operator namespace.
+# This is needed to let Proxy read the ServiceAccounts and Secrets in this namespace.
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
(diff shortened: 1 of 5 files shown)


MatousJobanek:
Ok, we need to discuss this. BTW, the content that is in this PR is not sufficient for creating a new NSTemplateTier.
In addition to that, there are other things I'm really concerned about:
* The most significant one is that this approach brings a lot of duplication and will make maintenance of the appstudio tiers very difficult. The NSTemplateTiers were not designed for such a usage.
* The move to the RH cluster is a manual operation and it would need to include also promotion to this specific tier (manually again)
* We need to set some boundaries in what we want to manage by NSTempalateTiers and what not. I'm not fine with bringing dependency on the external 3rd-party project into our code.
 
From the document that is linked in the jira, there is
> For the other Redhat internal users, if they also want to use these tokens, the following steps can help to achieve that. 

which means that it's an optional thing which is not necessary to be preesnt. Why users/admins couldn't create this resource on when they needed on demand?
Also, why this cannot by managed by a RHTAP component which is responsible for this?

jinqi7:
@MatousJobanek  I can understand your concerns.  I have two questions here - firstly, for study purpose, what's missed for creating a new NStemplateTier in the PR?  Secondly, we need to identify if the tenant is a Redhat internal user,  then we decide if we grant the permission of accessing the snyk_token to it.  How can we achieve it? Please feel free to give some suggestion. Thanks

MatousJobanek:
> I can understand your concerns. I have two questions here - firstly, for study purpose, what's missed for creating a new NStemplateTier in the PR? 

Creating template for a namespace is not enough - this defines what needs to be applied inside the namespace. There are other templates for:
* cluster-wide resources
* particular roles available in the namespace/workspace

All these templates needs to be then referenced in the `tier.yaml` file with the `NSTemplateTier` which is actually the representation of the tier. 

Just FYI, there is additional logic around it, but that should be automatically applied for the new tier as well. That logic:
* generates `metadata.yaml` file with the SHAs of the commits when any of the tier was modified for the last time
* generates TierTemplates based on the templates mentioned above and creates them in a cluster
* processes the NSTemplateTier with the references of the last TierTemplate refs
* creates/updates the NSTemplateTier in the cluster

There are also some unit test that needs to be updated as well.

> Secondly, we need to identify if the tenant is a Redhat internal user, then we decide if we grant the permission of accessing the snyk_token to it. How can we achieve it?

Let's differentiate between tenant vs workspace. The `ns_tenant.yaml` template is applied to all namespaces that are provisioned using the tier. This means that the object/resource will be present in the workspace/namespace by default.
The rh1 cluster is dedicated for rh-internal workspaces, which means that any namespace in the cluster is rh-internal. 
However, any tenant (user) can get access to these workspaces - this is done "manually" via managing SpaceBinding at the workspace level. So even non-rh account can get access to the rh-internal workspace if someone grants that access to the user. You have no way to figure out if the user is rh one or not at the namespace level.

TBH, this requires a longer discussion about the purpose of the change and the expected usage.

Also, there are a few additional questions I have:

Why is this so special for rh-internal workspaces? Is this only about sharing one token between multiple workspaces? Do we expect that the external users would do something similar or the same? If not then why? And if yes, then how?
 
How is the `ExternalSecret` API supposed to be used by the users who have access to the namespace? Is the `ExternalSecret` API exposed to the external users as well?

MatousJobanek (changes requested):
We should address this use-case in a more generic way. It's related to my questions that I pasted in my previous message. 

One of the main things of RHTAP (that were also mentioned by Burr during the keynote) is that the external users/customers will be able to experience the same CI/CD flow that is gonna by used by RH teams.
That being said, consider this situation:
* a bigger customer will start using RHTAP offering 
* they will create multiple workspaces for several teams/orgs which are part of the customer's structure
* they will want to use & share the same Snyk token between their workspaces

How would they do it?

alexeykazakov:
I guess we can close this PR now?

PR #845, opened 2023-08-09
Switch to cluster-scoped Argo CD (#560)

keithchong:
@jgwest , see https://issues.redhat.com/browse/GITOPSRVCE-560

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio-env/ns_env.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio-env/ns_env.yaml
@@ -12,7 +12,6 @@ objects:
       openshift.io/requester: ${SPACE_NAME}
     labels:
       name: ${SPACE_NAME}-env
-      argocd.argoproj.io/managed-by: gitops-service-argocd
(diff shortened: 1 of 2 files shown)


MatousJobanek:
@keithchong thank you for your PR.
TBH, I always thought that namespace-scoped Argo CD was the main goal and the state we would like to get to. I understand that there are some performance-related challenges, but from the PR you referenced it doesn't seem like that the label would be a problem. So my question is, do we really have to remove the label? 
There is one thing to keep in mind - the logic of member-operators won't remove the label from the already existing namespaces. It works fine with adding the missing labels, but as you can imagine, it's hard to decide which label would need to be removed. That being said - you would need to perform manual removal of all labels from all namespaces in all member clusters.

MatousJobanek:
@keithchong @jgwest have you seen my comment above? https://github.com/codeready-toolchain/host-operator/pull/845#issuecomment-1672818214

MatousJobanek:
Closing this PR as there was no response to my questions in the last month

PR #846, opened 2023-08-09
feat: create SpaceBindingRequest controller skeleton

mfrancisc:
For now the SpaceBindingRequest controller is just reconciling `SpaceBindingRequest` resources without performing any action. This PR just sets up the controller and enables it for e2e testing, the main logic will be implemented next.

Jira: https://issues.redhat.com/browse/ASC-395

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/773


controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -0,0 +1,147 @@
...
+				RestConfig: &rest.Config{
+					Host: c.APIEndpoint,
+					TLSClientConfig: rest.TLSClientConfig{
+						CAData: []byte("ASDf33=="),
+					},
+				},
+			},
+			RESTClient: restClient,
+			Client:     c.Client,
+		}
```
MatousJobanek, line 113: I'm wondering why do actually need this gock mocker, the RESTClient and the RestConfig? We don't need to call any special REST endpoint as we do in SpaceRequest for getting the tokens, right?
wouldn't be this sufficient?
```suggestion
		clusters[c.Name] = cluster.Cluster{
			Config: &commoncluster.Config{
				Type:              commoncluster.Member,
				OperatorNamespace: c.OperatorNamespace,
				OwnerClusterName:  test.MemberClusterName,
			},
			Client: c.Client,
		}
```

mfrancisc:
@MatousJobanek I've applied your suggestion, that was indeed not needed. I've forgot to clean that up when copying from the SpaceRequest controller. Thanks!

MatousJobanek (approved):

PR #848, opened 2023-08-14
feat: Implement SpaceBindingRequest controller main logic

mfrancisc:
This PR adds main logic for handling SpaceBindingRequests, the logic supports:
- create/update/delete of SpaceBindingRequests and associated SpaceBindings
- validation of MUR, SpaceRole

Jira: https://issues.redhat.com/browse/ASC-397

Related PRs: https://github.com/codeready-toolchain/api/pull/372

Paired with:

### DIFF
--- a/controllers/spacebindingrequest/spacebindingrequest_controller.go
+++ b/controllers/spacebindingrequest/spacebindingrequest_controller.go
@@ -2,10 +2,17 @@ package spacebindingrequest
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/pkg/cluster"
+	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
+	"github.com/codeready-toolchain/toolchain-common/pkg/spacebinding"
+	"github.com/go-logr/logr"
 	errs "github.com/pkg/errors"
+	"github.com/redhat-cop/operator-utils/pkg/util"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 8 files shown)


mfrancisc:
/retest

update e2e PR

test/spacerequest/spacerequest.go
```diff
@@ -61,3 +63,21 @@ func WithStatusNamespaceAccess(namespaceAccess toolchainv1alpha1.NamespaceAccess
 		spaceRequest.Status.NamespaceAccess = []toolchainv1alpha1.NamespaceAccess{namespaceAccess}
 	}
 }
+
+func NewNamespace(spacename string) *corev1.Namespace {
+	labels := map[string]string{
+		toolchainv1alpha1.TypeLabelKey:     "tenant",
+		toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
+	}
+	if spacename != "nospace" {
```
drpaneas, line 72: What happens if `== nospace` ? Is it still a valid namespace? Do think this ever returns an invalid namespace? I mean should we return `(*corev1.Namespace, error)` or not necessary?

test/spacebindingrequest/spacebindingrequest.go
```diff
@@ -34,3 +36,16 @@ func WithSpaceRole(spaceRole string) Option {
...
+func WithDeletionTimestamp() Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		now := metav1.NewTime(time.Now())
+		spaceBindingRequest.DeletionTimestamp = &now
+	}
+}
+
+func WithFinalizer() Option {
+	return func(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) {
+		spaceBindingRequest.Finalizers = append(spaceBindingRequest.Finalizers, toolchainv1alpha1.FinalizerName)
```
drpaneas, line 49: what if the `spaceBindingRequest` object is not `nil` before applying the finalizer? Would that be a problem? If so:

```go
if spaceBindingRequest != nil {
	spaceBindingRequest.Finalizers = append(spaceBindingRequest.Finalizers, toolchainv1alpha1.FinalizerName)
} else {
	// Handle the case where the input spaceBindingRequest is nil, possibly with logging or an error return.
		}
```

go.mod
```diff
@@ -114,3 +114,5 @@ require (
 )
 
 go 1.19
+
+replace github.com/codeready-toolchain/api => github.com/mfrancisc/api v0.0.0-20230814083608-6d8633a3bcf3
```
drpaneas, line 118: Please update the relevant api repo first and then update thus replicate statement, instead.

controllers/spacerequest/spacerequest_controller_test.go
```diff
@@ -853,7 +852,7 @@ func TestDeleteSpaceRequest(t *testing.T) {
 	err := apis.AddToScheme(scheme.Scheme)
 	require.NoError(t, err)
 	appstudioTier := tiertest.AppStudioTier(t, tiertest.AppStudioTemplates)
-	srNamespace := newNamespace("jane")
+	srNamespace := spacerequesttest.NewNamespace("jane")
```
drpaneas, line 855: Not your code per se --  can we put "jane" in a const? Seems it's been used in many place, to avoid any typos in the future.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -79,6 +123,518 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			member1 := NewMemberClusterWithClient(member1Client, "member-1", corev1.ConditionTrue)
+			hostClient := test.NewFakeClient(t)
+			ctrl := newReconciler(t, hostClient, member1)
+
+			// when
+			_, err := ctrl.Reconcile(context.TODO(), requestFor(sbr))
+
+			// then
+			require.EqualError(t, err, "failed to remove finalizer: mock error")
+		})
```
drpaneas, line 637: great work on testing and mocking this :D +1

test/spacerequest/spacerequest.go
drpaneas: yes it makes sense. Sorry I missed the `test` part ... sigh...

test/spacebindingrequest/spacebindingrequest.go
mfrancisc: Unless I'm missing something, it cannot be nil, since those `Option` functions are being passed to the `NewSpaceBindingRequest` function, which takes care of creating the SpaceBindingRequest object: https://github.com/codeready-toolchain/host-operator/blob/master/test/spacebindingrequest/spacebindingrequest.go#L13-L19

Of course if somehow misused you could break them, but those are "only" used for unit testing, and we might want to modify all our `With...` helper in this case (also the ones in toolchain-common), since they all follow the same style, eg: https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/test/usersignup/usersignup.go

go.mod
mfrancisc: This is done only for testing (unit and e2e) the PR with the updated version of the API repo, once all the PRs are approved and passing all the tests/checks, we:
- merge the API pr
- update other PRs with the new version
- merge other repos with new master API version

we also have a CI check in place which prevents from merging if there is any `replace` statement, see: https://github.com/codeready-toolchain/host-operator/actions/runs/5857146355/job/15878356586?pr=848

So this will be replaced once all the tests are passing and all the PRs are approved.

Hopefully this answers you question/concern. But plz let me know if I misunderstood.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: I can't take credit for this , I was just good at copy pasting and following same approach I saw in other unit tests.
The repository mature and full of great examples on how to solve many of these "tricky" problems with testing and not only. 

Kudos to the team 🏅

controllers/spacerequest/spacerequest_controller_test.go
mfrancisc: Yeah, thanks for pointing this out, it doesn't look very clean, there are usually some strings that can be extracted to const or variable in our testing code, just didn't saw this aligned with the other unit tests for other controllers. 

Or maybe we could put the string into a variable instead of const ? 🤔

test/spacerequest/spacerequest.go
mfrancisc: No worries! My pleasure getting more feedback/reviews and discussing alternative solutions! Thanks for checking!

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 	logger.Info("spacebindingrequest found", "member cluster", memberClusterWithSpaceBindingRequest.Name)
 
-	// todo implement logic
+	if util.IsBeingDeleted(spaceBindingRequest) {
```
alexeykazakov, line 102: It's not part of this PR but I missed the previous PR which introduced member cluster lookup. So I'm using this opportunity to point out that we can improve the lookup mechanism :)
Let's say there are two member clusters m-1 and m-2. And m-1 cluster is having an outage and temporally down. The reconcile won't work for all space requests even from m-2 if m-1 happens to be first in r.MemberClusters.

What we could do to mitigate it is too save the error in line 82 above but proceed with the lookup. And only if no SBR found in other members then return the saved error. Otherwise just log the error but do not return it. In that case m-2 won't be affected by the m-1 outage.

Feel free to address it in a separate PR though.

mfrancisc: Yeah, good point! The logic can be improved as you suggested, also I can check if it can be extracted to a reusable function since I see it's similar to what I'm using in spacerequest_controller and soon also in spacebindingcleanup_controller (for the SBR cleanup task) 🤔 

Thanks for your suggestion, I'll address it in a separate PR 👍

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		logger.Info("spaceBindingRequest is being deleted")
+		return reconcile.Result{}, r.ensureSpaceBindingDeletion(logger, memberClusterWithSpaceBindingRequest, spaceBindingRequest)
+	}
+	// Add the finalizer if it is not present
+	if err := r.addFinalizer(logger, memberClusterWithSpaceBindingRequest, spaceBindingRequest); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	createdOrUpdated, err := r.ensureSpaceBinding(logger, memberClusterWithSpaceBindingRequest, spaceBindingRequest)
```
MatousJobanek, line 111: does it make sense to return the boolean value in this case?
creation and update is one-time operation and we don't have to wait until the SpaceBinding is provisioned. 
It either fails for any reason (which means that it would return an error) or it succeeds, which means that we can mark it as done.

mfrancisc:
/retest

infra

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	// set SBR labels on spacebinding
+	spaceBinding.Labels[toolchainv1alpha1.SpaceBindingRequestLabelKey] = spaceBindingRequest.GetName()
+	spaceBinding.Labels[toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey] = spaceBindingRequest.GetNamespace()
+	logger.Info("creating spacebinding", "spaceBinding.Name", spaceBinding.Name)
+	if err := r.setStatusProvisioning(memberCluster, spaceBindingRequest); err != nil {
+		return err
+	}
+	if err := r.Client.Create(context.TODO(), spaceBinding); err != nil {
+		return errs.Wrap(err, "unable to create SpaceBinding")
+	}
```
MatousJobanek, line 315: btw, we should take care of conflicts with the already existing (default) SpaceBindings. Or with those that are created by admins. Such SpaceBindings are immutable for end users, so it would be good to set corresponding status in SBR CR.
It's fine if you do it in a new PR.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -79,6 +124,518 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			_, err = ctrl.Reconcile(context.TODO(), requestFor(badSBR))
+
+			// then
+			// an error should be returned
+			require.EqualError(t, err, "MasterUserRecord cannot be blank")
+			// spaceBindingRequest exists with config and finalizer
+			spacebindingrequesttest.AssertThatSpaceBindingRequest(t, badSBR.GetNamespace(), badSBR.GetName(), member1.Client).
+				HasSpecSpaceRole("admin").
+				HasSpecMasterUserRecord(""). // empty
+				HasFinalizer()
```
MatousJobanek, line 146: verify that it has corresponding reason, status, and message set

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -79,6 +124,518 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			_, err = ctrl.Reconcile(context.TODO(), requestFor(badSBR))
+
+			// then
+			// an error should be returned
+			require.EqualError(t, err, "SpaceRole cannot be blank")
+			// spaceBindingRequest exists with config and finalizer
+			spacebindingrequesttest.AssertThatSpaceBindingRequest(t, badSBR.GetNamespace(), badSBR.GetName(), member1.Client).
+				HasSpecSpaceRole(""). // empty
+				HasSpecMasterUserRecord(janeMur.Name).
+				HasFinalizer()
```
MatousJobanek, line 169: the same here

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -28,25 +36,62 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			// spaceBindingRequest exists with config and finalizer
+			spacebindingrequesttest.AssertThatSpaceBindingRequest(t, sbr.GetNamespace(), sbr.GetName(), member1.Client).
+				HasSpecSpaceRole("admin").
+				HasSpecMasterUserRecord(janeMur.Name).
+				HasConditions(spacebindingrequesttestcommon.Ready()).
+				HasFinalizer()
+			// there should be 1 spacebinding that was created from the SpaceBindingRequest
+			spacebindingtest.AssertThatSpaceBinding(t, test.HostOperatorNs, janeMur.Name, janeSpace.Name, hostClient).Exists().
+				HasLabelWithValue(toolchainv1alpha1.SpaceBindingRequestLabelKey, sbr.GetName()). // check expected labels are there
+				HasLabelWithValue(toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey, sbr.GetNamespace())
```
MatousJobanek, line 93: we expect that the labels should match the values in spec - this could be part of the `AssertThatSpaceBinding` function.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -79,6 +124,518 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			_, err := ctrl.Reconcile(context.TODO(), requestFor(sbrNoFinalizer))
+
+			// then
+			require.NoError(t, err)
+			spacebindingrequesttest.AssertThatSpaceBindingRequest(t, sbrNoFinalizer.GetNamespace(), sbrNoFinalizer.GetName(), member1.Client).HasNoFinalizers() // spaceBindingRequest is gone
+		})
+	})
+
+	t.Run("failure", func(t *testing.T) {
+		t.Run("SpaceBinding resource is already being deleted", func(t *testing.T) {
```
MatousJobanek, line 569: let's make clear why it's a failure case:
```suggestion
		t.Run("SpaceBinding resource is already being deleted for more than 2 minutes", func(t *testing.T) {
```

MatousJobanek (commented):
Nice work :+1:

mfrancisc:
/retest

infra again ..

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	return nil
+}
+
+// getSpaceBinding retrieves the spacebinding created by the spacebindingrequest
+func (r *Reconciler) getSpaceBinding(spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) (*toolchainv1alpha1.SpaceBinding, error) {
+	spaceBindings := &toolchainv1alpha1.SpaceBindingList{}
+	spaceBindingLabels := runtimeclient.MatchingLabels{
+		toolchainv1alpha1.SpaceBindingRequestLabelKey:          spaceBindingRequest.GetName(),
+		toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey: spaceBindingRequest.GetNamespace(),
+	}
```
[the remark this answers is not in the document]
mfrancisc: I'm not 100% sure I understand how I would use it, since the spacebinding now is being created by this utility function https://github.com/codeready-toolchain/toolchain-common/blob/c909c72d8f627a9f72917531eded2a5e765faea1/pkg/spacebinding/spacebinding.go#L14 

Maybe I can add some utility function like `WithSpaceBindingRequestLabels` in that package and use that instead of setting the labels directly as I'm doing now when creating the object https://github.com/codeready-toolchain/host-operator/pull/848/files#diff-e48d49741db80226348ae5ffd08100dd10afa2e83e5d18b2a94776dfdfb616f3R304-R305

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	logger.Info("ensuring spacebinding")
+
+	// find space from namespace labels
+	space, err := r.getSpace(memberCluster, spaceBindingRequest)
+	if err != nil {
+		return false, r.setStatusFailedToCreateSpaceBinding(logger, memberCluster, spaceBindingRequest, err)
+	}
+	// space is being deleted
+	if util.IsBeingDeleted(space) {
+		return false, errs.New("space is being deleted")
```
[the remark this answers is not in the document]
mfrancisc: Makes sense! Thanks! done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d24598c352faf223469ee1b913d9a6fc3993772

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}
+
+	// validate MUR
+	mur, err := r.getMUR(spaceBindingRequest)
+	if err != nil {
+		return false, r.setStatusFailedToCreateSpaceBinding(logger, memberCluster, spaceBindingRequest, err)
+	}
+	// mur is being deleted
+	if util.IsBeingDeleted(mur) {
+		return false, errs.New("mur is being deleted")
```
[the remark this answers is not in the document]
mfrancisc: Makes sense! Thanks! done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d24598c352faf223469ee1b913d9a6fc3993772

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	for actual := range nsTemplTier.Spec.SpaceRoles {
+		if spaceBindingRequest.Spec.SpaceRole == actual {
+			isRoleValid = true
+		}
+	}
+	if !isRoleValid {
+		return fmt.Errorf("invalid role '%s' for space '%s'", spaceBindingRequest.Spec.SpaceRole, space.Name)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
mfrancisc: Agree this makes is shorter, thanks! Done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d24598c352faf223469ee1b913d9a6fc3993772

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: If I'm not mistaken this is already coverd by the fact that the name of the SpaceBinding is computed here https://github.com/codeready-toolchain/toolchain-common/blob/c909c72d8f627a9f72917531eded2a5e765faea1/pkg/spacebinding/spacebinding.go#L24

So in case an SBR tries to create a SB resource for mur and space that already has a SpaceBinding, the Create will fail saying that the name already exists and it will be set in the SBR status : https://github.com/codeready-toolchain/host-operator/pull/848/files#diff-e48d49741db80226348ae5ffd08100dd10afa2e83e5d18b2a94776dfdfb616f3R250-R252

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: Done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d24598c352faf223469ee1b913d9a6fc3993772

Thanks!

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: Done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d24598c352faf223469ee1b913d9a6fc3993772

Thanks!

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: I'm not sure I'm understanding this . The `AssertThatSpaceBinding` function is being used by other tests like `usersignup`, `spacebindingcleanup`. If I move this check inside the function it will not work there, but maybe I'm not understanding this correctly.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: Done in https://github.com/codeready-toolchain/host-operator/commit/0d24598c352faf223469ee1b913d9a6fc3993772

Thanks!

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+}
+
+// updateExistingSpaceBinding updates the spacebinding with the config from the spaceBindingRequest.
+// returns true/nil if the spacebinding was updated
+// returns false/nil if the spacebinding was already up-to-date
+// returns false/err if something went wrong or the spacebinding is being deleted
+func (r *Reconciler) updateExistingSpaceBinding(logger logr.Logger, spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest, spaceBinding *toolchainv1alpha1.SpaceBinding) (bool, error) {
+	// check if spacebinding is being deleted
+	if util.IsBeingDeleted(spaceBinding) {
+		return false, fmt.Errorf("cannot update SpaceBinding because it is currently being deleted")
```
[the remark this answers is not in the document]
mfrancisc: Done in https://github.com/codeready-toolchain/host-operator/pull/848/commits/0d077fdc7829953cfc5ce89b87be1a06b3e2c325 

I've also moved things around so that any error returned from `ensureSpaceBinding` will end up with calling `setStatusFailedToCreateSpaceBinding` passing this error.

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -92,7 +99,360 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+
+	// update SpaceRole and MUR
+	logger.Info("updating spaceBinding", "spaceBinding.Name", spaceBinding.Name)
+	spaceBinding.Spec.SpaceRole = spaceBindingRequest.Spec.SpaceRole
+	spaceBinding.Spec.MasterUserRecord = spaceBindingRequest.Spec.MasterUserRecord
+	// update also the MUR label
+	spaceBinding.Labels[toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey] = spaceBindingRequest.Spec.MasterUserRecord
+	err := r.Client.Update(context.TODO(), spaceBinding)
+	if err != nil {
+		return false, errs.Wrap(err, "unable to update SpaceRole and MasterUserRecord fields")
```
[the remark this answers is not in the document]
mfrancisc: Done in https://github.com/codeready-toolchain/host-operator/commit/0d077fdc7829953cfc5ce89b87be1a06b3e2c325

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: Good point, thanks! 

I got rid of the boolean in https://github.com/codeready-toolchain/host-operator/commit/0d077fdc7829953cfc5ce89b87be1a06b3e2c325

It was a left over from copying SpaceRequest controller I guess 🤦‍♂️

mfrancisc:
/retest

infra

controllers/spacebindingrequest/spacebindingrequest_controller.go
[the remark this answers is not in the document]
MatousJobanek: never mind - I made the comment when I started reviewing the PR. When I saw the rest of the code, I agree that it doesn't make any sense.

mfrancisc:
/retest

updated e2e PR

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: Ok, then let's do it in separate PR, because if we want to specify if the SB that already exists is or not managed by SBR controller , we need to get it and check the labels. Not sure if we want to provide such information to the user/client?

mfrancisc:
/retest

flaky? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/848/pull-ci-codeready-toolchain-host-operator-master-e2e/1693521960136347648/build-log.txt

mfrancisc:
/retest

updated e2e PR

PR #853, opened 2023-08-17
Update admin spacerole role and rolebinding to indicate level

ralphbean:
When a cluster admin inspects a namespace, it would be helpful if the names of the roles and rolebindings consistently reflected their level.

Both the contributor and maintainer roles and rolebindings indicate their level in their names (contributor and maintainer respectively).

Here, do the same with admin for parity.

### DIFF
--- a/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml
@@ -9,7 +9,7 @@ objects:
   kind: Role
   metadata:
     namespace: ${NAMESPACE}
-    name: appstudio-user-actions
+    name: appstudio-admin-user-actions
   rules:
   - apiGroups:
     - appstudio.redhat.com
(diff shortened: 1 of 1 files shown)


alexeykazakov:
@mfrancisc please take a look.

mfrancisc:
Looks good to me 👍  

I'm not sure why we avoided the keyword `admin` , maybe there was a reason behind this 🤷‍♂️ 

As @alexeykazakov pointed out, could you please create a PR also in toolchain-e2e with the same branch name. 
Basically updating the role name here : https://github.com/codeready-toolchain/toolchain-e2e/blob/82d14a8198f213081614ef91fe7579bc20eaa314/testsupport/tiers/checks.go#L1466

and rolebinding here: https://github.com/codeready-toolchain/toolchain-e2e/blob/82d14a8198f213081614ef91fe7579bc20eaa314/testsupport/tiers/checks.go#L1732-L1733

alexeykazakov:
I don't remember any specific reason besides the fact that originally there was only one role. We didn't have "admin", "contributor", "maintainer". So it didn't have any name suffix for it. Then we introduced two more roles but didn't renamed the original one.

mfrancisc:
> I don't remember any specific reason besides the fact that originally there was only one role.

Ok gotcha , that makes sense indeed! Thanks!

alexeykazakov (approved):

PR #855, opened 2023-08-24
refactor: introduce LookupMember cluster function

mfrancisc:
- address comment from https://github.com/codeready-toolchain/host-operator/pull/848#discussion_r1294048180 
- extract the member lookup logic to function that is now reused in both spacerequest and sapacebindingrequest controllers.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"reflect"
 
 	commoncluster "github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	errs "github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


pkg/cluster/cluster.go
```diff
@@ -14,3 +22,51 @@ type Cluster struct {
...
+		} else if err != nil && errors.IsNotFound(err) {
+			// if we exited with a notFound error
+			// it means that we couldn't find the Object on any of the given member clusters
+			// let's just return false (as not found) and no error
+			return memberClusterWithObject, false, nil
+		}
+	}
+	// Object found, return the member cluster that has it
+	return memberClusterWithObject, true, nil
+}
```
alexeykazakov, line 72: I tried to simplify the code and also fixed some minor issues. Take a look:
```suggestion
func LookupMember(memberClusters map[string]Cluster, request ctrl.Request, object runtimeclient.Object) (Cluster, bool, error) {
	var memberClusterWithObject Cluster
	var getError error
	var found bool
	for _, memberCluster := range memberClusters {
		err := memberCluster.Client.Get(context.TODO(), types.NamespacedName{
			Namespace: request.Namespace,
			Name:      request.Name,
		}, object)
		if err != nil {
			if !errors.IsNotFound(err) {
				// Error reading the object
				getError = err // save the error and return it only later in case the Object was not found on any of the clusters
			}

			//  Object not found on current member cluster
			continue
		}

		// save the member cluster on which the Object was found
		memberClusterWithObject = memberCluster
		found = true
		break // exit once found
	}

	if !found {
		if getError != nil {
			// if we haven't found the Object, and we got an error while trying to retrieve it
			// then let's return the error
			return memberClusterWithObject, false, errs.Wrap(getError, fmt.Sprintf("unable to get the current %T", object))
		}
		// Not found. No Error.
		return nil, false, nil
	}
	// Object found, return the member cluster that has it. Also if we got an error when trying to connec to at least one of them member then return it too.
	return memberClusterWithObject, true, getError
}
```

ranakan19: l like having the variable `found` to observe if the cluster was found, and use that variable in the return statement. It makes it easier to read as well as removes hardcoding the return values.

alexeykazakov: OK. I updated the suggestion.

mfrancisc: At first I thought about the same, just didn't want to introduce a new variable (`found`), but if it makes the code easier to read let's keep it. 

If we want to return `nil` instead of `memberClusterWithObject` I would need to return a pointer and change the signature of all the functions that gets it as input to accept a pointer, I suggest we do this in a separate PR. 

plz check https://github.com/codeready-toolchain/host-operator/pull/855/commits/e618eac3a659e96ac6d7863fbba58feb743611ee

Thanks for all the suggestions!

alexeykazakov: OK. You can keep the memberSluster object as is in the signature. Doesn't have to be a pointer. However please fix the descritpion then. It says it will return nil if not found:
> returns nil/false/nil in case the Object was NOT found on any of the given member clusters
> returns nil/false/error in case the Object was NOT found on any of the given member clusters...

alexeykazakov (approved):
Looks good. Thanks for addressing the comments! I'm approving the PR but please fix the lookup method description comment before merging it.

pkg/cluster/cluster.go
mfrancisc: Ok thanks for pointing that out, it returns an empty struct and not a nil indeed. I've changed it here: https://github.com/codeready-toolchain/host-operator/pull/855/commits/09cb5e87138939915d64006d0e9a11de6184a206 I guess this is what you meant.

PR #857, opened 2023-08-29
feat: update SpaceBinding cleanup controller to delete SBR resources

mfrancisc:
Right now the SpaceBindingRequest controller will create back the SB resource deleted by the controller. 
This PR updates the cleanup logic to check for SBR resources and delete those, as a consequence the associated SB will be deleted as well by the SpaceBindingRequestController.

Jira: https://issues.redhat.com/browse/ASC-414

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/785


controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -94,8 +96,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			// got error while searching for SpaceBindingRequest CR
+			return err
+		}
+		// Just log the error but proceed because we did find the member anyway
+		logger.Error(err, "error while searching for SpaceBindingRequest")
+	} else if !found {
+		// let's just log the info
+		logger.Info("unable to find SpaceBindingRequest", "SpaceBindingRequest.Name", sbrAssociated.name, "SpaceBindingRequest.Namespace", sbrAssociated.namespace)
+		// try and delete the SpaceBinding since SBR was not found
+		return r.DeleteSpaceBinding(sbrAssociated.spaceBinding)
```
MatousJobanek, line 135: The deletion itself should be done by the SBR controller (as part of the finalizer), but it's probably fine to have this here to take care of potential zombies :+1:

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -94,8 +96,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}
+
+	// delete the SBR
+	if err := memberClusterWithSpaceBindingRequest.Client.Delete(context.TODO(), spaceBindingRequest); err != nil {
+		return errs.Wrapf(err, "unable to delete the SpaceBindingRequest")
+	}
+	return nil
```
MatousJobanek, line 142: I would also schedule another reconcile in let's say 10 seconds, to be sure that the controller would get back to the SpaceBinding again - to be sure that the deletion was completed.

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -94,8 +96,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 func (r *Reconciler) deleteSpaceBinding(logger logr.Logger, spaceBinding *toolchainv1alpha1.SpaceBinding) error {
 	logger.Info("deleting the SpaceBinding")
+
+	// check if spaceBinding was created from SpaceBindingRequest,
+	// in that case we must delete the SBR and will take care of deleting the SpaceBinding as well
+	spaceBindingReqeuestAssociated := checkSpaceBindingRequestAssociated(spaceBinding)
+	if spaceBindingReqeuestAssociated.found {
+		return r.deleteSpaceBindingRequest(logger, spaceBindingReqeuestAssociated)
```
MatousJobanek, line 104: Let's also check if the SBR functionality is enabled in ToolchainConfig - if it's disabled but there is still SB created via SBR, then I would return an error.

MatousJobanek (commented):
Looks good, only a few suggestions

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -94,8 +96,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 func (r *Reconciler) deleteSpaceBinding(logger logr.Logger, spaceBinding *toolchainv1alpha1.SpaceBinding) error {
 	logger.Info("deleting the SpaceBinding")
+
+	// check if spaceBinding was created from SpaceBindingRequest,
+	// in that case we must delete the SBR and will take care of deleting the SpaceBinding as well
```
rajivnathan, line 101:
```suggestion
	// check if spaceBinding was created from SpaceBindingRequest,
	// in that case we must delete the SBR and then the SBR controller will take care of deleting the SpaceBinding
```

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -94,8 +96,71 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	spaceBindingReqeuestAssociated := checkSpaceBindingRequestAssociated(spaceBinding)
+	if spaceBindingReqeuestAssociated.found {
+		return r.deleteSpaceBindingRequest(logger, spaceBindingReqeuestAssociated)
+	}
+
+	// otherwise delete the SpaceBinding resource directly ...
+	return r.DeleteSpaceBinding(spaceBinding)
+}
+
+func (r *Reconciler) DeleteSpaceBinding(spaceBinding *toolchainv1alpha1.SpaceBinding) error {
```
rajivnathan, line 111: it's a minor thing but I guess this function doesn't need to be exported so we could name it `deleteSpaceBindingDirectly` or something so that we don't export the function when it doesn't need to be. This wouldn't impact much other than potentially unit tests I guess.

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
rajivnathan: +1

rajivnathan (commented):
Nice!

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
mfrancisc: This should not happen , but just in case!

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
mfrancisc: Good point, done in https://github.com/codeready-toolchain/host-operator/pull/857/commits/9b0996d46106b741e974909d6f023e48f7e6b588

Thanks for your suggestions!

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
mfrancisc: I was thinking at the scenario in which for some reason we want to disable the SBR functionality (maybe temporarily), if we do this the cleanup of the SBRs will not be performed automatically and a bunch of errors will pop up in our logs.

Maybe it's enough to just log the error (a warning would be more suited IMO)  but still perform the cleanup ? WDYT ?  

Just wanted to better understand why we should not clean them up.

mfrancisc: > If the SBR functionality is disabled, then the SBR controller is not running, which means that the SBR CR will stay in a Terminating state.

Good point! If we restarted the operator then yes the controller will not be running.

> How about doing both - calling the delete, but at the same time either returning or logging an error?
I think this would be easier to implement, because I "just" need to return an error after delete if the config is disabled. Also it covers the case in which we disabled the config by mistake or there was some issue with config and was unintentionally turned off without restarting the operator, the cleanup will still be done by the controller.

MatousJobanek: ok, let's do that :+1:

mfrancisc: Actually logging the error was easier than returning it, done in https://github.com/codeready-toolchain/host-operator/pull/857/commits/98af28b3ecdfb02ba2c883facc2331982b8c3258

I can return it, but I thought there was no additional value in doing that compared to just logging it, since the resources will be reconciled anyway once we restart the operators (and once every N hours if I'm not mistaken). 
But please let me know if I missed anything and we should return it instead. 

Please also check new PR for unit testing utilities: https://github.com/codeready-toolchain/toolchain-common/pull/323

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -73,9 +80,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			if requeueAfter, err := r.deleteSpaceBinding(logger, spaceBinding); err != nil {
+				return ctrl.Result{}, err
+			} else if requeueAfter > 0 {
+				return ctrl.Result{
+					Requeue:      true,
+					RequeueAfter: requeueAfter,
+				}, nil
+			}
+			// spacebinding deleted and no reconcile needed
+			return ctrl.Result{}, nil
```
MatousJobanek, line 92: from the `Result` documentation:
```
// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
RequeueAfter time.Duration
```
you could simplify it:
```suggestion
			requeueAfter, err := r.deleteSpaceBinding(logger, spaceBinding)
			return ctrl.Result{
				RequeueAfter: requeueAfter,
			}, err
```

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -84,18 +100,109 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
```
MatousJobanek, line 112: the same
```suggestion
			requeueAfter, err := r.deleteSpaceBinding(logger, spaceBinding)
			return ctrl.Result{
				RequeueAfter: requeueAfter,
			}, err
```

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
mfrancisc: Ha nice, thanks for pointing that out!

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
```diff
@@ -34,12 +43,15 @@ func TestDeleteSpaceBinding(t *testing.T) {
...
+		fakeClient := test.NewFakeClient(t, sbLaraRedhatAdmin, sbJoeRedhatView, sbLaraIbmEdit, laraMur, joeMur, ibmSpace)
+		reconciler := prepareReconciler(t, fakeClient)
 
 		// when
-		_, err := reconciler.Reconcile(context.TODO(), request)
+		res, err := reconciler.Reconcile(context.TODO(), requestFor(sbLaraRedhatAdmin))
 
 		// then
+		require.False(t, res.Requeue)
+		require.True(t, res.RequeueAfter == 0) // no requeue
```
MatousJobanek, line 54: why not 
```suggestion
		require.Equal(t, res.RequeueAfter, time.Duration(0)) // no requeue
```

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
```diff
@@ -34,12 +43,15 @@ func TestDeleteSpaceBinding(t *testing.T) {
...
-		reconciler, request, fakeClient := prepareReconciler(t, sbLaraRedhatAdmin, sbJoeRedhatView, sbLaraIbmEdit, laraMur, joeMur, ibmSpace)
+		fakeClient := test.NewFakeClient(t, sbLaraRedhatAdmin, sbJoeRedhatView, sbLaraIbmEdit, laraMur, joeMur, ibmSpace)
+		reconciler := prepareReconciler(t, fakeClient)
 
 		// when
-		_, err := reconciler.Reconcile(context.TODO(), request)
+		res, err := reconciler.Reconcile(context.TODO(), requestFor(sbLaraRedhatAdmin))
 
 		// then
+		require.False(t, res.Requeue)
```
MatousJobanek, line 53: You don't need that when you check the `RequeueAfter`

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
```diff
@@ -64,12 +79,15 @@ func TestDeleteSpaceBinding(t *testing.T) {
...
+		require.True(t, res.RequeueAfter == 0) // no requeue
```
MatousJobanek, line 90: Could you please update also the other places?

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
mfrancisc: what happens if the controller returns:
```
	return ctrl.Result{
		Requeue:      True,
		RequeueAfter: 0,
	}
```
I was thinking that we should cover this case. But I can remove it.

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
mfrancisc: missed those 🤦‍♂️ . Done https://github.com/codeready-toolchain/host-operator/pull/857/commits/355026b34c21e9c54bcf209a3b61533c8bac5387

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
MatousJobanek: it would immediately schedule another reconcile (with no delay).
but I would skip this check as we don't set anything and we keep the value False by default. Let's do the check only when we set True for some use-cases.

MatousJobanek: or, if you would like to keep it there, you could do:
```go
var (
	noRequeue = reconcile.Result{
		RequeueAfter: 0 * time.Second,
	}
	requeue = reconcile.Result{
		RequeueAfter: 10 * time.Second,
	}
)


...
...
// for the test when no requeue is expected
require.Equal(t, res, noRequeue)

...
...
// for the test when requeue is expected
require.Equal(t, res, requeue)
```

in this way you would easily check both values

mfrancisc: Hmm yeah, this second option makes sense, thanks! But let's keep it simple and remove it if it doesn't provide any additional value in kind of test coverage.

mfrancisc: Done, plz check https://github.com/codeready-toolchain/host-operator/pull/857/commits/390a54d71416dbd910a56b5cce82757d1b514cc8

MatousJobanek (approved):

mfrancisc:
/retest 

Flaky ? https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/857/pull-ci-codeready-toolchain-host-operator-master-e2e/1699336544403329024

PR #863, opened 2023-09-18
Propagated identity claims from UserSignup to MasterUserRecord

sbryzak:
This PR propagates a subset of the identity claims in UserSignup to MasterUserRecord.

Fixes https://issues.redhat.com/browse/SANDBOX-382


alexeykazakov (approved):

MatousJobanek:
@sbryzak I'm sorry, but this won't update the already existing MURs. You need to change also the https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/masteruserrecord.go#L9-L25 right?
Let me know if I'm missing some small detail

MatousJobanek:
The update in the `migrateOrFixMurIfNecessary` function is not important only because of this initial step of introducing the new struct, but also in the future cases when we would update the new struct by adding new fields.

PR #867, opened 2023-09-22
fix: check conflicts with Space name for complaintUsername

MatousJobanek:
https://issues.redhat.com/browse/SANDBOX-369

This should help us to workaround the issue when users cannot reactivate because of having Space stuck in a terminating state https://redhat-internal.slack.com/archives/CHK0J6HT6/p1695253339361769

paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/798

### DIFF
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -530,19 +530,27 @@ func (r *Reconciler) generateCompliantUsername(config toolchainconfig.ToolchainC
 	for i := 2; i < 101; i++ { // No more than 100 attempts to find a vacant name
 		mur := &toolchainv1alpha1.MasterUserRecord{}
 		// Check if a MasterUserRecord exists with the same transformed name
-		namespacedMurName := types.NamespacedName{Namespace: instance.Namespace, Name: newUsername}
-		err := r.Client.Get(context.TODO(), namespacedMurName, mur)
+		namespacedName := types.NamespacedName{Namespace: instance.Namespace, Name: newUsername}
+		err := r.Client.Get(context.TODO(), namespacedName, mur)
 		if err != nil {
 			if !errors.IsNotFound(err) {
 				return "", err
 			}
-			// If there was a NotFound error looking up the mur, it means we found an available name
-			return newUsername, nil
+			space := &toolchainv1alpha1.Space{}
+			err = r.Client.Get(context.TODO(), namespacedName, space)
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					return "", err
+				}
+				// If there was a NotFound error looking up the mur as well as space, it means we found an available name
+				return newUsername, nil
+			}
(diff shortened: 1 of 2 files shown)


controllers/usersignup/usersignup_controller.go
```diff
@@ -530,19 +530,27 @@ func (r *Reconciler) generateCompliantUsername(config toolchainconfig.ToolchainC
...
-			return newUsername, nil
+			space := &toolchainv1alpha1.Space{}
+			err = r.Client.Get(context.TODO(), namespacedName, space)
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					return "", err
+				}
+				// If there was a NotFound error looking up the mur as well as space, it means we found an available name
+				return newUsername, nil
+			}
```
xcoulon, line 547: Your code is valid but I believe it would be more readable (and understandable) with something like this:
```suggestion
			if err = r.Client.Get(context.TODO(), namespacedName, space); err != nil {
				if errors.IsNotFound(err) {
					// If there was a NotFound error looking up the mur as well as space, it means we found an available name
					return newUsername, nil
				}
				return "", err
			}
```

xcoulon (approved):
Looks good, just added a suggestion to improve readability (IMHO)

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -3488,14 +3488,13 @@ func TestDeathBy100Signups(t *testing.T) {
...
 
-			for i := 2; i <= 100; i++ {
-				initObjs = append(initObjs, &toolchainv1alpha1.MasterUserRecord{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      fmt.Sprintf("%s-%d", testusername.replacedCompliantUsername, i),
-						Namespace: test.HostOperatorNs,
-						Labels:    map[string]string{toolchainv1alpha1.MasterUserRecordOwnerLabelKey: uuid.Must(uuid.NewV4()).String()},
-					},
-				})
+			for i := 2; i <= 100; i += 2 {
```
rajivnathan, line 3491: I think without the context of this PR it might be a bit harder to understand this test right away so maybe we could add a short comment to explain it?
```suggestion
			// stagger the numbering for MURs and Spaces so that one of them will always be missing. eg. MURs will not be found on odd numbers and Spaces not found on even numbers until it makes 100 attempts
			for i := 2; i <= 100; i += 2 {
```

rajivnathan (approved):
Nice change!

controllers/usersignup/usersignup_controller_test.go
MatousJobanek: thanks

MatousJobanek:
/retest
fix https://github.com/codeready-toolchain/toolchain-e2e/pull/798/commits/aef1a68a5a8182bb908b8e580986225fcf8f5b1c

alexeykazakov (approved):
Nice!

PR #868, opened 2023-09-22
Set approved to false after user sign up deactivated

mmulholla:
Code a tests updated. Let me know if I should look to adding/modifying an e2e test.

https://github.com/codeready-toolchain/toolchain-e2e/pull/801


controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2205,11 +2205,12 @@ func TestUserSignupDeactivatedAfterMURCreated(t *testing.T) {
 		AssertMetricsCounterEquals(t, 0, metrics.UserSignupUniqueTotal)
 
 		// Confirm the status has been set to Deactivated
+		fmt.Println(userSignup.Status.Conditions)
```
alexeykazakov, line 2208:
```suggestion
```

alexeykazakov (commented):
Looks good! Just one minor comment. And yes, we need to update our e2e-tests to reflect this change.

controllers/usersignup/usersignup_controller.go
```diff
@@ -224,6 +224,11 @@ func (r *Reconciler) handleDeactivatedUserSignup(logger logr.Logger, config tool
 		return reconcile.Result{}, err
 	}
 
+	err = r.updateStatus(logger, userSignup, r.setStatusDeactivatedApproval)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
MatousJobanek, line 230: I would rather do the change of both conditions at the same time - doing two updates of the Status immediately after each other is not needed.
Could you please update the `setStatusDeactivated` so it sets both conditions?

MatousJobanek (commented):
Thanks a lot for your first contribution @mmulholla 
As @alexeykazakov mentioned, you need to update e2e tests. Keep in mind that the PR in toolchain-e2e repo has to be created using the same branch name in your fork as this one: `RemoveUserSignUpApprovalAfterDeactivation ` so our e2e test setup logic can pair both PRs together

controllers/usersignup/usersignup_controller.go
ranakan19: +1

mmulholla:
@alexeykazakov I understand, I do have 3 commits, just have to do a force push after rebase.

alexeykazakov:
> just have to do a force push after rebase.

Why? You don't really have to do a force push after a rebase.

MatousJobanek:
> Why? You don't really have to do a force push after a rebase.

you always need to do force-push when you do "rebase" - it changes the history of the commits. You don't need to force-push after merging as this only adds a new commit at HEAD.

alexeykazakov:
Yes, yes. Poor wording on my end. What I meant is merging of course. Not rebase. Sorry for the confusion.
My point you always can update your branch (my using merges and resolving conflicts when needed) without any need for force push.

controllers/usersignup/usersignup_controller.go
mmulholla: Done

MatousJobanek (approved):
Looks good, thanks a lot for fixing this long-pending weirdness in the user's lifecycle :+1: :1st_place_medal: :rocket: 

This will also help us in the following tasks when we will implement additional metrics for measuring the time taken to provision a reactivated account (you know, since the time the account was approved) cc @ranakan19

mfrancisc (approved):
Nice job 👍

ranakan19 (approved):

PR #869, opened 2023-09-27
fix: forward context in UserSignup controller

filariow:
1. Forwards the Reconcile's context to called requiring functions
3. Retrieve the logger from context instead of passing it to each function
4. Removes a `//nolint: unparam` comment in `controllers/usersignup/status_updater.go`.


mfrancisc (approved):
Nice improvement 👍

alexeykazakov (approved):

PR #870, opened 2023-09-27
SANDBOX-202: Differentiation between notification emails from different environments

Kartikey-star:
Updating the notifications about toolchain status content to contain the source as well. 
Reference: https://issues.redhat.com/browse/SANDBOX-202

### DIFF
--- a/controllers/toolchainstatus/toolchainstatus_controller.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller.go
@@ -12,8 +12,8 @@ import (
 	"time"
 
 	"github.com/codeready-toolchain/toolchain-common/pkg/client"
+	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
-	"k8s.io/client-go/rest"
 
 	"github.com/codeready-toolchain/host-operator/controllers/toolchainconfig"
 	notify "github.com/codeready-toolchain/toolchain-common/pkg/notification"
@@ -72,8 +72,8 @@ const (
 )
 
 const (
-	adminUnreadyNotificationSubject  = "ToolchainStatus has been in an unready status for an extended period"
-	adminRestoredNotificationSubject = "ToolchainStatus has now been restored to ready status"
+	adminUnreadyNotificationSubject  = "ToolchainStatus has been in an unready status for an extended period for %v"
+	adminRestoredNotificationSubject = "ToolchainStatus has now been restored to ready status for %v"
 )
 
 type toolchainStatusNotificationType string
@@ -485,9 +485,17 @@ func getAPIEndpoint(clusterName string, memberClusters []*cluster.CachedToolchai
 	return ""
 }
 
+func removeSchemeFromURL(proxyURL string) (string, error) {
+	url, err := url.Parse(proxyURL)
+	if err != nil {
+		return "", err
+	}
+	return url.Hostname(), nil
+}
+
 func (r *Reconciler) sendToolchainStatusNotification(ctx context.Context,
 	toolchainStatus *toolchainv1alpha1.ToolchainStatus, status toolchainStatusNotificationType) error {
-
+	logger := log.FromContext(ctx)
 	config, err := toolchainconfig.GetToolchainConfig(r.Client)
 	if err != nil {
 		return errs.Wrapf(err, "unable to get ToolchainConfig")
@@ -496,20 +504,24 @@ func (r *Reconciler) sendToolchainStatusNotification(ctx context.Context,
 	tsValue := time.Now().Format("20060102150405")
 	contentString := ""
 	subjectString := ""
+	domain := ""
+	if domain, err = removeSchemeFromURL(toolchainStatus.Status.HostRoutes.ProxyURL); err != nil {
+		logger.Error(err, fmt.Sprintf("Error while parsing proxyUrl %v", toolchainStatus.Status.HostRoutes.ProxyURL))
+		domain = toolchainStatus.Status.HostRoutes.ProxyURL
+	}
 	switch status {
 	case unreadyStatus:
 		toolchainStatus = toolchainStatus.DeepCopy()
 		toolchainStatus.ManagedFields = nil // we don't need these managed fields in the notification
-
-		clusterURLs := ClusterURLs(toolchainStatus)
+		clusterURLs := ClusterURLs(logger, toolchainStatus)
 		contentString, err = GenerateUnreadyNotificationContent(clusterURLs, ExtractStatusMetadata(toolchainStatus))
 		if err != nil {
 			return err
 		}
-		subjectString = adminUnreadyNotificationSubject
+		subjectString = fmt.Sprintf(adminUnreadyNotificationSubject, domain)
 	case restoredStatus:
 		contentString = "<div><pre>ToolchainStatus is back to ready status.</pre></div>"
-		subjectString = adminRestoredNotificationSubject
+		subjectString = fmt.Sprintf(adminRestoredNotificationSubject, domain)
 	default:
 		return fmt.Errorf("invalid ToolchainStatusNotification status type - %s", status)
 	}
@@ -520,7 +532,6 @@ func (r *Reconciler) sendToolchainStatusNotification(ctx context.Context,
 		WithSubjectAndContent(subjectString, contentString).
 		Create(config.Notifications().AdminEmail())
 
-	logger := log.FromContext(ctx)
 	if err != nil {
 		logger.Error(err, fmt.Sprintf("Failed to create toolchain status %s notification resource", status))
 		return err
@@ -530,20 +541,27 @@ func (r *Reconciler) sendToolchainStatusNotification(ctx context.Context,
 	return nil
 }
 
-func ClusterURLs(instance *toolchainv1alpha1.ToolchainStatus) map[string]string {
-	cfg, err := rest.InClusterConfig()
-	if err != nil {
-		result := map[string]string{}
+func ClusterURLs(logger logr.Logger, instance *toolchainv1alpha1.ToolchainStatus) map[string]string {
 
-		for _, mbr := range instance.Status.Members {
-			result["Member cluster"] = mbr.ClusterName
+	if instance.Status.HostRoutes.ProxyURL != "" {
+		var domain string
+		var err error
+		if domain, err = removeSchemeFromURL(instance.Status.HostRoutes.ProxyURL); err != nil {
+			logger.Error(err, fmt.Sprintf("Error while parsing proxyUrl %v", instance.Status.HostRoutes.ProxyURL))
+			domain = instance.Status.HostRoutes.ProxyURL
 		}
+		return map[string]string{
+			"Cluster URL": domain,
+		}
+	}
 
-		return result
+	result := map[string]string{}
+
+	for _, mbr := range instance.Status.Members {
+		result["Member cluster"] = mbr.ClusterName
 	}
-	return map[string]string{
-		"Cluster URL": cfg.Host,
-	}
+
+	return result
 }
 
 type ComponentNotReadyStatus struct {
--- a/controllers/toolchainstatus/toolchainstatus_controller.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller.go
@@ -548,10 +548,10 @@ func ClusterURLs(logger logr.Logger, instance *toolchainv1alpha1.ToolchainStatus
 		var err error
 		if domain, err = removeSchemeFromURL(instance.Status.HostRoutes.ProxyURL); err != nil {
 			logger.Error(err, fmt.Sprintf("Error while parsing proxyUrl %v", instance.Status.HostRoutes.ProxyURL))
-			domain = instance.Status.HostRoutes.ProxyURL
-		}
-		return map[string]string{
-			"Cluster URL": domain,
+		} else {
+			return map[string]string{
+				"Cluster URL": domain,
+			}
 		}
 	}
 
--- a/controllers/toolchainstatus/toolchainstatus_controller.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller.go
@@ -507,7 +507,6 @@ func (r *Reconciler) sendToolchainStatusNotification(ctx context.Context,
 	domain := ""
 	if domain, err = removeSchemeFromURL(toolchainStatus.Status.HostRoutes.ProxyURL); err != nil {
 		logger.Error(err, fmt.Sprintf("Error while parsing proxyUrl %v", toolchainStatus.Status.HostRoutes.ProxyURL))
-		domain = toolchainStatus.Status.HostRoutes.ProxyURL
 	}
 	switch status {
 	case unreadyStatus:
--- a/controllers/toolchainstatus/toolchainstatus_controller_test.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller_test.go
@@ -1725,7 +1725,7 @@ func TestGenerateUnreadyNotificationContent(t *testing.T) {
 }
 
 func TestRemoveSchemeFromURL(t *testing.T) {
-	t.Run("test proxy url domain extraction", func(t *testing.T) {
+	t.Run("test proxy url domain extraction validURL", func(t *testing.T) {
 		// when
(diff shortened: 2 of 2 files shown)


controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -479,6 +480,16 @@ func getAPIEndpoint(clusterName string, memberClusters []*cluster.CachedToolchai
 	return ""
 }
 
+func fetchDomainFromUrl(logger logr.Logger, proxyUrl string) string {
+	url, err := url.Parse(proxyUrl)
+	if err != nil {
+		logger.Error(err, fmt.Sprintf("Error while parsing the proxyURL"))
```
mfrancisc, line 486: should we return the error here , as it might not be able to proceed with the rest of the function?

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -524,7 +535,7 @@ func (r *Reconciler) sendToolchainStatusNotification(logger logr.Logger,
 }
 
 func ClusterURLs(instance *toolchainv1alpha1.ToolchainStatus) map[string]string {
-	cfg, err := rest.InClusterConfig()
+	_, err := rest.InClusterConfig()
```
mfrancisc, line 538: if we don't use `cfg` anymore should we avoid invoking this function?

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1724,6 +1724,14 @@ func TestGenerateUnreadyNotificationContent(t *testing.T) {
 	})
 }
 
+func TestfetchDomainFromUrl(t *testing.T) {
+	t.Run("test proxy url domain extraction", func(t *testing.T) {
+		domain := fetchDomainFromUrl(logger, "https://api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com")
+		require.Equal(t, "openshiftapps.com", domain)
```
mfrancisc, line 1730: I think removing the scheme `https://` should be good enough . As we need also the other URL parts in order to identify the environment. In this case `stone-stg-host` is what defines the RHTAP environment.

So this should be good enough as expected value:
```
api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com
```

mfrancisc (commented):
Nice 👍 

I saw it's still a Draft but I left few minor comments anyway.

controllers/toolchainstatus/toolchainstatus_controller.go
Kartikey-star: The reason i still have this call is because there is a logic to populate the map with `Member Cluster`, if there is an errir to retrieve InClusterConfig. Just wanted to retain that logic. https://github.com/codeready-toolchain/host-operator/blob/master/controllers/toolchainstatus/toolchainstatus_controller.go

controllers/toolchainstatus/toolchainstatus_controller_test.go
Kartikey-star: Yeah makes sense

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -490,20 +498,23 @@ func (r *Reconciler) sendToolchainStatusNotification(logger logr.Logger,
 	tsValue := time.Now().Format("20060102150405")
 	contentString := ""
 	subjectString := ""
+	domain := ""
+	if domain, err = removeSchemeFromUrl(toolchainStatus.Status.HostRoutes.ProxyURL); err != nil {
+		return err
```
rajivnathan, line 503: I think in this case we would still want to send the notification even if there was an issue parsing the proxy URL so how about we log the error and set domain to the ClusterURL instead. That way the notification would still be sent to our mailing list.

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1724,6 +1724,15 @@ func TestGenerateUnreadyNotificationContent(t *testing.T) {
 	})
 }
 
+func TestRemoveSchemeFromUrl(t *testing.T) {
+	t.Run("test proxy url domain extraction", func(t *testing.T) {
+		domain, err := removeSchemeFromUrl("https://api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com")
+		require.Equal(t, "api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com", domain)
+		require.Nil(t, err)
```
rajivnathan, line 1731: Minor thing but our tests are usually sectioned by 'given', 'when', 'then' comments. With them it's easier to distinguish between test setup, what's being tested and the assertions.
```suggestion
		// when
		domain, err := removeSchemeFromUrl("https://api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com")
		// then
		require.Equal(t, "api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com", domain)
		require.Nil(t, err)
```

controllers/toolchainstatus/toolchainstatus_controller.go
mfrancisc: Right, but since we now want to use `toolchainStatus.Status.HostRoutes.ProxyURL` instead of `cfg` maybe we could check if `toolchainStatus.Status.HostRoutes.ProxyURL` is set here instead and if we can parse it using `removeSchemeFromUrl`. If not then we fallback on the logic that populates the map with `Member Cluster` logic. WDYT?

controllers/toolchainstatus/toolchainstatus_controller.go
mfrancisc: +1 
we could set it to ClusterURL or leave the `domain` empty as it is now

controllers/toolchainstatus/toolchainstatus_controller.go
Kartikey-star: Yeah that can de done.

controllers/toolchainstatus/toolchainstatus_controller.go
Kartikey-star: Yeah makes sense, instead of returning here, will log the error and set the domain to ClusterURL

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -524,19 +535,25 @@ func (r *Reconciler) sendToolchainStatusNotification(logger logr.Logger,
...
-	if err != nil {
-		result := map[string]string{}
 
-		for _, mbr := range instance.Status.Members {
-			result["Member cluster"] = mbr.ClusterName
+	if instance.Status.HostRoutes.ProxyURL != "" {
+		var domain string
+		var err error
+		if domain, err = removeSchemeFromUrl(instance.Status.HostRoutes.ProxyURL); err != nil {
+			domain = instance.Status.HostRoutes.ProxyURL
```
alexeykazakov, line 543: Let's log the error here.

mfrancisc: I guess it's a small corner case , but I'm wondering if we should still rely on the `.Status.HostRoutes.ProxyURL` when there is an error removing the scheme ( it probably means that the URL is invalid). It shouldn't break the email delivery part. 

Could you please add a unit test for this error case?

rajivnathan: I guess `domain = instance.Status.HostRoutes.ProxyURL` is a mistake? It's supposed to set domain to the ClusterURL?

rajivnathan: +1 to adding a unit test for the error case

Kartikey-star: Yeah makes sense.Added a test to pass incorrect url and remove the setting of domain to proxyURL in case of error

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -490,20 +497,24 @@ func (r *Reconciler) sendToolchainStatusNotification(logger logr.Logger,
 	tsValue := time.Now().Format("20060102150405")
 	contentString := ""
 	subjectString := ""
+	domain := ""
+	if domain, err = removeSchemeFromURL(toolchainStatus.Status.HostRoutes.ProxyURL); err != nil {
+		logger.Error(err, fmt.Sprintf("Error while parsing proxyUrl %v", toolchainStatus.Status.HostRoutes.ProxyURL))
+		domain = toolchainStatus.Status.HostRoutes.ProxyURL
```
mfrancisc, line 503: I guess we want to not set the domain here when is invalid right?  as you did on line https://github.com/codeready-toolchain/host-operator/pull/870/files#diff-dbe76ee9dfabe95c78e8d2e7c3efdf6526bfd4719168b8529c518129cb83a6acR543
```suggestion
```

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1724,6 +1724,26 @@ func TestGenerateUnreadyNotificationContent(t *testing.T) {
...
+
+		// when
+		domain, err = removeSchemeFromURL("incorrect$%url")
+
+		// then
+		require.Equal(t, "", domain)
+		require.NotNil(t, err)
+
+	})
```
mfrancisc, line 1743: thanks for adding the unit test! 

Could you please split it into 2 subtests, maybe something like "valid URL" and "incorrect URL?

controllers/toolchainstatus/toolchainstatus_controller.go
```diff
@@ -490,20 +497,23 @@ func (r *Reconciler) sendToolchainStatusNotification(logger logr.Logger,
...
 		toolchainStatus.ManagedFields = nil // we don't need these managed fields in the notification
-
-		clusterURLs := ClusterURLs(toolchainStatus)
+		clusterURLs := ClusterURLs(logger, toolchainStatus)
 		contentString, err = GenerateUnreadyNotificationContent(clusterURLs, ExtractStatusMetadata(toolchainStatus))
 		if err != nil {
 			return err
 		}
-		subjectString = adminUnreadyNotificationSubject
+		subjectString = fmt.Sprintf(adminUnreadyNotificationSubject, domain)
```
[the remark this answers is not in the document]
alexeykazakov: Yep. We need a unit test for that.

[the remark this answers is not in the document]
Kartikey-star: Sure , adding a test case for that.

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1297,7 +1297,230 @@ func TestToolchainStatusNotifications(t *testing.T) {
...
+								require.NoError(t, err)
+								assert.Equal(t, requeueResult, res)
+								// Confirm restored notification is not created
+								assertToolchainStatusNotificationNotCreated(t, fakeClient, restoredStatusNotification)
+								// Confirm the unready notification has been created
+								notification := assertToolchainStatusNotificationCreated(t, fakeClient)
+								require.True(t, strings.HasPrefix(notification.ObjectMeta.Name, "toolchainstatus-unready-"))
+								require.Len(t, notification.ObjectMeta.Name, 38)
+								require.NotNil(t, notification)
+								assert.Equal(t, "ToolchainStatus has been in an unready status for an extended period for ", notification.Spec.Subject)
```
mfrancisc, line 1523: @Kartikey-star thanks for adding the test!

How about adding just this subtest onto the existing `TestToolchainStatusNotifications` at the same level of "Toolchain status not ready again for extended period, notification is created" and maybe naming this one "Toolchain status not ready again for extended period, notification is created with invalid ProxyURL" ? In order to avoid some duplication.

Also maybe some additional common logic can be extracted to function.

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -1785,6 +2031,15 @@ func proxyRoute() *routev1.Route {
...
 }
 
+func proxyRouteInvalid() *routev1.Route {
+	return &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "api",
+			Namespace: test.HostOperatorNs,
+		},
+	}
+}
```
mfrancisc, line 2041: to trigger the url parsing error you need to specify an invalid URL as well. A blank one is not enough. 
```suggestion
	return &routev1.Route{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "api",
			Namespace: test.HostOperatorNs,
		},
		Spec: routev1.RouteSpec{
			Host: ":invalidhost",
			Port: &routev1.RoutePort{
				TargetPort: intstr.FromInt(8081),
			},
			TLS: &routev1.TLSConfig{
				Termination: routev1.TLSTerminationEdge,
			},
		},
	}
```

Kartikey-star: Yeah this works, I was trying to set the Host to `invalidhost` instead of `:invalidhost`, hence the error was not reachable .

controllers/toolchainstatus/toolchainstatus_controller_test.go
Kartikey-star: Thanks, that will make it a lot better

mfrancisc (approved):
Looks good to me 👍 

mfrancisc:
/retest

@sbryzak strange verify migration test error: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/870/pull-ci-codeready-toolchain-host-operator-master-e2e/1709451270005198848/build-log.txt

propagated fields seems to not match.
```
        -expected
        +actual
          any(
        - 	&v1alpha1.UserAccountSpec{
        - 		UserID:      "579c1f96-4f92-4cf4-8367-2272a0abd2b0",
        - 		OriginalSub: "original_sub_mig-m2-user",
        - 		PropagatedClaims: v1alpha1.PropagatedClaims{
        - 			Sub:         "579c1f96-4f92-4cf4-8367-2272a0abd2b0",
        - 			UserID:      "a171ee75-a982-44fd-8727-b5953adf6971",
        - 			AccountID:   "123a4bce-618a-4c24-8881-f5b901baf2dd",
        - 			OriginalSub: "original_sub_mig-m2-user",
        - 			Email:       "testuser-268f983a-0a02-49cf-8b07-5925b9e27249@test.com",
        - 		},
        - 	},
        + 	v1alpha1.UserAccountSpec{
        + 		UserID:      "579c1f96-4f92-4cf4-8367-2272a0abd2b0",
        + 		OriginalSub: "original_sub_mig-m2-user",
        + 		PropagatedClaims: v1alpha1.PropagatedClaims{
        + 			Sub:         "695be52e-50ab-47de-acd8-884879e22e6d",
        + 			UserID:      "718a2981-4ad4-4f46-ac99-768b972a4683",
        + 			AccountID:   "30179a1f-34ec-457c-b44b-5ee27795d6ce",
        + 			OriginalSub: "original_sub_mig-m2-user",
        + 			Email:       "testuser-9916f30a-240e-4181-8481-4a056373a28a@test.com",
        + 		},
        + 	},
          ) 
    user_assertions.go:68: 
        	Error Trace:	/tmp/toolchain-e2e/testsupport/user_assertions.go:68
        	            				/tmp/toolchain-e2e/testsupport/user_assertions.go:30
        	            				/tmp/toolchain-e2e/test/migration/verify/verify_migration_test.go:131
        	            				/tmp/toolchain-e2e/test/migration/verify/verify_migration_test.go:88
        	            				/tmp/toolchain-e2e/test/migration/verify/verify_migration_test.go:99
        	            				/usr/local/go/src/runtime/asm_amd64.s:1594
```

Kartikey-star:
@mfrancisc e2e passed now, was it flaky?

alexeykazakov:
Could you please resolve the conflict? Thanks.

alexeykazakov:
Please do not push new commits with force in PRs. Keep adding new commits to what is already there instead. It makes PR reviews much easier. The reviewers could see what has changed since they looked at the PR last time. With forced push they have to review everything from scratch every time.

PR #874, opened 2023-10-02
Network policy for cnv

rajivnathan:
Add network policy required for CNV operator to work with Dev Sandbox user namespaces.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/804


alexeykazakov (approved):
The PR looks good. However I think we need to raise it to the CNV folks. They should probably consider adding a `network.openshift.io/policy-group:` label to their `openshift-virtualization-os-images` namespace. In that case they would be able to rename their image namespace name in the future without breaking any existing NetworkPolicies.

PR #876, opened 2023-10-03
Allow user signups to complete when the associated space is updating

sadlerap:
If we mark user signups as incomplete while a space is updating, then the registration controller will reject the login.  As a workaround, allow user signups to progress while the associated space is updating.

Resolves https://issues.redhat.com/browse/SANDBOX-72.

### DIFF
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -357,7 +357,9 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 
 			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
 			reqLogger.Info("Checking whether Space is Ready", "Space", space.Name)
-			if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) {
+			if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
+				// don't mark the user signup as incomplete if the space is updating - wait for the space to finish first.
+				condition.IsFalseWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceUpdatingReason)) {
 				return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
 			}
 		}
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -357,8 +357,8 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 
 			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
 			reqLogger.Info("Checking whether Space is Ready", "Space", space.Name)
-			if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
-				condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)) {
+			if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) &&
+				!condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete) {
 				return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
 			}
 		}
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -358,8 +358,7 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
 			reqLogger.Info("Checking whether Space is Ready", "Space", space.Name)
 			if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
-				// don't mark the user signup as incomplete if the space is updating - wait for the space to finish first.
-				condition.IsFalseWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceUpdatingReason)) {
+				condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)) {
 				return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
 			}
 		}
--- a/controllers/usersignup/usersignup_controller_test.go
+++ b/controllers/usersignup/usersignup_controller_test.go
@@ -4385,23 +4385,64 @@ func TestUserSignupLastTargetClusterAnnotation(t *testing.T) {
 func TestUserSignupStatusNotReady(t *testing.T) {
 	member := NewMemberClusterWithTenantRole(t, "member1", corev1.ConditionTrue)
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
-	userSignup := commonsignup.NewUserSignup(
-		commonsignup.ApprovedManually(),
-		commonsignup.WithTargetCluster("member1"),
-		commonsignup.WithStateLabel(toolchainv1alpha1.UserSignupStateLabelValueNotReady),
-		commonsignup.WithoutAnnotation(toolchainv1alpha1.SkipAutoCreateSpaceAnnotationKey))
 
-	mur := newMasterUserRecord(userSignup, "member1", deactivate30Tier.Name, "foo")
-	mur.Labels = map[string]string{toolchainv1alpha1.MasterUserRecordOwnerLabelKey: userSignup.Name}
-	space := spacetest.NewSpace(test.HostOperatorNs, "foo",
-		spacetest.WithSpecTargetCluster("member-1"),
-		spacetest.WithStatusTargetCluster("member-1"), // already provisioned on a target cluster
-		spacetest.WithFinalizer())
+	setup := func() (*toolchainv1alpha1.UserSignup, *toolchainv1alpha1.MasterUserRecord, *toolchainv1alpha1.Space, *toolchainv1alpha1.SpaceBinding) {
+		userSignup := commonsignup.NewUserSignup(
+			commonsignup.ApprovedManually(),
+			commonsignup.WithTargetCluster("member1"),
+			commonsignup.WithStateLabel(toolchainv1alpha1.UserSignupStateLabelValueNotReady),
+			commonsignup.WithoutAnnotation(toolchainv1alpha1.SkipAutoCreateSpaceAnnotationKey))
 
-	spacebinding := spacebindingtest.NewSpaceBinding("foo", "foo", "admin", userSignup.Name)
+		mur := newMasterUserRecord(userSignup, "member1", deactivate30Tier.Name, "foo")
+		mur.Labels = map[string]string{toolchainv1alpha1.MasterUserRecordOwnerLabelKey: userSignup.Name}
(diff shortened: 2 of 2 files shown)


controllers/usersignup/usersignup_controller.go
```diff
@@ -357,7 +357,9 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 
 			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
 			reqLogger.Info("Checking whether Space is Ready", "Space", space.Name)
-			if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) {
+			if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
+				// don't mark the user signup as incomplete if the space is updating - wait for the space to finish first.
+				condition.IsFalseWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceUpdatingReason)) {
 				return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
 			}
```
alexeykazakov, line 364: I wonder if it's safer to not update any conditions at all if the space in in the updating state. Something like this:
```
if condition.IsFalseWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceUpdatingReason) {
	// don't mark the user signup as incomplete if the space is updating - wait for the space to get ready first.
	return true, nil
}
if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason {
	return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
}
```
Basically if the space is being updated then we don't touch the usersignup conditions. It stays in whatever status it's already in. Trying to avoid a potential edge case situation when the UserSignup in some incomplete state already (some error state for example) and the space is being updated. We should not (re-)set the UserSignup complete condition to true just because space is being updated. We just should avoid setting already "complete" condition to "incomplete" when updating. I'm not sure if it can happen... But just to be on the safe side.

@MatousJobanek, @mfrancisc WDYT?

And if we do want to cover that use case then we will need to add a unit test for this too.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -4430,6 +4430,42 @@ func TestUserSignupStatusNotReady(t *testing.T) {
 
 	})
 
+	// If a space is updating, don't mark usersignups as "Incomplete", since
+	// they'll get rejected by the registration service
```
alexeykazakov, line 4434: "Rejected" word can be a bit confusing here. The usersignup is not being rejected per se. It will be marked as "Incomplete" which can cause issues like the kube API proxy requests being rejected or wrong user signup status in UI (pending provisioning instead of ready).

alexeykazakov (commented):
Looks good overall. Thanks!

controllers/usersignup/usersignup_controller.go
MatousJobanek: If the UserSignup is in an incomplete (broken) state already, then reflecting the status from the Space doesn't change on the UserSignup state much. It either eventually recover after the Space update or not. In that case, the state of the UserSignup should be still incomplete.

TBH, I wouldn't make the logic hard-coded so it would rely on the specific reason in the Space CR. As soon as the UserSignup ends up in a complete state (after creating MUR and Space CRs and waiting for their readiness), then I wouldn't reflect any updates in status of the Space in the UserSignup CR. The job of the UserSignup controller is done - it's complete and it should stay as it is no matter what is going on with the actual Space CR. This is also the reason why we named the condition as "complete" and not "ready"

1. UserSignup is approved
2. It creates MUR, Space, and SpaceBinding
3. It waits until Space is ready
4. When Space is ready then it marks the "complete" condition to true
5. If there is something going on with the Space CR later on, then the UserSignup should care about that - the "Complete" condition should stay "true"

mfrancisc: I like @MatousJobanek proposal to not reflect the Space status in the UserSignup once the UserSignup is "Complete" 👍

mfrancisc: But maybe we will need a way to notify the user in case there is an issue with the space he/she is trying to access.  I know a User might have access to different (work)spaces see RHTAP for example.

sadlerap: Just to make sure I'm understanding this correctly: we want to keep usersignups that have already completed, but mark all the other signups as incomplete (provided they're not already rejected for another reason, such as a deactivated/banned user).

alexeykazakov: If I understand @MatousJobanek the proposal is to stop setting the `complete` condition to `false` if the Space is not ready. I like the idea in general. However I have a question.
There are two options. We just do not set the `complete` condition at all until the space is ready. So the condition could be missing/unknown until space is ready.
Or we do something like this:
```
if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason && !USER_SIGNUP_COMPLETE_IS_TRUE {
	return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s is not ready", space.Name))
}
```
So, it does set it to incomplete with the `space nor ready` message but only if it's not complete yet. Just to provide more details why it's not complete yet. If it was set to compelete then we don't change it no matter what status of the space.

The other cases (banned, etc) stay intact.

sadlerap: I've implemented the second option with 97d45fe, though I can change it if we want to go with the first.

```go
if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
	condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)) {
	return true, r.updateIncompleteStatus(userSignup, fmt.Sprintf("space %s was not ready", space.Name))
}
```

mfrancisc (approved):
Looks good. Nice job 👍

controllers/usersignup/usersignup_controller.go
```diff
@@ -357,7 +357,8 @@ func (r *Reconciler) checkIfMurAlreadyExists(reqLogger logr.Logger, config toolc
 
 			// if the Space is not Ready, another reconcile will occur when space status is updated since this controller watches space (without a predicate)
 			reqLogger.Info("Checking whether Space is Ready", "Space", space.Name)
-			if !condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) {
+			if !(condition.IsTrueWithReason(space.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.SpaceProvisionedReason) ||
+				condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)) {
```
mfrancisc, line 361: just one minor comment about the form (not the content) - I find it easier to read using `AND` instead of `OR`, as @alexeykazakov wrote it in the previous comment:
```
if !condition1 && !condition2 {
  return error
}
``` 
but maybe it's just personal preference :)

mfrancisc:
Should we add e2e test case for this scenario ?

controllers/usersignup/usersignup_controller.go
sadlerap: Ultimately it's the same logic, but I'll adjust it.

Fixed in ae1e668.

controllers/usersignup/usersignup_controller.go
MatousJobanek: > But maybe we will need a way to notify the user in case there is an issue with the space he/she is trying to access. 

let's address this as a separate task. Such a logic could be placed either here or for example in registration-service (we could provide similar functionality for (work)space-lister too).
For the DevSandbox use-case, we would need some support on UI side for that too, to show more states or errors of the UserSignup resource.

> So, it does set it to incomplete with the space nor ready message but only if it's not complete yet. Just to provide more details why it's not complete yet. If it was set to compelete then we don't change it no matter what status of the space.

Yes, this is exactly what I meant. Let me update the flow by adding another line/step
1. UserSignup is approved
2. It creates MUR, Space, and SpaceBinding
3. It waits until Space is ready
4. When UserSignup controller gets reconciled and the Space is still not ready, then it updates the condition with `space <space_name> is not ready`
5. When Space is ready then it marks the "complete" condition to true
6. If there is something going on with the Space CR later on, then the UserSignup shouldn't care about that - the "Complete" condition should stay "true" 

so the UserSignup controller reflects that the Space is not ready, but only until it gets ready for the first time. The it marks the condition as complete and whatever happens to the Space is ignored by the controller

> The other cases (banned, etc) stay intact.

yes, exactly

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -4463,6 +4463,76 @@ func TestUserSignupStatusNotReady(t *testing.T) {
...
+		}}
+		r, req, _ := prepareReconcile(t, userSignup.Name, NewGetMemberClusters(member), userSignup, mur, space, spacebinding, commonconfig.NewToolchainConfigObjWithReset(t, testconfig.AutomaticApproval().Enabled(true)), baseNSTemplateTier, deactivate30Tier)
+		// when
+		res, err := r.Reconcile(context.TODO(), req)
+		require.NoError(t, err)
+		require.Equal(t, reconcile.Result{}, res)
+		// then
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+		require.NoError(t, err)
+		test.AssertConditionsMatch(t, userSignup.Status.Conditions,
```
MatousJobanek, line 4483: I know that you copied the previous test cases, so the bug was already there, but we shouldn't use the same object for retrieving the actual UserSignup resource, because it influences the following subtests. We should use separate object instead.
```suggestion
		actualUserSignup := &toolchainv1alpha1.UserSignup{}
		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, actualUserSignup)
		require.NoError(t, err)
		test.AssertConditionsMatch(t, actualUserSignup.Status.Conditions,
```
I would be really thankful if you could update also the previous two test cases. When you update it, you will see that this one will start failing - it's just because the UserSignup won't have the needed condition upfront.

Sorry about the inconvenience. In most of the cases we use custom assertions with fluent-api, however, we haven't got to it for UserSignup yet.

controllers/usersignup/usersignup_controller.go
MatousJobanek: in other words, exactly as @sadlerap did in his latest changes :pray:

alexeykazakov:
> Should we add e2e test case for this scenario ?

It's probably going to be tricky to cover it by e2e-tests. Do you have a specific test scenario in mind?

controllers/usersignup/usersignup_controller_test.go
sadlerap: I did some cleanup and refactoring of these tests in 583f076 - should remove the dependency between tests.

controllers/usersignup/usersignup_controller_test.go
sadlerap: I've updated these tests in both 97d45fe and 583f076, please re-review.

alexeykazakov (approved):
Thank you for addressing all the comments!

sadlerap:
Looks like codecov upload failed, but it doesn't look like an issue with my tests though.

alexeykazakov:
> Looks like [codecov upload failed](https://github.com/codeready-toolchain/host-operator/actions/runs/6421632076/job/17436997393?pr=876), but it doesn't look like an issue with my tests though.

Yeah... it fails from time to time. I restarted the job.

mfrancisc:
> It's probably going to be tricky to cover it by e2e-tests. Do you have a specific test scenario in mind?

@alexeykazakov It might not be trivial indeed, I was thinking if we could:
- create the usersignup and wait for the space to be provisioned
- update (or keep updating the namespace using a separte goroutine) the namespace so that nstemplateset and space `ConditionReady` switches to `false` (with `SpaceUpdatingReason`) 
- check that usersignup has still `UserSignupComplete` condition

alexeykazakov:
> It might not be trivial indeed, I was thinking if we could:

I think the unit tests cover enough. And as you said the e2e-tests would not be trivial and also not very reliable (you would need to catch the updating state).

MatousJobanek (approved):
great :+1: 
Thanks a lot for the fix and for improving the tests :1st_place_medal: :rocket:

PR #878, opened 2023-10-05
Adding Pod name for idle email notification

fbm3307:
Adding Pod name for idle email notification 
Related to https://github.com/codeready-toolchain/member-operator/pull/478
and issue https://issues.redhat.com/browse/SANDBOX-186


deploy/templates/notificationtemplates/sandbox/idlertriggered/notification.html
```diff
@@ -47,7 +47,7 @@
 
     <p>
         In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
-        application to zero (0). You can restart your application(s) by increasing the number of instances from the
+        application {{.PodName}} to zero (0). You can restart your application(s) by increasing the number of instances from the
```
mfrancisc, line 50: just a small reminder - based on https://github.com/codeready-toolchain/member-operator/pull/478#discussion_r1346958235 should we rename the variable to `.AppName` ?

fbm3307: will be changing it applicationname

fbm3307:
Updated the PR

deploy/templates/notificationtemplates/appstudio/idlertriggered/notification.html
```diff
@@ -46,7 +46,7 @@
 
     <p>
         In accordance with the usage terms, we have reduced the number of instances of your
-        application to zero (0). You can restart your application(s) by increasing the number of instances from the User Interface.
+        application {{.Apptype}} {{.Appname}} to zero (0). You can restart your application(s) by increasing the number of instances from the User Interface.
```
alexeykazakov, line 49: Let's use CammelCase as we mentioned in the other PR.

fbm3307: Done

mfrancisc:
Not related to this change - but since this change depends on the member operator change this means both PRs need to be merged and deploy at the same time right? 

If that is true then could we move those idle notification assets to member operator (if it's the one that is using them) ? And make this process simpler.

mfrancisc:
@fbm3307 I think you also need to update the test , please check : https://github.com/codeready-toolchain/host-operator/actions/runs/6466713541/job/17555291564?pr=878

fbm3307:
> @fbm3307 I think you also need to update the test , please check : https://github.com/codeready-toolchain/host-operator/actions/runs/6466713541/job/17555291564?pr=878

Updated it

alexeykazakov:
> Not related to this change - but since this change depends on the member operator change this means both PRs need to be merged and deploy at the same time right?

Not really. The member operator change should go first. It just populate the new template params which wouldn't be used in the currently deployed host operator templates (from the master branch). So the templates won't be broken.
Then when the member operator PR is merged and deployed we can merge this host PR which introduces new templates which would now use the params populated by the member operator.

mfrancisc:
> Not really. The member operator change should go first.

Got it, the template will not be broken as soon as the member operator PR goes in for first. Thanks for explaining that 👍 

I had a quick chat with @MatousJobanek , and he also explained me some of the historical reasons and told me that there are plans for decoupling more the html templates from the operators. 

One quick idea could be to add the html inside the Notification CR, something like:
```
kind: Notification
.....
spec:
  context:
    emailBodyHTML: <p>Yehaw!</p>
```
so that host-operator responsibility is just to send/manage those notifications, and html templates can be stored outside (in member-operator or any other external operator that wants to create a Notification CR ) .

MatousJobanek (commented):
Looks all good, let's wait until the other PR in member operator is approved and merged

MatousJobanek:
> One quick idea could be to add the html inside the Notification CR, something like

We already support that for custom email notifications (that have content that cannot be specified as a template - like the ToolchainStatus one). 
But for these cases, we specify only the template name because adding the html template in a spec of the notification CR is a duplication of the same thing that is already present in the host cluster. 
However, we could change the notification controller so it updates the status with the actual content of what was sent via Mailgun - that would make testing and investigation of any issues much easier. But this is something to be done as a separate task of course :-) I created an epic for the whole effort of customizing the email templates  [SANDBOX-433](https://issues.redhat.com/browse/SANDBOX-433) It could be done as part of this effort

alexeykazakov (approved):

PR #879, opened 2023-10-05
ci: bump actions to silence node12 warnings

sadlerap:
As of [June 2023][1], GitHub no longer uses node v12 within its actions runner with the intention of removing support for this version in the future.  They have been encouraging action authors to switch to node v16 and [encouraging users][2] to use the new versions of the actions when they become available.

Bump the versions of the affected actions to bring this in line.

[1]: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
[2]: https://github.com/codeready-toolchain/host-operator/actions/runs/6421632076

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -20,7 +20,7 @@ jobs:
         go-version: 1.19.x
     
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
     
     - name: generate
       run: |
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):
Thanks!

alexeykazakov:
I guess we need to update other repos too?

sadlerap:
@alexeykazakov I believe so - I can start that process for the other repositories.

alexeykazakov:
>  I can start that process for the other repositories.

That would be great. Thank you!

PR #880, opened 2023-10-05
Synchronize propagated claims when updated in UserSignup

sbryzak:
This PR synchronizes changes made to the PropagatedClaims in UserSignup to MUR and UserAccount

Fixes https://issues.redhat.com/browse/SANDBOX-397

Related e2e test: https://github.com/codeready-toolchain/toolchain-e2e/pull/813


controllers/masteruserrecord/masteruserrecord_controller_test.go
```diff
@@ -171,6 +171,93 @@ func TestCreateUserAccountSuccessful(t *testing.T) {
 		})
 }
 
+func TestUserAccountSynchronizeSuccessfulWhenPropagatedClaimsModified(t *testing.T) {
```
alexeykazakov, line 174: Hm... I would expect that we already have a test which tests MUR->UserAccount synchronization which we just need to update to include a PropogatedClaim check. Instead of creating a fully new test. Are you saying we don't have such a test?

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2047,6 +2047,70 @@ func TestUserSignupWithExistingMURDifferentUserIDOK(t *testing.T) {
 	})
 }
 
+func TestUserSignupPropagatedClaimsSynchronizedToMURWhenModified(t *testing.T) {
```
alexeykazakov, line 2050: And the same question here.

alexeykazakov (commented):
Overall look good. But I have some questions regarding the tests.

controllers/masteruserrecord/masteruserrecord_controller_test.go
sbryzak: I don't think we have a general test.  There is `TestSyncMurStatusWithUserAccountStatuses` but it's specifically testing status synchronization.  There is also `TestCreateSynchronizeOrDeleteUserAccountFailed` which tests various failure scenarios.  If you're aware of any other tests that I may have missed please let me know.

controllers/usersignup/usersignup_controller_test.go
sbryzak: We have a huge number of UserSignup unit tests, I did go through them but was unable to identify anything close to what I wanted to test with this new functionality.

controllers/masteruserrecord/masteruserrecord_controller_test.go
alexeykazakov: Hm... ok. Just wanted to avoid too much of code duplication.

alexeykazakov (approved):
Looks good overall. Was hoping that the test could be organized a bit better to avoid code duplication where it makes sense to keep the code cleaner.

sbryzak:
Thanks for the approval.  I agree that it would be nice to have less code duplication but there's an incredible amount of boilerplate environment that's now required to be setup before a basic reconciliation will even pass successfully, with the cluster setup, counters/metrics, toolchain status etc all having to be established first.  We do have a lot of reusable code too, such as the following block which is used countless times to reduce the overall amount of code.

```// Confirm that the properties still match
	uatest.AssertThatUserAccount(t, "ricky", memberClient).
		Exists().
		MatchMasterUserRecord(mur).
		HasLabelWithValue(toolchainv1alpha1.TierLabelKey, "deactivate30").
		HasAnnotationWithValue(toolchainv1alpha1.SSOUserIDAnnotationKey, "999888").
		HasAnnotationWithValue(toolchainv1alpha1.SSOAccountIDAnnotationKey, "777666").
		HasSpec(toolchainv1alpha1.UserAccountSpec{
			UserID:           mur.Spec.UserID,
			Disabled:         false,
			OriginalSub:      mur.Spec.OriginalSub,
			PropagatedClaims: mur.Spec.PropagatedClaims,
		})```

File: controllers/usersignup/usersignup_controller_test.go (hunk 2 above)
Comment (MatousJobanek) [reply controllers/usersignup/usersignup_controller_test.go:2050]: you updated the `migrateOrFixMurIfNecessary` function in `masteruserrecord.go` file, there is also `masteruserrecord_test.go` file that contains unit-tests for it, one of them is `TestMigrateMurIfNecessary` which would be cool to update as well when you changed the logic of the function

Comment (MatousJobanek) [reply controllers/usersignup/usersignup_controller_test.go:2050]: btw, the `usersignup_controller_test.go` file has a test called `TestMigrateMur` - to find it was as easy as looking for "migrate" word in the file

File: controllers/usersignup/usersignup_controller_test.go [hunk 3]
```diff
@@ -2047,6 +2047,70 @@ func TestUserSignupWithExistingMURDifferentUserIDOK(t *testing.T) {
...
+	res, err := r.Reconcile(context.TODO(), req)
+
+	// then
+	require.NoError(t, err)
+	require.Equal(t, reconcile.Result{}, res)
+
+	// Lookup the user signup again
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
+	require.NoError(t, err)
+	assert.Equal(t, "approved", userSignup.Labels[toolchainv1alpha1.UserSignupStateLabelKey])
```
MatousJobanek, line 2077: there is custom assertion you could use for that
```
AssertThatUserSignup(...).HasLabel(...)
```

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2047,6 +2047,70 @@ func TestUserSignupWithExistingMURDifferentUserIDOK(t *testing.T) {
...
+	murtest.AssertThatMasterUserRecords(t, r.Client).HaveCount(1)
+
+	key := types.NamespacedName{
+		Namespace: test.HostOperatorNs,
+		Name:      userSignup.Name,
+	}
+	instance := &toolchainv1alpha1.UserSignup{}
+	err = r.Client.Get(context.TODO(), key, instance)
+	require.NoError(t, err)
+	assert.Equal(t, "approved", instance.Labels[toolchainv1alpha1.UserSignupStateLabelKey])
```
MatousJobanek, line 2089: this seems to be a duplication of the previous lines

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2047,6 +2047,70 @@ func TestUserSignupWithExistingMURDifferentUserIDOK(t *testing.T) {
...
+
+	mur := toolchainv1alpha1.MasterUserRecord{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, &mur)
+	require.NoError(t, err)
```
MatousJobanek, line 2093: you could simplify this whole part, no need to go through reconciles, look at the `TestMigrateMur` test how it is done there, or just move it there

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2047,6 +2047,70 @@ func TestUserSignupWithExistingMURDifferentUserIDOK(t *testing.T) {
...
+	res, err = r.Reconcile(context.TODO(), req)
+	require.NoError(t, err)
+	require.Equal(t, reconcile.Result{}, res)
+
+	// Reload the MUR
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, &mur)
+	require.NoError(t, err)
+
+	// Confirm the propagated claim has been updated in the MUR
+	require.Equal(t, "abc@def.com", mur.Spec.PropagatedClaims.Email)
```
MatousJobanek, line 2111: would be nice to not only verify the modified field, but also that all current MURs that don't have anything set in the propagatedClaims are properly migrated - the fields are populated

controllers/masteruserrecord/masteruserrecord_controller_test.go
MatousJobanek: again, the same as for the `masteruserrecord.go` and `masteruserrecord_test.go`. 
you modified `sync.go` and there is `sync_test.go` file with unit tests covering the functions from the `sync.go` - they should be updated as well.

PR #881, opened 2023-10-06
Do not try to delete UserSignups multiple times

alexeykazakov:
It turned out that our UserSignup cleanup controller tries to delete the same UserSignup twice. As the result we increment our abandoned signup count twice per each deleted signup. I guess it's because we do not check if the resource is already being deleted and try to delete it again.

### DIFF
--- a/controllers/usersignupcleanup/usersignup_cleanup_controller_test.go
+++ b/controllers/usersignupcleanup/usersignup_cleanup_controller_test.go
@@ -119,6 +119,10 @@ func TestUserCleanup(t *testing.T) {
 		statusErr := &apierrors.StatusError{}
 		require.True(t, errors.As(err, &statusErr))
 		require.Equal(t, fmt.Sprintf("usersignups.toolchain.dev.openshift.com \"%s\" not found", key.Name), statusErr.Error())
+
+		t.Run("deletion is not initiated twice", func(t *testing.T) {
+			alreadyDeletedSignupIgnored(t, userSignup)
+		})
 	})
 
 	t.Run("without phone verification initiated", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


controllers/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -251,7 +263,22 @@ func TestUserCleanup(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, deleted)
 	})
+}
 
+func alreadyDeletedSignupIgnored(t *testing.T, userSignup *toolchainv1alpha1.UserSignup) {
+	// Now let's simulate the situation when the signup is already being deleted and reconcile it again.
```
xcoulon, line 269: could we also check that the Prom metric is not incremented again?

alexeykazakov: Yep. Missed that. Fixed here: https://github.com/codeready-toolchain/host-operator/pull/881/commits/672d77306b84faec9201d088c45881b59410bd9b
Thanks.

mfrancisc (approved):
Good catch 👍 

Thanks for fixing the issue.

controllers/usersignupcleanup/usersignup_cleanup_controller_test.go
xcoulon: thanks!

xcoulon (approved):

controllers/usersignupcleanup/usersignup_cleanup_controller_test.go
```diff
@@ -251,7 +263,26 @@ func TestUserCleanup(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, deleted)
 	})
+}
+
+func alreadyDeletedSignupIgnored(t *testing.T, userSignup *toolchainv1alpha1.UserSignup) {
+	// Now let's simulate the situation when the signup is already being deleted and reconcile it again.
+	nw := corev1.Now()
+	userSignup.DeletionTimestamp = &nw
+	r, req, _ := prepareReconcile(t, userSignup.Name, userSignup)
```
MatousJobanek, line 272: btw, it's kind of weird to call `prepareReconcile` here again since it resets the metrics and all the values that were set before in the previous tests, but I understand that for the purpose of the test it shouldn't matter. It's just not clear why the values are 0 when some of them were 1 before that

rajivnathan (approved):

PR #882, opened 2023-10-09
update notification just containing cluster domain name

Kartikey-star:
While preparing the toolchain notification,we should use  only the cluster domain not containing route prefix of the proxy.
For example current notification is `ToolchainStatus has been in an unready status for an extended period for api-toolchain-host-operator.apps.stone-stg-host.qc0p.p1.openshiftapps.com`.

Can be optimized to `ToolchainStatus has been in an unready status for an extended period for stone-stg-host.qc0p.p1.openshiftapps.com`

### DIFF
--- a/controllers/toolchainstatus/toolchainstatus_controller_test.go
+++ b/controllers/toolchainstatus/toolchainstatus_controller_test.go
@@ -53,6 +53,7 @@ const (
 	defaultRegistrationServiceName    = "registration-service"
 	unreadyStatusNotification         = "toolchainstatus-unready"
 	restoredStatusNotification        = "toolchainstatus-restore"
+	hostProxyUrl                      = "https://api-toolchain-host-operator.apps.host-cluster"
 )
 
 type fakeHTTPClient struct {
@@ -203,7 +204,7 @@ func TestToolchainStatusConditions(t *testing.T) {
 				conditionReady(toolchainv1alpha1.ToolchainStatusRegServiceReadyReason),
 				conditionReady(toolchainv1alpha1.ToolchainStatusDeploymentUpToDateReason),
 			)). // also regservice is not up-to-date since we return the same mocked github commit
-			HasHostRoutesStatus("https://api-toolchain-host-operator.host-cluster", hostRoutesAvailable())
+			HasHostRoutesStatus(hostProxyUrl, hostRoutesAvailable())
 	})
 
 	t.Run("HostOperator tests", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ranakan19 (approved):
Looks good!

alexeykazakov (approved):

controllers/toolchainstatus/toolchainstatus_controller_test.go
```diff
@@ -656,7 +657,7 @@ func TestToolchainStatusConditions(t *testing.T) {
 				HasHostOperatorStatus(hostOperatorStatusReady()).
 				HasMemberClusterStatus(memberCluster("member-1", ready()), memberCluster("member-2", ready())).
 				HasRegistrationServiceStatus(registrationServiceReady(conditionReady(toolchainv1alpha1.ToolchainStatusRegServiceReadyReason), conditionReadyWithMessage(toolchainv1alpha1.ToolchainStatusDeploymentRevisionCheckDisabledReason, "access token key is not provided"))).
-				HasHostRoutesStatus("http://api-toolchain-host-operator.host-cluster/api", hostRoutesAvailable())
+				HasHostRoutesStatus("http://api-toolchain-host-operator.apps.host-cluster/api", hostRoutesAvailable())
```
mfrancisc, line 660: just curious , is there a reason for not using `hostProxyURL` here? 
```suggestion
				HasHostRoutesStatus(hostProxyURL, hostRoutesAvailable())
```

mfrancisc (approved):
Looks good 👍 

controllers/toolchainstatus/toolchainstatus_controller_test.go
Kartikey-star: We are specifying the route spec. at L645, so that changes the host Proxy URL to be appended with api. Hence I kept it the way it was

Kartikey-star:
is there something I need to do so as to pass `codecov/changes` failure?

controllers/toolchainstatus/toolchainstatus_controller_test.go
mfrancisc: Oh my bad, I missed that is expecting `http` (and not `https`) and the `/api` endpoint. Sorry for that 👍

mfrancisc:
> is there something I need to do so as to pass codecov/changes failure?

I don't think there is anything to do about it. Not sure what "1 file has unexpected changes not visible in diff" means.

rajivnathan (approved):

PR #890, opened 2023-10-17
rename metrics label

ranakan19:
having same labels for the 2 services may confuse Prometheus, thus renaming the label for metrics Service.
Also naming the port 8082 for metrics.

related PR: https://github.com/codeready-toolchain/registration-service/pull/363

### DIFF
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -226,7 +226,7 @@ objects:
         run: proxy-metrics
     spec:
       ports:
-        - name: "8082"
+        - name: proxy-metrics
           protocol: TCP
           port: 80
           targetPort: metrics
--- a/deploy/registration-service/registration-service.yaml
+++ b/deploy/registration-service/registration-service.yaml
@@ -99,6 +99,7 @@ objects:
                 - containerPort: 8080
                 - containerPort: 8081
                 - containerPort: 8082
+                  name: metrics
               command:
                 - registration-service
               imagePullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


deploy/registration-service/registration-service.yaml
```diff
@@ -222,13 +223,13 @@ objects:
       namespace: ${NAMESPACE}
       labels:
         provider: codeready-toolchain
-        run: registration-service
+        run: proxy-metrics
     spec:
       ports:
         - name: "8082"
```
xcoulon, line 229:
```suggestion
        - name: metrics
```
we can name the port on the service with `metrics`, too

alexeykazakov (approved):

deploy/registration-service/registration-service.yaml
ranakan19: wouldn't it have to be "metrics" then?
But I like having the name of the port and the target-port different, less confusing imo. Like in this example - https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports - I wouldn't have understood this example if the service port name and target port name were the same. 
Not saying its the right thing to do, but its just me and my brain.

ranakan19: done in [4367241](https://github.com/codeready-toolchain/host-operator/pull/890/commits/4367241f5db752741c30ff757e0088549be0cffd)

ranakan19: done in [4367241](https://github.com/codeready-toolchain/host-operator/pull/890/commits/4367241f5db752741c30ff757e0088549be0cffd)

xcoulon (approved):

PR #891, opened 2023-10-17
impl new subspace naming strategy

sadlerap:
This should prevent subspace names from growing longer from the level of subspace nesting.

### DIFF
--- a/controllers/spacerequest/spacerequest_controller.go
+++ b/controllers/spacerequest/spacerequest_controller.go
@@ -192,7 +192,8 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 		return nil, false, errs.Wrap(err, "failed to retrieve spaces")
 	}
 
-	if len(subSpaceList.Items) == 0 {
+	length := len(subSpaceList.Items)
+	if length == 0 {
 		// no spaces found, let's create it
 		logger.Info("creating subSpace")
 		if err := r.setStatusProvisioning(ctx, memberCluster, spaceRequest); err != nil {
@@ -204,7 +205,7 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 			return nil, false, r.setStatusFailedToCreateSubSpace(ctx, memberCluster, spaceRequest, err)
 		}
 		return subSpace, true, nil // a subSpace was created
-	} else if len(subSpaceList.Items) > 1 {
+	} else if length > 1 {
 		err := fmt.Errorf("Found %v matching subspaces, expected 1", len(subSpaceList.Items))
 		logger.Error(err, "Too many matching subspaces")
 		return nil, false, err
--- a/controllers/spacerequest/spacerequest_controller.go
+++ b/controllers/spacerequest/spacerequest_controller.go
@@ -205,7 +205,7 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 		}
 		return subSpace, true, nil // a subSpace was created
 	} else if len(subSpaceList.Items) > 1 {
-		err := fmt.Errorf("Found %v matching subspaces", len(subSpaceList.Items))
+		err := fmt.Errorf("Found %v matching subspaces, expected 1", len(subSpaceList.Items))
 		logger.Error(err, "Too many matching subspaces")
 		return nil, false, err
 	}
@@ -366,7 +366,7 @@ func (r *Reconciler) deleteSubSpace(ctx context.Context, parentSpace *toolchainv
 		toolchainv1alpha1.SpaceRequestNamespaceLabelKey: spaceRequest.GetNamespace(),
 	}); err != nil {
 		// failed to get subSpace
-		return false, err
+		return false, errs.Wrap(err, "failed to list sub spaces")
 	}
 
 	// should be a unique subspace
--- a/controllers/spacerequest/spacerequest_controller.go
+++ b/controllers/spacerequest/spacerequest_controller.go
@@ -347,12 +347,8 @@ func (r *Reconciler) ensureSpaceDeletion(ctx context.Context, memberClusterWithS
 		// finalizer was already removed, nothing to delete anymore...
 		return nil
 	}
-	// find parent space from namespace labels
-	parentSpace, err := r.getParentSpace(ctx, memberClusterWithSpaceRequest, spaceRequest)
-	if err != nil {
-		return err
-	}
-	if isBeingDeleted, err := r.deleteSubSpace(ctx, parentSpace, spaceRequest); err != nil {
+
+	if isBeingDeleted, err := r.deleteSubSpace(ctx, spaceRequest); err != nil {
 		return r.setStatusTerminatingFailed(ctx, memberClusterWithSpaceRequest, spaceRequest, err)
 	} else if isBeingDeleted {
 		if err := r.setStatusTerminating(ctx, memberClusterWithSpaceRequest, spaceRequest); err != nil {
@@ -374,7 +370,7 @@ func (r *Reconciler) ensureSpaceDeletion(ctx context.Context, memberClusterWithS
 // returns true/nil if the deletion of the subSpace was triggered
 // returns false/nil if the subSpace was already deleted
 // return false/err if something went wrong
-func (r *Reconciler) deleteSubSpace(ctx context.Context, parentSpace *toolchainv1alpha1.Space, spaceRequest *toolchainv1alpha1.SpaceRequest) (bool, error) {
+func (r *Reconciler) deleteSubSpace(ctx context.Context, spaceRequest *toolchainv1alpha1.SpaceRequest) (bool, error) {
 	subSpace, err := r.getSubSpace(ctx, spaceRequest)
 	if err != nil {
 		return false, err
--- a/controllers/spacerequest/spacerequest_controller.go
+++ b/controllers/spacerequest/spacerequest_controller.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -161,6 +160,33 @@ func (r *Reconciler) addFinalizer(ctx context.Context, memberCluster cluster.Clu
 	return nil
 }
 
+// fetches the subspace corresponding to the space request given.
+//
+// returns:
+//   - nil, err if an error occurred retrieving spaces or more than one matching subspace was found
+//   - nil, nil if no subspaces exist
+//   - space, nil if a matching space was found
+func (r *Reconciler) getSubSpace(ctx context.Context, spaceRequest *toolchainv1alpha1.SpaceRequest) (*toolchainv1alpha1.Space, error) {
+	subSpaceList := &toolchainv1alpha1.SpaceList{}
+	if err := r.Client.List(ctx, subSpaceList,
+		runtimeclient.InNamespace(r.Namespace),
+		runtimeclient.MatchingLabels{
+			toolchainv1alpha1.SpaceRequestLabelKey:          spaceRequest.GetName(),
+			toolchainv1alpha1.SpaceRequestNamespaceLabelKey: spaceRequest.GetNamespace(),
+		}); err != nil {
+		return nil, errs.Wrap(err, "failed to list subspaces")
+	}
+
+	length := len(subSpaceList.Items)
+	if length >= 2 {
+		return nil, errs.Errorf("Expected 1 matching subspace for spaceRequest %v in namespace %v, found %d", spaceRequest.GetName(), spaceRequest.GetNamespace(), length)
+	} else if length == 0 {
+		return nil, nil
+	}
+
+	return &subSpaceList.Items[0], nil
+}
+
 func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Cluster, spaceRequest *toolchainv1alpha1.SpaceRequest) (*toolchainv1alpha1.Space, bool, error) {
 	logger := log.FromContext(ctx)
 	logger.Info("ensuring subSpace")
@@ -181,19 +207,12 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 	}
 
 	// create if not found on the expected target cluster
-	subSpaceList := &toolchainv1alpha1.SpaceList{}
-	if err := r.Client.List(ctx, subSpaceList,
-		client.InNamespace(r.Namespace),
-		client.MatchingLabels{
-			toolchainv1alpha1.SpaceRequestLabelKey:          spaceRequest.GetName(),
-			toolchainv1alpha1.SpaceRequestNamespaceLabelKey: spaceRequest.GetNamespace(),
-		}); err != nil {
-		// failed to list subSpaces
-		return nil, false, errs.Wrap(err, "failed to retrieve spaces")
+	subSpace, err := r.getSubSpace(ctx, spaceRequest)
+	if err != nil {
+		return nil, false, err
 	}
 
-	length := len(subSpaceList.Items)
-	if length == 0 {
+	if subSpace == nil {
 		// no spaces found, let's create it
 		logger.Info("creating subSpace")
 		if err := r.setStatusProvisioning(ctx, memberCluster, spaceRequest); err != nil {
@@ -205,13 +224,8 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 			return nil, false, r.setStatusFailedToCreateSubSpace(ctx, memberCluster, spaceRequest, err)
 		}
 		return subSpace, true, nil // a subSpace was created
-	} else if length > 1 {
-		err := fmt.Errorf("Found %v matching subspaces, expected 1", len(subSpaceList.Items))
-		logger.Error(err, "Too many matching subspaces")
-		return nil, false, err
 	}
 
-	subSpace := &subSpaceList.Items[0]
 	logger.Info("subSpace already exists")
 	updated, err := r.updateExistingSubSpace(ctx, spaceRequest, subSpace)
 	return subSpace, updated, err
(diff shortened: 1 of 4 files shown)


pkg/space/space.go
```diff
@@ -59,5 +61,14 @@ func NewSubSpace(spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *tool
 
 // SubSpaceName generates a name for a subSpace based on parentSpace name and spacerequest UID.
 func SubSpaceName(parentSpace *toolchainv1alpha1.Space, spacerequest *toolchainv1alpha1.SpaceRequest) string {
-	return parentSpace.GetName() + string(spacerequest.UID[:5]) // we only get the first 5 chars from UID so we can keep the name within the length limits
+	parentSpaceName := parentSpace.GetName()
+
+	// if the parent space is itself a subspace, then we need to strip its
+	// identifier out to prevent length limitations from kicking in
+	if parentSpace.Spec.ParentSpace != "" {
```
alexeykazakov, line 68: Let's also check that the parent space name is actually longer than 6 characters. Just to be 100% on the safe side.

pkg/space/space.go
```diff
@@ -59,5 +61,14 @@ func NewSubSpace(spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *tool
...
+		// take of 6 characters to include the dash
+		parentSpaceName = string(parentSpaceName[:len(parentSpaceName)-6])
+	}
+
+	return fmt.Sprintf("%v-%v", parentSpaceName, string(spacerequest.UID[:5]))
```
alexeykazakov, line 73:
```suggestion
	return fmt.Sprintf("%v-%v", parentSpaceName, string(spacerequest.UID[:5])) // we only get the first 5 chars from UID so we can keep the name within the length limits
```

alexeykazakov (commented):
Looks good! Just a couple of minor comments.

pkg/space/space.go
sadlerap: Fixed in 46b3f1e.

pkg/space/space.go
sadlerap: Fixed in 46b3f1e.

sadlerap:
Hmmm, looks like uploading coverage info to codecov.io is [failing](https://github.com/codeready-toolchain/host-operator/actions/runs/6550083733/job/17788529292).  Not sure what's going on, since it looks like we're getting a 404, but [their status page](https://status.codecov.com/) is all green.

alexeykazakov:
Codecove upload fails pretty often unfortunately. Re-running the job usually helps.

pkg/space/space.go
```diff
@@ -59,5 +61,15 @@ func NewSubSpace(spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *tool
 
 // SubSpaceName generates a name for a subSpace based on parentSpace name and spacerequest UID.
 func SubSpaceName(parentSpace *toolchainv1alpha1.Space, spacerequest *toolchainv1alpha1.SpaceRequest) string {
-	return parentSpace.GetName() + string(spacerequest.UID[:5]) // we only get the first 5 chars from UID so we can keep the name within the length limits
+	parentSpaceName := parentSpace.GetName()
+
+	// if the parent space is itself a subspace, then we need to strip its
+	// identifier out to prevent length limitations from kicking in
+	if parentSpace.Spec.ParentSpace != "" && len(parentSpaceName) > 6 {
+		// take of 6 characters to include the dash
```
mfrancisc, line 69: minor thing - maybe adding a small example to make it clear to the reader that we are removing the UID suffix from the parent space name before using it for the sub space name creation. WDYT ? 
```suggestion
		// take of 6 characters to include the dash
		// for example if the parentSpaceName is "parentSpace-01234" then we keep only "parenSpace", removing the 5char UID suffix from the end. 
```

mfrancisc (approved):
Nice 👍 . Thanks for improving this 🙏 

pkg/space/space.go
sadlerap: sounds good to me - I've adjusted some of the comments with a7f983f, hopefully things are a bit clearer.

mfrancisc: looks good ! thanks !

pkg/space/space.go
```diff
@@ -59,5 +61,18 @@ func NewSubSpace(spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *tool
...
+	if parentSpace.Spec.ParentSpace != "" && len(parentSpaceName) > 6 {
+		// take off 6 characters to include the dash.  for example,
+		// "parentspace-12345" becomes "parentspace" when we strip the
+		// 5-character UID suffix and the "-"
+		parentSpaceName = string(parentSpaceName[:len(parentSpaceName)-6])
+	}
+
+	// only get the first 5 chars from the spacerequest's UID so we can keep
+	// the name within length limits.
+	return fmt.Sprintf("%v-%v", parentSpaceName, string(spacerequest.UID[:5]))
```
MatousJobanek, line 77: how do you want to ensure that the already existing sub-spaces will be re-used after this change is merged?
1. there is a SpaceRequest CR in RHTAP instance
2. for the SpaceRequest CR, there is a sub-space `johny123` created
3. you merge the PR
4. the reconciler is kicked and it goes to this section https://github.com/codeready-toolchain/host-operator/blob/932b437277941dce8cc27b1e319931fd980d3cb0/controllers/spacerequest/spacerequest_controller.go#L183-L187
5. it uses this new format and expects `johny-123`, it doesn't find any, so it creates a new sub-space

the same would happen for the spaces whose parent has a name longer than 6

pkg/space/space.go
```diff
@@ -59,5 +61,18 @@ func NewSubSpace(spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *tool
 
 // SubSpaceName generates a name for a subSpace based on parentSpace name and spacerequest UID.
 func SubSpaceName(parentSpace *toolchainv1alpha1.Space, spacerequest *toolchainv1alpha1.SpaceRequest) string {
-	return parentSpace.GetName() + string(spacerequest.UID[:5]) // we only get the first 5 chars from UID so we can keep the name within the length limits
+	parentSpaceName := parentSpace.GetName()
+
+	// if the parent space is itself a subspace, then we need to strip its UID
+	// suffix to prevent length limitations from kicking in
+	if parentSpace.Spec.ParentSpace != "" && len(parentSpaceName) > 6 {
```
MatousJobanek, line 68: out of curiosity, why did you decide to use only 6 first characters?

MatousJobanek (changes requested):
Hi, thanks a lot for your PR. :+1: :1st_place_medal: 
I think that the code doesn't cover the case with already existing sub-spaces. Could you please change that and also cover such use-case in unit-tests as well as in e2e tests?
Let me know if I missed some important detail .

pkg/space/space_test.go
```diff
@@ -53,3 +53,30 @@ func TestNewSubSpace(t *testing.T) {
...
+	// given
+	srClusterRoles := []string{commoncluster.RoleLabel(commoncluster.Tenant)}
+	sr := spacerequesttest.NewSpaceRequest("jane", "jane-tenant",
+		spacerequesttest.WithTierName("appstudio"),
+		spacerequesttest.WithTargetClusterRoles(srClusterRoles))
+	parentSpace := spacetest.NewSpace(test.HostOperatorNs, "parentSpace")
+	subSpace := NewSubSpace(sr, parentSpace)
+
+	// when
+	subSubSpace := NewSubSpace(sr, subSpace)
```
MatousJobanek, line 67: Let's not use the same SpaceRequest for both SubSpaces to make sure that the subSubSpace doesn't have the same name as its parent (subSpace)

pkg/space/space_test.go
```diff
@@ -53,3 +53,30 @@ func TestNewSubSpace(t *testing.T) {
...
+
+	// then
+	expectedSubSubSpace := spacetest.NewSpace(test.HostOperatorNs, SubSpaceName(subSpace, sr),
```
MatousJobanek, line 70: based on the changes from the previous comment, let's verify that it's using UID from the right SpaceRequest
```suggestion
	expectedSubSubSpace := spacetest.NewSpace(test.HostOperatorNs, fmt.Sprintf("parentspace-%s", sr.UID[:5]),
```

pkg/space/space_test.go
```diff
@@ -53,3 +53,30 @@ func TestNewSubSpace(t *testing.T) {
...
+		spacetest.WithTierName("appstudio"),
+		spacetest.WithSpecTargetClusterRoles([]string{cluster.RoleLabel(cluster.Tenant)}),
+		spacetest.WithLabel(toolchainv1alpha1.SpaceRequestLabelKey, sr.GetName()),
+		spacetest.WithLabel(toolchainv1alpha1.SpaceRequestNamespaceLabelKey, sr.GetNamespace()),
+		spacetest.WithLabel(toolchainv1alpha1.ParentSpaceLabelKey, subSpace.GetName()),
+	)
+	assert.Equal(t, expectedSubSubSpace, subSubSpace)
+
+	// also assert that names don't grow in length as we increase nesting
+	assert.Equal(t, len(subSpace.Name), len(subSubSpace.Name))
```
MatousJobanek, line 81: and also that the names are different 
```suggestion
	assert.Equal(t, len(subSpace.Name), len(subSubSpace.Name))
	assert.NotEqual(t, subSpace.Name, subSubSpace.Name)
```

pkg/space/space.go
mfrancisc: good point ! But do we have any existent subSubSpace ? 

I mean the `parentSpaceName` will be shortened only if it's a subSpace it self, not only if the name is longer than 6:
`if parentSpace.Spec.ParentSpace != "" && len(parentSpaceName) > 6`

I thought right now we are using this feature only for 1 level of inheritance : `parentSpace --> subSpace` , and not yet `parentSpace --> subSpace --> subSubSpace` but maybe I'm wrong.

sadlerap: In this particular case, could we list with a label selector for the correct parent space?
```go
subSpaceList := &toolchainv1alpha1.SpaceList{}
err := r.Client.List(ctx, subSpaceList,
    client.InNamespace(r.Namespace),
    client.MatchingLabels{
        toolchainv1alpha1.SpaceRequestLabelKey:          spaceRequest.GetName(),
        toolchainv1alpha1.SpaceRequestNamespaceLabelKey: spaceRequest.GetNamespace(),
        toolchainv1alpha1.ParentSpaceLabelKey:           parentSpace.GetName(),
    })
```

We'd have to handle the case where we find more than one matching space, but I can add a test for that too.

pkg/space/space_test.go
sadlerap: Fixed in 35eb34a.

pkg/space/space_test.go
sadlerap: Fixed in 35eb34a.

pkg/space/space_test.go
sadlerap: Fixed in 35eb34a.

sadlerap:
I'll look into writing some e2e tests for this.  If we have a subspace, we can't guarantee that it follows the new naming convention established here, so I imagine most places where we try to get a subspace by its name will need to turn into a list call with a label selector for its parent.

pkg/space/space.go
MatousJobanek: > good point ! But do we have any existent subSubSpace ?

not of subSubSpace, but of subSpace yes. There is one more change in the logic of generating the name apart from the `if` statement that you mentioned above - originally, we didn't add the name between the prefix (parent name) and suffix (UID of the SpaceRequest CR). so the result was for example `parentspace1a2b3cd`. For better readability, we would like to add a hyphen between the prefix and suffix so its' `parentspace-1a2b3cd`, which however brings the challenge of keeping the backward compatibility.

The list for checking the existence should work. We had it there even before but we used it in a combination with generating the name nondeterministically. We modified it to the `Get` (as it is now) because we saw that the informer cache of the client wasn't sometimes updated quickly enough so the `List` returned an empty list even when one Space was already created. However, since we didn't use deterministic approach to generating the names, the we sometimes ended up with multiple Spaces per SpaceRequest CR.
Using the `List` should be fine, because it uses a fixed format for the names, so if the cache is not updated with the latest Space CR, then the next attempt to create the Space CR should fail with conflict, which totally OK. As a benefit, we don't have to rely on the naming format, but just on the labels set at Space CR.

MatousJobanek:
As for the e2e tests, take a look at the migration tests https://github.com/codeready-toolchain/toolchain-e2e/tree/master/test/migration We don't have a case for sub-spaces there yet, so if you add one, then it should cover exactly what we need. Just follow the format that is used there.

pkg/space/space.go
mfrancisc: > There is one more change in the logic

My bad ! I missed that we are also adding the `-` between parentName and suffix 🤦‍♂️ . I don't remember why we haven't added that in the first place.

And indeed, if we want to change the naming convention of subSpaces as well, we could use `List` by labels while keeping the name deterministic.

sadlerap: I've replaced most of the direct fetches with list calls in 563be3d (and added some tests in ff9eec2).  As far as I can tell, the only place that uses `SubSpaceName` anymore is some tests, and I'm half-tempted to refactor those tests to make those calls indirect, though I'll probably punt that off for a future PR if I do.

pkg/space/space_test.go
```diff
@@ -53,3 +54,34 @@ func TestNewSubSpace(t *testing.T) {
...
+	assert.NotEqual(t, subSpace.Name, subSubSpace.Name)
+}
```
mfrancisc, line 87: We should rename sr and expect that as name and namespace in the labels:
```suggestion
func TestNewSubSubSpace(t *testing.T) {
	// given
	srClusterRoles := []string{commoncluster.RoleLabel(commoncluster.Tenant)}
	sr := spacerequesttest.NewSpaceRequest("jane", "jane-tenant",
		spacerequesttest.WithTierName("appstudio"),
		spacerequesttest.WithTargetClusterRoles(srClusterRoles))
	parentSpace := spacetest.NewSpace(test.HostOperatorNs, "parentSpace")
	subSpace := NewSubSpace(sr, parentSpace)
	// create another spacerequest in the subSpace namespace
	sr2 := spacerequesttest.NewSpaceRequest("jane2", subSpace.GetName()+"-tenant",
		spacerequesttest.WithTierName("appstudio"),
		spacerequesttest.WithTargetClusterRoles(srClusterRoles))

	// when
	subSubSpace := NewSubSpace(sr2, subSpace)

	// then
	expectedSubSubSpace := spacetest.NewSpace(test.HostOperatorNs, fmt.Sprintf("%s-%s", parentSpace.Name, sr2.UID[:5]),
		spacetest.WithSpecParentSpace(subSpace.GetName()),
		spacetest.WithTierName("appstudio"),
		spacetest.WithSpecTargetClusterRoles([]string{cluster.RoleLabel(cluster.Tenant)}),
		spacetest.WithLabel(toolchainv1alpha1.SpaceRequestLabelKey, sr2.GetName()),
		spacetest.WithLabel(toolchainv1alpha1.SpaceRequestNamespaceLabelKey, sr2.GetNamespace()),
		spacetest.WithLabel(toolchainv1alpha1.ParentSpaceLabelKey, subSpace.GetName()),
	)
	assert.Equal(t, expectedSubSubSpace, subSubSpace)

	// also assert that names don't grow in length as we increase nesting
	assert.Equal(t, len(subSpace.Name), len(subSubSpace.Name))
	assert.NotEqual(t, subSpace.Name, subSubSpace.Name)
}
```

mfrancisc (commented):
few minor things

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -180,27 +181,37 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
...
+		subSpace, err := r.createNewSubSpace(ctx, spaceRequest, parentSpace)
+		if err != nil {
+			// failed to create subSpace
+			return nil, false, r.setStatusFailedToCreateSubSpace(ctx, memberCluster, spaceRequest, err)
+		}
+		return subSpace, true, nil // a subSpace was created
+	} else if length > 1 {
+		err := fmt.Errorf("Found %v matching subspaces, expected 1", len(subSpaceList.Items))
+		logger.Error(err, "Too many matching subspaces")
+		return nil, false, err
```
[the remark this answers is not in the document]
sadlerap: I like it - I'll see if I can come up with something more satisfying.

pkg/space/space_test.go
sadlerap: Fixed in bec33a7.

controllers/spacerequest/spacerequest_controller_test.go
```diff
@@ -1039,6 +1101,48 @@ func TestDeleteSpaceRequest(t *testing.T) {
...
+			hostClient := test.NewFakeClient(t, appstudioTier, parentSpace, subSpace1, subSpace2)
+			ctrl := newReconciler(t, hostClient, member1)
+
+			// when
+			_, err := ctrl.Reconcile(context.TODO(), requestFor(sr))
+
+			// then
+			require.EqualError(t, err, "too many subspaces")
+		})
+
```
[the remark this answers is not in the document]
sadlerap: I took this test out in eeb9e65, since it wasn't doing anything for coverage.

controllers/spacerequest/spacerequest_controller.go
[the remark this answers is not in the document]
sadlerap: I've adjusted this in eeb9e65.

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -23,6 +23,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
sadlerap: Fixed in eeb9e65.

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -160,6 +160,33 @@ func (r *Reconciler) addFinalizer(ctx context.Context, memberCluster cluster.Clu
 	return nil
 }
 
+// fetches the subspace corresponding to the space request given.
+//
+// returns:
+//   - nil, err if an error occurred retrieving spaces or more than one matching subspace was found
+//   - nil, nil if no subspaces exist
+//   - space, nil if a matching space was found
+func (r *Reconciler) getSubSpace(ctx context.Context, spaceRequest *toolchainv1alpha1.SpaceRequest) (*toolchainv1alpha1.Space, error) {
```
mfrancisc, line 169: minor thing feel free to leave it as is - but how about `listSubSpace`? Since it's actually listing and not getting.

mfrancisc (approved):
Nice job 👍 

sadlerap:
/hold

I'd like https://github.com/codeready-toolchain/toolchain-e2e/pull/826 to merge before we merge this - the tests there make sure that this PR doesn't break anything.

Local testing seems to indicate things work, but I'll have higher confidence once it lands.

MatousJobanek (approved):
Looks good :+1: :rocket: Thanks a lot for the additional changes :1st_place_medal: :pray: 
I have just one comment regarding one unit test

MatousJobanek:
> I'd like https://github.com/codeready-toolchain/toolchain-e2e/pull/826 to merge before we merge this - the tests there make sure that this PR doesn't break anything.
Local testing seems to indicate things work, but I'll have higher confidence once it lands.

It's always the best practice to create both PRs from branches that have the same name so our e2e logic can pair them and run e2e tests from the e2e PR (instead of master). In this way, you can verify that the changes in e2e tests work as expected with the changes from the operator.

controllers/spacerequest/spacerequest_controller_test.go
```diff
@@ -441,6 +441,25 @@ func TestCreateSpaceRequest(t *testing.T) {
...
+			hostClient := test.NewFakeClient(t, appstudioTier, parentSpace)
+			ctrl := newReconciler(t, hostClient, member1)
+
+			// when
+			_, err := ctrl.Reconcile(context.TODO(), requestFor(sr))
+
+			// then
+			require.EqualError(t, err, "error updating status: mock error")
+		})
+
```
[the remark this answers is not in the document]
sadlerap: This test asserts that errors are propogated when an error occurs during status updates.  I added it to give full coverage to the `ensureSpace` method, which got refactored.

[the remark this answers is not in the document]
sadlerap: That said, the name of this test is certainly wrong, which I've fixed in 50e206a.

sadlerap:
> It's always the best practice to create both PRs from branches that have the same name so our e2e logic can pair them and run e2e tests from the e2e PR (instead of master). In this way, you can verify that the changes in e2e tests work as expected with the changes from the operator and you can merge both PRs together.

I wasn't aware there was automation set up for this case.  In the future, I'll do this, but I'll leave the currently-open PRs for now.

Is there any kind of contribution guide where this is documented?

PR #898, opened 2023-10-19
fix: add check for spacebinding existence in SBR controller

mfrancisc:
SpaceBindingRequest controller - add check for SpaceBinding existence for a given MUR and Space so that the error can be reflected in the SpaceBindingRequest conditions.

Jira: https://issues.redhat.com/browse/ASC-430


alexeykazakov (approved):

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -280,18 +280,29 @@ func (r *Reconciler) updateSpaceBinding(logger logr.Logger, spaceBinding *toolch
...
-		return err
-	}
-	if err := r.Client.Create(context.TODO(), spaceBinding); err != nil {
-		return errs.Wrap(err, "unable to create SpaceBinding")
+	// check if there is already a SpaceBinding created with same name (there could be already a SpaceBinding for the same space and MUR generated by system admins for example)
+	existingSpaceBinding := &toolchainv1alpha1.SpaceBinding{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: r.Namespace,
+		Name:      spaceBinding.GetName(),
+	}, existingSpaceBinding)
```
MatousJobanek, line 288: I would rather do the check somewhere else.
Before we go to this function we call
https://github.com/codeready-toolchain/host-operator/blob/b15e44ce05474814ad8388f76d639ee486b53e0e/controllers/spacebindingrequest/spacebindingrequest_controller.go#L99-L102

which Lists all existing SpaceBindings using the SpaceBindingRequest labels. 
https://github.com/codeready-toolchain/host-operator/blob/b15e44ce05474814ad8388f76d639ee486b53e0e/controllers/spacebindingrequest/spacebindingrequest_controller.go#L149-L156

instead of this, we should list them using the Space and MUR labels and then validate:
1. if there is no found, then let's create it
2. if there is one found, then check if it contains the SpaceBindingRequest labels. if not then return an error
3. if there are more than one found (which shouldn't really happen), then fail too (no matter who manages them) that it cannot properly apply the requested permissions because of multiple SBs to be present for the same combination of MUR & Space.

With this, we won't have to rely on the naming format to be the same everywhere (here as well as in sandbox-cli).

WDYT?

mfrancisc: yes that would work too, good point! Let me take a look and update it. 

I went this path because was the one requiring less code , but indeed not relying on the naming convention is a valid point!

mfrancisc: Please take a look at https://github.com/codeready-toolchain/host-operator/pull/898/commits/b53159786a5acc3d3df68ea0cc5eae3bbd1b5f0d when you have some time.

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -161,6 +161,32 @@ func (r *Reconciler) getSpaceBinding(ctx context.Context, spaceBindingRequest *t
...
+			return nil, fmt.Errorf("A SpaceBinding for Space %s and MUR %s already exists. But it doesn't have the expected SpaceBindingRequestNamespace label set: %s", space.GetName(), spaceBindingRequest.Spec.MasterUserRecord, toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey)
+		}
+		// check if spaceBinding has correct reference to spaceBindingRequest
+		// this should never happen, but you never know ¯\_(ツ)_/¯
+		if sbrLabel != spaceBindingRequest.GetName() {
+			return nil, fmt.Errorf("expected SpaceBindingRequest label: %s, got: %s", spaceBindingRequest.GetName(), sbrLabel)
+		}
+		if sbrNamespaceLabel != spaceBindingRequest.GetNamespace() {
+			return nil, fmt.Errorf("expected SpaceBindingRequestNamespace label: %s, got: %s", spaceBindingRequest.GetNamespace(), sbrNamespaceLabel)
+		}
```
MatousJobanek, line 187: We don't need to deal with each of the situation separately, so we could simplify this part into:
```suggestion
		sbrLabel := spaceBindings.Items[0].Labels[toolchainv1alpha1.SpaceBindingRequestLabelKey]
		sbrNamespaceLabel := spaceBindings.Items[0].Labels[toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey]
		if sbrLabel != spaceBindingRequest.GetName() || sbrNamespaceLabel != spaceBindingRequest.GetNamespace() {
			return nil, fmt.Errorf("A SpaceBinding for Space '%s' and MUR '%s' already exists, but it's not managed by this SpaceBindingRequest CR. It's not allowed to create multiple SpaceBindings for the same combination of Space and MasterUserRecord", space.GetName(), spaceBindingRequest.Spec.MasterUserRecord)
		}
```

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -209,19 +235,20 @@ func (r *Reconciler) addFinalizer(ctx context.Context, memberCluster cluster.Clu
...
+	spaceBinding, err := r.getSpaceBinding(ctx, spaceBindingRequest, memberCluster)
+	if err != nil {
+		return err
+	}
 
 	// find space from namespace labels
 	space, err := r.getSpace(ctx, memberCluster, spaceBindingRequest)
 	if err != nil {
 		return err
 	}
```
MatousJobanek, line 251: the `getSpace` is called also at the very beginning of the `getSpaceBinding` method. so you could swap it and drop the extra call from `getSpaceBinding`.
```suggestion
	// find space from namespace labels
	space, err := r.getSpace(ctx, memberCluster, spaceBindingRequest)
	if err != nil {
		return err
	}

	// create spacebinding if not found for given spaceBindingRequest
	spaceBinding, err := r.getSpaceBinding(ctx, spaceBindingRequest, space)
	if err != nil {
		return err
	}
```

MatousJobanek (commented):
looks good, just two suggestions for simplifying the code

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: Thanks for you suggestion. I agree , this will make the code shorter and the error message more "end user friendly". But I think it will make a bit more difficult to understand what actually went wrong, is there some mismatch/misconfiguration with labels or is there actually because of another spacebinding that is not managed by the SpaceBindingRequest. But we can find that out by investigating a bit the SpaceBinding and SBR objects in the cluster eventually.

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: Good point!  Thanks!

I've opted for returning also the space object from `getSpaceBinding` in https://github.com/codeready-toolchain/host-operator/pull/898/commits/c848d1e49b3fc3972bface45c27b2ad335ff894b

Mainly because I'm calling `getSpaceBinding` also from `ensureSpaceBindingDeletion` as one of the first things at the beginning of the `reconcile` function , and I don't have the space there yet. I think it's handy to retrieve the space from within  `getSpaceBinding`. The other option would be to call both `getSpace` and `getSpaceBinding` from the `reconcile` and pass the `spacebinding` and `space` objects to `ensureSpaceBindingDeletion` or `ensureSpaceBinding`. But this might require some reshuffle of the tests. 

Please let me know WDYT.

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -450,6 +451,54 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+			sbrForDuplicatedSpaceBinding := spacebindingrequesttest.NewSpaceBindingRequest("jane", "jane-tenant",
+				spacebindingrequesttest.WithMUR(janeMur.Name),
+				spacebindingrequesttest.WithSpaceRole("admin"))
+
+			// given
+			spaceBinding1 := spacebindingcommon.NewSpaceBinding(janeMur, janeSpace, "john") // there is already an admin generated SpaceBinding
+			spaceBinding2 := *spaceBinding1                                                 // there is another spacebinding (this should not happen)
+			spaceBinding2.Name = "somerandom name"                                          // the name doesn't matter since spacebindings are retrieved using lables
+			member1 := NewMemberClusterWithClient(test.NewFakeClient(t, sbrNamespace, sbrForDuplicatedSpaceBinding), "member-1", corev1.ConditionTrue)
+			hostClient := test.NewFakeClient(t, base1nsTier, janeSpace, janeMur, spaceBinding1, &spaceBinding2)
```
MatousJobanek, line 465: Just a detail, in such cases you can use the deepCopy function:
```suggestion
			spaceBinding2 := spaceBinding1.DeepCopy()                                       // there is another spacebinding (this should not happen)
			spaceBinding2.Name = "somerandom name"                                          // the name doesn't matter since spacebindings are retrieved using lables
			member1 := NewMemberClusterWithClient(test.NewFakeClient(t, sbrNamespace, sbrForDuplicatedSpaceBinding), "member-1", corev1.ConditionTrue)
			hostClient := test.NewFakeClient(t, base1nsTier, janeSpace, janeMur, spaceBinding1, spaceBinding2)
```

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -450,6 +451,54 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
...
+		})
+
+		t.Run("SpaceBinding managed by this SpaceBindingRequest CR", func(t *testing.T) {
+			// we have an SBR that will try to create the same SpaceBinding
+			sbrForDuplicatedSpaceBinding := spacebindingrequesttest.NewSpaceBindingRequest("jane", "jane-tenant",
+				spacebindingrequesttest.WithMUR(janeMur.Name),
+				spacebindingrequesttest.WithSpaceRole("admin"))
+
+			// given
+			spaceBinding := spacebindingcommon.NewSpaceBinding(janeMur, janeSpace, "john") // there is already an admin generated SpaceBinding
```
MatousJobanek, line 486: this is better for reading
```suggestion
		t.Run("SpaceBinding not managed by this SpaceBindingRequest CR", func(t *testing.T) {
			// given
			spaceBinding := spacebindingcommon.NewSpaceBinding(janeMur, janeSpace, "john") // there is already an admin generated SpaceBinding
			// this SBR will fail for the conflict with the already existing SpaceBinding
			sbrForDuplicatedSpaceBinding := spacebindingrequesttest.NewSpaceBindingRequest("jane", "jane-tenant",
				spacebindingrequesttest.WithMUR(janeMur.Name),
				spacebindingrequesttest.WithSpaceRole("admin"))

```

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
```diff
@@ -450,6 +451,54 @@ func TestCreateSpaceBindingRequest(t *testing.T) {
 				HasFinalizer()
 		})
 
+		t.Run("SpaceBinding for given MUR and Space already exists ", func(t *testing.T) {
```
MatousJobanek, line 454:
```suggestion
		t.Run("There are more than one SpaceBinding for the given MUR and Space", func(t *testing.T) {
```

MatousJobanek (approved):
Looks good :+1: 
Thanks for the improvement :1st_place_medal:

controllers/spacebindingrequest/spacebindingrequest_controller_test.go
mfrancisc: good point! thanks!

PR #902, opened 2023-10-23
ASC-402: Copy the SA token of sub-spaces only for specific tiers

Kartikey-star:
Acceptance Criteria
The SpaceRequest_controller copies the secret only for specific tiers like appstudio-env one
It is possible to create a sub-space using any other tier like appstudio or base1ns
If the SA token copy part is not supported for the provisioned tier, then the SpaceRequest.Status has corresponding information about it

### DIFF
--- a/controllers/spacerequest/spacerequest_controller.go
+++ b/controllers/spacerequest/spacerequest_controller.go
@@ -111,7 +111,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
 	}
 
-	if tier.Spec.SpaceRequestConfig != nil {
+	if tier.Spec.SpaceRequestConfig != nil && tier.Spec.SpaceRequestConfig.ServiceAccountName != "" {
 		// ensure there is a secret that provides admin access to each provisioned namespaces of the subSpace
 		if err := r.ensureSecretForProvisionedNamespaces(ctx, memberClusterWithSpaceRequest, spaceRequest, subSpace, tier.Spec.SpaceRequestConfig.ServiceAccountName); err != nil {
 			return reconcile.Result{}, r.setStatusFailedToCreateSubSpace(ctx, memberClusterWithSpaceRequest, spaceRequest, err)
@@ -175,7 +175,7 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 		return nil, nil, false, errs.New("parentSpace is being deleted")
 	}
 
-	var tier toolchainv1alpha1.NSTemplateTier
+	var tier *toolchainv1alpha1.NSTemplateTier
 
 	// validate tierName
 	if tier, err = r.validateNSTemplateTier(ctx, spaceRequest.Spec.TierName); err != nil {
@@ -198,14 +198,14 @@ func (r *Reconciler) ensureSpace(ctx context.Context, memberCluster cluster.Clus
 				// failed to create subSpace
 				return nil, nil, false, r.setStatusFailedToCreateSubSpace(ctx, memberCluster, spaceRequest, err)
 			}
-			return subSpace, &tier, true, nil // a subSpace was created
+			return subSpace, tier, true, nil // a subSpace was created
 		}
 		// failed to create subSpace
 		return nil, nil, false, r.setStatusFailedToCreateSubSpace(ctx, memberCluster, spaceRequest, err)
 	}
 	logger.Info("subSpace already exists")
 	updated, err := r.updateExistingSubSpace(ctx, spaceRequest, subSpace)
-	return subSpace, &tier, updated, err
+	return subSpace, tier, updated, err
 }
 
 func (r *Reconciler) createNewSubSpace(ctx context.Context, spaceRequest *toolchainv1alpha1.SpaceRequest, parentSpace *toolchainv1alpha1.Space) (*toolchainv1alpha1.Space, error) {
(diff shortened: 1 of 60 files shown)


deploy/templates/nstemplatetiers/appstudio-env/ns_env.yaml
```diff
@@ -25,7 +25,7 @@ objects:
   kind: ServiceAccount
   metadata:
     namespace: ${SPACE_NAME}-env
-    name: namespace-manager
+    name: ${SERVICE_ACCOUNT_NAME}
```
mfrancisc, line 28: i think it is fine to hard code the name here and in the tier.yaml. Also because unless I'm missing something, you are not populating this variable `SERVICE_ACCOUNT_NAME` right now.

deploy/templates/nstemplatetiers/appstudio-env/tier.yaml
```diff
@@ -20,10 +20,13 @@ objects:
         templateRef: ${MAINTAINER_TEMPL_REF}
       contributor:
         templateRef: ${CONTRIBUTOR_TEMPL_REF}
+    spaceRequestConfig:
+      serviceAccountName: ${SERVICE_ACCOUNT_NAME}
```
mfrancisc, line 24: see comment above, we could hard code the name here I guess.

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -104,18 +104,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	}
 
-	// ensure there is a secret that provides admin access to each provisioned namespaces of the subSpace
-	if err := r.ensureSecretForProvisionedNamespaces(ctx, memberClusterWithSpaceRequest, spaceRequest, subSpace); err != nil {
-		return reconcile.Result{}, r.setStatusFailedToCreateSubSpace(ctx, memberClusterWithSpaceRequest, spaceRequest, err)
+	if tier.Spec.SpaceRequestConfig != nil {
+		// ensure there is a secret that provides admin access to each provisioned namespaces of the subSpace
+		if err := r.ensureSecretForProvisionedNamespaces(ctx, memberClusterWithSpaceRequest, spaceRequest, subSpace, tier.Spec.SpaceRequestConfig.ServiceAccountName); err != nil {
+			return reconcile.Result{}, r.setStatusFailedToCreateSubSpace(ctx, memberClusterWithSpaceRequest, spaceRequest, err)
+		}
```
mfrancisc, line 118: Unless I've missed it, I don't see if we check also that `tier.Spec.SpaceRequestConfig.ServiceAccountName` is not empty. It can be done here or inside `ensureSecretForProvisionedNamespaces`. 

Probably we should also check that the SA with the given name exists.

mfrancisc (commented):
Looks good 👍 

I've added few minor comments. Also could you please add some tests.

controllers/spacerequest/spacerequest_controller.go
Kartikey-star: At Line 114 i am checking if the tier.Spec.SpaceRequestConfig is not nil,is that fine?

Kartikey-star: At L488 we are calling the method to create a token for the service account,( `generateKubeConfig` function CreateTokenRequest called)... So if the service account does not exists this should return error. Do you think that is good enough

mfrancisc: > At L488 we are calling the method to create a token for the service account,( generateKubeConfig function CreateTokenRequest called)...

I think that might be enough, maybe it's not worth trying to get the SA resource. 👍

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -483,11 +485,11 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
 	return nil
 }
 
-func (r *Reconciler) generateKubeConfig(subSpaceTargetCluster cluster.Cluster, namespace string) (*api.Config, error) {
+func (r *Reconciler) generateKubeConfig(subSpaceTargetCluster cluster.Cluster, namespace, serviceAccountName string) (*api.Config, error) {
 	// create a token request for the admin service account
 	token, err := restclient.CreateTokenRequest(subSpaceTargetCluster.RESTClient, types.NamespacedName{
 		Namespace: namespace,
-		Name:      toolchainv1alpha1.AdminServiceAccountName,
+		Name:      serviceAccountName,
```
mfrancisc, line 492: I see there are few other occurrences to be replaced, line 508 and 511. Could you please replace those as well?

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -237,9 +239,9 @@ func (r *Reconciler) updateExistingSubSpace(ctx context.Context, spaceRequest *t
 }
 
 // validateNSTemplateTier checks if the provided tierName in the spaceRequest exists and is valid
-func (r *Reconciler) validateNSTemplateTier(ctx context.Context, tierName string) error {
+func (r *Reconciler) validateNSTemplateTier(ctx context.Context, tierName string) (*toolchainv1alpha1.NSTemplateTier, error) {
 	if tierName == "" {
-		return fmt.Errorf("tierName cannot be blank")
+		return &toolchainv1alpha1.NSTemplateTier{}, fmt.Errorf("tierName cannot be blank")
```
mfrancisc, line 244: minor, but we could just return nil here
```suggestion
		return nil, fmt.Errorf("tierName cannot be blank")
```

controllers/spacerequest/spacerequest_controller_test.go
```diff
@@ -392,6 +392,42 @@ func TestCreateSpaceRequest(t *testing.T) {
 		})
 	})
 
+	appstudioTier.Spec.SpaceRequestConfig.ServiceAccountName = "manager"
+	t.Run("failure service account not present", func(t *testing.T) {
```
mfrancisc, line 396: this is testing the case in which the service account doesn't exist. For this I would rather create a new tier that has a service account which doesn't exist or just configure gock at line 403 to return a different SA name, something like:
```
commontest.SetupGockForServiceAccounts(t, member1.APIEndpoint, types.NamespacedName{
				Name:      "another-sa-name",
				Namespace: "jane-env",
			})
```

Also, unless I'm missing something, I don't see a test that:
- uses a tier which has `.Spec.SpaceRequestConfig` unset
- verifies that spacerequest is provisioned with namespace access and no secretRef
- the subspace is ready with the given tiername from the spacerequest

mfrancisc (commented):
Looks good! 

I have few minor comments.

mfrancisc:
@Kartikey-star since you were assigned to work on new effort, I took the liberty to fetch the changes from your PR here and  open a new PR https://github.com/codeready-toolchain/host-operator/pull/972 which has:
- rebase with master 
- fix the conflicts
- add the remaining changes  

I'll work on doing the same for the [toolchain-e2e tests PR](https://github.com/codeready-toolchain/toolchain-e2e/pull/827) 
I hope you don't mind. Thanks a lot for your help here and great work!

PR #904, opened 2023-10-24
Increase resource quota for appstudilo tier

alexeykazakov:
Just for builds.
Requested by Brain.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/825


amfred:
This is good but I think we need to double the pod count as well, from 150 to 300.

ranakan19 (approved):

PR #905, opened 2023-10-25
Update UserSignup status with HomeSpace

baijum:
Ref. https://issues.redhat.com/browse/SANDBOX-425


alexeykazakov (commented):
Any tests?

baijum:
> Any tests?

I have added a unit test here:
2b5552ab0765a5ce2560f1a924a3779b2f08b6ac

Also, send an e2e test:
https://github.com/codeready-toolchain/toolchain-e2e/pull/829

controllers/usersignup/status_updater.go
```diff
@@ -440,3 +440,8 @@ func (u *StatusUpdater) updateStatusConditions(ctx context.Context, userSignup *
 	}
 	return u.Client.Status().Update(ctx, userSignup)
 }
+
+func (u *StatusUpdater) updateStatusHomeSpace(ctx context.Context, userSignup *toolchainv1alpha1.UserSignup, spaceName string) error {
+	userSignup.Status.HomeSpace = spaceName
+	return u.Client.Status().Update(ctx, userSignup)
```
alexeykazakov, line 446: I'm not sure if it's needed and maybe the client already takes care of it but just to be on the safe side:

```suggestion
	if userSignup.Status.HomeSpace != spaceName {
		userSignup.Status.HomeSpace = spaceName
		return u.Client.Status().Update(ctx, userSignup)
	}
	return nil // Nothing changed
```

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -4669,6 +4669,33 @@ func TestUserReactivatingWhileOldSpaceExists(t *testing.T) {
 	})
 }
 
+func TestHomeSpace(t *testing.T) {
```
alexeykazakov, line 4672: I would expect that there are already existing tests which covers space creation. Can we just update those tests to make sure that the home space is set in the status instead of creating a new test for this case?

alexeykazakov (commented):
Thanks for adding tests. I added a few comments. Could you please also add tests to cover the error handling? See the codecov comments.

controllers/usersignup/usersignup_controller_test.go
baijum: Done. fa1cb589eb086f3e637b32baecd03e16b1231a0c

baijum:
> Could you please also add tests to cover the error handling?

Done. 9941c5687835e9cc651e19af45cb23ed864f54fa

controllers/usersignup/usersignup_controller.go
```diff
@@ -700,6 +700,11 @@ func (r *Reconciler) ensureSpace(
 		return nil, false, err
 	}
 
+	err = r.updateStatusHomeSpace(ctx, userSignup, space.Name)
+	if err != nil {
+		return nil, false, err
+	}
```
MatousJobanek, line 706: two things:
1. if the update of the UserSignup.Status fails for any reason, then the HomeSpace wouldn't be populated (Space already exists)
2. it's related to the previous item, this part of the code won't be executed for the already existing UserSignups (they already have the Space CRs created). We need to update also all existing UserSignups.

MatousJobanek (commented):
Could you please update also e2e tests? Verifying that the HomeSpace field is populated where expected?

controllers/usersignup/status_updater_test.go
```diff
@@ -47,4 +47,17 @@ func TestUpdateStatus(t *testing.T) {
 		require.Error(t, err)
 		assert.Equal(t, "failed to create namespace: oopsy woopsy", err.Error())
 	})
+
+	t.Run("home space status not updated", func(t *testing.T) {
```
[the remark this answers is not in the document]
baijum: Done. 514dad05af8930140851436688da0fe3dae7f1c8

controllers/usersignup/status_updater_test.go
```diff
@@ -47,4 +47,17 @@ func TestUpdateStatus(t *testing.T) {
...
+		// given
+		spaceName := "foo"
+		userSignup.Status.HomeSpace = "foo"
```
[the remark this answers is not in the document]
baijum: Fixed.  514dad05af8930140851436688da0fe3dae7f1c8

controllers/usersignup/usersignup_controller.go
baijum: How to update all existing UserSignups?

MatousJobanek: I think that moving it out of the `ensureSpace` method should do the job

baijum: Done. 9a425156d49334bd94a4da0abbd595bc3aba3a34

controllers/usersignup/usersignup_controller.go
```diff
@@ -361,9 +361,12 @@ func (r *Reconciler) checkIfMurAlreadyExists(
 			if err != nil {
 				return true, r.wrapErrorWithStatusUpdate(ctx, userSignup, r.setStatusFailedToCreateSpace, err, "error creating Space")
 			} else if created {
+				err = r.updateStatusHomeSpace(ctx, userSignup, space.Name)
+				if err != nil {
+					return true, err
+				}
```
MatousJobanek, line 367: the `updateStatusHomeSpace` should be called not only when a space is created, but also when it is already present

controllers/usersignup/usersignup_controller.go
```diff
@@ -357,13 +357,18 @@ func (r *Reconciler) checkIfMurAlreadyExists(
...
+			}
 			// if there was an error or the space was created then return to complete the reconcile, another reconcile will occur when space is created since this controller watches spaces
 			if err != nil {
 				return true, r.wrapErrorWithStatusUpdate(ctx, userSignup, r.setStatusFailedToCreateSpace, err, "error creating Space")
 			} else if created {
 				return true, nil
 			}
```
MatousJobanek, line 371: while this is fine logical-wise, it's pretty hard to read. could you reorganize it to something like this:
```suggestion
			if err != nil {
				return true, r.wrapErrorWithStatusUpdate(ctx, userSignup, r.setStatusFailedToCreateSpace, err, "error creating Space")
			}
			if err := r.updateStatusHomeSpace(ctx, userSignup, space.Name); err != nil {
				return true, err
			}
			// if the space was just created then return to complete the reconcile, another reconcile will occur when space is created since this controller watches spaces
			if created {
				return true, nil
			}
```

MatousJobanek (commented):
Just one formatting comment, otherwise, it looks good :+1:

MatousJobanek (approved):
Cool, thanks a lot @baijum

PR #922, opened 2023-11-07
feat: create temporary spacecebindingrequest migration controller

mfrancisc:
This PR introduces a TEMPORARY controller that will run only on the environments on which the SpaceBindingRequest feature is enabled.

The controller will convert all the SpaceBindings created via sandbox-cli into SpaceBindingRequests. The only spacebindings that will not be converted to SpaceBindingRequests are the ones of the owners of the Spaces.

The owner of the Space is verified by comparing the UserSignup owner reference from the MUR with the one in the `SpaceCreatorLabelKey` label value from the Space.

Once we remove the ability to create spacebindings via sandbox-cli and we've checked that all spacebindings have been converted the controller can be disabled/deleted. 

Jira: https://issues.redhat.com/browse/ASC-432

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/834

### DIFF
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -164,7 +164,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 
 	// if the SBR was found ( was created from the previous reconcile loop), we can now delete the SpaceBinding object
-	if err := r.Client.Delete(ctx, spaceBinding); err != nil {
+	if err := r.Client.Delete(ctx, spaceBinding); err != nil && !errors.IsNotFound(err) {
 		return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
 	}
 
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -109,7 +109,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		logger.Error(errs.New("space has no SpaceCreatorLabelKey set"), "the spacebindings for this space will not be migrated", "space name", space.Name)
 		return ctrl.Result{}, nil
 	}
+
 	usersignup := mur.ObjectMeta.OwnerReferences[0]
+	if usersignup.Kind != "UserSignup" {
+		return ctrl.Result{}, errs.New(fmt.Sprintf("owner reference for mur %s is of kind %s. Expected UserSignup", murName, usersignup.Kind))
+	}
 	if space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey] == usersignup.Name {
 		return reconcile.Result{}, nil
 	}
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -133,10 +133,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      sbrName,
 			Namespace: defaultNamespace,
-			Labels: map[string]string{
-				toolchainv1alpha1.SpaceBindingSpaceLabelKey:            space.GetName(),
-				toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey: mur.GetName(),
-			},
 		},
 		Spec: toolchainv1alpha1.SpaceBindingRequestSpec{
 			MasterUserRecord: mur.GetName(),
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -153,7 +153,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
 		return ctrl.Result{Requeue: true}, nil
 	}
-	// if the SBR was found ( was created from the previous reconcile loop), we can now delete the SpaceBinding object
+	// if the SBR was found (was created from the previous reconcile loop), we can now delete the SpaceBinding object
 	if err := r.Client.Delete(ctx, spaceBinding); err != nil && !errors.IsNotFound(err) {
 		return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
 	}
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -130,7 +130,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	// get the home namespace from space
 	defaultNamespace := getDefaultNamespace(space.Status.ProvisionedNamespaces)
 
-	// create the SpaceBindingRequest resource
+	// construct a SpaceBindingRequest object
 	sbrName := mur.GetName() + "-" + spaceRole
 	sbr := &toolchainv1alpha1.SpaceBindingRequest{
 		ObjectMeta: metav1.ObjectMeta{
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -1,0 +1,186 @@
+package spacebindingrequestmigration
+
+import (
+	"context"
+	"fmt"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/host-operator/controllers/spacebindingrequest"
+	"github.com/codeready-toolchain/host-operator/pkg/cluster"
+	errs "github.com/pkg/errors"
+	"github.com/redhat-cop/operator-utils/pkg/util"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Reconciler reconciles a SpaceBindingRequest object
+type Reconciler struct {
+	Client         runtimeclient.Client
+	Scheme         *runtime.Scheme
+	Namespace      string
+	MemberClusters map[string]cluster.Cluster
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
+	// since it's mandatory to add a primary resource when creating a new controller,
+	// we add the SpaceBindingRequest CR even if there should be no reconciles triggered from the host cluster,
+	// only from member clusters (see watches below)
+	// SpaceBindingRequest owns spacebindings so events will be triggered for those from the host cluster.
+	b := ctrl.NewControllerManagedBy(mgr).
+		For(&toolchainv1alpha1.SpaceBindingRequest{}).
+		Watches(&source.Kind{Type: &toolchainv1alpha1.SpaceBinding{}},
+			handler.EnqueueRequestsFromMapFunc(spacebindingrequest.MapSpaceBindingToSpaceBindingRequest()),
+		)
+
+	// Watch SpaceBindingRequests in all member clusters and all namespaces.
+	for _, memberCluster := range memberClusters {
+		b = b.Watches(
+			source.NewKindWithCache(&toolchainv1alpha1.SpaceBindingRequest{}, memberCluster.Cache),
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}))
+	}
+	return b.Complete(r)
+}
+
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=spacebindingrequests,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=spacebindingrequests/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=spacebindingrequests/finalizers,verbs=update
+
+// Reconcile converts all the SpaceBindings created using the sandbox-cli to SpaceBindingRequests
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("reconciling SpaceBindingRequestMigration")
+
+	// Fetch the SpaceBinding instance
+	spaceBinding := &toolchainv1alpha1.SpaceBinding{}
+	err := r.Client.Get(ctx, request.NamespacedName, spaceBinding)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, errs.Wrapf(err, "unable to get spacebinding")
+	}
+	if util.IsBeingDeleted(spaceBinding) {
+		logger.Info("the SpaceBinding is already being deleted")
+		return reconcile.Result{}, nil
+	}
+	// check if spaceBinding was created from SpaceBindingRequest,
+	// in that case we can skip it
+	if hasSpaceBindingRequest(spaceBinding) {
+		return reconcile.Result{}, nil
+	}
+
+	spaceName := types.NamespacedName{Namespace: spaceBinding.Namespace, Name: spaceBinding.Spec.Space}
+	space := &toolchainv1alpha1.Space{}
+	if err := r.Client.Get(ctx, spaceName, space); err != nil {
+		if errors.IsNotFound(err) {
+			// space was deleted
+			return reconcile.Result{}, nil
+		}
+		// error while reading space
+		return ctrl.Result{}, errs.Wrapf(err, "unable to get the bound Space")
+	}
+
+	murName := types.NamespacedName{Namespace: spaceBinding.Namespace, Name: spaceBinding.Spec.MasterUserRecord}
+	mur := &toolchainv1alpha1.MasterUserRecord{}
+	if err := r.Client.Get(ctx, murName, mur); err != nil {
+		if errors.IsNotFound(err) {
+			// mur was deleted
+			return reconcile.Result{}, nil
+		}
+		// error while reading MUR
+		return ctrl.Result{}, errs.Wrapf(err, "unable to get the bound MUR")
+	}
+
+	// skip workspace creator spacebinding
+	// the controller will convert only spacebindings created by system admins using the sandbox-cli.
+	// If the creator label on the space matches the usersignup reference on the MUR then this is the owner of the space
+	// and the spacebinding should not be migrated.
+	usersignup := mur.ObjectMeta.OwnerReferences[0]
+	if space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey] == usersignup.Name {
+		return reconcile.Result{}, nil
+	}
+
+	// get the spaceRole
+	spaceRole := spaceBinding.Spec.SpaceRole
+
+	// get member cluster name where the space was provisioned
+	targetCluster := space.Status.TargetCluster
+	memberCluster, memberClusterFound := r.MemberClusters[targetCluster]
+	if !memberClusterFound {
+		return ctrl.Result{}, errs.New(fmt.Sprintf("unable to find member cluster: %s", targetCluster))
+	}
+
+	// get the home namespace from space
+	defaultNamespace := getDefaultNamespace(space.Status.ProvisionedNamespaces)
+
+	// create the SpaceBindingRequest resource
+	sbrName := mur.GetName() + "-" + spaceRole
+	sbr := &toolchainv1alpha1.SpaceBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sbrName,
+			Namespace: defaultNamespace,
+			Labels: map[string]string{
+				toolchainv1alpha1.SpaceCreatorLabelKey:                 "spacebindingrequestmigration",
+				toolchainv1alpha1.SpaceBindingSpaceLabelKey:            space.GetName(),
+				toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey: mur.GetName(),
+			},
+		},
+		Spec: toolchainv1alpha1.SpaceBindingRequestSpec{
+			MasterUserRecord: mur.GetName(),
+			SpaceRole:        spaceRole,
+		},
+	}
+
+	// check if SpaceBindingRequest doesn't exist already
+	err = memberCluster.Client.Get(context.TODO(), types.NamespacedName{
+		Namespace: defaultNamespace,
+		Name:      sbrName,
+	}, &toolchainv1alpha1.SpaceBindingRequest{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// create the sbr resource
+			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
+			}
+
+			// delete the SpaceBinding
+			if err := r.Client.Delete(ctx, spaceBinding); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
+			}
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func getDefaultNamespace(provisionedNamespaces []toolchainv1alpha1.SpaceNamespace) string {
+	for _, namespaceObj := range provisionedNamespaces {
+		if namespaceObj.Type == "default" {
+			return namespaceObj.Name
+		}
+	}
+	return ""
+}
+
+func hasSpaceBindingRequest(spaceBinding *toolchainv1alpha1.SpaceBinding) bool {
+	_, sbrNameFound := spaceBinding.Labels[toolchainv1alpha1.SpaceBindingRequestLabelKey]
+	return sbrNameFound
+}
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -96,25 +96,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, errs.Wrapf(err, "unable to get the bound MUR")
 	}
 
-	// skip workspace creator spacebinding
-	// the controller will convert only spacebindings created by system admins using the sandbox-cli.
-	// If the creator label on the space matches the usersignup reference on the MUR then this is the owner of the space
-	// and the spacebinding should not be migrated.
-	if len(mur.ObjectMeta.OwnerReferences) == 0 {
-		return ctrl.Result{}, errs.New("MasterUserRecord has no UserSignup owner reference")
+	// error when mur has no owner label (should not happen in prod)
+	if _, ok := mur.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey]; !ok {
+		return ctrl.Result{}, errs.New("mur has no MasterUserRecordOwnerLabelKey set")
 	}
-	// skip spaces with no creator label
+	// error when space has no creator label (should not happen in prod)
 	if _, ok := space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey]; !ok {
-		// let's log an error, since this should not happen in production
-		logger.Error(errs.New("space has no SpaceCreatorLabelKey set"), "the spacebindings for this space will not be migrated", "space name", space.Name)
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, errs.New("space has no SpaceCreatorLabelKey set")
 	}
 
-	usersignup := mur.ObjectMeta.OwnerReferences[0]
-	if usersignup.Kind != "UserSignup" {
-		return ctrl.Result{}, errs.New(fmt.Sprintf("owner reference for mur %s is of kind %s. Expected UserSignup", murName, usersignup.Kind))
-	}
-	if space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey] == usersignup.Name {
+	// skip workspace creator spacebinding
+	// the controller will convert only spacebindings created by system admins using the sandbox-cli.
+	// If the creator label on the space matches the  owner label on the MUR then this is the owner of the space
+	// and the spacebinding should not be migrated.
+	if space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey] == mur.Labels[toolchainv1alpha1.MasterUserRecordOwnerLabelKey] {
 		return reconcile.Result{}, nil
 	}
 
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -35,8 +35,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[strin
 	// only from member clusters (see watches below)
 	// SpaceBindingRequest owns spacebindings so events will be triggered for those from the host cluster.
 	b := ctrl.NewControllerManagedBy(mgr).
-		For(&toolchainv1alpha1.SpaceBindingRequest{}).
-		Watches(&source.Kind{Type: &toolchainv1alpha1.SpaceBinding{}}, &handler.EnqueueRequestForObject{})
+		For(&toolchainv1alpha1.SpaceBinding{})
 
 	// Watch SpaceBindingRequests in all member clusters and all namespaces.
 	for _, memberCluster := range memberClusters {
@@ -107,6 +106,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	if mur.ObjectMeta.OwnerReferences == nil || len(mur.ObjectMeta.OwnerReferences) == 0 {
 		return ctrl.Result{}, errs.New("MasterUserRecord has no UserSignup owner reference")
 	}
+	// skip spaces with no creator label
+	if _, ok := space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey]; !ok {
+		// let's log an error, since this should not happen in production
+		logger.Error(errs.New("space has no SpaceCreatorLabelKey set"), "the spacebindings for this space will not be migrated", "space name", space.Name)
+		return ctrl.Result{}, nil
+	}
 	usersignup := mur.ObjectMeta.OwnerReferences[0]
 	if space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey] == usersignup.Name {
 		return reconcile.Result{}, nil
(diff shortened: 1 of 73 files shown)


controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
+	// since it's mandatory to add a primary resource when creating a new controller,
+	// we add the SpaceBindingRequest CR even if there should be no reconciles triggered from the host cluster,
+	// only from member clusters (see watches below)
+	// SpaceBindingRequest owns spacebindings so events will be triggered for those from the host cluster.
+	b := ctrl.NewControllerManagedBy(mgr).
+		For(&toolchainv1alpha1.SpaceBindingRequest{}).
```
alexeykazakov, line 38: Shouldn't it watch `SpaceBindings` in the host? And not `SBRs` in members?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	targetCluster := space.Status.TargetCluster
+	memberCluster, memberClusterFound := r.MemberClusters[targetCluster]
+	if !memberClusterFound {
+		return ctrl.Result{}, errs.New(fmt.Sprintf("unable to find member cluster: %s", targetCluster))
+	}
+
+	// get the home namespace from space
+	defaultNamespace := getDefaultNamespace(space.Status.ProvisionedNamespaces)
+
+	// create the SpaceBindingRequest resource
```
alexeykazakov, line 128:
```suggestion
	// construct a SpaceBindingRequest object
```

Just to make it clear that we are not creating an actual CR yet.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	sbrName := mur.GetName() + "-" + spaceRole
+	sbr := &toolchainv1alpha1.SpaceBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sbrName,
+			Namespace: defaultNamespace,
+			Labels: map[string]string{
+				toolchainv1alpha1.SpaceCreatorLabelKey:                 "spacebindingrequestmigration",
```
alexeykazakov, line 135: I wonder if it can cause issues in the future if we use a non-existing username as the create label... Maybe it will be safer to use the space owner as the creator?

WDYT?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,183 @@
...
+		Name:      sbrName,
+	}, &toolchainv1alpha1.SpaceBindingRequest{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// create the sbr resource
+			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
+			}
+
+			// delete the SpaceBinding
```
alexeykazakov, line 158: There is a race condition. We create an SBR which will trigger a SB creation and we delete the SB here too. We might end up deleting the SB created by the SBR controller. The question is if our SBR controller also watches SBs and will restore the deleted SB. Could you double check that?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: Yeah I was unsure about this too. I added something custom just to be able to distinguish the SpaceBindingRequests we created via this controller from the "regular" ones. Maybe I can use another label , and not   `SpaceCreatorLabelKey` ?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: Initially I was:
- deleting the SpaceBinding
- creating the SpaceBindingRequest

but after discussing this with @MatousJobanek , we decided to invert those, so that we avoid the case in which the SBR is not created and the SpaceBinding was already deleted, which means the user will not have access anymore.

If we go with the current order, could happen that:
- we create a SpaceBindingRequest
- the SBR stays in not provisioned since there is already a SpaceBinding for the same MUR and Space
- the migration controller deletes the existing SpaceBinding
- the SBR controller provisions a new SpaceBinding

> The question is if our SBR controller also watches SBs and will restore the deleted SB. Could you double check that?
yes, this is done here https://github.com/codeready-toolchain/host-operator/blob/567e867d43619a8a7d4f6b63ebf8e0d296be5f38/controllers/spacebindingrequest/spacebindingrequest_controller.go#L46 where we trigger a reconcile on the SpaceBindingRequest if the SpaceBinding is updated/deleted.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: makes sense! Thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: > Shouldn't it watch SpaceBindings in the host? 

I've repaced the `For` resource to use SpaceBindings, see https://github.com/codeready-toolchain/host-operator/pull/922/commits/038c6142d80ea7cd0d38f212c1e3b9d9a7f55438

> And not SBRs in members?

I thought that watching for SBR changes would be good, in case something changes the migration controller will reconcile it. Worst case it should not harm since the spacebinding will have an SBR already and will exit here https://github.com/codeready-toolchain/host-operator/pull/922/files#diff-d49007122db69f04ba9af1be1154a194e84eaaef10df1cb1caa211e54835a63bR76-R78

Also IIRC we need both a primary (`For`) and a secondary resource (`Watches`) , but I might be wrong.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
alexeykazakov: We could use another label for that but do we really need to make it even more complicated? IMO it should be fine if we create these SBRs "on the behalf of" the owner.
@MatousJobanek WDYT?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
alexeykazakov: > - the migration controller deletes the existing SpaceBinding
> - the SBR controller provisions a new SpaceBinding

And in parallel you are deleting the SpaceBinding in this controller. So theoretically it may happen that the SBR controller will ensure the SpaceBinding right before you are deleting it. This is the race condition I'm talking about. But since the SBR controller watches the SB resources it will restore it if it's deleted. So we are good here. Thanks for the confirmation.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
alexeykazakov: OK. The For() and Watches look good now since we actually have SBs as the primary resources and SBRs as secondary for this controller. However the description is still incorrect. I will create a new comment for that.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,188 @@
...
+	Namespace      string
+	MemberClusters map[string]cluster.Cluster
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
+	// since it's mandatory to add a primary resource when creating a new controller,
+	// we add the SpaceBindingRequest CR even if there should be no reconciles triggered from the host cluster,
+	// only from member clusters (see watches below)
+	// SpaceBindingRequest owns spacebindings so events will be triggered for those from the host cluster.
```
alexeykazakov, line 36: The description is not accurate. We watch the SB in the host as the primary resource and SBRs in members as secondary. The actual code is correct. We just need to update the description.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,188 @@
...
+		Name:      sbrName,
+	}, &toolchainv1alpha1.SpaceBindingRequest{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// create the sbr resource
+			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
+			}
+
+			// delete the SpaceBinding
```
alexeykazakov, line 163:
```suggestion
			// delete the SpaceBinding. The SpaceBindingRequest controller will create a new SpaceBinding for the SBR we created above.
```

alexeykazakov (approved):
Looks good overall. Just a couple of minor comments. However I didn't review it carefully enough. Please make sure you get the PR reviewed by someone else too before merging it.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: Good point. I've updated the comment as well https://github.com/codeready-toolchain/host-operator/pull/922/commits/d09374cfce39f3a9bd470735ea770348ecd347fb

Thanks!

mfrancisc:
> Please make sure you get the PR reviewed by someone else too before merging it.

Thanks for your review 
Yes , I'll wait for more reviews as this is kind of "delicate work". So let's make sure it will work as we expect.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
MatousJobanek: let's just drop the label - it's not used in SBRs anywhere. 
```suggestion
```

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	// construct a SpaceBindingRequest object
+	sbrName := mur.GetName() + "-" + spaceRole
+	sbr := &toolchainv1alpha1.SpaceBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sbrName,
+			Namespace: defaultNamespace,
+			Labels: map[string]string{
+				toolchainv1alpha1.SpaceCreatorLabelKey:                 "spacebindingrequestmigration",
+				toolchainv1alpha1.SpaceBindingSpaceLabelKey:            space.GetName(),
+				toolchainv1alpha1.SpaceBindingMasterUserRecordLabelKey: mur.GetName(),
```
MatousJobanek, line 139: We don't use these labels in case of SBRs anywhere either, right? I would drop them too, to not confuse anyone.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,185 @@
...
+			// create the sbr resource
+			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
+			}
+
+			// delete the SpaceBinding. The SpaceBindingRequest controller will create a new SpaceBinding for the SBR we created above.
+			if err := r.Client.Delete(ctx, spaceBinding); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
+			}
+			return ctrl.Result{}, nil
```
MatousJobanek, line 164: this should be executed when the SBR is already present. The controller watches both SBs and SBRs, so you can create the SBR and then just return and wait for the next reconcile triggered by the creation. 

This will also cover the situation when the deletion call fails for any reason - I'm talking abous a situation:
1. SpaceBinding is not migrated
2. the controller creates SBR CR
3. the `Delete` call fails for any reason
4. the controller returns and error
5. in the next reconcile, the controller doesn't get to this code because the SBR is already present.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller_test.go
```diff
@@ -0,0 +1,332 @@
...
+				HasLabelWithValue(toolchainv1alpha1.SpaceCreatorLabelKey, "somevalue").
+				Exists()
+		})
+
+		t.Run("spacebinding is being deleted", func(t *testing.T) {
+			// given
+			member1 := NewMemberClusterWithClient(test.NewFakeClient(t), "member-1", corev1.ConditionTrue)
+			// the spacebinding is being deleted
+			hostClient := test.NewFakeClient(t, janeMur, janeSpace, sbForCreator, sbForJohn)
+			ctrl := newReconciler(t, hostClient, member1)
```
MatousJobanek, line 206: you forgot to modify the SpaceBinding by adding the deletion timestamp
plus there is missing the john's MUR
```suggestion
			sbForJohn := spacebindingtest.NewSpaceBinding(johnMur.Name, janeSpace.Name, "admin", janeMur.GetName(), spacebindingtest.WithDeletionTimestamp())
			hostClient := test.NewFakeClient(t, janeMur, janeSpace, sbForCreator, johnMur, sbForJohn)
```

controllers/spacebindingrequestmigration/spacebindingrequest_spacebinding_mapper.go
```diff
@@ -0,0 +1,53 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// MapSpaceBindingRequestToSpaceBinding returns an event for the spacebinding that owns.
+func MapSpaceBindingRequestToSpaceBinding(cl runtimeclient.Client) func(spaceBindingREquest runtimeclient.Object) []reconcile.Request {
+	mapperLog := ctrl.Log.WithName("NSTemplateTierToSpaceMapper")
+	return func(obj runtimeclient.Object) []reconcile.Request {
+		if spaceBindingRequest, ok := obj.(*toolchainv1alpha1.SpaceBindingRequest); ok {
```
MatousJobanek, line 19: it's just a detail, technically speaking, we don't need to cast the object and validate its type, right?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: no , it is not used. The main usage was potentially for manual investigation. Dropped. Thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: I'm sorry, but If I'm not wrong, this way I'll be deleting the SpaceBinding owned by the SBR and not the one that was migrated right? 

I mean , if I defer the Delection of the SpaceBinding only when there's a SpaceBindingRequest associated to it, I'll not be deleting the intended objects , but I'm sure I'm missing something here 🤔

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller_test.go
```diff
@@ -0,0 +1,332 @@
...
+					Name:       "batman", // but the usersignup matches the space creator name
+				},
+			}
+			sbForBatman := spacebindingtest.NewSpaceBinding(batmanMur.GetName(), batmanSpace.GetName(), "admin", "batman")
+			member1 := NewMemberClusterWithClient(test.NewFakeClient(t), "member-1", corev1.ConditionTrue)
+			hostClient := test.NewFakeClient(t, batmanSpace, batmanMur, sbForBatman)
+			ctrl := newReconciler(t, hostClient, member1)
+
+			// when
+			_, err = ctrl.Reconcile(context.TODO(), requestFor(sbForJohn))
```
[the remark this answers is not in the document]
mfrancisc: 🤦‍♂️  thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller_test.go
```diff
@@ -0,0 +1,332 @@
...
+		})
+
+		t.Run("spacebinding has spacebindingrequest", func(t *testing.T) {
+			// given
+			// the spacebinding has a spacebindingrequest
+			sbrForJohn := spacebindingrequesttest.NewSpaceBindingRequest("john-admin", "jane-tenant", spacebindingrequesttest.WithLabel(toolchainv1alpha1.SpaceCreatorLabelKey, "somevalue"))
+			sbForJohnWithSBR := spacebindingtest.NewSpaceBinding(johnMur.Name, janeSpace.Name, "admin", janeMur.GetName(), spacebindingtest.WithSpaceBindingRequest(sbrForJohn))
+			// let's not load the mur object
+			member1 := NewMemberClusterWithClient(test.NewFakeClient(t, sbrForJohn), "member-1", corev1.ConditionTrue)
+			hostClient := test.NewFakeClient(t, janeMur, janeSpace, sbForJohnWithSBR)
```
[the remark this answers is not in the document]
mfrancisc: thanks

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller_test.go
mfrancisc: 🤦 good point! Thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_spacebinding_mapper.go
mfrancisc: correct! I've removed the cast in https://github.com/codeready-toolchain/host-operator/pull/922/commits/c86c05cade77cd893076a8a0a9088b0842015273

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: Not used right! Removed , please see https://github.com/codeready-toolchain/host-operator/pull/922/commits/532ca3aed1ef475bb9381fd31e1287715d2d73bb

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
MatousJobanek: If the SpaceBinding is already managed by SBR, then you shouldn't get to this point - you return from the reconcile much earlier: 
https://github.com/codeready-toolchain/host-operator/blob/f5ff6c7540fb694fd43b2403da30b589cac81595/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go#L71-L75

mfrancisc: As discussed, I have:
- forced a requeue after the creation of the SBR 
- moved the deletion of the migrated SpaceBinding object at the end, only after we checked that the SBR was already created

please check https://github.com/codeready-toolchain/host-operator/pull/922/commits/bf6f19d1839abd949dfda81ec4cf54c0f792227f

Thanks for your suggestion!

MatousJobanek (approved):
Looks great, thanks for moving the delete call :+1:

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,182 @@
...
+			Namespace: defaultNamespace,
+		},
+		Spec: toolchainv1alpha1.SpaceBindingRequestSpec{
+			MasterUserRecord: mur.GetName(),
+			SpaceRole:        spaceRole,
+		},
+	}
+
+	// check if SpaceBindingRequest doesn't exist already
+	err = memberCluster.Client.Get(context.TODO(), types.NamespacedName{
```
sadlerap, line 144: Use the context provided as a part of reconciliation.

```suggestion
	err = memberCluster.Client.Get(ctx, types.NamespacedName{
```

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,182 @@
...
+			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
+				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
+			}
+
+			// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
+			return ctrl.Result{Requeue: true}, nil
+		}
+		// Error reading the object
+		return ctrl.Result{}, err
+	}
```
sadlerap, line 160: Suggestion: maybe use [the `CreateOrUpdate` function from `controllerutil`][1]?  Up to you if you want to leave it as-is, but it might make this part of reconciliation easier to follow.

[1]: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/controller/controllerutil#CreateOrUpdate

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: thanks! I'm too used to this TODO context 🤦

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: Nice suggestion. I haven't used this function before tbh, we have few use case where it could improve readability 🤔 
  
In this specific case, I don't actually need to update anything. 
In my case is important to know if the SBR already exists so I can delete the SpaceBinding (see line 163). If it doesn't exist I need to create it and force a reconcile of the same object ( see line 156 ) 
I guess I could still do this by checking the `op` returned if it's `created` (force reconcile) or one between `updated ` or `unchanged` (delete the SB). But still it wouldn't look obvious to me what is going on. WDYT?

sadlerap: I refactored it to see how it'd turn out, and this is what I wound up with:

<details><summary>Reconcile method epilogue</summary>
<p>

```go

	// construct a SpaceBindingRequest object
	sbrName := mur.GetName() + "-" + spaceRole
	sbr := &toolchainv1alpha1.SpaceBindingRequest{
			Name:      sbrName,
			Namespace: defaultNamespace,
		},
	}

	result, err := controllerutil.CreateOrUpdate(ctx, memberCluster.Client, sbr, func() error {
		sbr.Spec = toolchainv1alpha1.SpaceBindingRequestSpec{
			MasterUserRecord: mur.GetName(),
			SpaceRole:        spaceRole,
		}
		return nil
	})

	if err != nil {
		// something happened when we tried to read or write the sbr
		return ctrl.Result{}, errs.Wrapf(err, "Failed to create or update space binding request %v", sbrName)
	}

	if result == controllerutil.OperationResultCreated {
		// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
		return ctrl.Result{Requeue: true}, nil
	}

	// if the SBR was found (was created from the previous reconcile loop), we can now delete the SpaceBinding object
	if err := r.Client.Delete(ctx, spaceBinding); err != nil {
		return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
	}

	return ctrl.Result{}, nil
}

```

</p>
</details> 

IMO it looks a bit nicer, but I'll leave it for you to decide.

If you're interested, here's the full diff:

<details><summary>The full diff</summary>
<p>

```diff
diff --git a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
index ae79c1d..c72df74 100644
--- a/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
+++ b/controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -134,32 +135,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 			Name:      sbrName,
 			Namespace: defaultNamespace,
 		},
-		Spec: toolchainv1alpha1.SpaceBindingRequestSpec{
+	}
+
+	result, err := controllerutil.CreateOrUpdate(ctx, memberCluster.Client, sbr, func() error {
+		sbr.Spec = toolchainv1alpha1.SpaceBindingRequestSpec{
 			MasterUserRecord: mur.GetName(),
 			SpaceRole:        spaceRole,
-		},
-	}
-
-	// check if SpaceBindingRequest doesn't exist already
-	err = memberCluster.Client.Get(ctx, types.NamespacedName{
-		Namespace: defaultNamespace,
-		Name:      sbrName,
-	}, &toolchainv1alpha1.SpaceBindingRequest{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// create the sbr resource
-			if err := memberCluster.Client.Create(ctx, sbr); err != nil {
-				return ctrl.Result{}, errs.Wrapf(err, "unable to create SpaceBindingRequest")
-			}
-
-			// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
-			return ctrl.Result{Requeue: true}, nil
 		}
-		// Error reading the object
-		return ctrl.Result{}, err
+		return nil
+	})
+
+	if err != nil {
+		// something happened when we tried to read or write the sbr
+		return ctrl.Result{}, errs.Wrapf(err, "Failed to create or update space binding request %v", sbrName)
 	}
 
-	// if the SBR was found ( was created from the previous reconcile loop), we can now delete the SpaceBinding object
+	if result == controllerutil.OperationResultCreated {
+		// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// if the SBR was found (was created from the previous reconcile loop), we can now delete the SpaceBinding object
 	if err := r.Client.Delete(ctx, spaceBinding); err != nil {
 		return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
 	}

```

</p>
</details>

mfrancisc: Nice job @sadlerap . I agree with you, this indeed looks nicer and easier to read. It is indeed a pattern we should use more often in our code.

My only "concern" , so to speak, is about the mutateFn, which will trigger an update of the object, and an extra reconcile if I'm not wrong. Given that this controller will reconcile all the spacebindings CRs (800 objects for prod), and convert those to SBRs, there will be potentially 800 extra reconcile loops when we deploy it, it could be a performance concern and cause some slowness to the migration task. But I might be wrong.

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	if mur.ObjectMeta.OwnerReferences == nil || len(mur.ObjectMeta.OwnerReferences) == 0 {
+		return ctrl.Result{}, errs.New("MasterUserRecord has no UserSignup owner reference")
+	}
+	// skip spaces with no creator label
+	if _, ok := space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey]; !ok {
+		// let's log an error, since this should not happen in production
+		logger.Error(errs.New("space has no SpaceCreatorLabelKey set"), "the spacebindings for this space will not be migrated", "space name", space.Name)
+		return ctrl.Result{}, nil
+	}
+	usersignup := mur.ObjectMeta.OwnerReferences[0]
```
filariow, line 112: should we check if OwnerReference's kind is UserSignup?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,182 @@
...
+			return ctrl.Result{Requeue: true}, nil
+		}
+		// Error reading the object
+		return ctrl.Result{}, err
+	}
+
+	// if the SBR was found ( was created from the previous reconcile loop), we can now delete the SpaceBinding object
+	if err := r.Client.Delete(ctx, spaceBinding); err != nil {
+		return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
+	}
```
filariow, line 165: if `err` here is `NotFound`, should we be ok with it?

```suggestion
	if err := r.Client.Delete(ctx, spaceBinding); err != nil {
	        if !errors.IsNotFound(err) {
			return ctrl.Result{}, errs.Wrapf(err, "unable to delete the SpaceBinding")
		}
	}
```

controllers/spacebindingrequestmigration/spacebindingrequest_spacebinding_mapper.go
```diff
@@ -0,0 +1,50 @@
...
+
+// MapSpaceBindingRequestToSpaceBinding returns an event for the spacebinding that owns.
+func MapSpaceBindingRequestToSpaceBinding(cl runtimeclient.Client) func(spaceBindingREquest runtimeclient.Object) []reconcile.Request {
+	mapperLog := ctrl.Log.WithName("NSTemplateTierToSpaceMapper")
+	return func(obj runtimeclient.Object) []reconcile.Request {
+		watchNamespace, err := commonconfig.GetWatchNamespace()
+		if err != nil {
+			mapperLog.Error(err, "cannot get watch namespace")
+			return []reconcile.Request{}
+		}
```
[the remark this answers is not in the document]
mfrancisc: good point. Done , please see https://github.com/codeready-toolchain/host-operator/pull/922/commits/73a342b637731cd526870bcc51f291202a4466fc 

thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		// error while reading MUR
+		return ctrl.Result{}, errs.Wrapf(err, "unable to get the bound MUR")
+	}
+
+	// skip workspace creator spacebinding
+	// the controller will convert only spacebindings created by system admins using the sandbox-cli.
+	// If the creator label on the space matches the usersignup reference on the MUR then this is the owner of the space
+	// and the spacebinding should not be migrated.
+	if mur.ObjectMeta.OwnerReferences == nil || len(mur.ObjectMeta.OwnerReferences) == 0 {
```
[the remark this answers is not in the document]
mfrancisc: I've tried it [here](https://go.dev/play/p/V23cBGs1SpQ?v=goprev) , but it gives an error 😕

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: why not. Please see https://github.com/codeready-toolchain/host-operator/pull/922/commits/f0603c97e02c415d54f08701727abdab24b277a8

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: good catch ! I've applied your first suggestion. please see https://github.com/codeready-toolchain/host-operator/pull/922/commits/c94803b6e8bad98a6f126c635d98ae601079ff8a

thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
filariow: I'm thinking we may look in the whole list of Owners to find the UserSignup entry. In this way we won't have the hard requirement that the UserSignup has to be the first of the list. WDYT?

something like the following:
```go
        // get the first UserSignup from the list of Owners
        // TODO: should we handle the scenario in which more than one UserSignup is set as owner?
	usersignup, err := func() (*metav1.OwnerReference, error) {
		for _, u := range mur.ObjectMeta.OwnerReferences {
			if u.Kind == "UserSignup" { // TODO: Group check?
				return &u, nil
			}
		}
		return nil, errs.New(fmt.Sprintf("owner reference of UserSignup not found for mur %s", murName))
	}()
	if err != nil {
		return ctrl.Result{}, err
	}
```

Other considerations on this piece of logic are:
- should we also check the group?
- should we handle the scenario in which more than one UserSignup is set as owner?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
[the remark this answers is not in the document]
filariow: that's not working because go can not understand which is the type of the `nil`, hence it's not able to compile.
Try with something similar to [this](https://go.dev/play/p/n2CZGzep_TJ?v=goprev) or [this](https://go.dev/play/p/zlG1KtbummT?v=goprev)

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
MatousJobanek: This is a resource that is fully managed by our controllers, I don't think that we can rely on the assumption that there is always one owner-refence which is UserSignup

mfrancisc: @filariow I can create a lookup function for sure. I don't think we have to check the group and handle the case of multiple usersignups as reference. 

TBH I thought it would be safe to assume there is only one UserSignup as reference.

mfrancisc: @MatousJobanek since the resource is fully managed by our controllers , I thought instead that we could rely on that assumption. But I can implement the lookup as @filariow suggested. WDYT?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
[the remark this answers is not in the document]
mfrancisc: hmm good point! Thanks for double checking! I've accepted your suggestions!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: If it doesn't post anything to k8s in case of `unchanged` result, then that's good. 
But I wouldn't add the check for changed objects inside the function.

Let me have a closer look at what it does and update the code eventually. Thanks again for looking into it!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
MatousJobanek: you can use the lookup logic, but let's not over-engineer it and be pragmatic here. 
Btw, the second option of getting the UserSignup name is using the `.../owner` label from the MUR resource: 
https://github.com/codeready-toolchain/host-operator/blob/ad4c06f14f96db10a0ff794f413a362f62c7196e/controllers/usersignup/masteruserrecord.go#L39
so you don't have to struggle with the owner-references and the corner cases, and just pick the value of the label :wink:

mfrancisc: @MatousJobanek @filariow could you PTAL at https://github.com/codeready-toolchain/host-operator/pull/922/commits/7d0e0b9306cdb44dba68aa762af355bda822be50 and see if this looks good to you?

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: @sadlerap I've implemented your first suggestion , please take a look at https://github.com/codeready-toolchain/host-operator/pull/922/commits/aaa038320bb7746cfb27eefa2ed5d9206eb90a9b 

thanks!

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
MatousJobanek: Looks good, thanks :+1:

mfrancisc:
/retest

🤔 
```
    signup_request.go:334: 
        	Error Trace:	/tmp/toolchain-e2e/testsupport/signup_request.go:334
        	            				/tmp/toolchain-e2e/testsupport/signup_request.go:237
        	            				/tmp/toolchain-e2e/testsupport/space/space.go:33
        	            				/tmp/toolchain-e2e/test/e2e/parallel/space_cleanup_test.go:204
        	            				/tmp/toolchain-e2e/test/e2e/parallel/space_cleanup_test.go:36
        	Error:      	Received unexpected error:
        	            	Post "https://registration-service-toolchain-host-20091012.apps.ci-op-gnhkrq4z-1de72.origin-ci-int-aws.dev.rhcloud.com/api/v1/signup?no-space=true": EOF
        	Test:       	TestSpaceAndSpaceBindingCleanup/for_SpaceBinding/when_space_is_deleted
        	Messages:   	error posting signup request.
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_host-operator/922/pull-ci-codeready-toolchain-host-operator-master-e2e/1726520391163187200/build-log.txt

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		// something happened when we tried to read or write the sbr
+		return ctrl.Result{}, errs.Wrapf(err, "Failed to create or update space binding request %v", sbrName)
+	}
+
+	if result == controllerutil.OperationResultCreated {
+		// let's requeue after we created the SBR, so that in next loop the migrated SpaceBinding object will be deleted
+		return ctrl.Result{Requeue: true}, nil
+	}
+	// if the SBR was found ( was created from the previous reconcile loop), we can now delete the SpaceBinding object
```
sadlerap, line 156:
```suggestion
	// if the SBR was found (was created from the previous reconcile loop), we can now delete the SpaceBinding object
```

sadlerap (commented):
one small nitpick, otherwise LGTM

controllers/spacebindingrequestmigration/spacebindingrequest_migration_controller.go
mfrancisc: thanks!

sadlerap (approved):

filariow (approved):
lgtm

PR #938, opened 2023-11-27
bugfix(spacebindingrequest): allow deletion of failed SpaceBindingRequest

mfrancisc:
This fixes the issue with SpaceBindingRequest not being deleted when there is already an existing SpaceBinding object not owned by the SBR resource.

Jira: https://issues.redhat.com/browse/ASC-442

### DIFF
--- a/controllers/spacebindingrequest/spacebindingrequest_controller.go
+++ b/controllers/spacebindingrequest/spacebindingrequest_controller.go
@@ -229,10 +229,8 @@ func (r *Reconciler) ensureSpaceBinding(ctx context.Context, memberCluster clust
 
 	// check if existing spacebinding was created from spaceBindingRequest
 	// and return an error in case it was not.
-	if spaceBinding != nil {
-		if ownedBySBR := sbrOwnsSpaceBinding(spaceBinding, spaceBindingRequest); !ownedBySBR {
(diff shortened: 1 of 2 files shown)


controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -260,6 +266,18 @@ func (r *Reconciler) ensureSpaceBinding(ctx context.Context, memberCluster clust
...
+// sbrOwnsSpaceBinding check if the SpaceBinding was created by the given SpaceBindingRequest resource
+// returns true if it was created by the SBR
+// returns false if it wasn't created by the SBR
+func sbrOwnsSpaceBinding(spaceBinding *toolchainv1alpha1.SpaceBinding, spaceBindingRequest *toolchainv1alpha1.SpaceBindingRequest) bool {
+	sbrLabel := spaceBinding.Labels[toolchainv1alpha1.SpaceBindingRequestLabelKey]
+	sbrNamespaceLabel := spaceBinding.Labels[toolchainv1alpha1.SpaceBindingRequestNamespaceLabelKey]
+	if sbrLabel != spaceBindingRequest.GetName() || sbrNamespaceLabel != spaceBindingRequest.GetNamespace() {
+		return false
+	}
+	return true
```
sadlerap, line 278:
```suggestion
	return sbrLabel == spaceBindingRequest.GetName() && sbrNamespaceLabel == spaceBindingRequest.GetNamespace()
```

sadlerap (commented):
One small nit, otherwise LGTM

controllers/spacebindingrequest/spacebindingrequest_controller.go
mfrancisc: indeed this is shorter/nicer! thanks a lot!

sadlerap (approved):

PR #943, opened 2023-12-01
Refactor MasterUserRecordController and tests to use only propagated claims

sbryzak:
This PR refactors the MasterUserRecordController to only use propagated claims when creating UserAccount etc.

Related toolchain-common PR: https://github.com/codeready-toolchain/toolchain-common/pull/345

Fixes https://issues.redhat.com/browse/SANDBOX-482

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/862


alexeykazakov (approved):

pkg/metrics/email_domain.go
```diff
@@ -20,9 +20,16 @@ var internalDomainPattern = regexp.MustCompile(`^.*@redhat.com$|^.*@((.+)\.)?ibm
...
-	if !exists {
-		log.Error(nil, "no email address found in annotations", "kind", obj.GetObjectKind().GroupVersionKind().Kind, "name", obj.GetName())
+	emailAddress := ""
+
+	switch obj := obj.(type) {
+	case *toolchainv1alpha1.MasterUserRecord:
+		emailAddress = obj.Spec.PropagatedClaims.Email
+	case *toolchainv1alpha1.UserSignup:
+		emailAddress = obj.Spec.IdentityClaims.Email
+	}
```
MatousJobanek, line 30: if you change the type of the parameter of the function to take the `PropagatedClaims`, then it would simplify the logic and it would be also clear what the function does - where the value takes from. There is no need for checking the types.
```
func GetEmailDomain(claims toolchainv1alpha1.PropagatedClaims) Domain {
	emailAddress := claims.Email
```

MatousJobanek (commented):
I don't see any changes regarding to the modified `UserID` function from the toolchain-common. It sets different value than it did before, so my question is, do we still need the function? Does it serve to our needs and do we use it properly?

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2490,6 +2473,11 @@ func TestUserSignupReactivateAfterDeactivated(t *testing.T) {
 		Spec: toolchainv1alpha1.UserSignupSpec{
 			Userid:   "UserID123",
 			Username: meta.Name,
+			IdentityClaims: toolchainv1alpha1.IdentityClaimsEmbedded{
+				PropagatedClaims: toolchainv1alpha1.PropagatedClaims{
+					Email: "john.doe@redhat.com",
+				},
+			},
```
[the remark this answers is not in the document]
sbryzak: I know that without the email address being set here, the black box mystery that is the metrics tests will fail.  I've spent some time trying to work out why this is the case, and my only suspicion is that it's related to the email domain function.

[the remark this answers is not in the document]
MatousJobanek: Talking about notifications - I see that the e2e tests are failing for the idling notifications. My guess is that it's because the idler controller is still using the email annotation for sending the notification. You need to update that part first before dropping it from the MUR: 
https://github.com/codeready-toolchain/member-operator/blob/f4220407cbae99fa7d3713504fef5f7dbd196ace/controllers/idler/idler_controller.go#L270-L271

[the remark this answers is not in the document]
sbryzak: @MatousJobanek thanks for pointing this out, I wasn't aware of the idler controller or what it did.  I've updated it to now use the propagated claims.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -2490,6 +2473,11 @@ func TestUserSignupReactivateAfterDeactivated(t *testing.T) {
 		Spec: toolchainv1alpha1.UserSignupSpec{
 			Userid:   "UserID123",
 			Username: meta.Name,
+			IdentityClaims: toolchainv1alpha1.IdentityClaimsEmbedded{
+				PropagatedClaims: toolchainv1alpha1.PropagatedClaims{
+					//Email: "john.doe@redhat.com",
+				},
+			},
```
alexeykazakov, line 2480: Do you need this empty IdentityClaims here?

sbryzak: We do need to set the email address here otherwise all the metrics tests fail, although I still haven't been able to work out why this is the case.  I've uncommented this line, I didn't intend to commit this change.

pkg/metrics/email_domain.go
sbryzak: This is a little simpler yes, however it would also mean that we lose details in the logs if the email address is not found.  We cannot determine the object type or the name from just the PropagatedClaims.

`log.Error(nil, "no email address found in object", "kind", obj.GetObjectKind().GroupVersionKind().Kind, "name", obj.GetName())`

MatousJobanek: fair enough :+1:

MatousJobanek (approved):
Thanks for answering my comments :+1:

PR #944, opened 2023-12-05
Our larger RHTAP tenants need more memory quota

amfred:
By making containers a little less CPU-hungry by default, we're now able to start up more containers.  But now we're running into memory quotas for the larger tenants.  Meanwhile, we still have plenty of actual capacity on our clusters.  

I also removed the CPU limit quota.  There's still a CPU request quota. 

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/861


deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
```diff
@@ -156,7 +155,6 @@ objects:
     limits:
     - type: "Container"
       default:
-        cpu: 2000m
```
hugares, line 159: If we want to remove the CPU limit from the LimitRange that is namespace wide, I think we need to remove cpu limits from all quotas, not only compute-build

alexeykazakov: Yes, we also would need to remove it from the NotTerminating quota too.
Also, could you test that dropping the CPU limit only from the quotas and the limitrange and keeping the requests there actually works?

alexeykazakov (commented):
Please also create a paired PR (with the same branch name as this one) for our e2e tests. See an example of such paired PR here: https://github.com/codeready-toolchain/toolchain-e2e/pull/853

amfred:
/hold
Let's see if your bots recognize that :-)

deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
amfred: Yeah.  As a first step can you give this one an ok-to-test to see what the automated tests do with "0" values, and we'll discuss how else we might need to test it in our next Infra team stand-up?

alexeykazakov: For testing I would just:
- take a clean dev cluster (using cluster bot for example)
- create a namespace and create the same quota and limit range as in this PR
- try to create a Pod
If it's created and scheduled successfully then we are good IMO.

Unfortunately our (sandbox) e2e tests to do not currently tests tiers by creating pods.

deploy/templates/nstemplatetiers/appstudio/ns_tenant.yaml
```diff
@@ -156,7 +154,6 @@ objects:
...
         memory: 2Gi
       defaultRequest:
         cpu: 10m
```
hugares, line 159: Should we put 200m as default cpu request, as it was tested by performance team?

amfred: Yes you're right.

amfred:
Yay, passing tests. 
image

amfred:
I was able to manually test with these settings and my PipelineRuns were able to complete: 
Default LimitRange settings:
```
spec:
  limits:
    - type: Container
      default:
        memory: 2Gi
      defaultRequest:
        cpu: 200m
        memory: 256Mi
compute-build ResourceQuota:
```
spec:
  hard:
    limits.memory: 512Gi
    requests.cpu: '24'
    requests.memory: 128Gi
  scopes:
    - Terminating
 ```
compute-deploy ResourceQuota:
```
spec:
  hard:
    limits.memory: 32Gi
    requests.cpu: 1750m
    requests.memory: 32Gi
  scopes:
    - NotTerminating
```

So I think we're OK to go with this change.

alexeykazakov:
@bamachrn please do not use force push in PRs. It makes PR review much more difficult. Now we have to review the entire PR from scratch instead of reviewing the new changes only.

config/crd/bases/toolchain.dev.openshift.com_spaceprovisionerconfigs.yaml
```diff
@@ -0,0 +1,125 @@
+---
```
alexeykazakov, line 1: I'm not sure what's going on but there is a bunch of unrelated changes in the PR which has been already merged into the master branch. Not sure why GitHub still showing them here...
Like this CRD for example.

alexeykazakov (changes requested):
Let's make sure the PR is clean and doesn't contain unrelated changes. Please take a look at my comment.

amfred:
@bamachrn What are the new PRs that replace this?

mfrancisc:
Should we close this ? As I see there is a new PR https://github.com/codeready-toolchain/host-operator/pull/969

PR #946, opened 2023-12-11
add support for Disable Inheritance in SpaceRequest

mmulholla:
Add support for SpaceRequest with disable inheritance support:
https://issues.redhat.com/browse/ASC-428

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/864


mmulholla:
@MatousJobanek 

Would like to understand SpaceRequest more. From what I can tell when a space is created from a space space request the only space binding it may get is the space binding from the parent, which with disable inheritance it will not get. This test then fails because a role is expected in the nstemplateset but there is none because the associated space has no space binding. So is the expected behavior that a space from a space request with disable inheritance has no space binding and therefore this must be added later. Or is it acceptable that the nstemplateset has no role and I change the test code to not expect on when disable inherttance is set? Of course I may have misunderstood things.

mfrancisc (commented):
Looks good 👍 

I have one question - should we support also update of the `disableInheritance` field ?

mmulholla:
> Looks good 👍
> 
> I have one question - should we support also update of the `disableInheritance` field ?

Not sure about update and what it involves. However as it was not part of the original requirement and I need to move to my new team I ask that if it is needed it is done separately and not by me.

mfrancisc:
> Not sure about update and what it involves. 

I'm not 100% sure either. @MatousJobanek might be able to help clarifying this.

But I think we should update the field here https://github.com/codeready-toolchain/host-operator/blob/33a4991bf3cdde8c51ee6b0e7bd4e5234dcba4ed/controllers/spacerequest/spacerequest_controller.go#L301-L302 
So that the value in the subSpace always reflects the one from SpaceRequest.

MatousJobanek:
@mmulholla your understanding is correct - as a result of such SBR there is gonna be a Space without any SpaceBinding at all. The purpose is that the such a sub-space should be managed by 3rd party controller running in the same cluster, so when the controller creates the SBR, then it will also manage who can have access to the sub-space.

>I have one question - should we support also update of the disableInheritance field ?

Having the field immutable can be actually considered as a feature in this case - just to make sure that no-one who has access to the namespace won't be able to modify the SBR so it would grant him/her-self the permissions via the SpaceBinding inheritance.

Take a look at this doc where I suggested the whole flow of provisioning & managing of the embargoed workspaces. The implementation on the side of the RHTAP 3rd party controllers will look a bit different (they decided to use cli for copying the application instead of a controller), but the main idea should stay the same. https://docs.google.com/document/d/18i7hGoSeVtk41UuIm7QJOpBmBRNbGizbkZOYBM9uijg/edit#heading=h.c956be86rfq4

mfrancisc:
> Having the field immutable can be actually considered as a feature in this case - just to make sure that no-one who has access to the namespace won't be able to modify the SBR so it would grant him/her-self the permissions via the SpaceBinding inheritance.

Ok I see, so this change is related to the embargoed workspace feature. Thanks for pointing that out. I see that we don't  want users with access to the parent space switching the `disableInheritance` field from `true` to `false`, which will grant themself access to the subSpace.

Maybe in the future we could have a validating webhook which will trigger an error when an update of the field is requested. Right now it will "silently fail" (without updating the field), which is fine as you said.

mfrancisc (approved):
Nice job 🚀 

Thank you!

MatousJobanek:
> Maybe in the future we could have a validating webhook which will trigger an error when an update of the field is requested. Right now it will "silently fail" (without updating the field), which is fine as you said.

BTW, as of now, RHTAP users don't have any permissions over SpaceRequest CRs, so they even cannot modify that even if we would support that.

pkg/space/space.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // NewSpace creates a space CR for a UserSignup object.
-func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string, compliantUserName, tier string) *toolchainv1alpha1.Space {
+func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string, compliantUserName, tier string, disableInheritance bool) *toolchainv1alpha1.Space {
```
MatousJobanek, line 12: Do we actually need this? the root-space should always have `DisableInheritance=false` - there is no point in disabling the inheritance there. Also, when I look at the places where you call this function, it always sets `false`, or am I missing something?

mfrancisc: @MatousJobanek I've addressed this in https://github.com/codeready-toolchain/host-operator/pull/946/commits/42d57507c680fd47418042c0f7dc4be07b7d19c2

MatousJobanek (approved):
Looks good :+1:

mfrancisc:
/retest

updated tests

filariow (approved):
lgtm

PR #953, opened 2024-01-02
bugfix: invalid number of secrets found

mfrancisc:
Bugfix for: https://issues.redhat.com/browse/RHTAPBUGS-980


alexeykazakov (approved):

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -486,7 +487,18 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
 
 		case len(secretList.Items) > 1:
 			// some unexpected issue causing to many secrets
-			return fmt.Errorf("invalid number of secrets found. actual %d, expected %d", len(secretList.Items), 1)
+			// this ca be caused by the client cache which is not up-to-date immediately after the first secret is created.
```
filariow, line 490:
```suggestion
			// this can be caused by the client cache which is not up-to-date immediately after the first secret is created.
```

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -486,7 +487,18 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
...
+			logger.Error(fmt.Errorf("invalid number of secrets found. actual %d, expected %d", len(secretList.Items), 1), "defaulting on the first secret in alphabetical order")
```
filariow, line 491: should we use log.Info here? something like the following

```suggestion
			logger.Info("invalid number of secrets found: defaulting on the first secret in alphabetical order", "actual", len(secretList.Items), "expected", 1)
```

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -486,7 +487,18 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
...
+			// Sort alphabetically the list of secrets
+			// so that in case of multiple secrets for the same namespace we always return the same one.
+			sort.Slice(secretList.Items, func(i, j int) bool {
+				return secretList.Items[i].Name < secretList.Items[j].Name
+			})
+			// let's default on the first secret, as they should all be valid, and they will all be deleted once the SpaceRequest resource is deleted.
+			namespaceAccess = append(namespaceAccess, toolchainv1alpha1.NamespaceAccess{
+				Name:      namespace.Name,
+				SecretRef: secretList.Items[0].Name,
+			})
```
filariow, line 501: should we cleanup by deleting the other secrets? What if we just delete the other secrets and reconcile again?

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -486,7 +487,18 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
...
 			// some unexpected issue causing to many secrets
-			return fmt.Errorf("invalid number of secrets found. actual %d, expected %d", len(secretList.Items), 1)
+			// this ca be caused by the client cache which is not up-to-date immediately after the first secret is created.
+			logger.Error(fmt.Errorf("invalid number of secrets found. actual %d, expected %d", len(secretList.Items), 1), "defaulting on the first secret in alphabetical order")
+			// Sort alphabetically the list of secrets
+			// so that in case of multiple secrets for the same namespace we always return the same one.
+			sort.Slice(secretList.Items, func(i, j int) bool {
+				return secretList.Items[i].Name < secretList.Items[j].Name
+			})
+			// let's default on the first secret, as they should all be valid, and they will all be deleted once the SpaceRequest resource is deleted.
```
filariow, line 497: if they are all valid we may consider to use the [ListOptions's Limit field](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/client/options.go#L416)

controllers/spacerequest/spacerequest_controller.go
mfrancisc: it's fine to use `Info` as well indeed. I haven't thought about the stacktrace. The main reason to use `Error` was that it is an error case actually :) , and I don't expect this to happen so often, but grepping the string should be easy as well. 

Thanks for the suggestion.

controllers/spacerequest/spacerequest_controller.go
mfrancisc: Yeah, it's a good idea indeed! 
But how do we know which secret we can delete?

I was thinking if it could happen that:
1) a first reconcile creates the first secret, let's say `secret-123` and sets the `SpaceRequest.Status.NamespaceAccess[0].SecretRef=secret-123`
 
2) meanwhile the "client" starts using this secret for accessing the namespace 

3) a second reconcile doesn't have this secret in the cache and creates a new secret `secret-456`, and updates the field to `SpaceRequest.Status.NamespaceAccess[0].SecretRef=secret-456`

4) the client might start using this new secret now, or it might still use the old one if it doesn't check constantly for changes on the SpaceRequest resource after it's provisioned 

how can we be sure about which secret to delete without creating "service interruption" ? 

Or should we just assume the client should "watch" after the SpaceRequest CR and always be using the secret that's present in the `SecretRef` field?

controllers/spacerequest/spacerequest_controller.go
mfrancisc: Nice idea, I'm now sure if there's a way to always return the same secret with the ListOptions mechanism. Do you know if it's possible?

I'm sorting the list of secrets, with the intent of returning the same secret and make it deterministic.

controllers/spacerequest/spacerequest_controller.go
filariow: We may implement the following piece of logic.

When creating the Secret (`case` at L446), if the field `SpaceRequest.Status.NamespaceAccess` contains an entry for the current namespace, use the entry's value for `SecretRef.Name` to as secret name, otherwise use the `GeneratedName` property for creating a new one.
This would recreate the same secret if it does not exist, or return an error if the secret already exists but we reached that place because the cache is not up to date.

When it comes to cleanup, we delete all the ones not in `SpaceRequest.Status.NamespaceAccess`.

WDYT?

controllers/spacerequest/spacerequest_controller.go
filariow: Unfortunately, I don't think there is any guarantee on that ordering. Empirically, I'd say results seems to be always returned sorted the same way.
The point here is: if we they are all valid, and we are not cleaning up, whichever we pick is the same.

If we are cleaning and picking just one, I'd say we should just pick the one in the status (`SpaceRequest.Status.NamespaceAccess`).

controllers/spacerequest/spacerequest_controller.go
filariow: > Yeah, it's a good idea indeed! But how do we know which secret we can delete?

For selecting the ones we can delete, I'd say we should pick the ones not in `SpaceRequest.Status.NamespaceAccess`

> Or should we just assume the client should "watch" after the SpaceRequest CR and always be using the secret that's present in the SecretRef field?

Yeah, that's a safe approach that could be used by clients

mfrancisc: > This would recreate the same secret if it does not exist, or return an error if the secret already exists but we reached that place because the cache is not up to date.

How about we create the secret (case at L446) only when `SecretRef.Name` is empty ? we don't really need to create the secret with the same name and trigger an error right ?

> When it comes to cleanup, we delete all the ones not in SpaceRequest.Status.NamespaceAccess

I think we could cleanup only when `SpaceRequest.Status.NamespaceAccess` has already a `SecretRef.Name` for the specific namespace and the listing found more then one. is this what you are suggesting?

controllers/spacerequest/spacerequest_controller.go
mfrancisc: TBH I'm not sure how much a performance improvement would be to limit the listing to just one secret. I don't think there could be more than 2 secrets per namespace ( worst case ). 

Plus if we want to cleanup the extra ones we need to retrieve them all right ?

controllers/spacerequest/spacerequest_controller.go
filariow: > How about we create the secret (case at L446) only when `SecretRef.Name` is empty ? we don't really need to create the secret with the same name and trigger an error right ?

Then if the secret is manually deleted, we won't recreate that. Is this a scenario we want/have to manage?
Also, as the Secret is created in a loop and SpaceRequest's status is updated after the loop is complete, we may have an error in the middle of the loop and find ourself with status not correctly up to date. This will lead to more than one secret created. We can fix this by, catching the error, updating the status and rethrowing the error.

> I think we could cleanup only when `SpaceRequest.Status.NamespaceAccess` has already a `SecretRef.Name` for the specific namespace and the listing found more then one. is this what you are suggesting?

controllers/spacerequest/spacerequest_controller.go
filariow: Yeah, I agree that the Limit field won't probably give us much advantages.

controllers/spacerequest/spacerequest_controller.go
mfrancisc: @filariow thanks for your suggestion. 

Please see https://github.com/codeready-toolchain/host-operator/pull/953/commits/81e18f2c048465b55e1214cff27ff0d6727dc741 , where I'm checking for an existing `secretRef` and if not found, because of cache delay or secret doesn't exist, I'm now trying to recreate it with the same name, otherwise it creates a new secret using the `generateName` field.

controllers/spacerequest/spacerequest_controller.go
```diff
@@ -500,6 +445,107 @@ func (r *Reconciler) ensureSecretForProvisionedNamespaces(ctx context.Context, m
...
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(spaceRequest, kubeConfigSecret, r.Scheme); err != nil {
+		return nil, errs.Wrap(err, "error setting controller reference for secret "+kubeConfigSecret.Name)
+	}
+	if err := memberClusterWithSpaceRequest.Client.Create(ctx, kubeConfigSecret); err != nil {
+		return nil, errs.Wrap(err, "error while creating secret")
+	}
+	log.FromContext(ctx).Info("Created Secret", "Name", kubeConfigSecret.Name)
```
rajivnathan, line 515: If it's the case when `GenerateName` is used, does this line log the secret's name or is `kubeConfigSecret.Name` empty?

controllers/spacerequest/spacerequest_controller_test.go
```diff
@@ -390,6 +325,172 @@ func TestCreateSpaceRequest(t *testing.T) {
...
+			// spacerequest exists with expected cluster roles and finalizer
+			spacerequesttest.AssertThatSpaceRequest(t, srNamespace.Name, sr.GetName(), member1.Client).
+				HasSpecTargetClusterRoles(srClusterRoles).
+				HasConditions(spacetest.Ready()).
+				HasStatusTargetClusterURL(member1.APIEndpoint).
+				HasNamespaceAccess([]toolchainv1alpha1.NamespaceAccess{{Name: "jane-env", SecretRef: "jane-xyz1"}}). // secret is the first one created
+				HasFinalizer()
+			// check that the secret is still there
+			secret := &corev1.Secret{}
+			err = member1Client.Get(context.TODO(), types.NamespacedName{Name: "jane-xyz1", Namespace: spaceRequest.GetNamespace()}, secret)
```
rajivnathan, line 425: Would it be worth verifying an additional secret is not created by doing a `List` and ensuring there is only 1 secret for the spacerequest?

rajivnathan (commented):
Nice 👍 , just a couple minor comments.

controllers/spacerequest/spacerequest_controller.go
mfrancisc: I'm almost sure that after the `Create` is executed the `Name` field will be populated even when using `GenerateName`.

controllers/spacerequest/spacerequest_controller_test.go
mfrancisc: Good point, I've replaced all the `Get` call with `List` to ensure that in https://github.com/codeready-toolchain/host-operator/pull/953/commits/63196275b0ccfb1cd120cff481a4219922af4c29.

I've also added a subtest to verify that additional reconciles of the same spacerequest will not create additional secret in https://github.com/codeready-toolchain/host-operator/pull/953/commits/4db1e8573229228b9f94cb88f20db942595209d6

filariow (approved):
lgtm

controllers/spacerequest/spacerequest_controller.go
rajivnathan: I knew that the created resource would have the `Name` field populated but I wasn't sure whether the `kubeConfigSecret` reference would have it by the time you logged it but looks like it does. 👍

PR #963, opened 2024-01-19
KSPACE-28: Implement the basic version of the SpaceProvisionerConfig controller.

metlos:
This only checks validity of the CR but doesn't integrate it into the space provisioning flow in any way.

Issue link: https://issues.redhat.com/browse/KSPACE-28
Associated toolchain-e2e PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/883
Associated toolchain-common PR: https://github.com/codeready-toolchain/toolchain-common/pull/357

### DIFF
--- a/controllers/spaceprovisionerconfig/mapper_test.go
+++ b/controllers/spaceprovisionerconfig/mapper_test.go
@@ -26,9 +26,9 @@ func TestFindingReferencingSpaceProvisionerConfigs(t *testing.T) {
 		ExpectedRequests     []reconcile.Request
 	}
 
-	spc0 := testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, "cluster1")
-	spc1 := testSpc.NewSpaceProvisionerConfig("spc1", test.HostOperatorNs, "cluster2")
-	spc2 := testSpc.NewSpaceProvisionerConfig("spc2", test.HostOperatorNs, "cluster1")
+	spc0 := testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1"))
+	spc1 := testSpc.NewSpaceProvisionerConfig("spc1", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster2"))
+	spc2 := testSpc.NewSpaceProvisionerConfig("spc2", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1"))
 
 	tests := map[string]testCase{
 		"find 1 referencing SpaceProvisionerConfig in many": {
@@ -66,7 +66,7 @@ func TestFindingReferencingSpaceProvisionerConfigs(t *testing.T) {
 
 	t.Run("empty references when listing space provisioner configs fails", func(t *testing.T) {
 		// given
-		cl := test.NewFakeClient(t, testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, "cluster1"))
+		cl := test.NewFakeClient(t, testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1")))
 		expectedErr := errors.New("expected list error")
 		cl.MockList = func(ctx context.Context, list runtimeclient.ObjectList, opts ...runtimeclient.ListOption) error {
 			if _, ok := list.(*toolchainv1alpha1.SpaceProvisionerConfigList); ok {
@@ -86,14 +86,14 @@ func TestFindingReferencingSpaceProvisionerConfigs(t *testing.T) {
 
 func TestMapToolchainClusterToSpaceProvisionerConfigs(t *testing.T) {
 	t.Run("maps existing spcs", func(t *testing.T) {
-		// this pretty much duplicates the test of finding the referencing space provisioner configs above but does it from
-		// the POV of the higher level function. Therefore, we don't need to replicate all the positive cases here, just
-		// one should be enough.
+		// this pretty much duplicates the test of finding the referencing space provisioner configs but does it from the POV
+		// of the higher level function. Therefore, we don't need to replicate all the positive cases here, just one should
+		//
 
 		// given
-		spc0 := testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, "cluster1")
-		spc1 := testSpc.NewSpaceProvisionerConfig("spc1", test.HostOperatorNs, "cluster2")
-		spc2 := testSpc.NewSpaceProvisionerConfig("spc2", test.HostOperatorNs, "cluster1")
+		spc0 := testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1"))
+		spc1 := testSpc.NewSpaceProvisionerConfig("spc1", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster2"))
+		spc2 := testSpc.NewSpaceProvisionerConfig("spc2", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1"))
 		cl := test.NewFakeClient(t, spc0, spc1, spc2)
 
 		// when
@@ -113,7 +113,7 @@ func TestMapToolchainClusterToSpaceProvisionerConfigs(t *testing.T) {
 
 	t.Run("interprets errors as empty result", func(t *testing.T) {
 		// given
-		cl := test.NewFakeClient(t, testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, "cluster1"))
+		cl := test.NewFakeClient(t, testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, testSpc.ReferencingToolchainCluster("cluster1")))
 		expectedErr := errors.New("expected list error")
 		cl.MockList = func(ctx context.Context, list runtimeclient.ObjectList, opts ...runtimeclient.ListOption) error {
 			if _, ok := list.(*toolchainv1alpha1.SpaceProvisionerConfigList); ok {
--- a/controllers/spaceprovisionerconfig/mapper_test.go
+++ b/controllers/spaceprovisionerconfig/mapper_test.go
@@ -7,73 +7,66 @@ import (
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
-	"github.com/stretchr/testify/assert"
+	testSpc "github.com/codeready-toolchain/toolchain-common/pkg/test/spaceprovisionerconfig"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func TestFindingReferencedSpaceProvisionerConfigs(t *testing.T) {
-	t.Run("find referenced provisioner configs", func(t *testing.T) {
-		// given
-		cl := test.NewFakeClient(t, &toolchainv1alpha1.SpaceProvisionerConfig{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "spc0",
-				Namespace: test.HostOperatorNs,
+func requestFromObject(obj runtimeclient.Object) reconcile.Request {
+	return reconcile.Request{
+		NamespacedName: runtimeclient.ObjectKeyFromObject(obj),
+	}
+}
+
+func TestFindingReferencingSpaceProvisionerConfigs(t *testing.T) {
+	type testCase struct {
+		ToolchainClusterName string
+		ExpectedRequests     []reconcile.Request
+	}
+
+	spc0 := testSpc.NewSpaceProvisionerConfig("spc0", test.HostOperatorNs, "cluster1")
+	spc1 := testSpc.NewSpaceProvisionerConfig("spc1", test.HostOperatorNs, "cluster2")
+	spc2 := testSpc.NewSpaceProvisionerConfig("spc2", test.HostOperatorNs, "cluster1")
+
+	tests := map[string]testCase{
+		"find 1 referencing SpaceProvisionerConfig in many": {
+			ToolchainClusterName: "cluster2",
+			ExpectedRequests: []reconcile.Request{
+				requestFromObject(spc1),
 			},
-			Spec: toolchainv1alpha1.SpaceProvisionerConfigSpec{
-				ToolchainCluster: "cluster1",
+		},
+		"find more referencing SpaceProvisionerConfigs in many": {
+			ToolchainClusterName: "cluster1",
+			ExpectedRequests: []reconcile.Request{
+				requestFromObject(spc0),
+				requestFromObject(spc2),
 			},
-		}, &toolchainv1alpha1.SpaceProvisionerConfig{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "spc1",
-				Namespace: test.HostOperatorNs,
-			},
-			Spec: toolchainv1alpha1.SpaceProvisionerConfigSpec{
-				ToolchainCluster: "cluster2",
-			},
-		}, &toolchainv1alpha1.SpaceProvisionerConfig{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "spc2",
-				Namespace: test.HostOperatorNs,
-			},
-			Spec: toolchainv1alpha1.SpaceProvisionerConfigSpec{
-				ToolchainCluster: "cluster1",
-			},
+		},
+		"find no referencing SpaceProvisionerConfigs in many": {
+			ToolchainClusterName: "cluster3",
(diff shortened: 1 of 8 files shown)


controllers/spaceprovisionerconfig/mapper_test.go
```diff
@@ -0,0 +1,63 @@
...
+			Name:      "spc2",
+			Namespace: test.HostOperatorNs,
+		},
+		Spec: toolchainv1alpha1.SpaceProvisionerConfigSpec{
+			ToolchainCluster: "cluster1",
+		},
+	})
+
+	// when
+	reqs, err := findReferencingProvisionerConfigs(context.TODO(), cl, runtimeclient.ObjectKey{Name: "cluster1", Namespace: test.HostOperatorNs})
```
MatousJobanek: Let's test 4 cases:
1. when there is one match
2. when there are multiple matches
3. when there is no match
4. when the list fails

we can theoretically join 1. and 2. together, but if it's not a big deal, then it's nice to have it separate.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// check that there exists a ToolchainCluster CR in the same namespace
+	toolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
+	toolchainClusterKey := runtimeclient.ObjectKey{Name: spaceProvisionerConfig.Spec.ToolchainCluster, Namespace: spaceProvisionerConfig.Namespace}
+	toolchainPresent := corev1.ConditionTrue
+	toolchainPresenceReason := toolchainv1alpha1.SpaceProvisionerConfigValidReason
+	if err := r.Client.Get(ctx, toolchainClusterKey, toolchainCluster); err != nil {
+		if !errors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("failed to get ToolchainCluster: %w", err)
+		}
```
MatousJobanek, line 81: Let's set the status when it fails for any type of error. We don't know what it failed and it's always good to have the visibility in the status of the CR. 
in this case, we can also include the content of the error as the message in the condition - something like this:
```go
	message := ""
	if err := r.Client.Get(ctx, toolchainClusterKey, toolchainCluster); err != nil {
		toolchainPresenceReason = toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotFoundReason
		toolchainPresent = corev1.ConditionFalse
		message = err.Error()
...
		toolchainv1alpha1.Condition{
			Type:    toolchainv1alpha1.ConditionReady,
			Status:  toolchainPresent,
			Reason:  toolchainPresenceReason,
			Message: message,
		})
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -0,0 +1,98 @@
...
+
+	spaceProvisionerConfig.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(spaceProvisionerConfig.Status.Conditions,
+		toolchainv1alpha1.Condition{
+			Type:   toolchainv1alpha1.ConditionReady,
+			Status: toolchainPresent,
+			Reason: toolchainPresenceReason,
+		})
+
+	if err := r.Client.Status().Update(ctx, spaceProvisionerConfig); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update the SpaceProvisionerConfig status: %w", err)
```
MatousJobanek, line 94: I would rather use `Wrap` function:
```go
errs "github.com/pkg/errors"

errs.Wrapf(err, "failed to update the SpaceProvisionerConfig status")
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get SpaceProvisionerConfig: %w", err)
+	}
+
+	if util.IsBeingDeleted(spaceProvisionerConfig) {
+		logger.Info("SpaceProvisionerConfig is being deleted - skipping reconciliation")
+		return ctrl.Result{}, nil
+	}
+
+	// check that there exists a ToolchainCluster CR in the same namespace
```
MatousJobanek, line 73:
```suggestion
	// check that there is a ToolchainCluster CR in the same namespace
```

controllers/spaceprovisionerconfig/mapper_test.go
```diff
@@ -0,0 +1,63 @@
...
+func TestToolchainClusterToSpaceProvisionerConfigMapper(t *testing.T) {
+	// given
+	cl := test.NewFakeClient(t, &toolchainv1alpha1.SpaceProvisionerConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "spc0",
+			Namespace: test.HostOperatorNs,
+		},
+		Spec: toolchainv1alpha1.SpaceProvisionerConfigSpec{
+			ToolchainCluster: "cluster1",
+		},
```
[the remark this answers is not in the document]
MatousJobanek: btw, this can be moved to toolchain-common so it can be re-used in toolchain-e2e tests if needed.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+
+		// then
+		assert.NoError(t, reconcileErr)
+		assert.Empty(t, spc.Status.Conditions)
+	})
+}
+
+func prepareReconcile(t *testing.T, spc *toolchainv1alpha1.SpaceProvisionerConfig, initObjs ...runtime.Object) (*Reconciler, reconcile.Request, *test.FakeClient) {
+	require.NoError(t, os.Setenv("WATCH_NAMESPACE", test.HostOperatorNs))
```
MatousJobanek, line 185: do we need this variable to be set for anything in these tests? it doesn't seem so
```suggestion
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		r, req, cl := prepareReconcile(t, spc)
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		getErr := cl.Get(context.TODO(), runtimeclient.ObjectKey{Name: "spc", Namespace: test.HostOperatorNs}, spc)
+
+		// then
+		require.NoError(t, reconcileErr)
+		require.NoError(t, getErr)
```
MatousJobanek, line 44: I would say that it's a good practice to check/handle the error immediately after the function/method call
```suggestion
		_, err := r.Reconcile(context.TODO(), req)

		// then
		require.NoError(t, err)
		err = cl.Get(context.TODO(), runtimeclient.ObjectKey{Name: "spc", Namespace: test.HostOperatorNs}, spc)
		require.NoError(t, err)
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+		r, req, cl := prepareReconcile(t, spc, &toolchainv1alpha1.ToolchainCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "cluster1",
+				Namespace: test.HostOperatorNs,
+			},
+		})
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		getErr := cl.Get(context.TODO(), runtimeclient.ObjectKey{Name: "spc", Namespace: test.HostOperatorNs}, spc)
```
MatousJobanek, line 68: the get calls can be simplified
```suggestion
		getErr := cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(spc), spc)
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+		r, req, cl := prepareReconcile(t, spc)
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		getErr := cl.Get(context.TODO(), runtimeclient.ObjectKey{Name: "spc", Namespace: test.HostOperatorNs}, spc)
+
+		// then
+		require.NoError(t, reconcileErr)
+		require.NoError(t, getErr)
+		assert.True(t, condition.IsFalse(spc.Status.Conditions, toolchainv1alpha1.ConditionReady))
```
MatousJobanek, line 45: I know that it's just a detail, mainly because we set only one "false reason" in the condition, but we could improve the validation so it checks also the capacity of the cluster in the future and set corresponding reasons based on the current state.
That being said, could you check also that the

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/stretchr/testify/require"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestSpaceProvisionerConfigValidation(t *testing.T) {
```
MatousJobanek, line 24: could you please add a test that checks the updates of the conditions:
* is true as part of the setup -> changes to false because the ToolchainCluster is gone
* is false as part of the setup -> changes to true because the ToolchainCluster was created

MatousJobanek (commented):
awesome, great job :+1: 
I left a few comments, but nothing critical, they are mostly refactoring proposals or test-related asks :smile:

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
metlos: `fmt.Errorf` is recommended in the `errors` package documentation (https://pkg.go.dev/errors@go1.21.6#pkg-overview). What are the advantages of using https://github.com/pkg/errors (btw that repo is archived and therefore, I assume, no longer maintained)?

MatousJobanek: whoops, good catch :grimacing: :man_facepalming: 
yup, let's keep the `fmt.Errorf` there then
sorry for confusion

controllers/spaceprovisionerconfig/mapper_test.go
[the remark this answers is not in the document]
metlos: https://github.com/codeready-toolchain/toolchain-common/pull/357

controllers/spaceprovisionerconfig/mapper_test.go
metlos: done

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
metlos: done, but please take a look, because I did it slightly differently to what you proposed here.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
metlos: should be done

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
metlos: Not sure what you suggest here but I expanded the tests to fully test values of all relevant fields of the condition.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
MatousJobanek: it's always better to test more things in unit tests than in e2e tests. 
As for the setup, what we do is that we create the object in the expected state (with the condition already set) as part of the "// given" section:
```go
spc := testSpc.NewSpaceProvisionerConfig("spc", test.HostOperatorNs, "cluster1", WithReadyCondition())
r, req, cl := prepareReconcile(t, spc, ...
```
or
```go
spc := testSpc.NewSpaceProvisionerConfig("spc", test.HostOperatorNs, "cluster1", WithUnreadyCondition("someReason", "some message"))
r, req, cl := prepareReconcile(t, spc, ...
```

in this way, you can easily verify the scenario in an isolated way. You have the CR in the given state -> what happens if under the given conditions during the reconcile.

controllers/spaceprovisionerconfig/mapper_test.go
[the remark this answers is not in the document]
MatousJobanek: no, there is no pairing at the level of go dependencies. 
what you would do is to add `replace`  clause into the go.mod file to point to the fork.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
MatousJobanek: created issue for changing that in the rest of our code https://issues.redhat.com/browse/KSPACE-32

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
metlos: The unit tests are now implemented.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -0,0 +1,233 @@
...
+		spc := NewSpaceProvisionerConfig("spc", test.HostOperatorNs, ReferencingToolchainCluster("non-existent"))
+		r, req, cl := prepareReconcile(t, spc)
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(spc), spc))
+
+		// then
+		assert.NoError(t, reconcileErr)
+		AssertThat(t, spc, Is(NotReadyWithReason(toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotFoundReason)))
```
MatousJobanek, line 36: related to my comment in the other PR:
https://github.com/codeready-toolchain/toolchain-common/pull/357#discussion_r1471088952
I just tried what would happen if I changed the test to
```
AssertThat(t, spc, Is(Ready()))
```
I know that it's wrong, but I wanted to know what the error would look like, where it actually fails and how easy it is to debug. The error looks like this:
```
--- FAIL: TestSpaceProvisionerConfigValidation (0.02s)
    --- FAIL: TestSpaceProvisionerConfigValidation/is_not_valid_when_non-existing_ToolchainCluster_is_referenced (0.01s)
        assertions.go:27: 
            	Error Trace:	/home/mjobanek/go-workspace/pkg/mod/github.com/metlos/toolchain-common@v0.0.0-20240125174611-654ac7c7e9a7/pkg/test/assertions/assertions.go:27
            	            				/home/mjobanek/go-workspace/src/github.com/codeready-toolchain/host-operator/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go:36
            	Error:      	Should be true
            	Test:       	TestSpaceProvisionerConfigValidation/is_not_valid_when_non-existing_ToolchainCluster_is_referenced
```
where the `assertions.go:27` line points to:
```
func AssertThat(t *testing.T, object client.Object, predicate Predicate[client.Object]) {
	assert.True(t, predicate.Matches(object))
}
```
so it's very hard to understand what was expected and what was the actual value.
TBH, it's fine in this particular case since there is only one expectation that wasn't met, but we should be very careful as soon as we have multiple assertions.

metlos: Excellent point. I think the solution here is to mark `AssertThat` as a helper method so that it is skipped when reporting the location of the error. Here's the PR: https://github.com/codeready-toolchain/toolchain-common/pull/360

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -0,0 +1,111 @@
...
+			Status:  toolchainPresent,
+			Reason:  toolchainPresenceReason,
+			Message: toolchainPresenceMessage,
+		})
+
+	if err := r.Client.Status().Update(ctx, spaceProvisionerConfig); err != nil {
+		if reportedError != nil {
+			logger.Info("failed to update the status (reported as failed reconciliation) with a previous unreported error during reconciliation", "unreportedError", reportedError)
+		}
+		reportedError = fmt.Errorf("failed to update the SpaceProvisionerConfig status: %w", err)
```
MatousJobanek, line 107: Thanks for changing this.
In most of the other controllers, we have some helper functions that help us with dealing with status updates and the error hanging
eg: 
https://github.com/codeready-toolchain/host-operator/blob/06236ebe82a9d4150226ac179b1f4bbdc1a0324f/controllers/space/space_controller.go#L672-L686

But it's totally fine as you did - it's a simple controller. If it would get more complicated, then we could extract the logic.

TBH, writing more generic helper functions for the status updates that could be used for all toolchain CRDs has been my plan for a long time already.

MatousJobanek (approved):
Awesome :+1: :rocket:

metlos:
/retest
new related changes in toolchain-e2e

PR #965, opened 2024-02-06
Make the idler template generic

rajivnathan:
This way the template can be reused when VMs are idled too.

https://issues.redhat.com/browse/SANDBOX-513


alexeykazakov (approved):

ranakan19 (approved):
LGTM

PR #966, opened 2024-02-07
reduce readiness failure threshold for registration service

mfrancisc:
Reduce number of errors:

```
{"level":"error","ts":"2024-02-07T02:33:06.081Z","logger":"registration-service","msg":"http: Server closed","timestamp":"Wed, 07 Feb 2024 02:33:06 +0000","commit":"b4a49ca","error":"http: Server closed","stacktrace":"github.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Errorf\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:183\ngithub.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Error\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:169\ngithub.com/codeready-toolchain/registration-service/pkg/log.Error\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:111\nmain.main.func2\n\t/home/runner/work/registration-service/registration-service/cmd/main.go:142"}
```

the readiness probe will fail earlier removing the service from load balancing and thus reducing the number of those errors.

Related to: https://github.com/codeready-toolchain/registration-service/pull/401

Jira: https://issues.redhat.com/browse/SANDBOX-521

NOTE: this will NOT solve the other error
```
unable to parse readiness status as bool: : strconv.ParseBool: parsing \"\": invalid syntax
```


metlos (commented):
The doc of the go's http server's `Shutdown` method says that as soon as it is called, the calls to `ListenAndServe` and other "serving" methods immediately return with the `http.ErrServerClosed` error, which I believe is the one we're seeing here.

So IMHO we're just reporting an error that is actually an expected behavior.

While the change to the readiness probe will make the registration service disappear faster from the load-balancer, as PR description says, IMHO the error message will still pop up in the logs.

The fix should IMHO be to just not log the error if it is the `ErrServerClosed`. I.e. change https://github.com/codeready-toolchain/registration-service/blob/master/cmd/main.go#L141

```diff
- if err := srv.HTTPServer().ListenAndServe(); err != nil {
+ if err := srv.HTTPServer().ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```

deploy/registration-service/registration-service.yaml
```diff
@@ -114,7 +114,7 @@ objects:
                 successThreshold: 1
                 timeoutSeconds: 1
               readinessProbe:
-                failureThreshold: 30
+                failureThreshold: 3
```
alexeykazakov, line 117:
```suggestion
                failureThreshold: 1
```

alexeykazakov (commented):
Let's set it to 1.
Also let's ignore the error as @metlos suggesting.

Please remember we can't merge this PR until we fix the host operator upgrade.

mfrancisc:
> The fix should IMHO be to just not log the error if it is the ErrServerClosed.

TBH I would prefer to keep logging all the errors , even if in this case is expected. By doing so we can confirm that the readiness works as expected, thus we should see `ErrServerClosed` logged maximum 3 times ( we have 3 replicas right now) at each registration service upgrade.

You can always filter out those errors when investigating the logs.

xcoulon:
> > The fix should IMHO be to just not log the error if it is the ErrServerClosed.
> 
> TBH I would prefer to keep logging all the errors , even if in this case is expected. By doing so we can confirm that the readiness works as expected, thus we should see `ErrServerClosed` logged maximum 3 times ( we have 3 replicas right now) at each registration service upgrade.
> 
> You can always filter out those errors when investigating the logs.

or you could check if `errors.Is(err, http.ErrServerClosed)` and print a friendly `info` msg saying that the server is closed or print the error otherwise?

mfrancisc:
> or you could check if errors.Is(err, http.ErrServerClosed) and print a friendly info msg saying that the server is closed or print the error otherwise?

I don't have a strong opinion on this, we can print it as `Info` if it really bothers us to see it as `Error` , but again I don't see the advantage of considering this as a special case when we can instead just filter the logs if needed. Also handling special cases might require changes in the future if the http library internal changes (for example they drop or change this type of error ).

metlos:
> > or you could check if errors.Is(err, http.ErrServerClosed) and print a friendly info msg saying that the server is closed or print the error otherwise?
> 
> I don't have a strong opinion on this, we can print it as `Info` if it really bothers us to see it as `Error` , but again I don't see the advantage of considering this as a special case when we can instead just filter the logs if needed. Also handling special cases might require changes in the future if the http library internal changes (for example they drop or change this type of error ).

Returning the `http.ErrServerClosed` is part of the documented behavior of `ListenAndServe` so changing that would basically be changing the API, which I don't think has much probability of happening. If it did change, there would either be many more changes coming along with it or there would have to be some severe security issue that would require this to change. We'd have to react to both of those anyway.

That said, we DO log the result of the server shutdown in the registration-service here: https://github.com/codeready-toolchain/registration-service/blob/master/cmd/main.go#L177 so that's why I suggested to not log anything in the `ListenAndServe` because it basically just reacts to the shutdown by returning `http.ErrServerClosed`.

But to be on the safe side and log everything I am for logging an info message as a reaction to `http.ErrServerClosed`.

(Also, sorry for starting a discussion about registration-service on the related PR of the host-operator - I got confused)

mfrancisc:
> But to be on the safe side and log everything I am for logging an info message as a reaction to http.ErrServerClosed.

I'm now logging this specific error as Info here https://github.com/codeready-toolchain/registration-service/pull/401/commits/e527590e0c11d8834f89714783dbc98470c37492. Thanks for the suggestions!

rajivnathan (approved):
Nice 👍

alexeykazakov (approved):

PR #974, opened 2024-02-27
Create templates with host ToolchainCluster-related manifests

mfrancisc:
Add SA, Role and RoleBinding for the `toolchaincluster` resource.

See: https://github.com/codeready-toolchain/toolchain-cicd/blob/34ea85fe21ae05c9bb1a0a2bbdc986b94c8d9e92/scripts/add-cluster.sh#L36

Jira: https://issues.redhat.com/browse/KSPACE-7

Related PRs:
- common: https://github.com/codeready-toolchain/toolchain-common/pull/365
- member: https://github.com/codeready-toolchain/member-operator/pull/537


pkg/templates/toolchaincluster/template.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, err
+	}
+	if gkv.Kind == "Role" {
+		return obj.(*rbac.Role), nil
+	}
+	return nil, nil
+}
+
+// GetRoleBindingFromTemplate returns the toolchaincluster RoleBinding object from the template content
+func GetRoleBindingFromTemplate(toolchainclusterFS *embed.FS, templateName string, into runtime.Object) (*rbac.RoleBinding, error) {
```
MatousJobanek, line 38: why do we actually have to care about the particular resources? Couldn't we just simply apply all resources no matter of their kinds? 
The only thing we have to rely on is the presence of the ServiceAccount to be able to get the needed token, but that's all.

deploy/templates/toolchaincluster/host-role.yaml
```diff
@@ -0,0 +1,25 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: toolchaincluster-host
+  namespace: toolchain-host-operator
```
MatousJobanek, line 5: The namespace shouldn't be hardcoded - it should be set based on where the host operator is running.

pkg/templates/toolchaincluster/template.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil, nil
+}
+
+func decodeTemplate(toolchainclusterFS *embed.FS, templateName string, into runtime.Object) (runtime.Object, *schema.GroupVersionKind, error) {
+	templateContent, err := toolchainclusterFS.ReadFile("templates/toolchaincluster/" + templateName)
+	if err != nil {
+		return nil, nil, err
+	}
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, gkv, err := decode(templateContent, nil, into)
```
MatousJobanek, line 55: with this we technically rely on the fact that there is one manifest per file and that the file name matches the expected kind. Have you considered using kustomize and building the set of objects from it?

deploy/templates/toolchaincluster/host-role.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - "spaces"
+  - "spacebindings"
+  - "tiertemplates"
+  - "toolchainconfigs"
+  - "toolchainclusters"
+  - "toolchainstatuses"
+  - "usersignups"
+  - "usertiers"
+  - "proxyplugins"
+  - "spaceprovisionerconfigs"
```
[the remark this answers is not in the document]
mfrancisc: makes sense. Thanks!

pkg/templates/toolchaincluster/template.go
mfrancisc: > why do we actually have to care about the particular resources? Couldn't we just simply apply all resources no matter of their kinds?

I'm not sure I'm following here, this logic is just converting the manifest yaml file that is embeded in the binary to a Golang type ( in this case is a RoleBinding ) . We will need the logic ( a controller I guess ) that will use all those functions and will actually apply/create those objects in the cluster. So it will need to know the kinds of the objects IMO.

mfrancisc: The alternative would be to just read the file content and apply it as is, without any sort of content validation, which in this case is done by checking the type of object we expect in the manifest.

deploy/templates/toolchaincluster/host-role.yaml
mfrancisc: good point. Done https://github.com/codeready-toolchain/host-operator/pull/974/commits/7965e579ad772c35f6cbca874055e15d329729d6 Thanks!

pkg/templates/toolchaincluster/template.go
mfrancisc: Also do we really think we will have other objects to deploy other than Role, RoleBinding and SA ? maybe ClusterRole and ClusterRoleBinding.

pkg/templates/toolchaincluster/template.go
mfrancisc: Do you have an example of using kustomize with embedded files. From what I know we use kustomize to generate static manifests at build time, not runtime. But I'm missing something for sure.

mfrancisc: Or you are suggesting to keep all objects in one file and generate the individual files at build time ?

pkg/templates/toolchaincluster/template.go
rajivnathan: If by content validation you're referring to decoding the template to get the object and gkv then I think that's only validating that it's a runtime object which is still generic? So that can be done the same way for all the objects and they can still be created/updated in a generic way.

I'm not sure that casting the object is equivalent to validating the object. If the template has some field that doesn't exist in the type then does it return an error or it's just ignored?

mfrancisc: > I'm not sure that casting the object is equivalent to validating the object. If the template has some field that doesn't exist in the type then does it return an error or it's just ignored?

That's true, it get's ignored, but you can unit test and check that required/expected fields are there. With generic types ( I believe you are referring to `runtime.Object` ) you can't check that as far as I know.

MatousJobanek: You can use `unstructured.Unstructured{}` instead, there is 
```
runtime.DefaultUnstructuredConverter.FromUnstructured
```
that should convert the type into the specific object if needed

pkg/templates/toolchaincluster/template.go
MatousJobanek: yeah, you are right with the kustomize, I did not realize it, not sure if they support bundled files, but never mind.
And yeah, the possibility of putting more objects in one file would be nice.
I didn't test it but something like that should work
```go
decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(templateContent), 100)
var objects []*unstructured.Unstructured
for {
	var rawExt runtime.RawExtension
	if err := decoder.Decode(&rawExt); err != nil {
		if err == io.EOF {
			break
		}
		return nil, nil, err
	}
	rawExt.Raw = bytes.TrimSpace(rawExt.Raw)
	if len(rawExt.Raw) == 0 || bytes.Equal(rawExt.Raw, []byte("null")) {
		continue
	}
	unstructuredObj := &unstructured.Unstructured{}
	_, _, err = scheme.Codecs.UniversalDeserializer().Decode(rawExt.Raw, nil, unstructuredObj)
	if err != nil {
		return nil, nil, err
	}
	objects = append(objects, unstructuredObj)
}
```

MatousJobanek: and btw, with this approach we don't need to rely on the filenames matching the kinds anymore - the logic can just read all the files from the folder and collect all objects that are defined in the files.

deploy/templates/toolchaincluster/host-role.yaml
MatousJobanek: cool, but couldn't this be set via a go-template variable? wouldn't it be a bit more intuitive?

pkg/templates/toolchaincluster/template.go
mfrancisc: Nice, I was not aware of this usage of `runtime.RawExtension`. Thanks!

I've implemented this and moved all the files in the same template. I've updated also the logic to scan recursively for all files in the given directory.

deploy/templates/toolchaincluster/host-role.yaml
mfrancisc: Makes sense, go-template variables are now supported. For now just `.Namespace`.

pkg/templates/toolchaincluster/template_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"testing"
+
+	"github.com/codeready-toolchain/host-operator/deploy"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadObjectsFromTemplates(t *testing.T) {
```
MatousJobanek, line 11: It would be nice to have some sanity check that all variables were properly set. 
If you move the function into toolchain-common, then you could use some dummy template and check the whole content of the decoded manifests.

pkg/templates/toolchaincluster/template.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// TemplateVariables contains all the available variables that are supported by the templates
+type TemplateVariables struct {
+	Namespace string
+}
+
+// LoadObjectsFromTemplates loads all the kubernetes objects from an embedded filesystem and returns a list of Unstructured objects that can be applied in the cluster.
+// The function will return all the objects it finds starting from the root of the embedded filesystem.
+func LoadObjectsFromTemplates(toolchainclusterFS *embed.FS, variables *TemplateVariables) ([]*unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
mfrancisc: Yes , I want to move it later to toolchain-common as soon as I start working at the member operator PR. My idea was to do it later in a follow up PR. WDYT?

pkg/templates/toolchaincluster/template_test.go
mfrancisc: Yes I can add some more tests using dummy templates once it's in toolchain-common.

mfrancisc: I've moved the functionality here: https://github.com/codeready-toolchain/toolchain-common/pull/365 as is. I've only renamed the function and added more tests. 

I'll remove the code from this PR, and leave only the templates.

MatousJobanek (approved):

PR #975, opened 2024-02-29
Update User Cleanup controller to be useful

sbryzak:
In order to keep our total UserSignups under control, this PR updates the UserCleanupController so that it will delete users that haven't been activated in the last 2 years, that only have one total activation and that aren't banned.

Fixes: https://issues.redhat.com/browse/SANDBOX-547

Note: the 2 years must be configured in the host operator configuration.

### DIFF
--- a/controllers/usersignupcleanup/usersignup_cleanup_controller.go
+++ b/controllers/usersignupcleanup/usersignup_cleanup_controller.go
@@ -25,6 +25,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	// MaxActivationsForCleanup is the maximum number of activations that a user may have in order to be considered
+	// for cleanup.  If the number of user activations exceeds this constant value, then cleanup will not occur.
+	MaxActivationsForCleanup = 1
+)
+
 type StatusUpdater func(userAcc *toolchainv1alpha1.UserSignup, message string) error
 
 // SetupWithManager sets up the controller with the Manager.
@@ -150,7 +156,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 				meetsActivationCriteria = true
 			} else {
 				// Otherwise it meets the criteria if we can determine there has only ever been one activation in total
-				meetsActivationCriteria = activationCount <= 1
+				meetsActivationCriteria = activationCount <= MaxActivationsForCleanup
 			}
 		}
 
@@ -205,16 +211,16 @@ func (r *Reconciler) isUserBanned(ctx context.Context, userSignup *toolchainv1al
 
 			hashIsValid := validateEmailHash(userSignup.Spec.IdentityClaims.Email, emailHashLbl)
 			if !hashIsValid {
-				err := fmt.Errorf("hash is invalid")
+				err := fmt.Errorf("email hash is invalid for UserSignup [%s]", userSignup.Name)
 				return banned, err
 			}
 		} else {
 			// If there isn't an email-hash label, then the state is invalid
-			err := fmt.Errorf("missing label at usersignup")
+			err := fmt.Errorf("missing email-hash label for UserSignup [%s]", userSignup.Name)
 			return banned, err
 		}
 	} else {
-		err := fmt.Errorf("missing email at usersignup")
+		err := fmt.Errorf("could not determine email address for UserSignup [%s]", userSignup.Name)
 		return banned, err
 	}
 	return banned, nil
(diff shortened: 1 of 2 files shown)


controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -136,7 +139,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			// Meets the criteria if there is no activations annotation
+			meetsActivationCriteria = true
+		} else {
+			activationCount, err := strconv.Atoi(activations)
+			if err != nil {
+				// If there was an error converting the activation count to an integer, then assume it meets the criteria for deletion
+				meetsActivationCriteria = true
+			} else {
+				// Otherwise it meets the criteria if we can determine there has only ever been one activation in total
+				meetsActivationCriteria = activationCount <= 1
```
alexeykazakov, line 153: Let's maybe move `1` to a const so it's easier to update in the feature. And down the road we might want to move it configuration. But for now it's fine to keep it as a const.

controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -155,6 +178,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			}
+
+			hashIsValid := validateEmailHash(userSignup.Spec.IdentityClaims.Email, emailHashLbl)
+			if !hashIsValid {
+				err := fmt.Errorf("hash is invalid")
+				return banned, err
+			}
+		} else {
+			// If there isn't an email-hash label, then the state is invalid
+			err := fmt.Errorf("missing label at usersignup")
```
alexeykazakov, line 213: Let's add the usersignup name to the error message.

controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -155,6 +178,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			return banned, err
+		}
+	} else {
+		err := fmt.Errorf("missing email at usersignup")
```
alexeykazakov, line 217: And here too.

alexeykazakov (commented):
Looks good overall. I have a few minor comments though. And any unit tests? :)

controllers/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Fixed

controllers/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Fixed

controllers/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Fixed

controllers/usersignupcleanup/usersignup_cleanup_controller.go
```diff
@@ -155,6 +184,52 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	return reconcile.Result{}, nil
 }
 
+func (r *Reconciler) isUserBanned(ctx context.Context, userSignup *toolchainv1alpha1.UserSignup) (bool, error) {
```
mfrancisc, line 187: I was wondering if checking the `Complete` condition would suffice, like we do in reg-service https://github.com/codeready-toolchain/registration-service/blob/1265c2318813f43cf0efa1c7bced960d1b0cffc1/pkg/signup/service/signup_service.go#L438-L445

or if there's anything different with host operator that we need to check also the banned resource and validate the email hash.

mfrancisc (commented):
Nice! I have one small comment only.

Was also wondering if we are going to need unit tests to validate this change?

controllers/usersignupcleanup/usersignup_cleanup_controller.go
sbryzak: Hey @mfrancisc , thanks for the review.  Unit tests have been added to the PR now.  We do a more thorough check to confirm that the user is banned in the host operator, as the BannedUser resource is the authorative source to determine whether a UserSignup is really banned.  Perhaps if the registration service is not checking this then it needs to be changed, it's been a long time since this code was written and I can't recall exactly why reg service was written this way.

alexeykazakov (commented):
Looks good! However it seems that there are a few potential errors which are not covered by tests. Could you add some tests for these if it's not too tricky?

alexeykazakov (approved):
Thanks for adding more tests! I did some test reorganization though just to reduce code duplication if you don't mind: https://github.com/codeready-toolchain/host-operator/pull/975/commits/bc27db23d09722426e0628cf56c441600ed402a6

PR #988, opened 2024-03-11
Modify DeactivationService to set scheduled deactivation time for active UserSignups

sbryzak:
This PR updates the deactivation controller so that it will manage the `UserSignup.Status.ScheduledDeactivationTimestamp` property, setting it appropriately as the controller is reconciled, also when the user tier changes, and when the usersignup is deactivated.

Related PR: https://github.com/codeready-toolchain/registration-service/pull/411


deploy/registration-service/registration-service.yaml
```diff
@@ -32,6 +32,7 @@ objects:
           - toolchain.dev.openshift.com
         resources:
           - masteruserrecords
+          - usertiers
```
alexeykazakov, line 35: You don't need to get tiers from reg-service with the new approach, do you?

sbryzak: Fixed

alexeykazakov (commented):
Looks good but I don't see any tests :)
And are you sure you want to combine the CRD/API update with the controller change? Usually we do it in separate PRs.

controllers/deactivation/deactivation_controller.go
```diff
@@ -199,6 +200,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		// It is not yet time to deactivate so requeue when it will be
 		requeueAfterExpired := time.Until(deactivationDueTime)
 
+		if usersignup.Status.ScheduledDeactivationTimestamp.IsZero() || !usersignup.Status.ScheduledDeactivationTimestamp.Time.Equal(deactivationDueTime) {
+			usersignup.Status.ScheduledDeactivationTimestamp = v1.NewTime(deactivationDueTime)
+			if err := r.Client.Status().Update(ctx, usersignup); err != nil {
```
MatousJobanek, line 205: Update of the UserSignup status should be managed by UserSignup controller, not by deactivation controller. Deactivation controller should manage only the spec of the UserSignup resource.

alexeykazakov: Ideally yes, but how do you suggest to make the UserSignup controller to set the `scheduledDeactivationTimestamp` status? I see two options:
- Rewrite our deactivation mechanism and introduce a new spec in UserSignup, which would tell the UserSignup controller when to deactivate instead of using our current "state" filed.
- Move the deactivation timestamp calculation from the deactivation controller to the UserSignup controller which would mean code duplication because the deactivation controller already does this job.

IMO adding the UserSignup status thing to the deactivation controller is a better option. Or you are thinking about something else?

MatousJobanek: We could have used an annotation instead. That would be totally OK, right?

MatousJobanek: as for the second option - which code duplication are you talking about? You can easily extract the logic that calculates the deactivation time based on the MUR.spec and call it from both places. In the case of usersignup, you could call it here: https://github.com/codeready-toolchain/host-operator/blob/188cb8ad8108505af07486656e3bbbdd8dd27615/controllers/usersignup/usersignup_controller.go#L382-L383
at this point you know that the user is approved, ready to be used, you have the latest MUR, and you know that when MUR is updated then you get to this line to reset the value in the status.

MatousJobanek: apart from that, the logic that is currently in the PR won't work - it would set the value only when the UserSignup is in the deactivating state, not when it's before the "deactivating" time (it calculates the `deactivationDueTime` based on the LastTransitionTime of the `UserSignupUserDeactivatingNotificationCreated` condition.
The controller exists earlier for the case when it's not in the "deactivating" state.

alexeykazakov: > apart from that, the logic that is currently in the PR won't work

This is why we need tests to prove that it works in all expected cases :)

I don't really like the annotation approach but we could extract the logic of calculating the deactivation timestamp indeed.

sbryzak: Logic for setting the scheduled deactivation time is now in the UserSignup controller.  I'm still uncertain whether we should be depending on this timestamp value in the deactivation controller itself, it currently recalculates everything independently and covers additional stuff like the deactivation notification calculation, so I'm tempted to just leave it as is.

alexeykazakov: The information about account deactivation is pretty critical. I don't like the idea of having two completely independent logics, one for the user's information and one for the actual deactivation. I like the idea of moving the calculation to the UserSignup controller while the deactivation controller uses this tiemstamp from the status instead of re-calculating it on it's own.

alexeykazakov: > Logic for setting the scheduled deactivation time is now in the UserSignup controller

I don't see it in this PR. Make sure you didn't forget to update the PR ;)

alexeykazakov (commented):
Shane, the current code in this PR is clearly broken. The deactivation controller doesn't check/calculate the expected deactivation timestamp and just deactivate the account right away.

No need to keep retraining to re-run e2e tests :) They will keep failing.

I really encourage you to add unit tests first which will catch most of the issues with the Pr before introducing e2e tests.

sbryzak:
@alexeykazakov yes it's broken (I forgot to restore a block of code in deactivation controller) but besides that I've recorded numerous e2e test failures which were occurring before I even pushed the changes to host operator.  I think there's still some serious test flakiness happening and I'm trying to pinpoint that, I have kept a bunch of log outputs which I will go through and summarise in a Google doc when I have the chance.

MatousJobanek:
@sbryzak Instead of creating a document could you please just push your work in the PRs? It's much easier to investigate the errors/flakiness when we see the changes in your PR and when we can take a look at the whole logs of the test execution and also of all pods.
Apart from that, I haven't seen any significant flakiness in other PRs, and it's surprising that it happens for your changes only. You maybe hit something that wasn't apparent before, but without being able to see the changes and the logs, it's almost impossible to investigate what is wrong.

controllers/deactivation/deactivation_controller.go
sbryzak: @alexeykazakov unfortunately it seems we have no choice about having two separate sources of logic for the deactivation.  If you recall, we were quite strict in our design with ensuring that the deactivation "countdown" only began once we notified the user that their account would be deactivated shortly, generally 3 days I believe before deactivation.  We contended with the possibility that for whatever reason, the user might not be notified due to unforeseen technical issues and due to this possibility their trial may be extended slightly as a result due to delayed issuing of their deactivation notification.  

If I recall correctly, this was to allow the user sufficient opportunity to make whatever preparations they deem necessary (backup or save whatever state they may have in their sandbox) before deactivation occurs.  This requirement unfortunately creates a situation where the actual deactivation time is at odds with the scheduled deactivation time which is calculated at the start of the trial period.  This is why I believe our best option might be to leave the scheduled deactivation time in the UserSignup as informational state only, and to continue having the deactivation controller calculate the actual deactivation time based on when the notification is created.

alexeykazakov: But in this case we will have incorrect deactivation timestamp in the UserSignup. If the deactivating notification was not created at least 3 days ago even if the deactivation is well past due then the deactivation timestamp will point to the past while the actual deactivation is still pending.
I looked at the deactivation controller more closely and I see there is many things which dictate when the user signup is actually going to be deactivated. Tier settings, deactivating notification status, if the user's email is in the list of the exceptions, etc. So I don't see how we can easily duplicate this logic in the UserSignup controller without effectively duplicating almost entire deactivation controller there. I also don't see any efficient way of extracting this logic to some common code which we could re-use in both controller.
So I think keeping this logic in the deactivation controller (which purpose is entirely to know when deactivation is supposed to happen) exclusively is the best approach. And having to update the UserSignup status (the new deactivation timestamp field) from the deactivation controller is relatively small price to pay.

sbryzak:
Damn you golangci-lint... team I'm going to need to refactor the reconciler in order to simplify it some more, will get this sorted out tomorrow.

controllers/deactivation/deactivation_controller.go
```diff
@@ -128,6 +130,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		if err := r.resetDeactivatingState(ctx, usersignup); err != nil {
 			return reconcile.Result{}, err
 		}
+		// Set the scheduled deactivation time to nil
```
alexeykazakov, line 133: Let's reset the scheduled time to nil also if the user's email is in the exclusion list (lines 104-111).

sbryzak: Done.

alexeykazakov: Looks good. You could move this set-to-nil code to a function though.

controllers/deactivation/deactivation_controller.go
```diff
@@ -164,6 +196,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	if !states.Deactivating(usersignup) {
 		states.SetDeactivating(usersignup, true)
 
+		// Before we update the UserSignup in order to set the deactivating state, we should reset the scheduled
+		// deactivation time if required just in case the current value is nil or has somehow changed.  Since the UserSignup
+		// controller is going to be reconciling immediately after setting the deactivating state then it will be
+		// creating a deactivating notification, meaning that the scheduled deactivation time that we set here is going
+		// to be extremely temporary (as it will be recalculated after the next reconcile), however it is done for correctness
+		scheduledDeactivationTime := (*mur.Status.ProvisionedTime).Add(deactivationTimeout)
```
alexeykazakov, line 204: Lines 204-210 seem to be 100% the same as lines 178-186. So it's a good candidate to move to a function.

controllers/deactivation/deactivation_controller.go
```diff
@@ -193,9 +240,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, nil
 	}
 
+	// We calculate the actual deactivation due time based on when the deactivating condition was set.  This may end up
+	// being significantly later than the scheduled deactivation time set in UserSignup.Status.ScheduledDeactivationTime, as
+	// in some rare circumstances the deactivating notification/status may fail to be set due to cluster downtime or
+	// other reasons.  Because of this, the scheduled deactivation time that is set in the UserSignup.Status should be
+	// treated as informational only.
 	deactivationDueTime := deactivatingCondition.LastTransitionTime.Time.Add(time.Duration(deactivatingNotificationDays*24) * time.Hour)
```
alexeykazakov, line 248: We actually should use this `deactivationDueTime` above before returning `reconcile.Result{}, nil` in line 240 as the scheduled deactivation time.
So if the actual deactivation is postponed because the deactivating notification condition is not `true` yet (so the notification has not been created yet) then we need to update the scheduled time too.

alexeykazakov (commented):
Looks good! I have a few minor comments though.

controllers/deactivation/deactivation_controller.go
sbryzak: I'm a bit confused by this one - if the deactivating condition is not found, or it is not in an expected state then it cannot be used as a basis for calculating the scheduled deactivation time.  In this case, the only other valid way of calculating it is based on the provisioned time, which we already do above.  Or am I missing something here?

controllers/deactivation/deactivation_controller.go
sbryzak: Done, this code has been refactored into a separate function.

controllers/deactivation/deactivation_controller.go
alexeykazakov: This is my thinking.
Let's say it's time to start the deactivation but the the deactivating condition is not set or is not in an expected state. Meaning we have to delay the deactivation. Something like this for example:
- Provisioning time: - 30 days (30 days ago).
- The original scheduled deactivation time: now (meaning 0 days left).
- The deactivating condition is still pending (there is an unexpected delay in the notification mechanism for example).
With the current code we will still show that there is no time left before the deactivation happens. While in the reality we won't start the deactivation for *at least* three more days. It maybe be longer than 3 days but it's impossible to know. So our best guess would be to use the same `now`  + `deactivatingNotificationDays` (which is currently 3 days).

controllers/deactivation/deactivation_controller.go
alexeykazakov: It looks like you forgot to commit it. I don't see any changes :)

sbryzak: oops, committed

rsoaresd:
Hello @sbryzak 👋

The last 3 runs failed with merge conflict in `go.mod` and `go.sum`

controllers/deactivation/deactivation_controller.go
```diff
@@ -227,14 +228,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		// timestamp plus the number of pre-deactivation days configured in the settings, BUT only if it is currently nil
+		// OR the calculated value is more than 24 hours later than the current value
 		deactivationDueTime := time.Now().Add(time.Duration(deactivatingNotificationDays) * 24 * time.Hour)
-		ts := v1.NewTime(deactivationDueTime)
-		usersignup.Status.ScheduledDeactivationTimestamp = &ts
-		if err := r.Client.Status().Update(ctx, usersignup); err != nil {
-			logger.Error(err, "failed to update usersignup status")
-			return reconcile.Result{}, err
+		if usersignup.Status.ScheduledDeactivationTimestamp == nil ||
+			usersignup.Status.ScheduledDeactivationTimestamp.Time.Before(deactivationDueTime.Add(-24*time.Hour)) {
```
alexeykazakov, line 235: As we discussed on the call it could be something shorter than 24h. 12h for example. But it's not a big deal if we use 24h.

controllers/deactivation/predicate.go
```diff
@@ -25,3 +29,58 @@ func (CreateAndUpdateOnlyPredicate) Delete(_ event.DeleteEvent) bool {
...
+	}
+	for _, c := range second {
+		if !ContainsCondition(first, c) {
+			return false
+		}
+	}
+	return true
+}
+
+// ContainsCondition returns true if the specified list of conditions contains the specified condition.
```
alexeykazakov, line 77:
```suggestion
// ContainsCondition returns true if the specified list of conditions contains the specified condition and the statuses of the conditions match.
```

controllers/deactivation/predicate.go
```diff
@@ -25,3 +29,58 @@ func (CreateAndUpdateOnlyPredicate) Delete(_ event.DeleteEvent) bool {
...
+			if !ConditionsMatch(objOld.Status.Conditions, objNew.Status.Conditions) {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func ConditionsMatch(first, second []toolchainv1alpha1.Condition) bool {
```
alexeykazakov, line 60: Looks like a good candidate to move to https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go

controllers/deactivation/predicate.go
```diff
@@ -25,3 +29,58 @@ func (CreateAndUpdateOnlyPredicate) Delete(_ event.DeleteEvent) bool {
 func (CreateAndUpdateOnlyPredicate) Generic(_ event.GenericEvent) bool {
 	return false
 }
+
+type GenerationOrConditionsChangedPredicate struct {
+	predicate.GenerationChangedPredicate
+}
+
+// Update implements default UpdateEvent filter for validating no generation change
```
alexeykazakov, line 37: Not only the generation change. It also takes into account the conditions.
Could you please fix the description and also add a description to `type GenerationOrConditionsChangedPredicate struct {` with an overall definition of what this predicate does?

alexeykazakov (commented):
Looks good overall. I have a few suggestions though.

controllers/deactivation/predicate.go
```diff
@@ -25,3 +30,30 @@ func (CreateAndUpdateOnlyPredicate) Delete(_ event.DeleteEvent) bool {
 func (CreateAndUpdateOnlyPredicate) Generic(_ event.GenericEvent) bool {
 	return false
 }
+
+type GenerationOrConditionsChangedPredicate struct {
+	predicate.GenerationChangedPredicate
+}
```
MatousJobanek, line 36: related to the other comments in the other PR, could you please explain why you want to introduce this predicate? What is the benefit of it?

MatousJobanek: but why doing it via a predicate and not in the controller logic?
btw, this is exactly why we shouldn't touch status from different controllers that rely on updates of the status. This is why the separation of concerns matters.

controllers/deactivation/deactivation_controller.go
```diff
@@ -188,14 +227,41 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	if !found || deactivatingCondition.Status != corev1.ConditionTrue ||
 		deactivatingCondition.Reason != toolchainv1alpha1.UserSignupDeactivatingNotificationCRCreatedReason {
 		// If the UserSignup has been marked as deactivating, however the deactivating notification hasn't been
-		// created yet, then wait - the notification should be created shortly by the UserSignup controller
-		// once the "deactivating" state has been set which should cause a new reconciliation to be triggered here
+		// created yet, then set the deactivation timestamp to a temporary value, calculated as being the current
+		// timestamp plus the number of pre-deactivation days configured in the settings, BUT only if it is currently nil
+		// OR the calculated value is more than 12 hours later than the current value
+		deactivationDueTime := time.Now().Add(time.Duration(deactivatingNotificationDays) * 12 * time.Hour)
```
MatousJobanek, line 233: shouldn't this be calculated since the notification was created instead? Also, why `* 12 * time.Hour`? Days has 24 hours if I'm not mistaken :thinking:

controllers/deactivation/predicate.go
sbryzak: We can't do it in the controller logic because we don't have a temporal point of reference.  The deactivation time is calculated as the provisioned time of the MUR plus the number of days in the user's tier but only when the user is not in a deactivating state.  When the user is in a deactivating state then it is calculated as the last transition time of the deactivating notification sent condition plus the number of deactivating notification days.  

Before the notification is actually sent we are left in a state of limbo when we have no way of calculating the scheduled deactivation time besides a calculation based on the current time, which is what the controller currently does in this PR.  The notification may fail to send for whatever hypothetical reason of your choice, but in the end this means that the user may be stuck in this state for an extended period of time.  In this case the worst that will happen is that the user sees they have "3" days left in the dashboard for their trial, and if it continues this way it will remain at 3 days until the notification is sent successfully.

There is already a lot of complex logic surrounding the deactivation process and Alexey and I have discussed the best way to approach this issue and this solution seems to provide the best experience for the user.

controllers/deactivation/deactivation_controller.go
MatousJobanek: ah, this is for the case when the notification is not created. but why such a value?

controllers/deactivation/predicate.go
MatousJobanek: put it simply:
* if in deactivating state, set it to 3 days since the notification was created
* if not and provisioned, set it to MUR provisioned time + number of days given by the tier
* in all other cases, either set it to nil (in some error cases or when user is deprovisioned), or keep the same value

in some error cases, sure, there could be some bad numbers, but let's be practical here and keep the code simple.
As for the case you mentioned above, if the notification is not send, then you can simply control what is set to the field directly in the method `setStatusDeactivatingNotificationCreationFailed`. In this way, it's much easier to control what is going on and to follow the states of the UserSignup resource.

controllers/deactivation/deactivation_controller.go
sbryzak: Fixed

controllers/deactivation/predicate.go
sbryzak: Done

controllers/deactivation/predicate.go
sbryzak: https://github.com/codeready-toolchain/toolchain-common/pull/397

controllers/deactivation/deactivation_controller.go
```diff
@@ -103,6 +111,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		for _, domain := range config.Deactivation().DeactivationDomainsExcluded() {
 			if strings.HasSuffix(usersignup.Spec.IdentityClaims.Email, domain) {
 				logger.Info("user cannot be automatically deactivated because they belong to the exclusion list", "domain", domain)
+
+				// Also set the Scheduled deactivation time to nil if it's not already
+				if usersignup.Status.ScheduledDeactivationTimestamp != nil {
+					usersignup.Status.ScheduledDeactivationTimestamp = nil
+					if err := r.Client.Status().Update(ctx, usersignup); err != nil {
```
alexeykazakov, line 118: Let's use the status updater here.

controllers/deactivation/deactivation_controller.go
```diff
@@ -128,6 +146,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		if err := r.resetDeactivatingState(ctx, usersignup); err != nil {
 			return reconcile.Result{}, err
 		}
+		// Set the scheduled deactivation time to nil
+		if usersignup.Status.ScheduledDeactivationTimestamp != nil {
```
alexeykazakov, line 150: You don't need to check if it's already nil or not. The status updater checks the old and new value.

controllers/deactivation/deactivation_controller.go
```diff
@@ -186,16 +228,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	// We calculate the actual deactivation due time based on when the deactivating condition was set.  This may end up
+	// being significantly later than the scheduled deactivation time set in UserSignup.Status.ScheduledDeactivationTime, as
+	// in some rare circumstances the deactivating notification/status may fail to be set due to cluster downtime or
+	// other reasons.  Because of this, the scheduled deactivation time that is set in the UserSignup.Status should be
+	// treated as informational only.
 	deactivationDueTime := deactivatingCondition.LastTransitionTime.Time.Add(time.Duration(deactivatingNotificationDays*24) * time.Hour)
 
 	if time.Now().Before(deactivationDueTime) {
+		// Update the ScheduledDeactivationTimestamp to the recalculated time base on when the deactivating notification was sent
+		if usersignup.Status.ScheduledDeactivationTimestamp == nil || !usersignup.Status.ScheduledDeactivationTimestamp.Time.Equal(deactivationDueTime) {
```
alexeykazakov, line 252: And you don't need to check the existing status value here either, do you? The updater will do it for you.

controllers/deactivation/deactivation_controller.go
```diff
@@ -221,6 +282,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	return reconcile.Result{}, nil
 }
 
+func setDeactivationTimestamp(ctx context.Context, r *Reconciler, mur *toolchainv1alpha1.MasterUserRecord,
+	userSignup *toolchainv1alpha1.UserSignup, deactivationTimeout time.Duration, logger logr.Logger) error {
+	scheduledDeactivationTime := (*mur.Status.ProvisionedTime).Add(deactivationTimeout)
+	if userSignup.Status.ScheduledDeactivationTimestamp == nil || !userSignup.Status.ScheduledDeactivationTimestamp.Time.Equal(scheduledDeactivationTime) {
+		ts := v1.NewTime(scheduledDeactivationTime)
+		userSignup.Status.ScheduledDeactivationTimestamp = &ts
+		if err := r.Client.Status().Update(ctx, userSignup); err != nil {
```
alexeykazakov, line 291: Use the status updater here too.

controllers/deactivation/deactivation_controller.go
sbryzak: Fixed

controllers/deactivation/deactivation_controller.go
sbryzak: Good catch, fixed

controllers/deactivation/deactivation_controller.go
sbryzak: Fixed

controllers/deactivation/deactivation_controller.go
sbryzak: I've removed the function all together, after replacing the code to use statusUpdater it turned out to just be 2 lines of code which isn't worth putting in its own function.

alexeykazakov (commented):
Looks good to me. We just need to add the status check in our e2e tests. We don't need to cover all the use cases but let's add the timestamp status check into at least some existing tests.

controllers/deactivation/deactivation_controller.go
```diff
@@ -48,6 +50,7 @@ type Reconciler struct {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+// nolint: gocyclo
```
MatousJobanek, line 53: why?

controllers/deactivation/deactivation_controller.go
```diff
@@ -128,6 +142,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		if err := r.resetDeactivatingState(ctx, usersignup); err != nil {
 			return reconcile.Result{}, err
 		}
+		// Set the scheduled deactivation time to nil
+		if err := statusUpdater.SetScheduledDeactivationStatus(ctx, usersignup, nil); err != nil {
```
MatousJobanek, line 146: what is the expected UX for:
1. user won't be deactivated (is either excluded or is assigned to non-deactivating tier)
2. the value cannot be not calculated for the account for some reasons

Do you want to show the same message/value for these two cases?

controllers/usersignup/status_updater.go
```diff
@@ -448,3 +449,11 @@ func (u *StatusUpdater) updateStatusHomeSpace(ctx context.Context, userSignup *t
 	}
 	return nil // Nothing changed
 }
+
+func (u *StatusUpdater) SetScheduledDeactivationStatus(ctx context.Context, userSignup *toolchainv1alpha1.UserSignup, scheduledTime *metav1.Time) error {
+	if userSignup.Status.ScheduledDeactivationTimestamp != scheduledTime {
```
MatousJobanek, line 454:
```suggestion
	if !userSignup.Status.ScheduledDeactivationTimestamp.Equal(scheduledTime) {
```

sbryzak: Fixed, thanks.

controllers/deactivation/deactivation_controller.go
sbryzak: I guess since we can't actually differentiate between these two cases currently then a generic message or indicator would be most appropriate, perhaps for the "days remaining" box/field in the dashboard it can just display a `-` symbol or something.

controllers/deactivation/deactivation_controller.go
alexeykazakov: Have you tried to re-enable it now? The code is less complex now then the initial one.

controllers/deactivation/deactivation_controller.go
alexeykazakov: Yeah. Or `N/A`. Which can mean both `Not Available` and `Not Applicable` :)

alexeykazakov (approved):

controllers/deactivation/deactivation_controller.go
sbryzak: Yes, the linter still doesn't like it.

sbryzak:
Merging now as I am on PTO from tomorrow.  Codecov is lying about many of the lines not being covered, I added additional tests and confirmed with the debugger that it is indeed hitting those lines it's reporting as uncovered in the tests.

controllers/deactivation/deactivation_controller.go
MatousJobanek: have you considered to refactor the code and split the method into multiple ones? TBH, the linter is correct that the method is really complex and should be simplified/refactored.

PR #994, opened 2024-03-18
CRD yaml make generate updates when upgrading controller-gen in api

ranakan19:
crd changes from running make generate in api when controller-gen version updated

DO-NOT Merge 
(All the PRs related to version updates need to be merged at once, but these PR are for early feedback so that there are not a lot of changes to review at once)

### DIFF
--- a/config/crd/bases/toolchain.dev.openshift.com_tiertemplates.yaml
+++ b/config/crd/bases/toolchain.dev.openshift.com_tiertemplates.yaml
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
-  creationTimestamp: null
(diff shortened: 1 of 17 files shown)


alexeykazakov (approved):

MatousJobanek (approved):

config/crd/bases/toolchain.dev.openshift.com_tiertemplates.yaml
```diff
@@ -76,6 +75,8 @@ spec:
                       can be included in the output.
                     type: string
                   metadata:
+                    description: 'metadata is the standard object''s metadata. More
```
rajivnathan, line 78: Extra comma? Not sure whether it'd effect anything but just thought I'd raise it

```suggestion
                    description: 'metadata is the standard object's metadata. More
```

alexeykazakov: It's a generated yaml. Not sure what API go code exactly is the source of this string but we shouldn't modify it in the CRD yaml directly :)

PR #995, opened 2024-03-19
Use `go list` in the go.mod replacements check to improve reliability

metlos:
Use `go list -m all` to get us the full machine-readable list of modules instead of manually parsing go.mod using grep to correctly check forbidden replacements in the go.mod.


rajivnathan (approved):
Nice 👍

ranakan19 (approved):

alexeykazakov (approved):

PR #998, opened 2024-03-22
add PublicViewer support

filariow:
linked to https://github.com/codeready-toolchain/api/pull/410

### DIFF
--- a/controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
+++ b/controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/codeready-toolchain/api/api/v1alpha1"
-	"github.com/codeready-toolchain/host-operator/controllers/toolchainconfig"
 	"github.com/codeready-toolchain/host-operator/pkg/apis"
 	"github.com/codeready-toolchain/host-operator/pkg/cluster"
 	. "github.com/codeready-toolchain/host-operator/test"
@@ -45,8 +44,8 @@ func TestDeleteSpaceBinding(t *testing.T) {
 	joeMur := masteruserrecord.NewMasterUserRecord(t, "joe")
 
 	t.Run("publicViewer config is enabled", func(t *testing.T) {
-		publicViewerConfig := toolchainconfig.PublicViewerConfig{
-			PublicViewerConfig: toolchainv1alpha1.PublicViewerConfig{
+		publicViewerConfig := commonconfig.PublicViewerConfig{
+			Config: toolchainv1alpha1.PublicViewerConfig{
 				Enabled:  true,
 				Username: "public-viewer",
 			},
@@ -82,8 +81,8 @@ func TestDeleteSpaceBinding(t *testing.T) {
 	})
 
 	t.Run("publicViewer config is disabled", func(t *testing.T) {
-		publicViewerConfig := toolchainconfig.PublicViewerConfig{
-			PublicViewerConfig: toolchainv1alpha1.PublicViewerConfig{
+		publicViewerConfig := commonconfig.PublicViewerConfig{
+			Config: toolchainv1alpha1.PublicViewerConfig{
 				Enabled:  false,
 				Username: "public-viewer",
 			},
--- a/controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
+++ b/controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/host-operator/controllers/toolchainconfig"
 	"github.com/codeready-toolchain/host-operator/pkg/apis"
 	"github.com/codeready-toolchain/host-operator/pkg/cluster"
 	. "github.com/codeready-toolchain/host-operator/test"
@@ -23,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/stretchr/testify/require"
 	"k8s.io/client-go/kubernetes/scheme"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -341,10 +341,11 @@ func main() { // nolint:gocyclo
...
 	if err = (&spacebindingcleanup.Reconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		Namespace:      namespace,
-		MemberClusters: clusterScopedMemberClusters,
+		Client:             mgr.GetClient(),
+		Scheme:             mgr.GetScheme(),
+		Namespace:          namespace,
+		MemberClusters:     clusterScopedMemberClusters,
+		PublicViewerConfig: crtConfig.PublicViewer(),
```
mfrancisc, line 348: I'm wondering if we should pass this here or read it at "Reconcile" time? 

In this way any change to this config would need a restart of the operator. I don't think there would be many changes but just something to consider.

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
```diff
@@ -34,13 +35,88 @@ func TestDeleteSpaceBinding(t *testing.T) {
...
 	sbLaraIbmEdit := sb.NewSpaceBinding("lara", "ibm", "edit", "signupC")
+	sbPublicViewerRedhatView := sb.NewSpaceBinding("public-viewer", "redhat", "view", "signupD")
 
 	redhatSpace := spacetest.NewSpace(test.HostOperatorNs, "redhat")
 	ibmSpace := spacetest.NewSpace(test.HostOperatorNs, "ibm")
 
 	laraMur := masteruserrecord.NewMasterUserRecord(t, "lara")
 	joeMur := masteruserrecord.NewMasterUserRecord(t, "joe")
 
+	t.Run("publicViewer config is enabled", func(t *testing.T) {
```
mfrancisc, line 46: minor - but should we create a test for 
```
Config: toolchainv1alpha1.PublicViewerConfig{
				Enabled:  true,
				Username: "",
			},
```
when the Username is empty ? WDYT ?

controllers/spacebindingrequest/spacebindingrequest_controller.go
```diff
@@ -94,20 +96,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-			logger.Error(errStatus, "error updating SpaceBindingRequest status")
+	if r.PublicViewerConfig.IsPublicViewer(spaceBindingRequest.Spec.MasterUserRecord) {
+		// set ready condition on spaceBindingRequest
+		err = r.updateStatus(ctx, spaceBindingRequest, memberClusterWithSpaceBindingRequest, toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.SpaceBindingRequestInvalidReason,
+			Message: fmt.Sprintf("%s is reserved and can not be used in SpaceBinding's MasterUserRecord", r.PublicViewerConfig.Username()),
+		})
+	} else {
```
mfrancisc, line 107: I haven't checked the other PRs yet ( member, reg-serv) but I'm not sure I'm following here, how is "public-viewer" getting access to the Spaces, isn't it through the usage of SpaceBindingRequests? 

I see you are blocking SBRs for "public-viewer" here 🤔 , but I'm sure I'm missing something.

controllers/usersignup/usersignup_controller.go
```diff
@@ -72,6 +73,7 @@ type Reconciler struct {
 	Scheme         *runtime.Scheme
 	SegmentClient  *segment.Client
 	ClusterManager *capacity.ClusterManager
+	PublicViewer   commonconfig.PublicViewerConfig
```
mfrancisc, line 76: Is this going to be used in follow-up PRs ? 
It doesn't seem to be used right now.

mfrancisc (commented):
Thanks for creating the PRs for the public-viewer effort 🙏  

I have few comments.

main.go
MatousJobanek: definitely at the reconcile time - we shouldn't need to restart the container when we change the configuration.
There is also a logic that wraps the ToolchainConfig, does some caching and provides the default values if nothing is set.

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
filariow: good catch. What would be here the expected behavior? panic?
To prevent this from happening I'd propose to have some default and validation, like follows

```diff
// PublicViewer used to enable public-viewer support
// +k8s:openapi-gen=true
type PublicViewerConfig struct {
	// +kubebuilder:default:=false
	Enabled bool `json:"enabled"`

	// +optional
+	// +kubebuilder:validation:MinLength=4
	// +kubebuilder:validation:MaxLength=63
	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	// +kubebuilder:default:=public-viewer
	Username string `json:"username,omitempty"`
}
```

filariow: WDYT?

controllers/usersignup/usersignup_controller.go
filariow: Seems something that I forgot to remove 😄 thanks for noticing it!

main.go
filariow: @MatousJobanek, just to be sure, like what's done [here](https://github.com/codeready-toolchain/host-operator/blob/de95481affea496ceb043113a347f64a61abe42c/controllers/usersignup/usersignup_controller.go#L123).

out of curiosity - why not doing this with [k8s volume mounting](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) and just fetch the data from filesystem?

controllers/spacebindingrequest/spacebindingrequest_controller.go
filariow: the actual implementation relies on the direct creation of a SpaceBinding on the Host cluster.
The check introduced here, is used to avoid users granted the permission to share the workspace to make the workspace publicly visible by creating SBRs for the `public-viewer` user

controllers/usersignup/usersignup_controller.go
filariow: removed in 84e431dcac88feaec16665874fb880c19c4a397f

main.go
mfrancisc: > like what's done [here](https://github.com/codeready-toolchain/host-operator/blob/de95481affea496ceb043113a347f64a61abe42c/controllers/usersignup/usersignup_controller.go#L123)?

yes that's one example of how to fetch the config/default values

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
mfrancisc: TBH maybe we could remove the `Enabled` field and just use the `username`, if provided then the feature is enabled. This will reduce the amount of permutations to check/test. Would also simplify a bit the configuration.

mfrancisc: The default validation using kubebuilder markers works fine, but also just checking that the `len(username) > 0` , in case it's empty we can assume that the feature is disabled.

sadlerap: I'd like to have validation where possible, but if we want to deduce enablement from the `Username` field only, then we're going to have to change these validation rules.

Maybe change the validation pattern to this?
```go
// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?)?$`
```

This will make the empty string valid, but still validate that usernames must be shaped like a username if they're set.

PR #1006, opened 2024-04-03
make NSTemplateTier generator generic & move to common

MatousJobanek:
depends on https://github.com/codeready-toolchain/toolchain-common/pull/381
* moves most of the `nstemplatetier_generator.go` into toolchain-common
* loads the metadata & templates from the assets and passes them as a parameter of the generic `GenerateTiers` function
* drops the test templates and related tests

[KUBESAW-59](https://issues.redhat.com/browse/KUBESAW-59)


mfrancisc (approved):
Great Job 👍 

I like how the new `nstemplatetiers.GenerateTiers` function looks, only nit is that the client needs to depend on a couple of packages potentially:
- the new `nstemplatetiers`
- the `commonclient` 

Do you see other implementations of the `ensureObject` function or maybe we could embed that in the `nstemplatetiers.GenerateTiers` function ?

fbm3307 (approved):
LGTM

PR #1011, opened 2024-04-16
add viewer role for appstudio

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>


sadlerap (commented):
PR needs a rebase, but the changes look ok.

test/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
```diff
@@ -0,0 +1,200 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: appstudio-spacerole-viewer
```
MatousJobanek, line 4: We don't keep the test templates in host-operator anymore. Let's remove this file.

deploy/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
```diff
@@ -0,0 +1,191 @@
...
+    - watch
+  - apiGroups:
+    - results.tekton.dev
+    resources:
+    - results
+    - records
+    - logs
+    verbs:
+    - get
+    - list
```
MatousJobanek, line 72: `- watch` is missing there intentionally?

deploy/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
```diff
@@ -0,0 +1,191 @@
...
+    - watch
+  - apiGroups:
+    - ''
+    resources:
+    - configmaps
+    verbs:
+    - get
+    - list
+    - watch
```
MatousJobanek, line 142: IIRC, this should be granted automatically via the aggregated `view` ClusterRole

MatousJobanek (commented):
Looks good, but could you please drop the test template - it's not used in host-operator anymore.
The other comments are just FYI - I saw that it was taken from the already existing templates, so in the future it would be nice to fix that also there. 

BTW, in the other templates are some obsolete/copy-paste left-overs that would be nice to clean as well, but that's for a different PR.

deploy/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
filariow: should we remove it from here and from https://github.com/codeready-toolchain/host-operator/blob/ea6e92e187b044a11527bdbb33e369f086befa3d/deploy/templates/nstemplatetiers/appstudio/spacerole_contributor.yaml#L138 ? If that's the case, I'd say we can merge this PR as it is and remove from both the roles in a subsequent PR. WDYT?

test/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
filariow: done in a4ff3f73c63567c5d56606e8e8b3f0e8423bae8c

deploy/templates/nstemplatetiers/appstudio/spacerole_viewer.yaml
MatousJobanek: yup, sure. 
as I mentioned in the description of the review - that was just FYI

MatousJobanek (approved):

alexeykazakov (approved):

PR #1020, opened 2024-04-29
add test for 'kubesaw' prefix

filariow:
JIRA: https://issues.redhat.com/browse/ASC-531

As 'kube' is already a forbidden prefix, there is no need of explicitly add 'kubesaw'.
Adding it in test to prevent future changes to break this behavior.

Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/controllers/usersignup/usersignup_controller_test.go
+++ b/controllers/usersignup/usersignup_controller_test.go
@@ -3929,7 +3929,8 @@ func TestUsernameWithForbiddenPrefix(t *testing.T) {
 	require.Len(t, config.Users().ForbiddenUsernamePrefixes(), 5)
 	names := []string{"-Bob", "-Dave", "Linda", ""}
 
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):
Looks good. But please add a comment to the code explaining why it's here.
Basically the same description you have in the PR. Something like:
```
As 'kube' is already a forbidden prefix, so "kubesaw" is covered but testing it explicitly would prevent future changes from breaking this behavior.
```

MatousJobanek (approved):

filariow:
> Looks good. But please add a comment to the code explaining why it's here. Basically the same description you have in the PR. Something like:
> 
> ```
> As 'kube' is already a forbidden prefix, so "kubesaw" is covered but testing it explicitly would prevent future changes from breaking this behavior.
> ```

done in 20266e09712c24be883904fd4029258e45b6a8b7

PR #1026, opened 2024-05-07
Removing Owners File

fbm3307:
- Removing the Owners File .The openshift-ci uses CODEOWNERS file , so the OWNERS file is not needed now
- Updating the API, Toolchain-common dependency to latest

This is similar to
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/566
- Toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/967
- Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/428


MatousJobanek (approved):
Thanks :+1: 
The list should technically match what we have in CODEOWNERS. But TBH, I'm not 100% sure if we need this file at all there. Given the way we use prow, it could be technically OK to completely remove the file.
WDYT? @rsoaresd

rsoaresd:
> Thanks 👍 The list should technically match what we have in CODEOWNERS. But TBH, I'm not 100% sure if we need this file at all there. Given the way we use prow, it could be technically OK to completely remove the file. WDYT? @rsoaresd

@MatousJobanek, since, at the moment, the openshift-ci bot is not using the OWNERS file to restrict who should approve and who should review it, maybe there is no point in keeping the file since we already have CODEOWNERS.

mfrancisc (approved):

fbm3307:
@rsoaresd @MatousJobanek , if that is the case, i can remove the Owners file from here and rest of the repos.
I initially thought of raising the PR with that, but was not sure if Owners file was being used anywhere else other than PR reviews

alexeykazakov (approved):

alexeykazakov:
Looks like another flaky test: toolchain-e2e/test/e2e/space_autocompletion_test.go:141

/retest

PR #1033, opened 2024-05-16
SpaceProvisionerConfig should only be ready when underlying ToolchainCluster is ready

metlos:
Fix the SpaceProvisionerConfig controller to only set the SPC ready when the referenced ToolchainCluster is ready. This should make sure we never provision to offline clusters.

Associated PRs:
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/977
- api: https://github.com/codeready-toolchain/api/pull/424

### DIFF
--- a/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
+++ b/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
@@ -78,7 +78,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 func (r *Reconciler) determineReadyState(ctx context.Context, spc *toolchainv1alpha1.SpaceProvisionerConfig) (toolchainv1alpha1.Condition, error) {
 	toolchainCluster := &toolchainv1alpha1.ToolchainCluster{}
 	toolchainClusterKey := runtimeclient.ObjectKey{Name: spc.Spec.ToolchainCluster, Namespace: spc.Namespace}
-	toolchainPresent := corev1.ConditionFalse
+	var toolchainPresent corev1.ConditionStatus
 	toolchainPresenceReason := toolchainv1alpha1.SpaceProvisionerConfigValidReason
 	var reportedError error
 	toolchainPresenceMessage := ""
@@ -96,11 +96,11 @@ func (r *Reconciler) determineReadyState(ctx context.Context, spc *toolchainv1al
 		toolchainPresenceReason = toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotFoundReason
 		toolchainPresent = corev1.ConditionFalse
 	} else {
-		for _, c := range toolchainCluster.Status.Conditions {
-			if c.Type == toolchainv1alpha1.ToolchainClusterReady {
-				toolchainPresent = c.Status
-				break
-			}
+		readyCond, found := condition.FindConditionByType(toolchainCluster.Status.Conditions, toolchainv1alpha1.ConditionReady)
+		if !found {
+			toolchainPresent = corev1.ConditionFalse
+		} else {
+			toolchainPresent = readyCond.Status
 		}
 		if toolchainPresent != corev1.ConditionTrue {
 			toolchainPresenceReason = toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotReadyReason
--- a/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
+++ b/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
@@ -12,7 +12,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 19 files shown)


controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -67,7 +76,33 @@ func TestSpaceProvisionerConfigValidation(t *testing.T) {
...
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(spc), spc))
+
+		// then
+		assert.NoError(t, reconcileErr)
+		AssertThat(t, spc, Is(NotReadyWithReason(toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotReadyReason)))
+	})
+	t.Run("becomes ready when the referenced ToolchainCluster appears and is ready", func(t *testing.T) {
```
mfrancisc, line 105: should this be a subtest of the previous one ( "is not ready when existing not-ready ToolchainCluster is referenced " ) and use the same `spc` variable so that we check that a second reconcile would put that back to ready when toolchaincluster gets ready ?

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -102,6 +145,54 @@ func TestSpaceProvisionerConfigValidation(t *testing.T) {
 		assert.NoError(t, reconcileErr)
 		AssertThat(t, spc, Is(NotReadyWithReason(toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotFoundReason)))
 	})
+	t.Run("becomes not ready when the referenced ToolchainCluster is marked as not ready", func(t *testing.T) {
```
mfrancisc, line 148: similarly should this be a subtest of "is ready when existing ready ToolchainCluster is referenced" so that we first set the SPC to ready and we verify that the controller put's it back to non ready ?

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller_test.go
```diff
@@ -102,6 +145,54 @@ func TestSpaceProvisionerConfigValidation(t *testing.T) {
...
+
+		// when
+		_, reconcileErr := r.Reconcile(context.TODO(), req)
+		require.NoError(t, cl.Get(context.TODO(), runtimeclient.ObjectKeyFromObject(spc), spc))
+
+		// then
+		assert.NoError(t, reconcileErr)
+		AssertThat(t, spc, Is(NotReadyWithReason(toolchainv1alpha1.SpaceProvisionerConfigToolchainClusterNotReadyReason)))
+	})
+	t.Run("becomes not ready when the referenced ToolchainCluster no longer has ready condition", func(t *testing.T) {
```
mfrancisc, line 176: same here

mfrancisc (commented):
Nice Job 👍 

I have only few questions related to the unit tests.

mfrancisc (approved):
Looks good 👍 

Thanks for addressing my comments.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -91,23 +95,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-		if reportedError != nil {
-			logger.Info("failed to update the status (reported as failed reconciliation) with a previous unreported error during reconciliation", "unreportedError", reportedError)
+	} else {
+		for _, c := range toolchainCluster.Status.Conditions {
+			if c.Type == toolchainv1alpha1.ToolchainClusterReady {
+				toolchainPresent = c.Status
+				break
+			}
+		}
+		if toolchainPresent != corev1.ConditionTrue {
```
rajivnathan, line 105: I believe we can simplify this to
```suggestion
		if condition.IsNotTrue(toolchainCluster.Status.Conditions, toolchainv1alpha1.ToolchainClusterReady) {
			toolchainPresent = apiv1.ConditionFalse
```

rajivnathan (approved):
Looks great overall 👍 , just made a small suggestion

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
metlos: These are cluster conditions which are different from the "ordinary" conditions that have the utility functions in the `condition` package. Note that there is currently PR open to get rid of the cluster conditions and replace them with the "ordinary" ones.

rajivnathan: Ah, missed that 👍

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -71,10 +59,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, nil
 	}
 
-	// check that there is a ToolchainCluster CR in the same namespace
+	readyCondition, reportedError := r.determineReadyState(ctx, spaceProvisionerConfig)
+
+	spaceProvisionerConfig.Status.Conditions, _ = condition.AddOrUpdateStatusConditions(spaceProvisionerConfig.Status.Conditions,
+		readyCondition)
```
MatousJobanek, line 65: I know that it doesn't matter much in this case, but you could do:
```suggestion
	var updated bool
	spaceProvisionerConfig.Status.Conditions, updated = condition.AddOrUpdateStatusConditions(spaceProvisionerConfig.Status.Conditions,
		readyCondition)
	if !updated {
		return ctrl.Result{}, reportedError
	}
```
but not probably needed applying it in this case :-)

MatousJobanek (approved):
cool, thanks for the additional changes :+1:

PR #1035, opened 2024-05-21
Reload the usersignup if it is updated during the deactivation reconcile

sbryzak:
This simple PR reloads the UserSignup instance if its state is updated as part of the deactivation controller's reconciliation phase.


controllers/deactivation/deactivation_controller.go
```diff
@@ -139,9 +139,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	if deactivationTimeoutDays == 0 {
 		// If the usersignup was already set to deactivating then reset it to false
-		if err := r.resetDeactivatingState(ctx, usersignup); err != nil {
+		updated, err := r.resetDeactivatingState(ctx, usersignup)
+		if err != nil {
 			return reconcile.Result{}, err
 		}
+
+		// If the UserSignup was updated, reload it
+		if updated {
```
alexeykazakov, line 148: Maybe something like this instead?
```
usersignup, err := r.resetDeactivatingState(ctx, usersignup)

func (r *Reconciler) resetDeactivatingState(ctx context.Context, usersignup *toolchainv1alpha1.UserSignup) (*toolchainv1alpha1.UserSignup, error) {
    us := usersignup
    ...
    if updated {
        us = ... // Get the updated UserSignup
    }
    return us, nil
}

```

alexeykazakov (commented):
Why you update usersignup only in two cases but not in all cases when we update its status?

controllers/deactivation/deactivation_controller.go
alexeykazakov: Actually why not just simply exit the reconcile loop if we updated the status? The controller watches the entire usersignup including the status. So when we change it then it will trigger another reconcile loop.

sbryzak: I actually thought about doing it this way after, I think it will make for neater code.

sbryzak: BTW I don't believe it's necessary to reload the UserSignup after all updates, only when it's possible for multiple updates to occur in the same reconcile.

alexeykazakov (approved):
The code looks fine. I'm still not sure if we really want to reload the UserSignup instead of just exiting the reconcile loop. The controller will be triggered again it watches the status too and it will resume the rest of the loop. But I'm fine with reloading the UserSignup too.

MatousJobanek (changes requested):
Could you please explain why you want to reload the UserSignup? You have already loaded it once, if the UserSignup is outdated then it should fail during the update. 
Also, I don't see any updates in the unit tests.

controllers/deactivation/deactivation_controller.go
```diff
@@ -278,13 +281,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		}
+		var reloaded *toolchainv1alpha1.UserSignup
+		// Reload the UserSignup
+		if err := r.Client.Get(ctx, types.NamespacedName{
+			Namespace: userSignup.Namespace,
+			Name:      userSignup.Name,
+		}, reloaded); err != nil {
+			// If there was an error, return the existing UserSignup and the error
+			return userSignup, err
 		}
```
MatousJobanek, line 299: The original UserSignup object already has the values set, there is no need to reload what you have already set there.
We should rather really return here and reconcile again as it was before.

sbryzak: Agreed, have now refactored to work this way instead.

MatousJobanek (commented):
Looks good, but the unit tests need more love because they are becoming something else than just unit tests.

alexeykazakov:
@MatousJobanek could you please take a look at this PR again to make sure your comments are addressed now?

MatousJobanek (commented):
I'm sorry, but the changes didn't address my comments. The tests are not really unit tests, they keep re-instantiating the client in the middle of the test and/or reloading objects and re-calling `Reconcile` as part of the same test.

My assumption is that this was written in this way to save time and lines of code by re-using the previous states. I understand the intention, but this is not the way we (should) write unit tests. There should be clear:
1. setup of the test
2. call of the single function we want to test
3. verify the results

This is the approach that 98% of all unit tests can and should follow. There are always some exceptions where it makes sense to create sub-tests that reuse the state from previous tests, but they should not re-instantiate the client storage in the middle of the execution in any way.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -576,6 +650,17 @@ func TestReconcile(t *testing.T) {
...
+			_, err := r.Reconcile(context.TODO(), req)
+			require.NoError(t, err)
+
+			// Reload the userSignup
+			require.NoError(t, fakeClient.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
+			require.False(t, states.Deactivating(userSignupFoobar))
+
+			// Reconcile again
+			r, req, fakeClient = prepareReconcile(t, mur.Name, userTierNoDeactivation, mur, userSignupFoobar, config)
+
```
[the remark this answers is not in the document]
sbryzak: Moved into its own subtest, new client no longer created

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -553,8 +619,16 @@ func TestReconcile(t *testing.T) {
 			require.False(t, states.Deactivating(userSignupFoobar))
 			require.False(t, states.Deactivated(userSignupFoobar))
 
+			// Reconcile again
+			r, req, cl = prepareReconcile(t, mur.Name, userTierNoDeactivation, mur, userSignupFoobar, config)
+			res, err = r.Reconcile(context.TODO(), req)
+			require.NoError(t, err)
+
+			require.NoError(t, cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
+
```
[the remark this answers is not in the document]
sbryzak: Moved into separate subtest and removed creation of new client.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -501,25 +547,45 @@ func TestReconcile(t *testing.T) {
...
-			// The scheduled deactivation time should have also been updated, and should now expire in ~62 days
-			expected := time.Now().Add(62 * time.Hour * 24)
-			comparison := expected.Sub(userSignupFoobar.Status.ScheduledDeactivationTimestamp.Time)
+				// Confirm the deactivating state has been unset in the UserSignup
+				require.False(t, states.Deactivating(reloaded))
 
-			// accept if we're within 1 hour of the expected deactivation time
-			require.Less(t, comparison, time.Hour)
+				// Prepare to reconcile again, with the updated UserSignup
+				r, req, cl = prepareReconcile(t, mur.Name, userTier90, mur, reloaded, config)
```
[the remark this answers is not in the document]
sbryzak: Refactored slightly, however I'm not moving this to a standalone test as its a direct progression from the state we expect in the previous reconcile.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
...
+						// Reload the usersignup again
+						reloaded = &toolchainv1alpha1.UserSignup{}
+						err = cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, reloaded)
+						require.NoError(t, err)
+
+						scheduledDeactivationTime := reloaded.Status.ScheduledDeactivationTimestamp
+
+						// when
+						// Reconcile again
+						r, req, cl = prepareReconcile(t, mur.Name, userTier30, mur, reloaded, config)
```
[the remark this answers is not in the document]
sbryzak: Fixed

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
...
-			reloaded = &toolchainv1alpha1.UserSignup{}
-			err = cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, reloaded)
-			require.NoError(t, err)
-			require.Equal(t, scheduledDeactivationTime, reloaded.Status.ScheduledDeactivationTimestamp)
+				// Confirm the deactivating state has been unset in the UserSignup
+				require.False(t, states.Deactivating(reloaded))
+
+				t.Run("reconcile a third time to confirm the scheduled deactivation time is correctly set", func(t *testing.T) {
+					// given
+					r, req, cl = prepareReconcile(t, mur.Name, userTier30, mur, reloaded, config)
```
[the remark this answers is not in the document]
sbryzak: Refactored to remove line.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
...
-			require.Less(t, comparison, time.Hour)
+			cl.MockUpdate = nil
 
-			// Reload the usersignup
-			reloaded := &toolchainv1alpha1.UserSignup{}
-			err = cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, reloaded)
-			require.NoError(t, err)
+			t.Run("reconcile again to ensure the usersignup has been updated", func(t *testing.T) {
+				// when
+				_, err = r.Reconcile(context.TODO(), req)
```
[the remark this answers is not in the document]
sbryzak: Error case moved to separate test.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
...
+				default:
+					return cl.Client.Status().Update(ctx, obj)
+				}
+			}
 
-			// confirm that the scheduled deactivation time is ~30 days
-			expected := time.Now().Add(30 * time.Hour * 24)
-			comparison := expected.Sub(userSignupFoobar.Status.ScheduledDeactivationTimestamp.Time)
+			_, err := r.Reconcile(context.TODO(), req)
+			require.Error(t, err)
```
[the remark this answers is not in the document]
sbryzak: Updated

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
...
+					return errors.New("mock error")
+				default:
+					return cl.Client.Status().Update(ctx, obj)
+				}
+			}
 
-			// confirm that the scheduled deactivation time is ~30 days
-			expected := time.Now().Add(30 * time.Hour * 24)
-			comparison := expected.Sub(userSignupFoobar.Status.ScheduledDeactivationTimestamp.Time)
+			_, err := r.Reconcile(context.TODO(), req)
```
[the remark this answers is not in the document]
sbryzak: Updated

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -75,43 +75,89 @@ func TestReconcile(t *testing.T) {
 			r, req, cl := prepareReconcile(t, mur.Name, userTier30, mur, userSignupFoobar, config)
 			// when
```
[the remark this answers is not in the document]
sbryzak: Updated

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -86,76 +84,78 @@ func TestReconcile(t *testing.T) {
...
+				diff := expectedTime - actualTime
+				require.Truef(t, diff > 0 && diff < 2*time.Second, "expectedTime: '%v' is not within 2 seconds of actualTime: '%v' diff: '%v'", expectedTime, actualTime, diff)
+				assertThatUserSignupStateIsDeactivated(t, cl, username, false)
 
-					// confirm that the scheduled deactivation time is set
-					require.NoError(t, cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
-					require.NotNil(t, userSignupFoobar.Status.ScheduledDeactivationTimestamp)
+				// confirm that the scheduled deactivation time is set
+				require.NoError(t, cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
+				require.NotNil(t, userSignupFoobar.Status.ScheduledDeactivationTimestamp)
```
MatousJobanek: if you update the function `assertThatUserSignupStateIsDeactivated` so it returns the retrieved Usersignup:
```go
func assertThatUserSignupStateIsDeactivated(t *testing.T, cl *commontest.FakeClient, name string, expected bool) *toolchainv1alpha1.UserSignup {
	userSignup := &toolchainv1alpha1.UserSignup{}
	err := cl.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: operatorNamespace}, userSignup)
	require.NoError(t, err)
	require.Equal(t, expected, states.Deactivated(userSignup))
	return userSignup
}
```
then you could simplify this code:
```
userSignupFoobar = assertThatUserSignupStateIsDeactivated(t, cl, username, false)
```

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -552,21 +552,17 @@ func TestReconcile(t *testing.T) {
...
-				// Confirm the deactivating state has been unset in the UserSignup
-				require.False(t, states.Deactivating(reloaded))
+			// Reload the usersignup
+			err = cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar)
+			require.NoError(t, err)
 
-				// Prepare to reconcile again, with the updated UserSignup
-				r, req, cl = prepareReconcile(t, mur.Name, userTier90, mur, reloaded, config)
+			// Confirm the deactivating state has been unset in the UserSignup
+			require.False(t, states.Deactivating(userSignupFoobar))
```
MatousJobanek:
```suggestion
			// Confirm the deactivating state has been unset in the UserSignup
			assertThatUserSignupStateIsDeactivated(t, cl, userSignupFoobar.Name, false)
```

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -619,16 +615,18 @@ func TestReconcile(t *testing.T) {
...
-			// Reconcile again
-			r, req, cl = prepareReconcile(t, mur.Name, userTierNoDeactivation, mur, userSignupFoobar, config)
-			res, err = r.Reconcile(context.TODO(), req)
-			require.NoError(t, err)
+			require.NotNil(t, userSignupFoobar.Status.ScheduledDeactivationTimestamp)
 
-			require.NoError(t, cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar))
+			t.Run("ensure scheduled deactivation time is set to nil after reconciling again", func(t *testing.T) {
+				res, err = r.Reconcile(context.TODO(), req)
+				require.NoError(t, err)
```
MatousJobanek, line 622:
```suggestion
				// when
				res, err = r.Reconcile(context.TODO(), req)
				
				// then
				require.NoError(t, err)
```

MatousJobanek (approved):
Looks good, but please take a look at my comments and suggestions

controllers/deactivation/deactivation_controller_test.go
sbryzak: Fixed

controllers/deactivation/deactivation_controller_test.go
sbryzak: Done

controllers/deactivation/deactivation_controller_test.go
sbryzak: good suggestion, changed.

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -86,76 +84,78 @@ func TestReconcile(t *testing.T) {
...
-				require.False(t, states.Deactivating(reloaded))
+			// Reload the usersignup
+			err = cl.Get(context.TODO(), types.NamespacedName{Name: userSignupFoobar.Name, Namespace: operatorNamespace}, userSignupFoobar)
+			require.NoError(t, err)
 
-				t.Run("reconcile a third time to confirm the scheduled deactivation time is correctly set", func(t *testing.T) {
-					// given
-					r, req, cl = prepareReconcile(t, mur.Name, userTier30, mur, reloaded, config)
+			// Confirm the deactivating state has been unset in the UserSignup
+			require.False(t, states.Deactivating(userSignupFoobar))
```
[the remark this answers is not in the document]
sbryzak: Fixed

controllers/deactivation/deactivation_controller_test.go
```diff
@@ -65,18 +65,16 @@ func TestReconcile(t *testing.T) {
...
 
 			r, req, cl := prepareReconcile(t, mur.Name, userTier30, mur, userSignupFoobar, config)
-			// when
-			timeSinceProvisioned := time.Since(murProvisionedTime.Time)
 
-			// First throw an error when the deactivating status is set
+			// Throw an error when the deactivating status is set
 			cl.MockUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
 				switch obj.(type) {
 				case *toolchainv1alpha1.UserSignup:
```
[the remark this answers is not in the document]
sbryzak: Good catch, this should be state, not status. Fixed.

PR #1040, opened 2024-06-11
Escape html when setting message content for notifications

sbryzak:
This simple change should fix the issue with truncated notification content.

### DIFF
--- a/controllers/notification/mailgun_notification_delivery_service.go
+++ b/controllers/notification/mailgun_notification_delivery_service.go
@@ -3,6 +3,7 @@ package notification
 import (
 	"context"
 	"fmt"
+	"html"
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


rajivnathan (approved):
Nice catch! 👍

mfrancisc (approved):

controllers/notification/mailgun_notification_delivery_service.go
```diff
@@ -114,7 +115,7 @@ func (s *MailgunNotificationDeliveryService) Send(notification *toolchainv1alpha
 		message.SetReplyTo(s.ReplyToEmail)
 	}
 
-	message.SetHtml(body)
+	message.SetHtml(html.EscapeString(body))
```
MatousJobanek, line 118: but wouldn't this mean that the html fomat we have in our notification templates is not applied (since it's escaped)? I mean, we don't want to escape the body for all notifications, right? Only the body of the suspicious notifications created by the miner detector. we don't want to modify the expected behavior for the standard notifications sent to end-users.

sbryzak: You're right, this would likely impact templates that contain html tags.  Perhaps the best way to approach this is to escape the string when the notification is actually created, rather than in the delivery service.  Where are the miner detector notifications created from?

alexeykazakov (changes requested):
I agree with Matous. Let's move it to the place where we create the notification to avoid breaking html content.

PR #1041, opened 2024-06-12
fix: don't rely on certificates in ToolchainClusters

MatousJobanek:
[KUBESAW-121](https://issues.redhat.com/browse/KUBESAW-121)

depends on https://github.com/codeready-toolchain/toolchain-common/pull/403


alexeykazakov (approved):

fbm3307 (approved):

MatousJobanek:
/retest
after merging cicd PR

mfrancisc (approved):
Nice Job 👍 

So basically this makes sure we disable/enable tls verification in the SR token based on the member tls configuration right ?

MatousJobanek:
yes, exactly. if the caBundle is not present and it's running in dev/e2e environment (which means with a self-signed certificates), then we expect that it should use insecure connection. So technically, the SR kubeconfig now matches what is used in TC client.

PR #1043, opened 2024-06-18
Annotate captcha assessments

rajivnathan:
Tests pending.


controllers/usersignup/usersignup_controller.go
```diff
@@ -756,6 +765,71 @@ func (r *Reconciler) ensureSpaceBinding(
...
+		// set the annotated assessment value: FRAUDULENT or LEGITIMATE
+		userSignup.Annotations[toolchainv1alpha1.UserSignupCaptchaAnnotatedAssessmentAnnotationKey] = newAnnotationName
+	}
+	logger.Info("Assessment annotated successfully", "assessment_annotation", newAnnotationName, "response", response.String())
+}
+
+// shouldAnnotateCaptchaAssessment determines whether the previous assessment should be annotated
+// returns LEGITIMATE or FRAUDULENT if the assessment should be annotated
+// returns ANNOTATION_UNSPECIFIED otherwise
+func (r *Reconciler) shouldAnnotateCaptchaAssessment(userSignup *toolchainv1alpha1.UserSignup, newState string) recaptchapb.AnnotateAssessmentRequest_Annotation {
```
alexeykazakov, line 815: The function name is a bit misleading, isn't it? It doesn't determine if it should be annotated or not but rather always returns a new/expected assessment annotation, right?

controllers/usersignup/usersignup_controller.go
```diff
@@ -756,6 +765,71 @@ func (r *Reconciler) ensureSpaceBinding(
...
+func (r *Reconciler) annotateCaptchaAssessment(ctx context.Context, userSignup *toolchainv1alpha1.UserSignup, newState string) {
+	logger := log.FromContext(ctx)
+	assessmentID, assessmentIDFound := userSignup.Annotations[toolchainv1alpha1.UserSignupCaptchaAssessmentIDAnnotationKey]
+	if !assessmentIDFound {
+		// there's no assessment ID to annotate
+		return
+	}
+
+	newAssessmentAnnotation := r.shouldAnnotateCaptchaAssessment(userSignup, newState)
+	if newAssessmentAnnotation == recaptchapb.AnnotateAssessmentRequest_ANNOTATION_UNSPECIFIED {
```
alexeykazakov, line 780: Shouldn't we also check if the `newAssessmentAnnotation` is the same as the current one from the UserSignup resource? And if it's so then it was already submitted to recaptcha so we can/should ignore it now? Just to avoid duplicating the same assessment annotation for the same users multiple times.

controllers/usersignup/usersignup_controller.go
```diff
@@ -756,6 +765,71 @@ func (r *Reconciler) ensureSpaceBinding(
...
+		return
+	}
+	defer rclient.Close()
+
+	annotateRequest := &recaptchapb.AnnotateAssessmentRequest{
+		Name:       assessmentID,
+		Annotation: newAssessmentAnnotation,
+	}
+
+	response, err := rclient.AnnotateAssessment(gctx, annotateRequest)
```
alexeykazakov, line 798: I guess we could do it a go routine so we don't wait for the recapcha client here?

controllers/usersignup/usersignup_controller.go
rajivnathan: I was using `ANNOTATION_UNSPECIFIED` as the base case where we don't actually do the annotation. In the other function we check whether it's `ANNOTATION_UNSPECIFIED` and if so we skip annotating the assessment. But I'm fine to change it if it's confusing. Do you have a suggestion to make it more clear?

controllers/usersignup/usersignup_controller.go
rajivnathan: Yeah, good point. I think it technically shouldn't happen because it's called from the setStateLabel function which does the `if oldState == state ` check before continuing but it's a good idea to check the new assessment against the old one in case we end up moving the code around in the future.

rajivnathan: 94e1342413e18a430f97c78dbf77314264d01ffe

controllers/usersignup/usersignup_controller.go
rajivnathan: Good idea 3e1681db3961ab2e855896d306a2a6e9d30b93fd

controllers/usersignup/usersignup_controller.go
rajivnathan: I renamed the function to `getCaptchaAssessmentAnnotation`, let me know if you think another name would be better

alexeykazakov (approved):
Looks good. Thanks!

PR #1048, opened 2024-06-22
Add feature toggle annotations to Space & NSTemplateSet

alexeykazakov:
Part of https://issues.redhat.com/browse/SANDBOX-653

This PR introduces feature lottery drawing for Spaces. If there is features defined in the Tier section of the Configuration then the corresponding annotation added for all winning features to the Space during its creation. This annotation is also propagated to NSTemplateSet.
See more details here: https://docs.google.com/document/d/1YWaR94JTrVBBVoMSjgIb8jb2FkqKiT1myVib9Qv0BzU/edit#bookmark=id.k3fvz630dsa6

Requires https://github.com/codeready-toolchain/toolchain-common/pull/409

The member operator part is here: https://github.com/codeready-toolchain/member-operator/pull/580
This PR and member operator PR are related but independent.

There is no e2e tests yet. We will think about adding some later when other pieces of the entire flow are addressed in following PRs.


controllers/space/space_controller_test.go
```diff
@@ -42,228 +42,259 @@ func TestCreateSpace(t *testing.T) {
 	require.NoError(t, err)
 	base1nsTier := tiertest.Base1nsTier(t, tiertest.CurrentBase1nsTemplates)
 	t.Run("success", func(t *testing.T) {
-		// given
```
alexeykazakov, line 45: It seems that there are a lot of changes in this test but it's actually almost the same test but wrapped with a table for two usecases:
- with a feature toggle annotation in the Space
- without any

controllers/space/space_controller_test.go
```diff
@@ -42,228 +42,259 @@ func TestCreateSpace(t *testing.T) {
...
+				name:           "with feature toggles",
+				featureToggles: "feature-1,feature-2",
+			},
+		}
+		for _, testRun := range tests {
+			t.Run(testRun.name, func(t *testing.T) {
+				spaceOptions := []spacetest.Option{spacetest.WithSpecTargetCluster("member-1")}
+				if testRun.featureToggles != "" {
+					spaceOptions = append(spaceOptions, spacetest.WithAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.featureToggles))
 				}
```
alexeykazakov, line 64: This is the first change in this test: Adding the feature annotation to the Space depending on the specific test from the table.

controllers/space/space_controller_test.go
```diff
@@ -42,228 +42,259 @@ func TestCreateSpace(t *testing.T) {
...
+					HasTierName(base1nsTier.Name).
+					HasClusterResourcesTemplateRef("base1ns-clusterresources-123456new").
+					HasNamespaceTemplateRefs("base1ns-code-123456new", "base1ns-dev-123456new", "base1ns-stage-123456new")
+				if testRun.featureToggles != "" {
+					nsTmplSetAssertion = nsTmplSetAssertion.
+						HasAnnotationWithValue(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.featureToggles)
+				} else {
+					nsTmplSetAssertion = nsTmplSetAssertion.
+						DoesNotHaveAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey)
+				}
```
alexeykazakov, line 99: This is the second change. Check the NsTemplateSet feature annotation if it's present in the Space.
The rest of this test is not changed and just formatted to be moved to the table test wrapper.

pkg/space/space.go
```diff
@@ -29,6 +32,42 @@ func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string
...
+	return s
+}
+
+// addFeatureToggles does "lottery" drawing for all feature toggles defined in the configuration according to their weights.
+// And it adds the corresponding feature annotation to the space for features that won and should be enabled for the space.
+func addFeatureToggles(space *toolchainv1alpha1.Space, config toolchainconfig.ToolchainConfig) {
+	toggles := config.Tiers().FeatureToggles()
+	var winners string
+	for _, t := range toggles {
+		weight := int(t.Weight())
```
xcoulon, line 49: the logic looks good, but `weight` is a bit confusing, could we use something like `threshold` instead?
(I believe `weight` would be used in case of load balancers, to distribute the load between targets)

pkg/space/space.go
```diff
@@ -29,6 +32,42 @@ func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string
...
+		// We generate a random number between 0 and 100. If the number is equal to or lower than the weight
+		// then the feature wins.
+		// We don't use recommended crypto/rand here because we don't need crypto grade random generator
+		// and math/rand with pseudo random numbers, which is much cheaper, is sufficient, so we are disabling the linter
+		if weight == 100 || (weight > 0 && rand.Intn(100) <= weight) { // nolint:gosec
+			// Winner!
+			if winners == "" {
+				winners = t.Name() // First winner so far
+			} else {
+				winners = fmt.Sprintf("%s,%s", winners, t.Name()) // Add to the comma separated list of existing winners
```
xcoulon, line 59: or you could append each enabled feature to `var features = []string{}` and use `strings.Join(features, ",")` afterwards

xcoulon (commented):
looks good overall, just comments about the var names (naming is hard! 😅)

pkg/space/space_test.go
```diff
@@ -30,6 +32,70 @@ func TestNewSpace(t *testing.T) {
...
+			default:
+				assert.Fail(t, "unknown feature", feature)
+			}
+		}
+	}
+
+	// then
+	assert.Equal(t, 100, myCoolFeatureCount)
+	assert.Equal(t, 0, myNotSoCoolFeatureCount)
+	assert.True(t, mySoSoFeatureCount > 0 && mySoSoFeatureCount < 100, "Feature with weight 50 should be present at least once but less than 100 times", mySoSoFeatureCount)
```
xcoulon, line 96: it _might_ happen then the `mySoSoFeatureCount` is 0 or 100 (bad luck) 🤔

pkg/space/space.go
```diff
@@ -29,6 +32,42 @@ func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string
...
+func addFeatureToggles(space *toolchainv1alpha1.Space, config toolchainconfig.ToolchainConfig) {
+	toggles := config.Tiers().FeatureToggles()
+	var winners string
+	for _, t := range toggles {
+		weight := int(t.Weight())
+		// We generate a random number between 0 and 100. If the number is equal to or lower than the weight
+		// then the feature wins.
+		// We don't use recommended crypto/rand here because we don't need crypto grade random generator
+		// and math/rand with pseudo random numbers, which is much cheaper, is sufficient, so we are disabling the linter
+		if weight == 100 || (weight > 0 && rand.Intn(100) <= weight) { // nolint:gosec
```
filariow, line 54: For better testability, we could modify this `addFeatureToggles` to accept a function that produces random numbers.

```suggestion
// NewSpaceWithFeatureToggles is the same as NewSpace() but also does a feature toggle lottery drawing
// and adds the corresponding feature annotations for features which "won" and should be enabled for the space.
func NewSpaceWithFeatureToggles(
	userSignup *toolchainv1alpha1.
	targetClusterName string,
	compliantUserName, tier string,
	config toolchainconfig.ToolchainConfig,
) *toolchainv1alpha1.Space {
	// We don't use recommended crypto/rand here because we don't need crypto grade random generator
	// and math/rand with pseudo random numbers, which is much cheaper, is sufficient, so we are disabling the linter
	return newSpaceWithFeatureTogglesAndRandomizer(userSignup, targetClusterName, compliantUserName, config, rand.Intn)  // nolint:gosec
}

func newSpaceWithFeatureToggles(
	userSignup *toolchainv1alpha1.UserSignup,
	targetClusterName string,
	compliantUserName, tier string,
	config toolchainconfig.ToolchainConfig, 
	randIntn func(maxValue int) int,
) *toolchainv1alpha1.Space {
	s := NewSpace(userSignup, targetClusterName, compliantUserName, tier)
	addFeatureToggles(s, config, randIntn)
	return s
}

// addFeatureToggles does "lottery" drawing for all feature toggles defined in the configuration according to their weights.
// And it adds the corresponding feature annotation to the space for features that won and should be enabled for the space.
func addFeatureToggles(space *toolchainv1alpha1.Space, config toolchainconfig.ToolchainConfig, randIntn func(maxValue int) int) {
	toggles := config.Tiers().FeatureToggles()
	var winners string
	for _, t := range toggles {
		weight := int(t.Weight())
		// We generate a random number between 0 and 100. If the number is equal to or lower than the weight
		// then the feature wins.
		if weight == 100 || (weight > 0 && randIntn(100) <= weight) {
```

With this change, in tests we will provide a mock function and test the behaviors we are interested in:
* `weight = 0`
* `weight = 100`
* `weight < randIntn(100)`
* `weight == randIntn(100)`
* `weight > randIntn(100)`

pkg/space/space.go
filariow: +1

pkg/space/space.go
alexeykazakov: But it's not a threshold. It's a weight :)

Term "weight" is wildly used when there is a need of influence the probability of choosing a specific path / making a decision.

pkg/space/space_test.go
alexeykazakov: Yes. There is a potential flakiness. But probability of that is close to 1 out of 50. Maybe we could increase it to 1 out of 500 without compromising the performance too much... Do you have any idea how we could test a weight which is not 0 or 100 with guaranteed result?

pkg/space/space.go
```diff
@@ -29,6 +32,42 @@ func NewSpace(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string
...
+func NewSpaceWithFeatureToggles(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string, compliantUserName, tier string, config toolchainconfig.ToolchainConfig) *toolchainv1alpha1.Space {
+	s := NewSpace(userSignup, targetClusterName, compliantUserName, tier)
+	addFeatureToggles(s, config)
+	return s
+}
+
+// addFeatureToggles does "lottery" drawing for all feature toggles defined in the configuration according to their weights.
+// And it adds the corresponding feature annotation to the space for features that won and should be enabled for the space.
+func addFeatureToggles(space *toolchainv1alpha1.Space, config toolchainconfig.ToolchainConfig) {
+	toggles := config.Tiers().FeatureToggles()
```
filariow, line 46: it seems to me that the only field from the `ToolchainConfig` we are interested in is the list of `FeatureToggle`. I'd propose to pass just it instead of the whole `config`.

```suggestion
func NewSpaceWithFeatureToggles(userSignup *toolchainv1alpha1.UserSignup, targetClusterName string, compliantUserName, tier string, toggles []toolchainconfig.FeatureToggle) *toolchainv1alpha1.Space {
	s := NewSpace(userSignup, targetClusterName, compliantUserName, tier)
	addFeatureToggles(s, toggles)
	return s
}

// addFeatureToggles does "lottery" drawing for all feature toggles defined in the configuration according to their weights.
// And it adds the corresponding feature annotation to the space for features that won and should be enabled for the space.
func addFeatureToggles(space *toolchainv1alpha1.Space, toggles []toolchainconfig.FeatureToggle) {
```

filariow (commented):
largely LGTM. Just some ideas for enhancing tests

pkg/space/space.go
xcoulon: okay... that's a matter of semantic, I'm not totally convinced by the "weight" term. As you said, it's used to choose a path (in load balancers, for example). Here, it looks like it's a minimal _score_ to get the feature enabled.
But naming is hard, I won't block the PR for this somehow minor point.

pkg/space/space_test.go
xcoulon: > Do you have any idea how we could test a weight which is not 0 or 100 with guaranteed result?

```
mySoSoFeatureCount >= 0 && mySoSoFeatureCount =< 100
```
😂

controllers/space/space_controller.go
```diff
@@ -387,6 +386,15 @@ func NewNSTemplateSet(namespace string, space *toolchainv1alpha1.Space, bindings
...
 	nsTmplSet.Spec = NewNSTemplateSetSpec(space, bindings, tmplTier)
+
+	// propagate feature annotation from the space
+	winners, found := space.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	if found {
+		if nsTmplSet.Annotations == nil {
+			nsTmplSet.Annotations = make(map[string]string)
+		}
+		nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey] = winners
+	}
```
MatousJobanek, line 397: This will work only for the new NSTemplateSets, not for the already existing ones. If I want to enable/disable the feature toggle for some Space, then this should get propagated as well, but since we compare only 
https://github.com/codeready-toolchain/host-operator/blob/012f0cdf58e4fe53e37a2122ef5eb176009527ad/controllers/space/space_controller.go#L335-L336
then the annotation change won't get there.
Let's also make sure that it's tested in unit tests

pkg/space/space.go
metlos: +1 for weight from my side. As @alexeykazakov said, "weight" is widely used to influence the "importance" of some item in some list (i.e. weighted average, etc.).

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -255,7 +279,7 @@ func TestUserSignupCreateSpaceAndSpaceBindingOk(t *testing.T) {
 				// then
 				require.Equal(t, reconcile.Result{}, res)
 				switch testname {
-				case "without skip space creation annotation", "with skip space creation annotation set to false", "with social event":
+				case "without skip space creation annotation", "with skip space creation annotation set to false", "with social event", "with feature toggles":
```
mfrancisc, line 282: should we check that the annotation is still there after the second reconcile ?

mfrancisc (commented):
Great job 👏 

I have only few minor comments.

pkg/space/space.go
mfrancisc: +1 nice idea!

alexeykazakov: But in this case the test coverage would be lower because we are not testing the default random number generator which would be used in production. Unless I'm missing something.
What is the benefit of such approach?

controllers/toolchainconfig/configuration.go
```diff
@@ -318,6 +322,26 @@ func (d TiersConfig) DurationBeforeChangeTierRequestDeletion() time.Duration {
...
 
+func (d TiersConfig) FeatureToggles() []FeatureToggle {
+	toggles := make([]FeatureToggle, 0, len(d.tiers.FeatureToggles))
+	for _, t := range d.tiers.FeatureToggles {
+		toggles = append(toggles, FeatureToggle{t})
+	}
+	return toggles
+}
+
+type FeatureToggle struct {
```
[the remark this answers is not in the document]
alexeykazakov: We need a GetWieght() wrapper so we can enforce the default 100 value.

pkg/space/space_test.go
xcoulon: > Actually, probability of so-so feature getting enabled 100 times out of 100 is 0.5^100.

I believe it's (1/100)^100, but in any case, we could definitely call it "bad luck" if it happened. So yeah, let's keep the code as-is!

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -206,14 +206,27 @@ func TestUserSignupCreateSpaceAndSpaceBindingOk(t *testing.T) {
...
 		t.Run(testname, func(t *testing.T) {
 			// given
 			defer counter.Reset()
 
 			mur := newMasterUserRecord(userSignup, "member1", deactivate30Tier.Name, "foo")
 			mur.Labels = map[string]string{toolchainv1alpha1.MasterUserRecordOwnerLabelKey: userSignup.Name}
-			r, req, _ := prepareReconcile(t, userSignup.Name, spaceProvisionerConfig, userSignup, mur, baseNSTemplateTier, base2NSTemplateTier, deactivate30Tier, deactivate80Tier, event)
+
+			config := &toolchainv1alpha1.ToolchainConfig{}
+			switch testname {
```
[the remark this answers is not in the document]
alexeykazakov: I agree. I don't like the current approach either. But I didn't want to refactor these tests in this PR. Just adding new test case to existing ugly tests. We can refactor it in a separate PR.

pkg/space/space.go
xcoulon: again, I believe that "weight" is used when there's a choice between multiple paths, but it's fine, let's not waste too much time on this term.

pkg/space/space_test.go
alexeykazakov: The weight is 50. Meaning the probability of a single draw getting us "enabled" is 0.5
If we draw it 100 times then it's 0.5^100, isn't?

pkg/space/space.go
filariow: Yeah. you're right. We won't test the code running in production in these unit tests and coverage will be lower.

However, for unit tests that may be good. To me here we should be mostly interested into testing that given different values for `weight` and `randIntn` the logic behave as we expect. So long we ensure that `addFeatureToggles` is using the values generated from `randIntn func(maxValue int) int` as expected, and we know that `rand.Intn` is generating numbers following a distribution that's good for us, I'd say we are ok. 
Alternatively, we may fall into testing that `rand.Intn` is returning values that respects some criteria (distribution) we define, instead of testing the `addFeatureToggles` logic itself, and we may risk to have flaky tests.

Using the approach proposed here, we'll be more precise into testing our logic and we'll left out only a wrapper that's configuring our logic to use the go standard random number generator.  Indeed, we'll be able to test the `addFeatureToggles` logic independently from implementation details about how we generate the random numbers and we'll be able to test with precision the scenarios in which `weight < randIntn(100)`, `weight == randIntn(100)`, and `weight > randIntn(100)`.

I'm trying to think to a valid approach for testing the default behavior (`addFeatureToggles` + `math/rand.Intn`), but I'm not having any good idea.

filariow: Maybe we can implement the `mySoSoFeature` test to cover `addFeatureToggles` + `math/rand.Intn`? I mean this test case [here](https://github.com/codeready-toolchain/host-operator/blob/012f0cdf58e4fe53e37a2122ef5eb176009527ad/pkg/space/space_test.go#L96)

alexeykazakov: But if we make a mistake and instead of passing something else instead of `rand.Intn()` to that function in the runtime code then this test won't catch it.
So I still believe the current approach is superior because it gives a better result: catches more potential problems. And the risk of a flaky test here is close to 0.5^100 * 2. Which realistically speaking means 0.

filariow: That's good for me. Btw, if I'm not mistaken, refactoring we'll be able to test the behavior with and without `rand.Intn`, so we should be able to explicitly test the same behaviors we're already testing plus 3 more test cases (somehow implicitly tested if we do not refactor). So, when it comes to the completeness of tests, I feel like it's mostly equivalent.

That said, it seems to me that the difference between the two approaches is not that big, and I'd say we can just choose the one that we prefer without any harm.

pkg/space/space.go
alexeykazakov: Yep. That would look better. Done in https://github.com/codeready-toolchain/host-operator/pull/1048/commits/44ecb0a8f2e403069dfdc5de71403e845cca05f5

pkg/space/space.go
alexeykazakov: https://github.com/codeready-toolchain/host-operator/pull/1048/commits/461d542c5a2a88392e9d0c4caa00e0ad225a1fe4

controllers/usersignup/usersignup_controller_test.go
alexeykazakov: Added here: https://github.com/codeready-toolchain/host-operator/pull/1048/commits/6951200098796af4e8846e7dd9603bdeb9ac2931

controllers/space/space_controller.go
alexeykazakov: Great catch! Addressed in https://github.com/codeready-toolchain/host-operator/pull/1048/commits/601661b0569d662ebf2e3d9cd988fe8abc40e4cc

controllers/toolchainconfig/configuration.go
```diff
@@ -64,6 +64,10 @@ func ForceLoadToolchainConfig(cl runtimeclient.Client) (ToolchainConfig, error)
 	return newToolchainConfig(config, secrets), nil
 }
 
+func NewToolchainConfig(config *toolchainv1alpha1.ToolchainConfigSpec, secrets map[string]map[string]string) ToolchainConfig {
```
mfrancisc, line 67: this function doesn't seem to be used, unless I'm missing something

mfrancisc (approved):
Great Job 🚀 

Looks good to me!

controllers/toolchainconfig/configuration.go
alexeykazakov: Yep, you are right. It's not used anymore. Removed in https://github.com/codeready-toolchain/host-operator/pull/1048/commits/d53e5d4dfb9c7b9be812223f846f241a3107c3cd

rajivnathan (approved):
Looks great 👍

controllers/space/space_controller_test.go
```diff
@@ -1180,6 +1211,100 @@ func TestUpdateSpaceTier(t *testing.T) {
...
+					s.Annotations = make(map[string]string)
+					s.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey] = testRun.newSpaceFeatures
+				} else {
+					delete(s.Annotations, toolchainv1alpha1.FeatureToggleNameAnnotationKey)
+				}
+				err = hostClient.Update(context.TODO(), s)
+				require.NoError(t, err)
+
+				_, err = ctrl.Reconcile(context.TODO(), requestFor(s))
+				require.NoError(t, err)
```
MatousJobanek, line 1291: Calling the reconcile twice as part of one unit-test is not ideal (for the sake of isolation, and also for the sake of consistency). 
You can either split it into two tests (parent and sub-test), but this would mean that the first test duplicates one from the top. 
Or modify the setup so it defines the outdated NSTempateSet explicitly - it's easy to understand the setup and the purpose of the test, and it's consistent with the rest of the unit tests:
```suggestion
				// given
				spaceOptions := []spacetest.Option{spacetest.WithSpecTargetCluster("member-1")}
				if testRun.newSpaceFeatures != "" {
					spaceOptions = append(spaceOptions, spacetest.WithAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.newSpaceFeatures))
				}

				s := spacetest.NewSpace(test.HostOperatorNs, "oddity", spaceOptions...)

				nsTemplateSetOptions := []nstemplatetsettest.Option{nstemplatetsettest.WithReadyCondition(), nstemplatetsettest.WithReferencesFor(base1nsTier)}
				if testRun.oldSpaceFeatures != "" {
					nsTemplateSetOptions = append(nsTemplateSetOptions, nstemplatetsettest.WithAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.oldSpaceFeatures))
				}
				nsTemplateSet := nstemplatetsettest.NewNSTemplateSet("oddity", nsTemplateSetOptions...)
				hostClient := test.NewFakeClient(t, s, base1nsTier)
				member1Client := test.NewFakeClient(t, nsTemplateSet)
				member1 := NewMemberClusterWithClient(member1Client, "member-1", corev1.ConditionTrue)
				ctrl := newReconciler(hostClient, member1)

				// when
				_, err = ctrl.Reconcile(context.TODO(), requestFor(s))
				require.NoError(t, err)
```

MatousJobanek: you can also add a check that the NSTemplateSet is not updated when it's not supposed to be:
```go
				if testRun.oldSpaceFeatures == testRun.newSpaceFeatures {
					member1Client.MockUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
						return fmt.Errorf("shouldn't be called")
					}
				}
```

MatousJobanek: so the whole proposed code is:
```suggestion
				// given
				spaceOptions := []spacetest.Option{spacetest.WithSpecTargetCluster("member-1")}
				if testRun.newSpaceFeatures != "" {
					spaceOptions = append(spaceOptions, spacetest.WithAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.newSpaceFeatures))
				}

				s := spacetest.NewSpace(test.HostOperatorNs, "oddity", spaceOptions...)

				nsTemplateSetOptions := []nstemplatetsettest.Option{nstemplatetsettest.WithReadyCondition(), nstemplatetsettest.WithReferencesFor(base1nsTier)}
				if testRun.oldSpaceFeatures != "" {
					nsTemplateSetOptions = append(nsTemplateSetOptions, nstemplatetsettest.WithAnnotation(toolchainv1alpha1.FeatureToggleNameAnnotationKey, testRun.oldSpaceFeatures))
				}
				nsTemplateSet := nstemplatetsettest.NewNSTemplateSet("oddity", nsTemplateSetOptions...)
				hostClient := test.NewFakeClient(t, s, base1nsTier)
				member1Client := test.NewFakeClient(t, nsTemplateSet)
				if testRun.oldSpaceFeatures == testRun.newSpaceFeatures {
					member1Client.MockUpdate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.UpdateOption) error {
						return fmt.Errorf("shouldn't be called")
					}
				}
				member1 := NewMemberClusterWithClient(member1Client, "member-1", corev1.ConditionTrue)
				ctrl := newReconciler(hostClient, member1)
```

MatousJobanek (approved):
Just one comment regarding the unit test; otherwise, it looks good :rocket:

controllers/space/space_controller_test.go
alexeykazakov: Yes, I was using the first reconcile just as part of the setup. Not the test itself. My comment could be clearer I guess.
Anyway thanks for the suggestion. It's a bit clearer what the tests is doing here I think.

PR #1052, opened 2024-06-25
KUBESAW-124: Makefile target for dependency validations

fbm3307:
This is to create makefile target for dependency validations in each reporsitory so that this can be used to verify if there are any compilation issues after updating the dependencies of toolchain-common/api

Similar PRs
- Toolchain-Common - https://github.com/codeready-toolchain/toolchain-common/pull/410
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/581 
- Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/440 
- Toolchain-E2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1002
- KSTCL - https://github.com/kubesaw/ksctl/pull/43


mfrancisc (commented):
Looks great 👍 

Should we add `lint-go-code` as well ?

MatousJobanek (approved):

MatousJobanek:
> Should we add lint-go-code as well ?

good idea

fbm3307:
Added it

make/verify-dependencies.mk
```diff
@@ -0,0 +1,12 @@
+.PHONY: verify-dependencies
+## Runs commands to verify after the updated dependecies of toolchain-common/API(go mod replace), if the repo needs any changes to be made
+verify-dependencies: tidy generate vet build test lint
```
MatousJobanek, line 3: for this particular case, we should use go lint only
```suggestion
verify-dependencies: tidy generate vet build test lint-go-code
```
also, I don't think that the yaml files are lint-compatible right now

fbm3307: Yes.. i figured when i was testing it locally, will change it .

fbm3307: updated it

mfrancisc: yeah, as suggested only `lint-go-code` , the yaml linter might fail I've observed.

mfrancisc (approved):
Looks good 👍

MatousJobanek (approved):

make/verify-dependencies.mk
```diff
@@ -0,0 +1,12 @@
...
+## Runs commands to verify after the updated dependecies of toolchain-common/API(go mod replace), if the repo needs any changes to be made
+verify-dependencies: tidy generate vet build test lint-go-code
+
+.PHONY: tidy
+tidy: 
+	go mod tidy
+
+.PHONY: vet
+vet:
+	go vet ./...
```
MatousJobanek, line 11: btw, all these commands are related to go - you could maybe add it to the `go.mk` file instead of creating a new one - to not split the related targets between two files

fbm3307: done

PR #1056, opened 2024-07-04
Receive the updates from the API repo for the new fields in the ToolchainCluster status

metlos:
This trivially just updates the API and gets the updated CRD definitions.

Associated PRs:
- member-operator: https://github.com/codeready-toolchain/member-operator/pull/583
- api: https://github.com/codeready-toolchain/api/pull/433

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -129,4 +129,4 @@ require (
 
 go 1.20
 
-replace github.com/codeready-toolchain/api => github.com/metlos/api v0.0.0-20240703104250-40b2c66a959f
+replace github.com/codeready-toolchain/api => github.com/metlos/api v0.0.0-20240704185630-4a086a7f3529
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -128,3 +128,5 @@ require (
 )
 
 go 1.20
+
+replace github.com/codeready-toolchain/api => github.com/metlos/api v0.0.0-20240708084953-3d53a2121f9f
```
MatousJobanek, line 132: we should update the dependency as a separate PR - together with the changes in the code; otherwise, you merge api PR and then you would need to wait for another round of e2e tests. 
So ideally, in such PRs when we introduce only a new fields so it stays backward compatible with the logic in the code, the change would contain only the result of the `make generate` command.

MatousJobanek (approved):
I would separate update of the api from the changes in CRD to make the PR merges more fluent

PR #1061, opened 2024-07-19
Add annotation changed predicate for Space and NSTemplateSet watch

alexeykazakov:
To properly auto-update NSTemplateSets when we enable or disable a feature for Spaces we need to watch for Space and NSTemplateSet annotations. This PR adds `AnnotationChangedPredicate` to the Space and NSTemplateSet watchers in the Space controller.

This PR should unblock https://github.com/codeready-toolchain/toolchain-e2e/pull/1008 which is currently failing because of this. I can't pair https://github.com/codeready-toolchain/toolchain-e2e/pull/1008 to this PR though because it's already paired with https://github.com/codeready-toolchain/member-operator/pull/580

### DIFF
--- a/controllers/space/space_controller.go
+++ b/controllers/space/space_controller.go
@@ -47,10 +47,10 @@ type Reconciler struct {
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
 	b := ctrl.NewControllerManagedBy(mgr).
 		// watch Spaces in the host cluster
-		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
 		Watches(&source.Kind{Type: &toolchainv1alpha1.NSTemplateTier{}},
 			handler.EnqueueRequestsFromMapFunc(MapNSTemplateTierToSpaces(r.Namespace, r.Client)),
-			builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
 		Watches(&source.Kind{Type: &toolchainv1alpha1.SpaceBinding{}},
 			handler.EnqueueRequestsFromMapFunc(MapSpaceBindingToParentAndSubSpaces(r.Client)))
 	// watch NSTemplateSets in all the member clusters
--- a/controllers/space/space_controller.go
+++ b/controllers/space/space_controller.go
@@ -47,10 +47,10 @@ type Reconciler struct {
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
 	b := ctrl.NewControllerManagedBy(mgr).
 		// watch Spaces in the host cluster
-		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
+		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}))).
 		Watches(&source.Kind{Type: &toolchainv1alpha1.NSTemplateTier{}},
 			handler.EnqueueRequestsFromMapFunc(MapNSTemplateTierToSpaces(r.Namespace, r.Client)),
-			builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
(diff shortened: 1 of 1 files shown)


ranakan19 (approved):

alexeykazakov:
I'm not sure how this change affects the migration tests but it looks like there is conflict of two usersignups.
The test fails because it can't find MUR with name `mig-prov`. There are the following resources:
- UserSignup with name `mig-prov` but with `mig-prov-2` complaint username. With Ready status.
- MUR with name `mig-prov-2` (owned by the usersignup from above ^^^). With Ready status.
- Space with name `mig-prov-2` (obviously created for the same usersignup from above ^^^). With Ready status.
- Space with name `mig-prov` (there is no usersingup with that complaint username or MUR with that username). With Ready status.

So it looks like there was UserSignup with name `mig-prov` and complaint name `mig-prov` created at some point. But before it could be fully deleted another one was created and this is why the second one has `mig-prov-2` complained username.

I see that the UserSignup with `mig-prov` name is created twice.
- Once here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/migration/setup/setup_migration_test.go#L19
- And here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/migration/verify/verify_migration_test.go#L62

So I'm not quite sure how it's supposed to work. Signing up the same user multiple times without cleanup will cause such conflicts. What am I missing?

rajivnathan (approved):

controllers/space/space_controller.go
```diff
@@ -47,10 +47,10 @@ type Reconciler struct {
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, memberClusters map[string]cluster.Cluster) error {
 	b := ctrl.NewControllerManagedBy(mgr).
 		// watch Spaces in the host cluster
-		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
```
MatousJobanek, line 50: are you sure that this performs the logical `OR` operation on the predicates? I mean that the reconciliation is triggered either when the generation number is changed or when there is any changed in the annotations?

MatousJobanek: Taking a quick glance at the controller-runtime code, this looks more like it does the logical `AND` operation over the set of predicates - it filters out all events that don't pass at least one of the provided predicates. In other words, the reconciliation is triggered only when the generation number is changed and at the same time when there is some change in the annotations. https://github.com/kubernetes-sigs/controller-runtime/blob/6ffcf19a340d9e2ee7b7738114c4d256f67d5337/pkg/source/internal/eventsource.go#L87-L91

That being said, we will need to create our own predicate for this - something similar to what we have for UserSignups https://github.com/codeready-toolchain/host-operator/blob/master/controllers/usersignup/predicate.go

MatousJobanek: or there is even better solution - there is `predicate.Or` function
```suggestion
		For(&toolchainv1alpha1.Space{}, builder.WithPredicates(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}))).
```

controllers/space/space_controller.go
```diff
@@ -47,10 +47,10 @@ type Reconciler struct {
...
 		Watches(&source.Kind{Type: &toolchainv1alpha1.NSTemplateTier{}},
 			handler.EnqueueRequestsFromMapFunc(MapNSTemplateTierToSpaces(r.Namespace, r.Client)),
-			builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
```
MatousJobanek, line 53: why do you need to change the predicate for NSTemplateTier CRD?

controllers/space/space_controller.go
rsoaresd: I tested now with `OR` and it passed:

```
For(&toolchainv1alpha1.Space{}, builder.WithPredicates(
			predicate.Or(
				predicate.GenerationChangedPredicate{},
				predicate.AnnotationChangedPredicate{},
			))).
```

alexeykazakov: I should have checked it... I just assumed it's `OR` by default :grimacing:

controllers/space/space_controller.go
alexeykazakov: In case the annotation is deleted (or added) by accident on the NSTemplateSet side. Then the space controller would revert it. It should not happen because we don't actually usually touch NSTemplateSets manually but I guess that wouldn't hurt.

MatousJobanek: but you modified the watch of the NSTemplateTier, not NSTemplateSet. 
We do watch literally all changes in NSTemplateSet CRs (we don't have any predicate there) - see https://github.com/alexeykazakov/host-operator/blob/d8fa9d5139bb2294a840f089e2a37bb4c244a77d/controllers/space/space_controller.go#L57-L61

alexeykazakov: Oh... yes, it's `NSTemplateTier` indeed! Let me revert this. Good catch! https://github.com/codeready-toolchain/host-operator/pull/1063

PR #1068, opened 2024-07-26
remove webConsolePlugin from ToolchainClusterConfig

xcoulon:
Pendo has been decommisioned, we don't need this plugin anymore

see also:
- https://github.com/codeready-toolchain/api/pull/438
- https://github.com/codeready-toolchain/toolchain-common/pull/420 
- https://github.com/codeready-toolchain/toolchain-e2e/pull/1026 (standalone)
- 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,8 @@ module github.com/codeready-toolchain/host-operator
 
 require (
 	cloud.google.com/go/recaptchaenterprise/v2 v2.13.0
-	github.com/codeready-toolchain/api v0.0.0-20240717145630-bb67a632867a
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20240716065433-8604fe46b96a
+	github.com/codeready-toolchain/api v0.0.0-20240801160201-ed4387f19137
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20240801160555-2f18a82d9752
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v1.4.1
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -31,6 +31,10 @@ require (
 	sigs.k8s.io/controller-runtime v0.13.0
 )
 
+replace github.com/codeready-toolchain/api => github.com/xcoulon/api v0.0.0-20240725145329-0fc7541fe19e
+
+replace github.com/codeready-toolchain/toolchain-common => github.com/xcoulon/toolchain-common v0.0.0-20240725145759-056fcf98e945
```
alexeykazakov, line 36: @xcoulon as you know we don't update CRDs and dependencies to api/common in the same PR. The reason is that you can't merge api and CRD changes in this same time if you do it.
You would have to:

1. merge API
2. change your operator/CRD PR to update the dependency (remove `replace`)
3. wait for the operator/CRD PR to get green
4. merge it

steps 2 and 3 can take time (especially if there are problems with e2e tests, like infra outage) and you risk blocking others.
It's much safer to update CRDs only, so you can merge the API and the operator PR in the same time and then create a new PR to update the dependency. So you don't block anyone who might need to make a different change to API.

alexeykazakov (commented):
Looks good overall but please do not update the go.mod dependency in this PR.

go.mod
xcoulon: ok, got it! (I totally forgot about these constraints)

here are the PR which only deal with the removal in the CRD:
https://github.com/codeready-toolchain/host-operator/pull/1069
https://github.com/codeready-toolchain/member-operator/pull/589
https://github.com/codeready-toolchain/toolchain-e2e/pull/1021

ranakan19 (approved):

alexeykazakov (approved):

PR #1071, opened 2024-08-06
Update api dependency for Autoscler Buffer CPU config

alexeykazakov:
Required for https://github.com/codeready-toolchain/member-operator/pull/592, so the CPU configuration can be propagated from ToolchainConfig (host) to MemberOperatorConfig (member)


MatousJobanek (approved):
/lgtm

PR #1072, opened 2024-08-09
changes suggested by lint check

ranakan19:
Separate PR with linter changes for easier review, otherwise https://github.com/codeready-toolchain/host-operator/pull/1057 would have been even bigger to review.

pre-req - https://github.com/codeready-toolchain/host-operator/pull/1073 to be merged


pkg/metrics/metrics_test.go
```diff
@@ -27,7 +27,7 @@ func TestInitGauge(t *testing.T) {
 	m.Set(22)
 
 	// then
-	assert.Equal(t, float64(22), promtestutil.ToFloat64(m))
+	assert.InDelta(t, float64(22), promtestutil.ToFloat64(m), 0)
```
alexeykazakov, line 30: I'm curios why linter doesn't like `Equal()` here... :)

alexeykazakov (approved):
Looks good!

pkg/metrics/metrics_test.go
```diff
@@ -39,8 +39,8 @@ func TestInitGaugeVec(t *testing.T) {
 	m.WithLabelValues("member-2").Set(2)
 
 	// then
-	assert.Equal(t, float64(1), promtestutil.ToFloat64(m.WithLabelValues("member-1")))
-	assert.Equal(t, float64(2), promtestutil.ToFloat64(m.WithLabelValues("member-2")))
+	assert.InDelta(t, float64(1), promtestutil.ToFloat64(m.WithLabelValues("member-1")), 0)
+	assert.InDelta(t, float64(2), promtestutil.ToFloat64(m.WithLabelValues("member-2")), 0)
```
MatousJobanek, line 43: This doesn't test the same - the value should be `2`, not something between `0` and `2`. 
So I have the same question as Alexey - why is the `Equal` so problematic? We use it all over the place. And why the Delta makes is better? I see it rather as a regression.

MatousJobanek (commented):
Could you please revert the InDelta changes back to Equal?

pkg/metrics/metrics_test.go
xcoulon: the testifylint linter complains about the usage of `Equal` for float numbers and suggests we use `InDelta` or `InEpsilon` instead: https://github.com/Antonboom/testifylint?tab=readme-ov-file#float-compare

The `delta` arg is the margin of error of the value to check, not the lower-bound of a range (eg: `[0,2]`)
> InDelta asserts that the two numerals are within delta of each other.

We could also use this [wrapper func](https://github.com/codeready-toolchain/host-operator/blob/master/test/metric.go#L11), instead

xcoulon (approved):

pkg/metrics/metrics_test.go
alexeykazakov: If I'm reading it right then from the technical perspective those two assertion are 100% identical:
- `assert.Equal(t, float64(1), promtestutil.ToFloat64(m.WithLabelValues("member-1")))`
- `assert.InDelta(t, float64(1), promtestutil.ToFloat64(m.WithLabelValues("member-1")), 0)`

First checks if two float64 numbers are equal. The second check if two float64 numbers are within a delta `0` which means exactly equal. So it's just a trick to make the linter happy.
If we wanted to avoid the potential floating point rounding issues we would need to use something like
`assert.InDelta(t, float64(1), promtestutil.ToFloat64(m.WithLabelValues("member-1")), 0.0001)`

I didn't look very closely but It looks like we shouldn't really have a rounding problem in this case. We use floats (because of how Prometheus metrics work I guess?) but they really always round to integers. So I would rather use `Equal()` and suppress the linter so the code is easy to understand.
If for whatever reason we can expect the the actual metric can be slightly off (`1.0001` instead of `1.0`) for example then I would use `InDelta()` with some small delta. But I doubt we need it.

ranakan19: > So I would rather use Equal() and suppress the linter so the code is easy to understand.

I'd say using the wrapper function xavier suggested would keep the code easy to understand without suppressing the linter.
Or we could also use CollectAndCount which returns an int instead of float.
`assert.Equal(t, 1, promtestutil.CollectAndCount(m))`

ranakan19: @alexeykazakov I was thinking something similar when i wrote the delta to be 0 expecting that the two numbers would be exactly equal because of the rounding off. 
On reading further I understand that the concern here is the way a float variable is represented in memory i.e even though we expect the value to be 2 - it could be stored in memory as 1.999999999 as the closest float to 2. This explains:
1. Why the linter expects to use a InDelta or Inepsilon for comparing floats.
2. In the function InDelta/InEpsilon - the value of delta/epsilon should not be zero (it goes against the whole concept of using these separate functions for comparing float variable) 

So looks like wrapper function would indeed be the best solution here.

MatousJobanek: sorry about the confusion in my comment. I completely misread/misunderstood the `InDelta` function from the GH UI (without inspecting it locally). For some reason I thought that you set the high and low number, not the number and the delta from which it can differ.

MatousJobanek (approved):

pkg/metrics/metrics_test.go
MatousJobanek: I would prefer having the Equal there, but it's fine as it is - I'll keep the syntax of the `InDelta` assertion in mind for the future :-) 
I approved the PR

alexeykazakov: I think `assertion.Equal()` should handle floating point numbers issues OOTB :) But since it doesn't (I assume it doesn't) then yeah, let's use the `AssertMetricsCounterEquals()` wrapper.

ranakan19: oh well, looks like using the wrapper won't be straightforward - because of [this](https://github.com/codeready-toolchain/host-operator/blob/master/test/counter.go#L13) there is import cycling dependency - so will have to rearrange a few things.

alexeykazakov: Let's maybe move the wrappers to common?

ranakan19: yeah sure, do you want me to that in this same PR or maybe a separate one would be better because the wrapper is used in a lot of places.

alexeykazakov: A separate PR would be easy to handle I think.

ranakan19: here is the PR - https://github.com/codeready-toolchain/host-operator/pull/1073 
once those are merged I will update the current PR as well accordingly.

PR #1086, opened 2024-09-12
avoid index out-of-bounds in toolchainstatus controller

sadlerap:
Modifying the members array while we're indexing through it with a range is an unsafe operation.  Instead, create a new array and assign it after processing all elements.


alexeykazakov (approved):
Looks good! We could clean up the unit tests which check different cases of removed clusters though (there is a lot of code duplication). But let's address it in a different PR.
Thanks!

PR #1087, opened 2024-09-19
KUBESAW-170: have a unique label for our OLM based operator

fbm3307:
Have a unique label instead of default controller-manger, so that it becomes easier to perform operations to only our operators


MatousJobanek (approved):
/lgtm

PR #1099, opened 2024-10-10
format CRD description in CSV

MatousJobanek:
to match the format after running `make bundle`


metlos (approved):
Is there something to fix in the `make generate` target in the api so that we don't get this diff in the future?

PR #1113, opened 2024-12-16
feat: add ingress network policy with generic sandbox label

mfrancisc:
This PR introduces an ingress network policy with a generic sandbox label that can be used on any operator namespace that needs to communicate with the user namespace.


deploy/templates/nstemplatetiers/base/ns_dev.yaml
```diff
@@ -227,5 +227,21 @@ objects:
     policyTypes:
       - Ingress
 parameters:
+- name: SPACE_NAME
+  required: true
```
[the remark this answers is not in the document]
mfrancisc: ops .. thanks for catching this!

deploy/templates/nstemplatetiers/intellarge/ns_dev.yaml
```diff
@@ -252,6 +252,19 @@ objects:
                 kubernetes.io/metadata.name: redhat-ods-applications
     policyTypes:
       - Ingress
+- apiVersion: networking.k8s.io/v1
```
alexeykazakov, line 255: BTW, we need to delete all intel tiers :) But let's do it in a separate PR.

deploy/templates/nstemplatetiers/test/ns_dev.yaml
```diff
@@ -197,6 +197,19 @@ objects:
     podSelector: {}
     policyTypes:
     - Ingress
+- apiVersion: networking.k8s.io/v1
```
alexeykazakov, line 200: We don't have to modify the test tier. But it won't heart to add it here too.

alexeykazakov (approved):

deploy/templates/nstemplatetiers/intellarge/ns_dev.yaml
mfrancisc: got it, I'll open a new one, once those are merged!

deploy/templates/nstemplatetiers/test/ns_dev.yaml
mfrancisc: ok, makes sense. I wasn't sure, just wanted to be consitent.

deploy/templates/nstemplatetiers/intellarge/ns_dev.yaml
mfrancisc: should we keep them , just in case they might be used by the intel team for their own kubesaw instance ?

alexeykazakov: We can always revert the cleanup PRs later if needed. But most likely we won't need these tiers anytime soon.

mfrancisc:
/retest

update e2e PR

mfrancisc:
/retest

fixed test

PR #1115, opened 2025-01-06
KUBESAW-248: Drop NSTemplateTier.Status.Updates field

fbm3307:
This PR is to remove the NSTemplateTier.Status.Updates field(already optional in API) and its related code since it's not used anywhere

Paired and Related PR
- Toolchain-E2E - https://github.com/codeready-toolchain/toolchain-e2e/pull/1088


ranakan19 (approved):
LGTM

test/nstemplatetier/assertion.go
```diff
@@ -1,72 +0,0 @@
...
-}
-
-// AssertThatNSTemplateTier helper func to begin with the assertions on an NSTemplateTier
-func AssertThatNSTemplateTier(t test.T, name string, client runtimeclient.Client) *Assertion {
-	return &Assertion{
-		client:         client,
-		namespacedName: test.NamespacedName(test.HostOperatorNs, name),
-		t:              t,
-	}
-}
```
MatousJobanek, line 37: I know that the custom assertion doesn't make much sense when the status-related checks are removed, but keep in mind that the assertion is used & extended by the @mfrancisc's PR, so it would be fine to keep at least these three functions to minimize the conflicts in his PR

MatousJobanek (approved):
Thanks

test/nstemplatetier/assertion.go
fbm3307: yes thanks for this,  I did try to keep them, but the linter complained, so removed it , but can bring it back

mfrancisc: @MatousJobanek thanks for the heads up! 

@fbm3307 In case this is causing you issues, feel free to drop it. I can add it back , in case you merge this first 👍

fbm3307: @MatousJobanek  looks like the required golang ci test case failed here, i guess we shouldn't force merge ?@mfrancisc can add it in his PR ? WDYT?

PR #1116, opened 2025-01-07
KUBESAW-250: Updating GolangCiLint to v1.63.1

fbm3307:
Updating the GolangciLint to v1.63.1 across all repos fix the [Linter error](https://github.com/codeready-toolchain/host-operator/actions/runs/12649616851/job/35246418292?pr=1116#step:5:28) which were caught as a result of upgrading the linter

Similar PRs

- API - https://github.com/codeready-toolchain/api/pull/454
- Toolchain-common - https://github.com/codeready-toolchain/toolchain-common/pull/442
- Member-Operator - https://github.com/codeready-toolchain/member-operator/pull/612
- Registration-Service - https://github.com/codeready-toolchain/registration-service/pull/491
- Toolchain E2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1092
- Ksctl - https://github.com/kubesaw/ksctl/pull/97


MatousJobanek (approved):

pkg/capacity/manager.go
```diff
@@ -21,7 +21,7 @@ type (
 
 func hasNotReachedMaxNumberOfSpacesThreshold(counts counter.Counts) spaceProvisionerConfigPredicate {
 	return func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
+		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster]) // nolint:gosec
```
fbm3307, line 24: This fix is for the linter error . This error is thrown for integer to uint conversion
"pkg/capacity/manager.go:24:25: G115: **integer overflow conversion int -> uint (gosec)**
  		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])"
  		
  There are few other places where this error occurs, there is a fix for this as per [this issue ](https://github.com/securego/gosec/issues/1185)but it seems its not useful for us, There can be other option of using only Int or UInt and change it everywhere, i chose to use ignore this, if otherwise will try to see more for converting the whole thing

fbm3307 (commented):
The places where i have written  **// nolint:gosec** , it bcoz the linter from 1.60 and above are throwing **integer overflow conversion uint->Int** and vice versa error , and the fix they have provided seems not working for us, probably, the long term solutions would be to fix this overflow error by not having to convert and only have one ttype of var

pkg/capacity/manager.go
MatousJobanek: ok, I see. Please add a comments to the lines explaining why it is OK to ignore the linter - why the number cannot overflow the capacity of the type

fbm3307: updated it with comments

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -184,7 +184,8 @@ func (r *Reconciler) determineCapacityReadyState(spc *toolchainv1alpha1.SpacePro
 // It always knows this fact so returning a bool is ok, in contrast to determinMemoryUtilizationReadyState.
 func determineSpaceCountReadyState(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
 	max := spc.Spec.CapacityThresholds.MaxNumberOfSpaces
-	return max == 0 || max > uint(spc.Status.ConsumedCapacity.SpaceCount)
+	//the value of this is not going beyond 100 and it won't overflow, hence its okay to ignore the overflow linter error
+	return max == 0 || max > uint(spc.Status.ConsumedCapacity.SpaceCount) // nolint:gosec
```
MatousJobanek, line 188: the comment is not true for this case - the line is casting `spc.Status.ConsumedCapacity.SpaceCount` which should be `0` and more, it shouldn't ever go under `0`. 
The second option would be changing the comparison to:
```go
return max == 0 || int(max) > spc.Status.ConsumedCapacity.SpaceCount
```
stating that we don't expect that the max number of Space count would ever go above the max size of int

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -203,7 +204,8 @@ func determineMemoryUtilizationReadyState(spc *toolchainv1alpha1.SpaceProvisione
 
 	// the memory utilitzation is ok if it is below the threshold in all node types
 	for _, val := range spc.Status.ConsumedCapacity.MemoryUsagePercentPerNodeRole {
-		if uint(val) >= spc.Spec.CapacityThresholds.MaxMemoryUtilizationPercent {
+		//the value of this is not going beyond 100 and it won't overflow, hence its okay to ignore the overflow linter error
+		if uint(val) >= spc.Spec.CapacityThresholds.MaxMemoryUtilizationPercent { // nolint:gosec
```
MatousJobanek, line 208: this is not true either - it's fine to cast the val to `uint` because we don't expect that it would be negative.
But it can be modified so you cast `spc.Spec.CapacityThresholds.MaxMemoryUtilizationPercent` to int instead

pkg/capacity/manager.go
```diff
@@ -21,7 +21,8 @@ type (
 
 func hasNotReachedMaxNumberOfSpacesThreshold(counts counter.Counts) spaceProvisionerConfigPredicate {
 	return func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
+		//the value of this is not going beyond 100 and it won't overflow, hence its okay to ignore the overflow linter error
+		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster]) // nolint:gosec
```
MatousJobanek, line 25: the same

pkg/capacity/manager.go
```diff
@@ -45,7 +46,8 @@ func hasEnoughMemoryCapacity(status *toolchainv1alpha1.ToolchainStatus) spacePro
 func hasMemberEnoughMemoryCapacity(memberStatus toolchainv1alpha1.Member, threshold uint) bool {
 	if len(memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole) > 0 {
 		for _, usagePerNode := range memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole {
-			if uint(usagePerNode) >= threshold {
+			//the value of this is not going beyond 100 and it won't overflow, hence its okay to ignore the overflow linter error
+			if uint(usagePerNode) >= threshold { // nolint:gosec
```
MatousJobanek, line 50: the same

MatousJobanek (commented):
please fix the comments

pkg/capacity/manager.go
fbm3307: , I have changed it , but  i dont fully understand this, we are not expecting this to be negative?

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -203,7 +204,8 @@ func determineMemoryUtilizationReadyState(spc *toolchainv1alpha1.SpaceProvisione
 
 	// the memory utilitzation is ok if it is below the threshold in all node types
 	for _, val := range spc.Status.ConsumedCapacity.MemoryUsagePercentPerNodeRole {
-		if uint(val) >= spc.Spec.CapacityThresholds.MaxMemoryUtilizationPercent {
+		//the value of this is not going be less 0 or negative and it won't overflow, hence its okay to ignore the overflow linter error
```
MatousJobanek, line 207:
```suggestion
		// the MaxMemoryUtilizationPercent won't go over 100, so it's safe to cast it to int and to not worry about overflow error, thus ignoring the linter
```

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -184,7 +184,8 @@ func (r *Reconciler) determineCapacityReadyState(spc *toolchainv1alpha1.SpacePro
 // It always knows this fact so returning a bool is ok, in contrast to determinMemoryUtilizationReadyState.
 func determineSpaceCountReadyState(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
 	max := spc.Spec.CapacityThresholds.MaxNumberOfSpaces
-	return max == 0 || max > uint(spc.Status.ConsumedCapacity.SpaceCount)
+	//the value of we don't expect that the max number of Space count would ever go above the max size of int and it won't overflow, hence its okay to ignore the overflow linter error
```
MatousJobanek, line 187:
```suggestion
	// we don't expect that the max number of spaces would ever go above the max size of int, so we don't have to worry about the overflow error, hence its okay to ignore the linter here
```

pkg/capacity/manager.go
```diff
@@ -21,7 +21,8 @@ type (
 
 func hasNotReachedMaxNumberOfSpacesThreshold(counts counter.Counts) spaceProvisionerConfigPredicate {
 	return func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
+		//the value of this is not going to be negative and it won't overflow, hence its okay to ignore the overflow linter error
```
MatousJobanek, line 24:
```suggestion
		// the numbers of Spaces per cluster shouldn't be negative, so it's fine to cast it to uint without worrying about the overflow error, and thus also ignore the linter
```

pkg/capacity/manager.go
```diff
@@ -45,7 +46,8 @@ func hasEnoughMemoryCapacity(status *toolchainv1alpha1.ToolchainStatus) spacePro
 func hasMemberEnoughMemoryCapacity(memberStatus toolchainv1alpha1.Member, threshold uint) bool {
 	if len(memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole) > 0 {
 		for _, usagePerNode := range memberStatus.MemberStatus.ResourceUsage.MemoryUsagePerNodeRole {
-			if uint(usagePerNode) >= threshold {
+			//We do not expect this value to be negative and it won't overflow, hence its okay to ignore the overflow linter error
```
MatousJobanek, line 49:
```suggestion
			// the memory utilization threshold should be max 100, so it's fine to cast it to int without worrying about the overflow error, thus ignoring the linter here
```

pkg/capacity/manager.go
fbm3307: Okay thanks for explaining

pkg/space/space.go
```diff
@@ -46,7 +46,8 @@ func NewSpaceWithFeatureToggles(userSignup *toolchainv1alpha1.UserSignup, target
 func addFeatureToggles(space *toolchainv1alpha1.Space, toggles []toolchainconfig.FeatureToggle) {
 	var winners []string
 	for _, t := range toggles {
-		weight := int(t.Weight())
+		//the value of weight is not expected to go beyond 100 or negative , it won't overflow, hence its okay to ignore the overflow linter error
```
MatousJobanek, line 49: you are converting from uint (which cannot be negative on it's own) to int
```suggestion
		//the value of weight is not expected to go beyond 100, it won't overflow, hence its okay to ignore the overflow linter error
```

MatousJobanek (approved):

MatousJobanek (approved):

PR #1117, opened 2025-01-07
DO NOT MERGE: An attempt at restarting the controller manager without restarting the whole operator binary

metlos:
This is not meant for merging. 

It is an experiment at soft-restarting the controller manager within the process of the host operator so that we don't have to rely on external tooling (i.e. ksctl) to manually restart the operator when the set of the registered members changes. The operator itself could detect these occasions and "heal" itself be soft-restarting so that further reconciliations work with the up-to-date set of members. In the longer terms this could also give us the opportunity to consolidate the explicit set of member clusters established at the operator start with the "cluster cache" used in the toolchaincluster controller and health checks.

Let's discuss here whether this potential approach has some merit and whether this solution is actually what we'd want.

At the moment, the restarts are just triggered every minute. This is of course not how it would be working for real - we would request a restart only when a new member is detected or when an existing one is removed.

The frequent restarts are put in place so that we can judge the effects on the memory consumption and/or see any leaks that might be happening in the controller manager due to the way we restart it.

I have ran the experimental version of the host-operator in a cluster and was continuously measuring the memory consumption for 5 hours after creating 5000 user signups in the cluster. This means that a non-trivial amount of objects needed to be loaded in the cache managed by the controller manager and therefore we should be seeing the effects of the leaks if there were any.

So, in the experiment, the host-operator was reconciling 5000 user signups, restarting every minute, for 5 hours. The results are compiled in this spreadsheet: https://docs.google.com/spreadsheets/d/1maSd_KUUYBwA3U1d3UdSMPaQkEwXYmbhWg4duCBvfO0/edit?gid=0#gid=0.

The memory consumption seems quite stable to me:
obrazek

### DIFF
--- a/.github/workflows/ci-golang-sbom.yml
+++ b/.github/workflows/ci-golang-sbom.yml
@@ -27,9 +27,9 @@ jobs:
         make generate
     
     - name: Lint
-      uses: golangci/golangci-lint-action@v6
+      uses: golangci/golangci-lint-action@v7
       with:
-        version: v1.63.1
+        version: v2.0.2
         skip-pkg-cache: true
         skip-build-cache: true
         args: --config=./.golangci.yml --verbose
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,34 +1,66 @@
-run:
-  # timeout for analysis, e.g. 30s, 5m, default is 1m
-  timeout: 10m
-
+version: "2"
 linters:
   enable:
-    - megacheck
+    - asasalint
+    - asciicheck
+    - bidichk
+    - bodyclose
+    - durationcheck
+    - errchkjson
+    - errorlint
+    - exhaustive
+    - gocheckcompilerdirectives
+    - gochecksumtype
     - gocyclo
-    - gofmt
-    - revive
+    - gosec
+    - gosmopolitan
+    - loggercheck
+    - makezero
     - misspell
-  presets: # groups of linters. See https://golangci-lint.run/usage/linters/
-    - bugs
-    - unused
-  disable: 
-    - golint # deprecated, use 'revive'
-    - scopelint # deprecated, use 'exportloopref'
-    - contextcheck # too many false-positives
-    - noctx # not needed
-
-# all available settings of specific linters
-linters-settings:
-  unparam:
-    # Inspect exported functions, default is false. Set to true if no external program/library imports your code.
-    # XXX: if you enable this setting, unparam will report a lot of false-positives in text editors:
-    # if it's called for subdir of a project it can't find external interfaces. All text editor integrations
-    # with golangci-lint call it on a directory with the changed file.
-    check-exported: true
-  revive:
-    rules:
-      - name: dot-imports
-        disabled: true # we allow for dot-imports
-
- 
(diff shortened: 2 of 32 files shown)


pkg/restart/restart.go
```diff
@@ -0,0 +1,166 @@
...
+func (s *StartManager) Start(ctx context.Context) error {
+	for {
+		debug("Start: start loop enter")
+		startErr, finished := s.doStart(ctx)
+
+		if s.runningContext == nil {
+			// this actually means that the start failed, but we need a non-nil value for the runningContext
+			// so that we can select on it below
+			s.runningContext = context.TODO()
+		}
```
MatousJobanek, line 46: if the start fails, why do we actually care and do not return an error here?

pkg/restart/restart.go
```diff
@@ -0,0 +1,166 @@
...
+	if initErr != nil {
+		debug("doStart: InitializeManager failed")
+		err <- initErr
+		return err, finished
+	}
+	go func() {
+		debug("doStart: about to invoke mgr.Start()")
+		if startErr := mgr.Start(s.runningContext); startErr != nil {
+			debug("doStart: mgr.Start() returned error. Sending it.")
+			err <- startErr
```
MatousJobanek, line 135: if the manager cannot be started, then we can just simply force-kill, right?
```suggestion
			os.Exit(1)
```
There is nothing we could try to recover. Or is there any other cleanup we would need to do in this case?

MatousJobanek: I have one more note/question - I understand the usage of the channel in this case, but I don't understand why we need to use the channel for all other error cases returned from this method. Is it to only unify the error handling, or is there any other reason?

pkg/restart/restart.go
```diff
@@ -0,0 +1,166 @@
...
+				s.lock.Lock()
+				debug("Start: locked")
+				s.running = false
+				s.lock.Unlock()
+				debug("Start: unlocked")
+				debug("Start: context cancelled, running=false is set")
+
+				// now, let's see what caused the cancellation
+
+				if !errors.Is(context.Cause(s.runningContext), restartNeededError) {
```
MatousJobanek, line 71: checking the cause of the cancel function and comparing it is a bit weird to me - it reminds me catching exceptions in Java. Not sure if this is a best practice in Go though.

Instead of using "CancelCause Context" you could use the standard "Cancel context" and add another field to the `StartManager` stating if the restart was triggered or not (in the same way as you do it with the `running` field. And have another case when the root context itself:
```go
		case <-ctx.Done():
			debug("Start: context cancelled, cancellation from above")
			<-finished
			debug("Start: quitting")
```

MatousJobanek (commented):
Thanks a lot for this experimental PR, it looks very promising.

pkg/restart/restart.go
metlos: I am not a fan of panicking in "library" code.

As for the error handling - yes, the error is always returned in the channel so that the caller doesn't have to have special logic for when the "setup" of the start fails as opposed to the failure that happens async-ly during the startup and lifetime of the controller manager.

PR #1118, opened 2025-01-08
KUBESAW-250: Fixing the linter error Linter for v1.63.1

fbm3307:
Would be upgrading to golangci-lint version 1.63.1 . This PR is to fix the [Linter error](https://github.com/codeready-toolchain/host-operator/actions/runs/12649616851/job/35246418292?pr=1116#step:5:28) which were caught as a result of upgrading the linter

Related PR -
- Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/1116

### DIFF
--- a/pkg/capacity/manager.go
+++ b/pkg/capacity/manager.go
@@ -21,7 +21,7 @@ type (
 
 func hasNotReachedMaxNumberOfSpacesThreshold(counts counter.Counts) spaceProvisionerConfigPredicate {
 	return func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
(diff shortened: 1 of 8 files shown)


pkg/capacity/manager.go
```diff
@@ -21,7 +21,7 @@ type (
 
 func hasNotReachedMaxNumberOfSpacesThreshold(counts counter.Counts) spaceProvisionerConfigPredicate {
 	return func(spc *toolchainv1alpha1.SpaceProvisionerConfig) bool {
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
+		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster]) // nolint:gosec
```
MatousJobanek, line 24: can you give a bit more details on why and what linter error is ignored here?

fbm3307:
closing in favour of #1116

pkg/capacity/manager.go
fbm3307: i will put the comments in the other PR where i have moved the code now

PR #1109, opened 2024-12-06
Consumed capacity info in SpaceProvisionerConfig

metlos:
This updates the status of the `SpaceProvisionerConfig` with "consumed capacity" info copied from `ToolchainStatus`.

This info is also used for readiness computation of the SPC. The logic is more or less duplicated in the capacity manager so no behavioral changes should happen to the optimal cluster selection logic.

In the forthcoming PRs the capacity manager will be simplified to take advantage of the new info in the SPC status.

Paired PRs:
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1079

Note to self:
- [x] go through the comments in #1108 and make all of them are addressed

PR #1119, opened 2025-01-08
Member info in SPC status - capacity manager part

metlos:
This slightly simplifies the capacity manager by relying more on the ready status of the SPC (which is now based on more complex logic).

A whole bunch of tests was also updated to not have to initialize the count cache just for the capacity manager and instead set up the capacity manager to only use SPCs for determining the space counts.

This is a follow-up PR of https://github.com/codeready-toolchain/host-operator/pull/1109

Note that there are no new or updated e2e tests for this PR because it should not change any behavior whatsoever.


pkg/capacity/manager.go
```diff
@@ -6,57 +6,38 @@ import (
...
-		if threshold == 0 {
-			return true
-		}
-		for _, memberStatus := range status.Status.Members {
-			if memberStatus.ClusterName == spc.Spec.ToolchainCluster {
-				return hasMemberEnoughMemoryCapacity(memberStatus, threshold)
-			}
+		var spaceCount int
+		if spc.Status.ConsumedCapacity != nil {
+			spaceCount = spc.Status.ConsumedCapacity.SpaceCount
```
MatousJobanek, line 35: correct me if I'm wrong, but we shouldn't use the SPC for getting the number of provisioned Spaces

pkg/capacity/manager.go
```diff
@@ -234,3 +257,20 @@ func (b *ClusterManager) getOptimalTargetClusters(ctx context.Context, preferred
...
 }
+
+func updateSpaceCountFromCache(ctx context.Context, getSpaceCount SpaceCountGetter, spc *toolchainv1alpha1.SpaceProvisionerConfig) {
+	if spaceCount, ok := getSpaceCount(ctx, spc.Spec.ToolchainCluster); ok {
+		// the cached spaceCount is always going to be fresher than (or as fresh as) what's in the SPC
+		if spc.Status.ConsumedCapacity == nil {
+			spc.Status.ConsumedCapacity = &toolchainv1alpha1.ConsumedCapacity{}
+		}
+		spc.Status.ConsumedCapacity.SpaceCount = spaceCount
+	}
```
MatousJobanek, line 268: as discussed in the previous PR, let's not update the status of the SPC if we are not going to update the actual resource - there is no benefit in doing that

MatousJobanek (commented):
It looks like that the PR still needs some cleanup after the changes and discussions in the previous PR. I'm talking mainly about the usage of SPC status for storing the number of provisioned Spaces - we shouldn't do that in the capacity manager.

In other words, use SPC only to get the readiness, and the threshold for max number of Spaces, nothing else. The Space count should be taken from the cache only.

pkg/capacity/manager.go
metlos: We discussed that it is weird for a predicate to have side-effects which I completely agree with.

Now, the "side-effect" is done while getting the candidate optimal clusters SO THAT the value we update can later be used again in the rest of the computation. Otherwise we'd be forced to repeatedly read the value from the count cache while sorting the array of the candidate clusters. I stand by the fact that it is better do it once and re-use those values than to do that repeatedly. It is not only more performant but also more stable (think of concurrent updates of the count cache).

This behavior is documented in the docs of `getOptionalTargetClusters()` as well as a comment in the body of `GetOptionalTargetCluster()`.

I don't agree with the reasoning that we only should update the CRDs if we intend to update them. Those are data structures like any other - they're used to hold data. And if we need some kind of data be updated because we are going to need the updated values down the road, it is IMHO completely warranted to use the CRD for that. Note that these CRDs are local to the computation in `GetOptionalTargetCluster()` function. They don't escape that function because they are not stored in cluster nor in any "global" structure like the count cache. So there is no non-local "suprise" in having the updates in the CRDs.

controllers/usersignup/approval_test.go
```diff
@@ -261,14 +239,13 @@ func TestGetClusterIfApproved(t *testing.T) {
 
 	t.Run("automatic approval not enabled, user manually approved and second cluster has capacity", func(t *testing.T) {
 		// given
-		spc1 := hspc.NewEnabledValidTenantSPC("member1", spc.MaxNumberOfSpaces(2000), spc.MaxMemoryUtilizationPercent(62))
-		spc2 := hspc.NewEnabledValidTenantSPC("member2", spc.MaxMemoryUtilizationPercent(62))
-		fakeClient := commontest.NewFakeClient(t, toolchainStatus, spc1, spc2)
-		InitializeCounters(t, toolchainStatus)
+		spc1 := hspc.NewEnabledTenantSPC("member1")
+		spc2 := hspc.NewEnabledValidTenantSPC("member2")
```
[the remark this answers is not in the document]
metlos: I don't think that user approval tests should be testing the capacity manager. That's why I wanted to simplify these tests as much as possible. It is no concern to the approval process why a cluster is not capable of accepting a space.

I didn't want to be radical and wanted some kind of "middle ground" solution where I'd just simplify the capacity manager logic to at least not rely on the ToolchainStatus. But in hindsight I think I should have just modified the usersignup to be able to completely mock-out the capacity manager so that these tests actually just test what they're supposed to and not have to have a ton of setup.

I do think though that there shouldn't be a single threshold check in the approval tests. The approval only is concerned about clusters that can be deployed to. It should not have to know about the reasons why a cluster is not eligible, just whether it is or not.

controllers/usersignup/approval_test.go
```diff
@@ -48,13 +35,12 @@ func TestGetClusterIfApproved(t *testing.T) {
 			testconfig.AutomaticApproval().
 				Enabled(true),
 		)
-		spc1 := hspc.NewEnabledValidTenantSPC("member1", spc.MaxNumberOfSpaces(1000), spc.MaxMemoryUtilizationPercent(70))
-		spc2 := hspc.NewEnabledValidTenantSPC("member2", spc.MaxNumberOfSpaces(1000), spc.MaxMemoryUtilizationPercent(75))
-		fakeClient := commontest.NewFakeClient(t, toolchainStatus, toolchainConfig, spc1, spc2)
-		InitializeCounters(t, toolchainStatus)
```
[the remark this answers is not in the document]
MatousJobanek: based on the other comment below, we don't need to initialize the counters in this unit tests, it's fine if they stay just 0 for the cache.

controllers/usersignup/approval_test.go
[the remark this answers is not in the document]
MatousJobanek: no needed

controllers/usersignup/approval_test.go
```diff
@@ -48,13 +35,12 @@ func TestGetClusterIfApproved(t *testing.T) {
...
+		spc1 := hspc.NewEnabledValidTenantSPC("member1")
+		spc2 := hspc.NewEnabledValidTenantSPC("member2")
+		fakeClient := commontest.NewFakeClient(t, toolchainConfig, spc1, spc2)
 
 		// when
-		approved, clusterName, err := getClusterIfApproved(ctx, fakeClient, signup, capacity.NewClusterManager(test.HostOperatorNs, fakeClient))
+		approved, clusterName, err := getClusterIfApproved(ctx, fakeClient, signup, capacity.NewClusterManager(test.HostOperatorNs, fakeClient, hspc.GetSpaceCountFromSpaceProvisionerConfigs(fakeClient, test.HostOperatorNs)))
```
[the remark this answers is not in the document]
MatousJobanek: btw, how about having something like this:
```go 
type CountPerCluster func() (string, int)

func ClusterCount(cluster string, count int) CountPerCluster {
	return func() (string, int) {
		return cluster, count
	}
}

func InitializeCountersWith(t *testing.T, counts ...CountPerCluster) {
	commontest.SetEnvVarAndRestore(t, "WATCH_NAMESPACE", commontest.HostOperatorNs)
	counter.Reset()
	t.Cleanup(counter.Reset)

	var members []ToolchainStatusOption
	for _, clusterCount := range counts {
		cluster, count := clusterCount()
		members = append(members, WithMember(cluster, WithSpaceCount(count)))
	}

	toolchainStatus := NewToolchainStatus(members...)
	initializeCounters(t, commontest.NewFakeClient(t), toolchainStatus)
}
```
so it could be used:
```go
InitializeCountersWith(t, ClusterCount("member1", 0), ClusterCount("member2", 0))
```

controllers/usersignup/approval_test.go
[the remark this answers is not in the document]
metlos: I've reworked the code and the counts are initialized again in the approval_test.go.

controllers/usersignup/approval_test.go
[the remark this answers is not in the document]
metlos: The tests were slightly updated to not use any counts. The count cache is initialized empty and the tests only work with ready/unready SPCs.

pkg/capacity/manager.go
```diff
@@ -6,57 +6,38 @@ import (
...
-		numberOfSpaces := uint(counts.SpacesPerClusterCounts[spc.Spec.ToolchainCluster])
-		threshold := spc.Spec.CapacityThresholds.MaxNumberOfSpaces
-		return threshold == 0 || numberOfSpaces < threshold
-	}
-}
+	// SpaceCountGetter is a function useful for mocking the counts cache and can be passed
+	// to the NewClusterManager function. The returned tuple represents the actual number
+	// of the spaces found for given cluster and the fact whether the value was actually found or not
+	// (similar to the return value of the map lookup).
+	SpaceCountGetter func(ctx context.Context, clusterName string) (int, bool)
```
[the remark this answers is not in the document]
metlos: Good point. I simplified the code such that the count cache is mandatory again and this construct is not there anymore.

pkg/capacity/manager.go
metlos: This has been reworked. A custom struct is now used in the algorithm so the data flow is more clear.

pkg/capacity/manager.go
metlos: I introduced a custom struct to hold the data necessary for the computation, so that it is more clear how the data flows through the algorithm.

controllers/usersignup/approval_test.go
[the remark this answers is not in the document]
metlos: With the re-introduction of the counters inititialization and reworking of the approval_test.go to not actually require any counts to be present in the cache, I don't think this is needed anymore. A neat idea though.

pkg/capacity/manager_test.go
```diff
@@ -8,323 +8,171 @@ import (
...
+				}
+			} else if m2_eligible {
+				chosenMember = "member2"
+			} else {
+				chosenMember = ""
+			}
 
-		fakeClient := commontest.NewFakeClient(t, toolchainStatus, spc1, spc2, spc3)
-		InitializeCounters(t, toolchainStatus)
+			t.Run(fmt.Sprintf("choosing between member1(valid=%v,has_role=%v,preferred=%v) and member2(valid=%v,has_role=%v,preferred=%v) should yield %s", m1_valid, m1_has_role, m1_preferred, m2_valid, m2_has_role, m2_preferred, chosenMember), func(t *testing.T) {
```
MatousJobanek, line 85: It's maybe a lack of coffee this morning, but it looks very cryptic to me and it took me some time to understand what you were trying to achieve here. 
Let's not make a rocket science (for us as normal human being) from something that is simple in it's core. I've gave a shot and tried to write it in a more human-readable way:
```go

type combination struct {
	valid   bool
	hasRole bool
}

func (c combination) isEligible() bool {
	return c.valid && c.hasRole
}

func (c combination) spaceProvisionerConfig(name string) *toolchainv1alpha1.SpaceProvisionerConfig {
	opts := []spc.CreateOption{spc.ReferencingToolchainCluster(name)}
	if c.valid {
		opts = append(opts, spc.WithReadyConditionValid())
	}
	if c.hasRole {
		opts = append(opts, spc.WithPlacementRoles(spc.PlacementRole("tenant")))
	}

	return spc.NewSpaceProvisionerConfig(name, commontest.HostOperatorNs, opts...)
}

func generateCombinations() []combination {
	var combinations []combination
	for _, valid := range []bool{true, false} {
		for _, hasRole := range []bool{true, false} {
			combinations = append(combinations, combination{
				valid:   valid,
				hasRole: hasRole,
			})
		}
	}
	return combinations
}

func TestGetOptimalTargetCluster(t *testing.T) {
	// given
	ctx := context.TODO()
	for _, preferredMember := range []string{"", "member1", "member2"} {

		member1Combinations := generateCombinations()
		member2Combinations := generateCombinations()

		for _, member1 := range member1Combinations {
			for _, member2 := range member2Combinations {

				// here we define the expected test result, i.e. which member the algorithm should choose
				chosenMember := ""
				if member1.isEligible() {
					chosenMember = "member1"
					if member2.isEligible() && preferredMember == "member2" {
						chosenMember = "member2"
					}
				} else if member2.isEligible() {
					chosenMember = "member2"
				}

				t.Run(fmt.Sprintf("choosing between member1(%+v) and member2(%+v) when preferred is %s, should yield %s", member1, member2, preferredMember, chosenMember), func(t *testing.T) {
					// given
					spc1 := member1.spaceProvisionerConfig("member1")
					spc2 := member2.spaceProvisionerConfig("member2")
					status := test.NewToolchainStatus()
					test.InitializeCounters(t, status)
					fakeClient := commontest.NewFakeClient(t, spc1, spc2, status)
					cm := capacity.NewClusterManager(commontest.HostOperatorNs, runtimeclient.Client(fakeClient))

					// when
					clusterName, err := cm.GetOptimalTargetCluster(ctx, capacity.OptimalTargetClusterFilter{
						ClusterRoles:     []string{spc.PlacementRole("tenant")},
						PreferredCluster: preferredMember,
					})

					// then
					require.NoError(t, err)
					assert.Equal(t, chosenMember, clusterName)
				})
			}
		}
	}
```

MatousJobanek: btw, with this setup, it would be also much easier to add additional field into the set of combinations

pkg/capacity/manager_test.go
```diff
@@ -8,323 +8,171 @@ import (
...
-		// when
-		clusterName, err := capacity.NewClusterManager(commontest.HostOperatorNs, fakeClient).GetOptimalTargetCluster(
-			ctx,
-			capacity.OptimalTargetClusterFilter{
-				ToolchainStatusNamespace: commontest.HostOperatorNs,
-			},
-		)
+		spc1 := hspc.NewEnabledValidTenantSPC("member1", spc.MaxNumberOfSpaces(10000), spc.WithConsumedSpaceCount(700))
+		spc2 := hspc.NewEnabledValidTenantSPC("member2", spc.MaxNumberOfSpaces(1000), spc.WithConsumedSpaceCount(700))
+		spc3 := hspc.NewEnabledValidTenantSPC("member3", spc.MaxNumberOfSpaces(1000), spc.WithConsumedSpaceCount(200))
```
MatousJobanek, line 127: within the context of the capacity manager, we shouldn't care about the consumed capacity set in the SPC status
```suggestion
		spc1 := hspc.NewEnabledValidTenantSPC("member1", spc.MaxNumberOfSpaces(10000))
		spc2 := hspc.NewEnabledValidTenantSPC("member2", spc.MaxNumberOfSpaces(1000))
		spc3 := hspc.NewEnabledValidTenantSPC("member3", spc.MaxNumberOfSpaces(1000))
```

pkg/capacity/manager_test.go
```diff
@@ -8,323 +8,171 @@ import (
...
-		assert.Equal(t, "member2", clusterName)
-	})
-
-	t.Run("with two clusters but none of them has enough capacity - using the default memory values", func(t *testing.T) {
-		// given
-		spc1 := hspc.NewEnabledValidTenantSPC("member1", spc.MaxMemoryUtilizationPercent(1))
-		spc2 := hspc.NewEnabledValidTenantSPC("member2", spc.MaxMemoryUtilizationPercent(1))
-		fakeClient := commontest.NewFakeClient(t, toolchainStatus, spc1, spc2)
-		InitializeCounters(t, toolchainStatus)
+		test.InitializeCounters(t, status)
```
MatousJobanek, line 130: this doesn't actually test the space counter, does it? The ToolchainStatus is empty, which means that the counter will be empty as well
consider using the helper functions to simplify the initialization of the that I mentioned in the previous review:  https://github.com/codeready-toolchain/host-operator/pull/1119#discussion_r1910444037

pkg/capacity/manager_test.go
```diff
@@ -390,25 +211,27 @@ func TestGetOptimalTargetClusterInBatchesBy50WhenTwoClusterHaveTheSameUsage(t *t
...
+						spc.WithConsumedMemoryUsagePercentInNode("worker", 55),
+						spc.WithConsumedMemoryUsagePercentInNode("master", 60))
+					spc3 := hspc.NewEnabledValidTenantSPC("member3",
+						spc.MaxNumberOfSpaces(limit),
+						spc.WithConsumedSpaceCount(numberOfSpaces),
+						spc.WithConsumedMemoryUsagePercentInNode("worker", 55),
+						spc.WithConsumedMemoryUsagePercentInNode("master", 50))
+
+					status := test.NewToolchainStatus()
+					test.InitializeCounters(t, status)
```
MatousJobanek, line 232: the same also here and at other places, don't use the utilization from the SPC, but rather from the counter instead

pkg/capacity/manager.go
```diff
@@ -226,13 +217,32 @@ func (b *ClusterManager) getOptimalTargetClusters(ctx context.Context, preferred
...
+		spaceCount = 0
+	}
+
+	return provisionerCandidate{
+		clusterName:         spc.Spec.ToolchainCluster,
+		spaceCount:          spaceCount,
+		spaceCountThreshold: int(spc.Spec.CapacityThresholds.MaxNumberOfSpaces), //nolint:gosec // this just doesn't overflow there's no way of having > 2*10^9 namespaces in a cluster
+		isReady:             condition.IsTrue(spc.Status.Conditions, toolchainv1alpha1.ConditionReady),
+		placementRoles:      spc.Spec.PlacementRoles,
+	}
```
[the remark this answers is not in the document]
metlos: Makes sense to simplify the logic.

pkg/capacity/manager_test.go
metlos: I guess it is a matter of taste but I consider my version more readable and easy to understand because you can just read it top down instead of having to understand the whole setup and having to jump around the code to understand the whole picture.

I would opt for breaking up the logic into more functions and structs the way you do if it was used in more than 1 place but that's not the case here.

pkg/capacity/manager.go
```diff
@@ -205,18 +193,21 @@ func matches(spc *toolchainv1alpha1.SpaceProvisionerConfig, predicates []spacePr
...
-	matching := make([]toolchainv1alpha1.SpaceProvisionerConfig, 0, len(list.Items))
+	matching := make([]provisionerCandidate, 0, len(list.Items))
 
 	for _, spc := range list.Items {
-		spc := spc
-		if matches(&spc, predicates) {
-			matching = append(matching, spc)
+		spc := spc // to prevent the memory aliasing in a for-loop
+		candidate := provisionerCandidateFromSPC(&spc, counts)
+		if matches(&candidate, predicates) {
```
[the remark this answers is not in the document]
metlos: true. I updated it to pass by value.

pkg/capacity/manager_test.go
metlos: Btw `TestGetOptimalTargetClusterInBatchesBy50WhenTwoClusterHaveTheSameUsage` is written in the very same vein. 4 nested for-loops with non-trivial setup. But because it is linear and is in one place it makes more sense to do it all linearly than to break it up into pieces (or at least that's how I understand it).

mfrancisc: >  it took me some time to understand what you were trying to achieve here.

I had a similar thought as @MatousJobanek , as tried to explain in the old pr comment https://github.com/codeready-toolchain/host-operator/pull/1108#discussion_r1865764380

I do find @MatousJobanek proposal easier to follow and to extend in the future, In addition to that I would maybe add a few comments to  the `combinations` struct to give some more context about the `hasRole` and `valid` fields purpose.

Minor - I would rename `chosenMember` to something like  `expectedClusterName`,  and `clusterName` to something like `actualClusterName`, to make it easier to understand which was the computed variable compared to the constant and expected one.

metlos: Ok, it's 2:1, so I updated the code :)

MatousJobanek (approved):
Looks great :tada: :+1:  thanks a lot for addressing all comments :1st_place_medal: :pray:

PR #1128, opened 2025-01-23
don't process NSTemplateTier that is being deleted

MatousJobanek:
don't process NSTemplateTier that is being deleted, otherwise, the controller can create TTR CRs in the reconcile that was triggered by the "delete" call.

https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1737375318490079


mfrancisc (approved):
Great catch 🚀 

Thanks for fixing it!

PR #1131, opened 2025-01-24
Try to reduce the log verbosity of the SPC controller.

metlos:
This is done by introducing a filter for the watch that should ignore any changes to the TCs that are only caused by the changes in the timestamps of the conditions.

FYI, I didn't create any e2e test for this because frankly, I am not sure how that would look like. There should be no change in the behavior and I don't really see a way of testing "the reconciliation shouldn't happen if I do this".

JIRA: https://issues.redhat.com/browse/KUBESAW-256

### DIFF
--- a/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
+++ b/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
@@ -28,7 +28,7 @@ type Reconciler struct {
 
 var _ reconcile.Reconciler = (*Reconciler)(nil)
 
-func GetToolchainClusterConditions(obj runtimeclient.Object) []toolchainv1alpha1.Condition {
+func getToolchainClusterConditions(obj runtimeclient.Object) []toolchainv1alpha1.Condition {
 	tc, ok := obj.(*toolchainv1alpha1.ToolchainCluster)
 	if !ok {
 		return nil
@@ -42,18 +42,18 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(
 			&toolchainv1alpha1.ToolchainCluster{},
 			handler.EnqueueRequestsFromMapFunc(mapToolchainClusterToSpaceProvisionerConfigs(r.Client)),
-			builder.WithPredicates(predicate.Or(
+			builder.WithPredicates(
 				// NOTE: this is ignoring the changes in the Spec as well as the Status.OperatorNamespace or Status.APIEndpoint
 				//
 				// The spec of the TC contains just the reference to the secret with the connection information. We're not actually
-				// interested in that change here because the SPC is not interested HOW we connect to the cluster rather IF
+				// interested in that change here because the SPC is not interested in HOW we connect to the cluster but rather IF
 				// we're connected to the cluster.
 				//
 				// For the same reason we're not actually interested in the values of the Status.OperatorNamespace or Status.APIEndpoint.
 				//
 				// So in another words, we're only interested in the change of the ready condition of the TC.
-				&toolchainpredicate.ToolchainConditionChanged{GetConditions: GetToolchainClusterConditions, Type: toolchainv1alpha1.ConditionReady},
-			)),
+				&toolchainpredicate.ToolchainConditionChanged{GetConditions: getToolchainClusterConditions, Type: toolchainv1alpha1.ConditionReady},
+			),
 		).
 		// we use the same information as the ToolchainStatus specific for the SPCs. Because memory consumption is
 		// read directly out of the member clusters using remote connections, let's look for it only once
--- a/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
+++ b/controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
@@ -6,6 +6,7 @@ import (
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/controllers/toolchainconfig"
+	toolchainpredicate "github.com/codeready-toolchain/host-operator/pkg/predicate"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	"github.com/redhat-cop/operator-utils/pkg/util"
 	corev1 "k8s.io/api/core/v1"
@@ -27,12 +28,32 @@ type Reconciler struct {
 
 var _ reconcile.Reconciler = (*Reconciler)(nil)
 
+func GetToolchainClusterConditions(obj runtimeclient.Object) []toolchainv1alpha1.Condition {
+	tc, ok := obj.(*toolchainv1alpha1.ToolchainCluster)
+	if !ok {
+		return nil
+	}
+	return tc.Status.Conditions
+}
+
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&toolchainv1alpha1.SpaceProvisionerConfig{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(
 			&toolchainv1alpha1.ToolchainCluster{},
-			handler.EnqueueRequestsFromMapFunc(MapToolchainClusterToSpaceProvisionerConfigs(r.Client)),
+			handler.EnqueueRequestsFromMapFunc(mapToolchainClusterToSpaceProvisionerConfigs(r.Client)),
+			builder.WithPredicates(predicate.Or(
+				// NOTE: this is ignoring the changes in the Spec as well as the Status.OperatorNamespace or Status.APIEndpoint
+				//
+				// The spec of the TC contains just the reference to the secret with the connection information. We're not actually
+				// interested in that change here because the SPC is not interested HOW we connect to the cluster rather IF
+				// we're connected to the cluster.
+				//
+				// For the same reason we're not actually interested in the values of the Status.OperatorNamespace or Status.APIEndpoint.
+				//
+				// So in another words, we're only interested in the change of the ready condition of the TC.
+				&toolchainpredicate.ToolchainConditionChanged{GetConditions: GetToolchainClusterConditions, Type: toolchainv1alpha1.ConditionReady},
+			)),
 		).
 		// we use the same information as the ToolchainStatus specific for the SPCs. Because memory consumption is
 		// read directly out of the member clusters using remote connections, let's look for it only once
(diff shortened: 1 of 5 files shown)


pkg/mapper/change_filter.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+)
+
+// FilterIgnoredFields produces an event handler that wraps the provided even handler.
```
mfrancisc, line 12:
```suggestion
// FilterIgnoredFields produces an event handler that wraps the provided event handler.
```

pkg/mapper/conditions_changed.go
```diff
@@ -0,0 +1,12 @@
+package mapper
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+)
+
+// IgnoringTimeChangesInToolchainConditions wraps the provided event handler such that it only reacts
+// to updates discarding the changes in the condition timestamps. Only updates are affected, all other
+// event types are always forwarded.
+func IgnoringTimeChangesInToolchainConditions(eventHandler handler.EventHandler) handler.EventHandler {
```
mfrancisc, line 10: minor - should we keep the name more generic ? Just `IgnoringTimeChangesInConditions`, as i think those are not specific/custom toolchain conditions using these fields.

pkg/mapper/conditions_changed_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestIfConditionsChangedMapper(t *testing.T) {
```
mfrancisc, line 18: should we also have a scenario for multiple conditions, for example one where there is a change in the ignored fields and but also in one of the not ignored fields ( like the `Status` ) ?

mfrancisc (commented):
Really nice solution 👍 

mostly LGTM , just one minor comment regarding the tests.

controllers/spaceprovisionerconfig/spaceprovisionerconfig_controller.go
```diff
@@ -32,7 +33,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&toolchainv1alpha1.SpaceProvisionerConfig{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(
 			&toolchainv1alpha1.ToolchainCluster{},
-			handler.EnqueueRequestsFromMapFunc(MapToolchainClusterToSpaceProvisionerConfigs(r.Client)),
+			mapper.IgnoringTimeChangesInToolchainConditions(handler.EnqueueRequestsFromMapFunc(mapToolchainClusterToSpaceProvisionerConfigs(r.Client))),
```
MatousJobanek, line 36: could you please clarify why it's not implemented as a predicate? It looks like a wrapper of the mapper, but a predicate would be a better fit for this.

pkg/mapper/change_filter.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Generic implements handler.EventHandler.
+func (i *ifChangedIgnoring) Generic(ctx context.Context, evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	i.eventHandler.Generic(ctx, evt, q)
+}
+
+// Update implements handler.EventHandler.
+func (i *ifChangedIgnoring) Update(ctx context.Context, evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+	if !cmp.Equal(evt.ObjectOld, evt.ObjectNew, i.cmpOpts...) {
```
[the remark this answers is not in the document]
MatousJobanek: I'm wondering, what if we combine the already existing `predicate.GenerationChangedPredicate` with the check of the ready condition. So it would get triggered whenever the spec is updated or when there is a change in the status/reason/message of the ready condition.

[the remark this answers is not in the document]
MatousJobanek: btw, we already do something like these for example here https://github.com/codeready-toolchain/host-operator/blob/f35e54f81c9c3ceebac61f6675a76f1a7a8132b3/controllers/usersignup/predicate.go#L33-L34
and also here: https://github.com/codeready-toolchain/host-operator/blob/f35e54f81c9c3ceebac61f6675a76f1a7a8132b3/controllers/space/space_controller.go#L50

pkg/mapper/conditions_changed.go
metlos: Good suggestion, but I redid the implementation according to @MatousJobanek comments, so this actually no longer applies.

pkg/mapper/conditions_changed_test.go
metlos: Good point. I redid the impl as a predicate as suggested by @MatousJobanek so the tests are not completely different.

pkg/predicate/toolchainconditionchanged.go
```diff
@@ -0,0 +1,66 @@
...
+
+var _ predicate.Predicate = (*ToolchainConditionChanged)(nil)
+
+type (
+	GetToolchainConditions func(client.Object) []toolchainv1alpha1.Condition
+
+	// ToolchainConditionChanged is a predicate to check that some conditions changed in a toolchain object.
+	ToolchainConditionChanged struct {
+		// GetConditions is function used to obtain the conditions from an object. There is no generic way of
+		// doing that using the standard k8s api.
```
MatousJobanek, line 19: Yeah, I've thought about creating an interface on the top of all our API types that would enable us creating a generic API for the condition management. Right now, we technically re-do and copy-paste more or less the same code in all controllers. With an interface, we could have a set of generic functions, or a fluent API that could be reused without any other changes. Also the already existing functions in toolchain-common would work with the interface abstraction, not with the actual set of conditions.
But that's for later.

pkg/predicate/toolchainconditionchanged.go
```diff
@@ -0,0 +1,66 @@
...
+		return false
+	}
+
+	// we're intentionally ignoring changes to *Time fields of the conditions. If nothing else changed, those do not represent a factual
+	// change in the condition.
+	if oldCond.Status != newCond.Status || oldCond.Reason != newCond.Reason || oldCond.Message != newCond.Message {
+		return true
+	}
+
+	return false
```
MatousJobanek, line 65:
```suggestion
	return oldCond.Status != newCond.Status || oldCond.Reason != newCond.Reason || oldCond.Message != newCond.Message
```

pkg/predicate/toolchainconditionchanged_test.go
```diff
@@ -0,0 +1,141 @@
...
+		ev := event.UpdateEvent{
+			ObjectOld: tcOld,
+			ObjectNew: tcNew,
+		}
+
+		// when
+		result := predicate.Update(ev)
+
+		// then
+		assert.Equal(t, expectedResult, result)
```
MatousJobanek, line 67: Let's do the check for all methods
```suggestion
		t.Run("update", func(t *testing.T) {
			// when
			result := predicate.Update(ev)

			// then
			assert.Equal(t, expectedResult, result)
		})

		t.Run("create", func(t *testing.T) {
			// when
			result := predicate.Create(event.CreateEvent{Object: tcNew})

			// then
			assert.True(t, result)
		})

		t.Run("delete", func(t *testing.T) {
			// when
			result := predicate.Delete(event.DeleteEvent{Object: tcNew})

			// then
			assert.True(t, result)
		})

		t.Run("generic", func(t *testing.T) {
			// when
			result := predicate.Generic(event.GenericEvent{Object: tcNew})

			// then
			assert.True(t, result)
		})
```

pkg/predicate/toolchainconditionchanged_test.go
```diff
@@ -0,0 +1,141 @@
...
+		})
+	})
+
+	t.Run("detects when condition appears", func(t *testing.T) {
+		test(t, true, func(old, new *toolchainv1alpha1.ToolchainCluster) {
+			new.Status.Conditions = old.Status.Conditions
+			old.Status.Conditions = nil
+		})
+	})
+
```
MatousJobanek, line 117: I haven't found any case when the condition would be missing in both objects - something like this:
```suggestion
	t.Run("ignores missing conditions", func(t *testing.T) {
		test(t, false, func(old, new *toolchainv1alpha1.ToolchainCluster) {
			new.Status.Conditions = nil
			old.Status.Conditions = nil
		})
	})
```

MatousJobanek (commented):
Looks good, just small suggestions

pkg/predicate/toolchainconditionchanged.go
```diff
@@ -0,0 +1,62 @@
...
+func (t *ToolchainConditionChanged) Delete(event.DeleteEvent) bool {
+	return true
+}
+
+// Generic implements predicate.Predicate.
+func (t *ToolchainConditionChanged) Generic(event.GenericEvent) bool {
+	return true
+}
+
+func (t *ToolchainConditionChanged) Update(evt event.UpdateEvent) bool {
```
fbm3307, line 42: The update event could ideally also cover the create and delete event too,(which probably could in turn reduce the reconciles) but that would complicate the logic and probably not worth the effort just to cover the edge case
is my understanding correct here?

fbm3307 (approved):
Great work !!

pkg/predicate/toolchainconditionchanged.go
metlos: I agree. The TCs are not created/delete that often though, so I think we're fine with doing a couple more reconciles than absolutely necessary.

pkg/predicate/toolchainconditionchanged.go
metlos: applied

pkg/predicate/toolchainconditionchanged_test.go
metlos: applied.

pkg/predicate/toolchainconditionchanged_test.go
metlos: sharp eyes. applied.

MatousJobanek (approved):
Thanks :+1:

PR #1136, opened 2025-02-10
feat: cleanup revision entries from nstemplatetier

mfrancisc:
This PR cleans up entries from the `NSTemplateTier.Status.Revisions` as they were replaced by new TierTemplate revisions.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/1114

Jira:  https://issues.redhat.com/browse/KUBESAW-259


controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -109,15 +109,22 @@ func (r *Reconciler) ensureRevision(ctx context.Context, nsTmplTier *toolchainv1
...
-	// and check if there is any reference to a TierTemplate that is not in the Spec anymore
-	if ttrCreated {
+	tierRemoved := false
+	for tierTempalateKey := range nsTmplTier.Status.Revisions {
+		if !checkIfTierTemplateIsStillUsed(nsTmplTier, tierTempalateKey) {
+			// remove old tiermtemplate from revisions
+			delete(nsTmplTier.Status.Revisions, tierTempalateKey)
+			tierRemoved = true
+		}
+	}
```
MatousJobanek, line 123: do we actually need this new for loop and the `checkIfTierTemplateIsStillUsed` function? couldn't we slightly update the already existing code and achieve the same with it?
```go
revisions := map[string]string{}
for _, tierTemplateRef := range refs {
	// get the TierTemplate
	var tierTemplate toolchainv1alpha1.TierTemplate
	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: nsTmplTier.GetNamespace(), Name: tierTemplateRef}, &tierTemplate); err != nil {
		// something went wrong or we haven't found the TierTemplate
		return false, err
	}

	// check if there is TTR associated with this TierTemplate
	ttrCreatedLatest, ttrName, err := r.ensureTTRforTemplate(ctx, nsTmplTier, &tierTemplate)
	ttrCreated = ttrCreated || ttrCreatedLatest
	if err != nil {
		return false, err
	}
	revisions[tierTemplate.GetName()] = ttrName
}
nsTmplTier.Status.Revisions = revisions
```

mfrancisc: great suggestion! Pls see https://github.com/codeready-toolchain/host-operator/pull/1136/commits/7209d716457e6c326279ff0b620d8e818aa614f9 

Thanks!

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -395,6 +395,56 @@ func TestUpdateNSTemplateTier(t *testing.T) {
 				})
 		})
 	})
+
+	t.Run("when updating one tiertemplate the revisions field should be cleaned up from old entries", func(t *testing.T) {
```
fbm3307, line 399: Minor, should we also check the case, if we create new ttr, the status.revisions is updated , if its not too complicated to write?

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -107,17 +108,14 @@ func (r *Reconciler) ensureRevision(ctx context.Context, nsTmplTier *toolchainv1
...
-		// we need to update the status.revisions with the new ttrs
-		logger.Info("ttr created updating status")
+		revisions[tierTemplate.GetName()] = ttrName
+	}
+
+	// if ttr was created or entries were removed from revisions list
+	// we need to update the status.revisions
+	if ttrCreated || !reflect.DeepEqual(revisions, nsTmplTier.Status.Revisions) {
+		nsTmplTier.Status.Revisions = revisions
+		logger.Info("ttr created or tier removed, updating status")
```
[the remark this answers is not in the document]
mfrancisc: It's not just the revisions , we update whole `Status` field, since we trigger ` r.Client.Status().Update` the line after.

controllers/nstemplatetier/nstemplatetier_controller_test.go
mfrancisc: The TTRs are created by the nstemplatetier_controller, and the status.revisions update is being checked in most of the tests using :

```
tiertest.AssertThatNSTemplateTier(t, "base1ns", cl).
					HasStatusTierTemplateRevisions(tierTemplatesRefs)
``` 

see for example : https://github.com/codeready-toolchain/host-operator/blob/508b798e82cbdd5b706fde35afbcbb671ffa130a/controllers/nstemplatetier/nstemplatetier_controller_test.go#L129-L140

is this what you were looking for ?

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -395,6 +395,56 @@ func TestUpdateNSTemplateTier(t *testing.T) {
...
+
+		// then
+		require.NoError(t, err)
+		// revisions values should be different compared to the previous ones
+		// ensure the old revisions are not there anymore, and instead the update one is in the list
+		tierTemplatesRefs := []string{
+			"base1ns-admin-123456new", "base1nsupdate-clusterresources-123456updated", "base1ns-code-123456new", "base1ns-dev-123456new", "base1ns-edit-123456new", "base1ns-stage-123456new", "base1ns-viewer-123456new",
+		}
+		newNSTmplTier := tiertest.AssertThatNSTemplateTier(t, tierBeingUpdated.Name, cl).
+			HasStatusTierTemplateRevisions(tierTemplatesRefs).Tier()
```
MatousJobanek, line 419: btw, you need to update also the `HasStatusTierTemplateRevisions` so it checks that the length is the same. Currently, it only checks that the refs are present, but it doesn't catch the case when there are also other refs that shouldn't be there.

controllers/nstemplatetier/nstemplatetier_controller_test.go
fbm3307: No i mean with respect to the deletion ,
i was thinking maybe https://github.com/codeready-toolchain/host-operator/pull/1136/files#diff-b4dc433d5162cf28853736dd0fa3f879568b9c587546af8b321c9f304664d27cR116
if this condition can somehow be checked in unit test cases

controllers/nstemplatetier/nstemplatetier_controller.go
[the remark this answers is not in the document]
fbm3307: okay.. got it.. i understood it differently that since in the previous  line we are updating revisions we are loging for it

controllers/nstemplatetier/nstemplatetier_controller_test.go
fbm3307: i see matous's comment of not requiring ttrcreated variable.. if thats the case.. you dont require the case i am talking about

MatousJobanek: this is already covered in the unit-test. the test replaces TierTemplate reference with a new one, which means that it will trigger creation of new TTR CR, plus i twill require deletion of the old entries from the revisions map.

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -107,17 +108,14 @@ func (r *Reconciler) ensureRevision(ctx context.Context, nsTmplTier *toolchainv1
...
-	// and check if there is any reference to a TierTemplate that is not in the Spec anymore
-	if ttrCreated {
-		// we need to update the status.revisions with the new ttrs
-		logger.Info("ttr created updating status")
+		revisions[tierTemplate.GetName()] = ttrName
+	}
+
+	// if ttr was created or entries were removed from revisions list
+	// we need to update the status.revisions
+	if ttrCreated || !reflect.DeepEqual(revisions, nsTmplTier.Status.Revisions) {
```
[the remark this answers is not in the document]
mfrancisc: good point! Added in https://github.com/codeready-toolchain/host-operator/pull/1136/commits/a9916b1164f67999b1ae5278957d72b9be086148

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -429,13 +479,13 @@ func initTierTemplates(t *testing.T, withTemplateObjects []runtime.RawExtension,
...
-	adminRoleTierTemplate := createTierTemplate(t, "admin", withTemplateObjects, tierName)
-	viewerRoleTierTemplate := createTierTemplate(t, "viewer", withTemplateObjects, tierName)
-	editRoleTierTemplate := createTierTemplate(t, "edit", withTemplateObjects, tierName)
+	clusterResourceTierTemplate := createTierTemplate(t, "clusterresources", withTemplateObjects, tierName, "123456new")
+	codeNsTierTemplate := createTierTemplate(t, "code", withTemplateObjects, tierName, "123456new")
+	devNsTierTemplate := createTierTemplate(t, "dev", withTemplateObjects, tierName, "123456new")
+	stageNsTierTemplate := createTierTemplate(t, "stage", withTemplateObjects, tierName, "123456new")
+	adminRoleTierTemplate := createTierTemplate(t, "admin", withTemplateObjects, tierName, "123456new")
+	viewerRoleTierTemplate := createTierTemplate(t, "viewer", withTemplateObjects, tierName, "123456new")
+	editRoleTierTemplate := createTierTemplate(t, "edit", withTemplateObjects, tierName, "123456new")
```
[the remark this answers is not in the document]
mfrancisc: makes sense, not sure why I went with this path which is more complicated 🤦‍♂️ . Thanks for the suggestion . Applied in https://github.com/codeready-toolchain/host-operator/pull/1136/commits/a9916b1164f67999b1ae5278957d72b9be086148

controllers/nstemplatetier/nstemplatetier_controller_test.go
mfrancisc: you're right! Great catch! Thanks! 

Fixed in https://github.com/codeready-toolchain/host-operator/pull/1136/commits/a9916b1164f67999b1ae5278957d72b9be086148

MatousJobanek (approved):
Thanks

PR #1148, opened 2025-02-25
measure UserSignup provision time

MatousJobanek:
Measures the provision time of the UserSignups. 
The start-time is taken from the annotation set in reg-service, as the end-time is taken the moment when UserSignup is marked as Complete.
The provision time is measured only when:
* the user didn't have to complete the whole verification
* the UserSignup wasn't approved manually
* when the annotation is present

The annotation is removed to ensure that we don't record the provision time twice for the same user

TODO:
- [x] e2e tests

depends on:
https://github.com/codeready-toolchain/api/pull/464
https://github.com/codeready-toolchain/toolchain-common/pull/461
https://github.com/codeready-toolchain/registration-service/pull/515 (this needs to be merged first to properly test it in e2e tests)

paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/1126

[SANDBOX-957](https://issues.redhat.com/browse/SANDBOX-957)

### DIFF
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -380,6 +380,12 @@ func (r *Reconciler) checkIfMurAlreadyExists(
 			}
 		}
 
+		logger.Info("Checking whether MUR is Ready", "MUR", mur.Name)
+		if !condition.IsTrueWithReason(mur.Status.Conditions, toolchainv1alpha1.ConditionReady, toolchainv1alpha1.MasterUserRecordProvisionedReason) &&
+			!condition.IsTrue(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete) {
+			return true, r.updateIncompleteStatus(ctx, userSignup, fmt.Sprintf("MUR %s was not ready", mur.Name))
+		}
+
 		logger.Info("Setting UserSignup status to 'Complete'")
 
 		if err := r.updateStatus(ctx, userSignup, r.updateCompleteStatus(mur.Name)); err != nil {
--- a/controllers/usersignup/usersignup_controller.go
+++ b/controllers/usersignup/usersignup_controller.go
@@ -3,8 +3,8 @@ package usersignup
 import (
 	"context"
 	"fmt"
-
 	"strconv"
+	"time"
 
 	recaptcha "cloud.google.com/go/recaptchaenterprise/v2/apiv1"
 	recaptchapb "cloud.google.com/go/recaptchaenterprise/v2/apiv1/recaptchaenterprisepb"
@@ -382,10 +382,44 @@ func (r *Reconciler) checkIfMurAlreadyExists(
 
 		logger.Info("Setting UserSignup status to 'Complete'")
 
-		return true, r.updateStatus(ctx, userSignup, r.updateCompleteStatus(mur.Name))
+		if err := r.updateStatus(ctx, userSignup, r.updateCompleteStatus(mur.Name)); err != nil {
+			return true, err
+		}
+		if err := recordProvisionTime(ctx, r.Client, userSignup); err != nil {
+			return true, fmt.Errorf("unable to record provision time: %w", err)
+		}
+		return true, nil
 	}
 
 	return false, nil
+}
+
+func recordProvisionTime(ctx context.Context, cl runtimeclient.Client, userSignup *toolchainv1alpha1.UserSignup) error {
+	requestReceivedTime, ok := userSignup.Annotations[toolchainv1alpha1.UserSignupRequestReceivedTimeAnnotationKey]
+	// if annotation not present, then nothing to record
+	if !ok {
+		return nil
+	}
+	// delete annotation to make sure that we don't record the same UserSignup twice
+	delete(userSignup.Annotations, toolchainv1alpha1.UserSignupRequestReceivedTimeAnnotationKey)
+	if err := cl.Update(ctx, userSignup); err != nil {
+		return err
+	}
+	logger := log.FromContext(ctx)
+	parsedTime, err := time.Parse(time.RFC3339, requestReceivedTime)
+	if err != nil {
+		logger.Error(err, "unable to parse the request-received-time annotation", "value", requestReceivedTime)
+		return nil
+	}
+	_, phoneVerificationTriggered := userSignup.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey]
+	// don't measure provision time for cases when UserSignup was either approved manually or it required phone verification step
+	if states.ApprovedManually(userSignup) || phoneVerificationTriggered {
+		return nil
+	}
+	provisionTimeSeconds := time.Since(parsedTime).Seconds()
+	logger.Info("provision time", "time-in-seconds", provisionTimeSeconds)
+	metrics.UserSignupProvisionTimeHistogram.Observe(provisionTimeSeconds)
+	return nil
 }
 
 func (r *Reconciler) ensureNewMurIfApproved(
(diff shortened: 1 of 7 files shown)


alexeykazakov (commented):
Code-wise looks good but this metric measures the time between the Signup request (UserSignupRequestReceivedTimeAnnotationKey which I guess will be added later in a separate PR?) and the UserSingup set to the Complete status.

But https://docs.google.com/document/d/1xr1CVKIjcTdp5obpmTMGoNr7syFplnVArd-h5-yQnDk/edit?tab=t.0#heading=h.xd8t24f9nad talks about the time from the signup request to the namespace/user creation. Which basically means a combination of UserSignup complete status AND Space & Mur Ready statuses.
The UserSignup.Status.Complete doesn't necessarily mean everything has been provisioned.

MatousJobanek:
> Code-wise looks good but this metric measures the time between the Signup request (UserSignupRequestReceivedTimeAnnotationKey which I guess will be added later in a separate PR?) and the UserSingup set to the Complete status.

I guess that you already noticed the other PR in the reg-service: https://github.com/codeready-toolchain/registration-service/pull/515

> Which basically means a combination of UserSignup complete status AND Space & Mur Ready statuses.
The UserSignup.Status.Complete doesn't necessarily mean everything has been provisioned.

Look at this part of the code: https://github.com/codeready-toolchain/host-operator/blob/ec07825f0d844334fdc876155b554a2ed0eded08/controllers/usersignup/usersignup_controller.go#L375-L380
it actually waits until the Space is ready and provisioned, so this part is covered. Since MUR is created before and acts immediately on the presence of SpaceBinding and Space.Spec.TargetCluster, we could assume that it should be ready before the Space, but it's true that we don't check it explicitly. I added the check/wait here 84c6983d

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -4717,32 +4760,83 @@ func TestUserSignupStatusNotReady(t *testing.T) {
...
 		res, err := r.Reconcile(context.TODO(), req)
+
+		// then
 		require.NoError(t, err)
 		require.Equal(t, reconcile.Result{}, res)
-		// then
 		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: userSignup.Name, Namespace: req.Namespace}, userSignup)
 		require.NoError(t, err)
-		test.AssertConditionsMatch(t, userSignup.Status.Conditions, signupIncomplete...)
+		test.AssertConditionsMatch(t, userSignup.Status.Conditions, signupIncomplete("space")...)
```
mfrancisc, line 4818: Is this test and the subtest different compared to "until Space is provisioned" and  "when space is provisioned, but not mur" , they seem to be checking the same conditions, but I might have missed something.

controllers/usersignup/usersignup_controller_test.go
```diff
@@ -4837,20 +4931,124 @@ func TestUserReactivatingWhileOldSpaceExists(t *testing.T) {
...
+			client := test.NewFakeClient(t, userSignup)
+
+			// when
+			err := recordProvisionTime(context.TODO(), client, userSignup)
+
+			// then
+			require.NoError(t, err)
+			AssertThatUserSignup(t, test.HostOperatorNs, userSignup.Name, client).DoesNotHaveAnnotation(toolchainv1alpha1.UserSignupRequestReceivedTimeAnnotationKey)
+		}
+		for _, value := range []int{1, 2, 3, 5, 8, 13, 21, 34, 55, 89} {
```
mfrancisc, line 4950: very minor - I was wondering where the buckets are coming from, If I'm not mistaken those are defined and created in the `metrics` package, should we extract this to a variable in there ?

mfrancisc (commented):
Nice job 👍 

I have only few cosmetic comments.

controllers/usersignup/usersignup_controller.go
```diff
@@ -382,12 +382,46 @@
...
+	requestReceivedTime, ok := userSignup.Annotations[toolchainv1alpha1.UserSignupRequestReceivedTimeAnnotationKey]
+	// if annotation not present, then nothing to record
+	if !ok {
+		return nil
+	}
+	// delete annotation to make sure that we don't record the same UserSignup twice
+	delete(userSignup.Annotations, toolchainv1alpha1.UserSignupRequestReceivedTimeAnnotationKey)
+	if err := cl.Update(ctx, userSignup); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
MatousJobanek: Yeah, I was thinking about it too. For the simplicity, I decided to delete it and log the information. In this case, the presence of the annotation says if the given UserSignup was already recorded or not. If we would want to keep it, then we would need to add another field into UserSignup.Status to store the provision time.
Both options are valid, I decided to go with the simpler one. We can always change it later.
But if there would be more people asking for having the information stored in UserSignup, then I can do it in this PR.

controllers/usersignup/usersignup_controller_test.go
MatousJobanek: the buckets were defined by me :-) I decided to use the fibonacci sequence.
As for extracting the buckets into an exported variable. Yeah, we could do that, but assuming some sequence in the unit-tests is a valid approach too, so it starts failing (as an reminder) as soon as we decide to change the buckets in the metrics.

MatousJobanek: it would actually complicate the things - the particular cases use different types (`float64` vs `int` vs `uint64`), so I would keep it as it is

controllers/usersignup/usersignup_controller_test.go
MatousJobanek: good point, they are technically the same. The only diff is that the Space & MUR have specific condition set in the second case, but it doesn't really matter if they don't have any condition at all, or some specific set - in both cases they are not ready. 
removed the second test in 97eeacf9

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
mfrancisc: Sorry , I was just wondering if we should move those lines after line 427, not keeping the information. Just move it as last thing in the function.

controllers/usersignup/usersignup_controller_test.go
mfrancisc: Thanks for looking into it !

alexeykazakov:
> it actually waits until the Space is ready and provisioned, so this part is covered. Since MUR is created before and acts immediately on the presence of SpaceBinding and Space.Spec.TargetCluster, we could assume that it should be ready before the Space, but it's true that we don't check it explicitly. I added the check/wait here

OK. I missed when we started waiting for the Space to be ready. And with this change (wait for MUR) we can clean up our reg-service a bit: https://github.com/codeready-toolchain/registration-service/blob/60b898ffd19907210b19e9a43f0b26facabb1583/pkg/signup/service/signup_service.go#L398-L399
It still assumes that MUR can be NotReady when UserSignup is Complete. But it's minor/cosmetic and we can do it later.

alexeykazakov (approved):
Looks good!

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
MatousJobanek: ah, gotcha. 
I used this order intentionally to ensure that the provision time for the particular UserSignup will be recorded only once. If the update fails (because of a conflict) then the controller requeue a new reconcile and try to remove the annotation again. With this approach, the provision time will be recorded only when the removal of the annotation is successful. If the removal was at the end of the function and the update failed, then we would have two records for the same UserSignup.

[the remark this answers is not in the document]
mfrancisc: gotcha, makes sense! thanks for explaining that!

I was thinking about the case in which fails after the update and before the recording ( basically parsing the value ). But I think that is tolerated, more than having duplicated recordings.

mfrancisc (approved):
Nice Job 🚀

controllers/usersignup/usersignup_controller.go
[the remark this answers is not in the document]
MatousJobanek: If it fails for parsing the value, then there is some problem between reg-service and host-operator and we should only log the error and not return it nor try to parse it again. The resulting duration would be irrelevant and we shouldn't block the UserSignup because of some broken telemetry data.

MatousJobanek:
/retest
fix in e2e tests

PR #1149, opened 2025-02-25
feat: manage ready condition in NSTemplatetier

mfrancisc:
This PR adds/updates the Ready condition in the NSTemplateTier. 

depends on: https://github.com/codeready-toolchain/api/pull/465

jira: https://issues.redhat.com/browse/KUBESAW-251


controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -241,3 +255,38 @@ func NewTTR(tierTmpl *toolchainv1alpha1.TierTemplate, nsTmplTier *toolchainv1alp
...
+	var updated bool
+	tmplTier.Status.Conditions, updated = condition.AddOrUpdateStatusConditions(tmplTier.Status.Conditions, conditions...)
+	if !updated {
+		// Nothing changed
+		return nil
+	}
+	return r.Client.Status().Update(ctx, tmplTier)
+}
+
+func ReadyCondition() toolchainv1alpha1.Condition {
```
fbm3307, line 286: should we add this in toolchain-common instead of it here as we have for different conditions?

fbm3307 (commented):
Great work , just some minor suggestions/questions.

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -24,6 +26,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const UnableToEnsureRevisions = "UnableToEnsureRevisions"
+
+// NSTemplateTierProvisionedReason represents the reason for a successfully provisioned NSTemplateTier.
+const NSTemplateTierProvisionedReason = "Provisioned"
```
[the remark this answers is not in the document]
mfrancisc: I see that usually we have the `Reason` suffix, so I just followed the convetion. I've moved those to the api repo as we usually do for those reasons, see : https://github.com/codeready-toolchain/api/pull/465

controllers/nstemplatetier/nstemplatetier_controller.go
mfrancisc: I was looking into it, but unless I've missed something , I see that we usually construct those in the controllers, for example see space_controller. But maybe I've missed where we keep them in toolchain-common. As far as I can see in common we have the test ones only.

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -268,9 +275,18 @@ func TestReconcile(t *testing.T) {
 					_, err := r.Reconcile(context.TODO(), req)
 					// then
 					// we expect an error caused by the absence of the tiertemplate for the `code` namespace CR
-					require.ErrorContains(t, err, "tiertemplates.toolchain.dev.openshift.com \"base1ns-code-123456new\" not found")
+					errCause := "tiertemplates.toolchain.dev.openshift.com \"base1ns-code-123456new\" not found"
```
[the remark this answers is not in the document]
mfrancisc: We could, but unless I've missed something it will be same as this one, because it will have to return an error from the `ensureRevision` function.

controllers/nstemplatetier/nstemplatetier_controller.go
fbm3307: okay this can work.. if we don't use these condition outside of controller., it doesn't make sense to have it in toolchain-common.

i was actually referring to the conditions we set in toolchaincluster_controller.go https://github.com/codeready-toolchain/toolchain-common/blob/03e9ebab49c77d2c3906e452d9a16dd04d0f7444/controllers/toolchaincluster/toolchaincluster_controller.go#L130
and hence thought that similar conditions can be kept in toolchain-common controllers.

controllers/nstemplatetier/nstemplatetier_controller.go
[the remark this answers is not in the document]
fbm3307: Okay makes sense

controllers/nstemplatetier/nstemplatetier_controller_test.go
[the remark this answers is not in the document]
fbm3307: yea, i saw that, but i just thought if may be a comment or the test name could reflect what we are doing in this test, if we are checking condition etc.
but its just good to have thing can be done later.

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -221,3 +231,38 @@ func NewTTR(tierTmpl *toolchainv1alpha1.TierTemplate, nsTmplTier *toolchainv1alp
 
 	return ttr
 }
+
+func (r *Reconciler) setStatusFailed(ctx context.Context, tmplTier *toolchainv1alpha1.NSTemplateTier, reason string, cause error) error {
+	if err := r.updateStatus(
+		ctx,
+		tmplTier,
+		toolchainv1alpha1.Condition{
```
fbm3307, line 239: Minor, if we can move this condition out as a separate func or variable it will simplify the code, similar to `ReadyCondition()` you have done

controllers/nstemplatetier/nstemplatetier_controller_test.go
```diff
@@ -268,9 +275,18 @@ func TestReconcile(t *testing.T) {
...
+					require.ErrorContains(t, err, errCause)
 					// the revisions field also should remain empty
 					tiertest.AssertThatNSTemplateTier(t, "base1ns", cl).
+						HasConditions(
+							toolchainv1alpha1.Condition{
+								Type:    toolchainv1alpha1.ConditionReady,
```
fbm3307, line 284: Same here. if we can move this out?

fbm3307 (approved):
Looks Good, 
just good to have , for code readability

controllers/nstemplatetier/nstemplatetier_controller.go
mfrancisc: for toolchaincluster we have the conditions along side with the controller, since that is a common controller. In this case the controller code is stored in host operator repo. 

But if we see any need for those conditions outside this repo we can always extract them later 👍

controllers/nstemplatetier/nstemplatetier_controller.go
mfrancisc: makes sense , done in https://github.com/codeready-toolchain/host-operator/pull/1149/commits/45404cdf31313c45a6c6437aaf90bf4cc2199775

metlos (approved):

mfrancisc:
/retest

flaky subspace test : https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1149/pull-ci-codeready-toolchain-host-operator-master-e2e/1896912619076325376/build-log.txt 

looking at it

MatousJobanek (approved):

mfrancisc:
it looks like we might have a flaky unit test also:
```
     metric.go:40: 
            	Error Trace:	/home/runner/go/pkg/mod/github.com/codeready-toolchain/toolchain-common@v0.0.0-20250303095208-d379ee86d136/pkg/test/metrics/metric.go:40
            	            				/home/runner/work/host-operator/host-operator/controllers/usersignup/usersignup_controller_test.go:4907
            	Error:      	Not equal: 
            	            	expected: 0xd
            	            	actual  : 0xc
            	Test:       	TestRecordProvisionTime/should_be_in_histogram
```
will take a look at it as well.

fbm3307:
> it looks like we might have a flaky unit test also:
> 
> ```
>      metric.go:40: 
>             	Error Trace:	/home/runner/go/pkg/mod/github.com/codeready-toolchain/toolchain-common@v0.0.0-20250303095208-d379ee86d136/pkg/test/metrics/metric.go:40
>             	            				/home/runner/work/host-operator/host-operator/controllers/usersignup/usersignup_controller_test.go:4907
>             	Error:      	Not equal: 
>             	            	expected: 0xd
>             	            	actual  : 0xc
>             	Test:       	TestRecordProvisionTime/should_be_in_histogram
> ```
> 
> will take a look at it as well.

@mfrancisc , its because it checks the sequence too... Probably using `assert.elementsmatch` should work rather than `require.equal`

mfrancisc:
Not sure I'm following. 

TBH to me it seems that the bucket nr 13 had 12 as value. So basically it expected 13 but it got 12. Which seems like a timing issue when using the prometheus client. 

To me the only way to make the tests 100% reproducible would be to create a mock/stub instead of using the prometheus library. Thus the test would only assert that expected function were called with expected values.

PR #1153, opened 2025-03-07
KUBESAW-264: Implement TierTemplateRevision cleanup controller 

fbm3307:
This PR is to create a cleanup controller for TierTemplateRevision CRs so that it deletes all the TierTemplateRevisions CRs that are not used by any Space


controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,128 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr manager.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&toolchainv1alpha1.TierTemplateRevision{}).
+		Complete(r)
+}
+
+type Reconciler struct {
+	Client runtimeclient.Client
+	Scheme *runtime.Scheme
```
MatousJobanek, line 33: the Scheme doesn't seem to be used in the whole controller

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	// fetch the related NSTemplateTier tier
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: ttr.GetNamespace(),
+		Name:      tierName,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("NSTemplateTier not found")
+			return reconcile.Result{}, nil
+		}
```
MatousJobanek, line 64: We cannot return here, if the NSTemplateTier is removed, then we should eventually remove the TTR CRs as well

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("NSTemplateTier not found")
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current NSTemplateTier")
+		return reconcile.Result{}, fmt.Errorf("unable to get the current NSTemplateTier: %w", err)
+	}
+
+	// verify that the tier template revision which is unsused(not referenced by any tiers and whose creation date is older that 30secs
```
rsoaresd, line 80:
```suggestion
 // verify that the tier template revision which is unused(not referenced by any tiers and whose creation date is older that 30secs
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+		if err := r.Client.Delete(ctx, ttr); err != nil {
+			return reconcile.Result{}, fmt.Errorf("unable to delete the current Tier Template Revision %s: %w", ttr.Name, err)
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// VerifyUnusedTTR function verifies that the tier template revision's creation time stamp is older than 30sec,
+// it is not present/refernced in status.revisions field, and there are no outdated spaces.
```
rsoaresd, line 98:
```suggestion
// it is not present/referenced in status.revisions field, and there are no outdated spaces.
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	logger := log.FromContext(ctx)
+	//check if the ttr name is present status.revisions
+	for _, ttStatusRev := range nsTmplTier.Status.Revisions {
+		if ttStatusRev == rev.Name {
+			logger.Info("the revision is still being referenced in status.revisions")
+			return false, nil
+		}
+	}
+
+	// get the outdated matchig label to list outdated spaces
```
rsoaresd, line 111:
```suggestion
	// get the outdated matching label to list outdated spaces
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	// look-up all spaces associated with the NSTemplateTier which are outdated
+	spaces := &toolchainv1alpha1.SpaceList{}
+	if err := r.Client.List(ctx, spaces, runtimeclient.InNamespace(nsTmplTier.Namespace),
+		matchOutdated, runtimeclient.Limit(1)); err != nil {
+		return false, err
+	}
+
+	//If there has been an update on nstemplatetier, it might be in a process to update all the spaces.
+	// so we need to check that that there should not be any outdated spaces.
```
rsoaresd, line 125:
```suggestion
 // so we need to check that there should be no outdated spaces.
```

rsoaresd (commented):
Great work! 🚀

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// Delete the unused revision
+	if unusedRevisionBool {
+		if err := r.Client.Delete(ctx, ttr); err != nil {
+			return reconcile.Result{}, fmt.Errorf("unable to delete the current Tier Template Revision %s: %w", ttr.Name, err)
+		}
+	}
+
+	return reconcile.Result{}, nil
```
rsoaresd, line 94: Sorry if it is a "noob question", shouldn't we requeue here?

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// fetch the NSTemplateTier tier
+	ttr := &toolchainv1alpha1.TierTemplateRevision{}
+	if err := r.Client.Get(ctx, request.NamespacedName, ttr); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("TierTemplateRevision not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current TierTemplateRevision")
```
mfrancisc, line 45: minor - there is no need to log the errors if we return them, those will be logged by the controller runtime.

mfrancisc (commented):
Overall looks good, I still need to take a better look at the unit tests 👍 
Nice job 🚀

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: Actually we thought about having a requeue, this would anyways be reconciled by default when controller runtime does it every 5-7 or something, and this controller is good to have clean up so we do not want to overload the reconciliation and hence no requeue

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	//the ttr age should be greater than 30 seconds
+	if timeSinceCreation < deletionTimeThreshold {
+		requeAfter := deletionTimeThreshold - timeSinceCreation
+		return reconcile.Result{RequeueAfter: requeAfter, Requeue: true}, nil
+	}
+
+	//check if there is tier-name label available
+	tierName, ok := ttr.GetLabels()[toolchainv1alpha1.TierLabelKey]
+	if !ok {
+		logger.Info("tier-name label not found in tiertemplaterevision")
```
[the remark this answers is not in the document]
fbm3307: i had done that in the previous commit, but somehow i thought returning an error from controller is not worth

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: ttr.GetNamespace(),
+		Name:      tierName,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("NSTemplateTier not found")
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current NSTemplateTier")
+		return reconcile.Result{}, fmt.Errorf("unable to get the current NSTemplateTier: %w", err)
```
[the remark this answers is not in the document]
fbm3307: i do have unit test case for that, i was actually returning an error for not found ,
but due to this https://github.com/codeready-toolchain/host-operator/pull/1153#discussion_r1986787838 discussion it got me thinking if we need to remove the ttr if nstemplate tier is not found(means its deleted)? it doesn't make sense to reconcile and return an error if there is not related nstemlatetier itself. Not sure if that case does happen or not .

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+			return reconcile.Result{}, fmt.Errorf("unable to delete the current Tier Template Revision %s: %w", ttr.Name, err)
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// VerifyUnusedTTR function verifies that the tier template revision's creation time stamp is older than 30sec,
+// it is not present/refernced in status.revisions field, and there are no outdated spaces.
+func (r *Reconciler) VerifyUnusedTTR(ctx context.Context, nsTmplTier *toolchainv1alpha1.NSTemplateTier,
```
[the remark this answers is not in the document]
fbm3307: you are right,

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
MatousJobanek: I believe that it would be completely valid to delete such a TTR - if there is no label, then there is no ownership of it. 
I doubt that someone would monitor the logs to check if the cleanup controller is printing something out there

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
MatousJobanek: @mfrancisc is correct, you are missing else statement. With this code, it will return an error for TTRs whose NSTemplateTier is deleted, it will actually keep reconciling and won't ever get to the point when the controller could delete it

>  it doesn't make sense to reconcile and return an error if there is no related nstemlatetier itself. Not sure if that case does happen or not .

we discussed that in the call and it's also mentioned in my comment above - we want to delete TTRs whose NSTemplateTier is not present - and we should follow the same pre-conditions (age & all Spaces are up to date)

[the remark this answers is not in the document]
mfrancisc: yes we discussed this with @fbm3307 today in our 1:1 , we should delete the TTR and return. We can check the 

> we want to delete TTRs whose NSTemplateTier is not present - and we should follow the same pre-conditions (age & all Spaces are up to date)

we can check the age, not sure about the Spaces, the tier is gone so there is no way to compute the hash, but maybe I'm missing something.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	//there is no point in fetching  the NStemplateTier, if the TTR  is just created
+
+	//get the tier template revision creation time stamp and the duration
+	timeSinceCreation := time.Since(ttr.GetCreationTimestamp().Time)
+
+	//the ttr age should be greater than 30 seconds
+	if timeSinceCreation < deletionTimeThreshold {
+		requeAfter := deletionTimeThreshold - timeSinceCreation
+		return reconcile.Result{RequeueAfter: requeAfter, Requeue: true}, nil
```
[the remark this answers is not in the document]
mfrancisc: it's minor thing - I was referring to this: https://github.com/kubernetes-sigs/controller-runtime/blob/f80bc5dbf8f7d60b565179724bc1a77d2db7b342/pkg/reconcile/reconcile.go#L44-L46

there is no need to set `Requeue: true` when setting `RequeueAfter`. In the unit test you could just check that `result.RequeueAfter` when expected after the reconcilation.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
mfrancisc: makes sense to delete it, it's the easiest thing to implement probably.

The other option would be to return the error and have an alert for such cases. I don't expect this to happen - if happens it means we have some bug in our controllers or there was some incident caused by manual updates of the CRs.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
fbm3307: Yes, as francisc mentioned we are checking the age,  already, checking the outdated space is based on nstemplatier, so checking that precodition is not possible, if we dont have the nstemplatetier itself.
thats my understanding,
 i have directly deleted it now PTAL

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
fbm3307: yes but i am seeing it being set in spacecleanupcontroller to  https://github.com/codeready-toolchain/host-operator/blob/f14afd9d6f4b9a4a2ac00a973b26aeb48eca3ef4/controllers/spacecleanup/space_cleanup_controller.go#L114.
so i am thinking to keep it as i am unsure on why it. doesnt work for unit test case.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
fbm3307: Updated the logic to delete the TTR .PTAL

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
MatousJobanek: it won't work for unit-tests, it's a controller-runtime internal logic you cannot verify in our unit-tests. Just read the comment in the code @mfrancisc referenced:
> // RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
	// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
	RequeueAfter [time](https://pkg.go.dev/time).[Duration](https://pkg.go.dev/time#Duration)

There could be some places where it is configured in a wrong way, but that doesn't mean then it couldn't be done better in the new code.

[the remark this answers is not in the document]
fbm3307: > I did try not setting it to true and its written in the result definition also, 
i had already read it , but somehow i was more confused on why it didn't work in unit.
Thanks for explaining, changed it now

rsoaresd:
e2e tests failed with:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x266098b]
goroutine 13 [running]:
testing.tRunner.func1.2({0x2957540, 0x41ad2e0})
	/usr/local/go/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1548 +0x397
panic({0x2957540?, 0x41ad2e0?})
	/usr/local/go/src/runtime/panic.go:914 +0x21f
github.com/codeready-toolchain/toolchain-e2e/testsupport.(*SignupRequest).Execute(0xc0011b4d00, 0xc000103d40)
	/tmp/toolchain-e2e/testsupport/signup_request.go:252 +0x70b
github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel.TestE2EFlow(0xc000103d40?)
	/tmp/toolchain-e2e/test/e2e/parallel/e2e_test.go:113 +0x149a
testing.tRunner(0xc000103d40, 0x2d65420)
	/usr/local/go/src/testing/testing.go:1595 +0xff
created by testing.(*T).Run in goroutine 1
	/usr/local/go/src/testing/testing.go:1648 +0x3ad
FAIL	github.com/codeready-toolchain/toolchain-e2e/test/e2e/parallel	208.319s
```

If we check [ToolchainConfig](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_host-operator/1153/pull-ci-codeready-toolchain-host-operator-master-e2e/1899733891980201984/artifacts/e2e/codeready-toolchain-gather/artifacts/toolchainconfigs.toolchain.dev.openshift.com.json), we can see that `automaticApproval` is empty, which led to hit nil in `signup_request.go:252`

It's the first time that I see it. I also checked [here](https://search.dptools.openshift.org/?search=panic%3A+runtime+error%3A&maxAge=336h&context=1&type=build-log&name=.*codeready-toolchain.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job), and at least in the last 14 days, just happened today

fbm3307:
/test e2e
looks unrelated failure

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+		t.Run("revision not found", func(t *testing.T) {
+			r, req, _ := prepareReconcile(t, "base")
+			//when
+			res, err := r.Reconcile(context.TODO(), req)
+			//then
+			require.NoError(t, err)
+			require.Equal(t, controllerruntime.Result{}, res)
+
+		})
+		t.Run("NSTemplate Tier not found", func(t *testing.T) {
```
mfrancisc, line 170: just one suggestion to make it clear of what the behavior should be in this case
```suggestion
		t.Run("NSTemplate Tier not found - ttr gets deleted", func(t *testing.T) {
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+			}
+			//when
+			res, err := r.Reconcile(context.TODO(), req)
+			//then
+			require.EqualError(t, err, "unable to delete the current Tier Template Revision base1ns-clusterresources-123456new-ttrcr: some error cannot delete")
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+		})
+		t.Run("tier label not found", func(t *testing.T) {
+			ttr.Labels = map[string]string{}
```
mfrancisc, line 206: I wouldn't modify this object if it's being used by other tests, maybe we could create it's own ttr object

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+			//when
+			res, err := r.Reconcile(context.TODO(), req)
+			//then
+			require.NoError(t, err)
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).DoNotExist()
+		})
+
+		t.Run("tier label not found but error deleting ttr", func(t *testing.T) {
+			ttr.Labels = map[string]string{}
```
mfrancisc, line 217: same as above, I would modify the ttr object used by other subtests

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+				return fmt.Errorf("some error cannot delete")
+			}
+			//when
+			res, err := r.Reconcile(context.TODO(), req)
+			//then
+			require.EqualError(t, err, "unable to delete the current Tier Template Revision base1ns-clusterresources-123456new-ttrcr: some error cannot delete")
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+		})
+		t.Run("tier label not found", func(t *testing.T) {
```
mfrancisc, line 205:
```suggestion
		t.Run("tier label not found - ttr gets deleted", func(t *testing.T) {
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,145 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// Delete the unused revision
+	if isTTrUnused {
+		if err := r.deleteTTR(ctx, ttr); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
```
mfrancisc, line 98: minor code simplification 
```suggestion
	if isTTrUnused {
		return reconcile.Result{}, r.deleteTTR(ctx, ttr)
	}
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+	return ttr
+}
+
+func createSpace(nsTTier *toolchainv1alpha1.NSTemplateTier) *toolchainv1alpha1.Space {
+	testSpace := spacetest.NewSpace(test.HostOperatorNs, "oddity1",
+		spacetest.WithTierNameAndHashLabelFor(nsTTier),
+		spacetest.WithSpecTargetCluster("member-1"),
+		spacetest.WithStatusTargetCluster("member-1"),
+		spacetest.WithFinalizer(),
+		spacetest.WithCondition(spacetest.Ready()))
```
mfrancisc, line 274: it doesn't seem that we actually need those for the sake of the test
```suggestion
		spacetest.WithTierNameAndHashLabelFor(nsTTier))
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+	ttr := &toolchainv1alpha1.TierTemplateRevision{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:         operatorNamespace,
+			Name:              name,
+			Labels:            labels,
+			CreationTimestamp: crtime,
+		},
+		Spec: toolchainv1alpha1.TierTemplateRevisionSpec{
+			Parameters: nsTTier.Spec.Parameters,
+		},
```
mfrancisc, line 263: unless I'm wrong those are not actually used by test or important for the controller logic
```suggestion
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+			// then
+			require.NoError(t, err)
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+
+		})
+
+		t.Run("spaces are still being updated", func(t *testing.T) {
+			// given
+			nsTemplateTier.Status.Revisions = map[string]string{
```
mfrancisc, line 83: I would create a specific object if we modify it in the test
```suggestion
                         nsTemplateTier := tiertest.Base1nsTier(t, tiertest.CurrentBase1nsTemplates, tiertest.WithStatusRevisions())
			nsTemplateTier.Status.Revisions = map[string]string{
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+			// then
+			require.NoError(t, err)
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+		})
+
+		t.Run("Error while deleting the TTR", func(t *testing.T) {
+			// given
+			nsTemplateTier := tiertest.Base1nsTier(t, tiertest.CurrentBase1nsTemplates, tiertest.WithStatusRevisions())
```
mfrancisc, line 101: we could use the one from the parent test here
```suggestion
```

mfrancisc (commented):
Great job ! Mostly LGTM.

I have only few last minor comments.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
fbm3307: Done PTAL

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,145 @@
...
+		logger.Info("there are still some spaces which are outdated")
+		return false, nil
+	}
+	return true, nil
+}
+
+// deleteTTR function delete the TTR
+func (r *Reconciler) deleteTTR(ctx context.Context, ttr *toolchainv1alpha1.TierTemplateRevision) error {
+	if err := r.Client.Delete(ctx, ttr); err != nil {
+		return fmt.Errorf("unable to delete the current Tier Template Revision %s: %w", ttr.Name, err)
```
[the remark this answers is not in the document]
fbm3307: great catch, updated it and added the test case too for it

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := r.Client.Get(ctx, request.NamespacedName, ttr); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("TierTemplateRevision not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		logger.Error(err, "unable to get the current TierTemplateRevision")
+		return reconcile.Result{}, fmt.Errorf("unable to get the current TierTemplateRevision: %w", err)
+	}
+	//there is no point in fetching  the NStemplateTier, if the TTR  is just created
```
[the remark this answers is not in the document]
fbm3307: its just the explanation of the why we are checking the creation time stamp and related to the below checks.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+			// then
+			require.EqualError(t, err, "unable to delete the current Tier Template Revision base1ns-clusterresources-123456new-ttrcr: some error cannot delete")
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+		})
+
+		t.Run("Is Not Found Error-already deleted, while deleting the TTR", func(t *testing.T) {
+			// given
+			nsTemplateTier := tiertest.Base1nsTier(t, tiertest.CurrentBase1nsTemplates, tiertest.WithStatusRevisions())
```
mfrancisc, line 119:
```suggestion
```

mfrancisc (approved):
Great job 🚀 

Thanks for addressing my comments and for the additional changes 🙏

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	//there is no point in fetching  the NStemplateTier, if the TTR  is just created
+	//get the tier template revision creation time stamp and the duration
+	timeSinceCreation := time.Since(ttr.GetCreationTimestamp().Time)
+
+	//the ttr age should be greater than 30 seconds
+	if timeSinceCreation < DeletionTimeThreshold {
+		requeAfter := DeletionTimeThreshold - timeSinceCreation
+		return reconcile.Result{RequeueAfter: requeAfter}, nil
```
MatousJobanek, line 55: two things: typo and missing log
```suggestion
		requeueAfter := DeletionTimeThreshold - timeSinceCreation
		logger.Info("the TierTemplateRevision is not old enough", "requeue-after", requeueAfter)
		return reconcile.Result{RequeueAfter: requeueAfter}, nil
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	//check if there is tier-name label available
+	tierName, ok := ttr.GetLabels()[toolchainv1alpha1.TierLabelKey]
+	if !ok {
+		//if tier-name label is not found we should delete the TTR
+		if err := r.deleteTTR(ctx, ttr); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
```
MatousJobanek, line 65:
```suggestion
		return reconcile.Result{}, r.deleteTTR(ctx, ttr)
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: ttr.GetNamespace(),
+		Name:      tierName,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			// if there is no related nstemplate tier, we can delete the TTR directly
+			if err := r.deleteTTR(ctx, ttr); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
```
MatousJobanek, line 78:
```suggestion
			return reconcile.Result{}, r.deleteTTR(ctx, ttr)
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+		return reconcile.Result{}, nil
+	}
+	// fetch the related NSTemplateTier tier
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: ttr.GetNamespace(),
+		Name:      tierName,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			// if there is no related nstemplate tier, we can delete the TTR directly
```
MatousJobanek, line 74: it would be nice to add a finalizer to the NSTemplateTier to ensure that it's gonna be deleted only when there is no Space that would use it.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, fmt.Errorf("unable to get the current NSTemplateTier: %w", err)
+	}
+
+	// verify that the tier template revision which is unused(not referenced by any tiers
+	// and all the spaces are up-to-date)
```
MatousJobanek, line 85:
```suggestion
	// let's make sure that we delete the TTR only if it's not used
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	// Delete the unused revision
+	if isTTrUnused {
+		return reconcile.Result{}, r.deleteTTR(ctx, ttr)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// verifyUnusedTTR function verifies that the tier template revision
+// is not present/referenced in status.revisions field, and there are no outdated spaces.
```
MatousJobanek, line 100:
```suggestion
// verifyUnusedTTR function verifies that the TTR is not used (returns true if it's used).
// this is done by:
//   - checking the NSTemplateTier.status.revisions field, if the TTR is referenced there or not
//   - checking if all Spaces are up-to-date. In case there are some outdated space, we could risk that the TTR is still being used
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *Reconciler) verifyUnusedTTR(ctx context.Context, nsTmplTier *toolchainv1alpha1.NSTemplateTier,
+	rev *toolchainv1alpha1.TierTemplateRevision) (bool, error) {
+
+	logger := log.FromContext(ctx)
+	//check if the ttr name is present status.revisions
+	for _, ttStatusRev := range nsTmplTier.Status.Revisions {
+		if ttStatusRev == rev.Name {
+			logger.Info("the revision is still being referenced in status.revisions", "ttr.name", rev.Name)
```
MatousJobanek, line 108: two things:
* improve the message
* there is no need to reference the name of TTR - this is already included in the logger from the context
```suggestion
			logger.Info("the TierTemplateRevision is still being referenced in NSTemplateTier.Status.Revisions")
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,144 @@
...
+
+// deleteTTR function delete the TTR
+func (r *Reconciler) deleteTTR(ctx context.Context, ttr *toolchainv1alpha1.TierTemplateRevision) error {
+	if err := r.Client.Delete(ctx, ttr); err != nil {
+		if errors.IsNotFound(err) {
+			return nil // was already deleted
+		}
+		return fmt.Errorf("unable to delete the current Tier Template Revision %s: %w", ttr.Name, err)
+	}
+	return nil
```
MatousJobanek, line 143:
```suggestion
	log.FromContext(ctx).Info("The TierTemplateRevision has been deleted")
	return nil
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: i added the ttr name as a part of this comment , https://github.com/codeready-toolchain/host-operator/pull/1153#discussion_r1987731526 , changed it now

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: that would be a good point .

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+		})
+		t.Run("NSTemplate Tier not found - ttr gets deleted", func(t *testing.T) {
+			r, req, cl := prepareReconcile(t, ttr.Name, ttr, nsTemplateTier)
+			cl.MockGet = func(ctx context.Context, key types.NamespacedName, obj runtimeclient.Object, opts ...runtimeclient.GetOption) error {
+				if _, ok := obj.(*toolchainv1alpha1.NSTemplateTier); ok {
+					return errors.NewNotFound(schema.GroupResource{}, key.Name)
+				}
+				return cl.Client.Get(ctx, key, obj, opts...)
+			}
```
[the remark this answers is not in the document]
fbm3307: yeah, i had tried that before and somehow i ended up doing it this way, updated now PTAL

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,287 @@
...
+		res, err := r.Reconcile(context.TODO(), req)
+		//then
+		require.NoError(t, err)
+		require.Equal(t, controllerruntime.Result{}, res)
+		tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).DoNotExist()
+	})
+
+	t.Run("Failure", func(t *testing.T) {
+		nsTemplateTier := tiertest.Base1nsTier(t, tiertest.CurrentBase1nsTemplates, tiertest.WithStatusRevisions())
+		ttr := createttr(*nsTemplateTier, (nsTemplateTier.Spec.ClusterResources.TemplateRef + "-ttrcr"), metav1.NewTime(time.Now().Add(-time.Minute)))
```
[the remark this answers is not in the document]
fbm3307: no particular reason, removed it

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,287 @@
...
+			require.Equal(t, controllerruntime.Result{}, res)
+			tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).ExistFor(nsTemplateTier.Name)
+		})
+
+	})
+
+}
+
+func prepareReconcile(t *testing.T, name string, initObjs ...runtimeclient.Object) (*nstemplatetierrevisioncleanup.Reconciler, reconcile.Request, *test.FakeClient) {
+	os.Setenv("WATCH_NAMESPACE", test.HostOperatorNs)
```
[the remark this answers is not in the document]
fbm3307: no particular reason i just copy pasted from other test case file, removed it

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
mfrancisc: my bad, I forgot that it's in the context 🤦‍♂️ . Sorry for the confusion.

fbm3307: i blindly followed the suggestion rather than questioning /understanding it. 😕

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
mfrancisc: we can do that.

I've created a small story for this https://issues.redhat.com/browse/KUBESAW-268 as a reminder.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Delete the unused revision
+	if isTTrUnused {
+		return reconcile.Result{}, r.deleteTTR(ctx, ttr)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// verifyUnusedTTR function verifies that the TTR is not used (returns true if it's used).
```
MatousJobanek, line 93:
```suggestion
// verifyUnusedTTR function verifies that the TTR is not used (returns true if it's NOT used).
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
+		res, err := r.Reconcile(context.TODO(), req)
+		//then
+		require.NoError(t, err)
+		require.Equal(t, controllerruntime.Result{}, res)
+		tiertemplaterevision.AssertThatTTRs(t, cl, nsTemplateTier.GetNamespace()).DoNotExist()
+	})
+
+	t.Run("the creation timestamp is less than 30 sec", func(t *testing.T) {
+		// given
+		ttr := newTTR(*nsTemplateTier, (nsTemplateTier.Spec.ClusterResources.TemplateRef + "-ttr"), metav1.NewTime(time.Now().Add(-29*time.Second)))
```
MatousJobanek, line 45:
```suggestion
		ttr := newTTR(*nsTemplateTier, nsTemplateTier.Spec.ClusterResources.TemplateRef + "-ttr", metav1.NewTime(time.Now().Add(-29*time.Second)))
```

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	// fetch the related NSTemplateTier tier
+	tier := &toolchainv1alpha1.NSTemplateTier{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Namespace: ttr.GetNamespace(),
+		Name:      tierName,
+	}, tier); err != nil {
+		if errors.IsNotFound(err) {
+			// if there is no related nstemplate tier, we can delete the TTR directly
+			return reconcile.Result{}, r.deleteTTR(ctx, ttr)
+		}
```
MatousJobanek, line 74: Btw, if you added a watcher on NSTemplateTier CRD with a predicate that would allow deletion events only, then you could remove the TTR-specific cleanup logic from the e2e tests. 
Can be done as separate PR

MatousJobanek (approved):
Looks good

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := r.Client.Get(ctx, request.NamespacedName, ttr); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("TierTemplateRevision not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, fmt.Errorf("unable to get the current TierTemplateRevision: %w", err)
+	}
+
+	//there is no point in fetching  the NStemplateTier, if the TTR  is just created
```
metlos, line 48: For the benefit of future us, I'd like to see the explanation for why there is no point in fetching the ns template tier if the TTR is recent.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	//get the tier template revision creation time stamp and the duration
+	timeSinceCreation := time.Since(ttr.GetCreationTimestamp().Time)
+
+	//the ttr age should be greater than 30 seconds
+	if timeSinceCreation < DeletionTimeThreshold {
```
metlos, line 53: `DeletionTimeThreshold` -> `minTtrAge`? IMHO that would better reflect what the constant is used for. Also the constant seems to only be used in this file, so we can keep it package-private.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+//   - checking the NSTemplateTier.status.revisions field, if the TTR is referenced there or not
+//   - checking if all Spaces are up-to-date. In case there are some outdated space, we could risk that the TTR is still being used
+func (r *Reconciler) verifyUnusedTTR(ctx context.Context, nsTmplTier *toolchainv1alpha1.NSTemplateTier,
+	rev *toolchainv1alpha1.TierTemplateRevision) (bool, error) {
+
+	logger := log.FromContext(ctx)
+	//check if the ttr name is present status.revisions
+	for _, ttStatusRev := range nsTmplTier.Status.Revisions {
+		if ttStatusRev == rev.Name {
+			logger.Info("the TierTemplateRevision is still being referenced in NSTemplateTier.Status.Revisions")
```
metlos, line 104: Isn't this the expected case? IMHO we should not log this if it is.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	ttr := &toolchainv1alpha1.TierTemplateRevision{}
+	if err := r.Client.Get(ctx, request.NamespacedName, ttr); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("TierTemplateRevision not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, fmt.Errorf("unable to get the current TierTemplateRevision: %w", err)
+	}
+
```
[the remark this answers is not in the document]
MatousJobanek: good catch, but there is no finalizer which means that it should be deleted immediately.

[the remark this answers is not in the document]
MatousJobanek: and if we want to check if it's not being deleted, then we should use `util.IsBeingDeleted(ttr)` instead.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
MatousJobanek: I believe that this could help us with any future debugging of why some specific (old) TTR is not deleted, so I would keep it there. 
This controller won't get too many reconciles, so it shouldn't populate the logs too much.

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
fbm3307: okay just curious, if we don't have finalizers, then any particular reason on why we should still check `isBeingDeleted`, there won't be any delay as per my understanding, may be i missed some case ?

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: updated it

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: i would vote for keeping the logger

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
fbm3307: You meant the watcher in this controller itselt?

fbm3307:
/test e2e

looks unrelated flaky test

controllers/nstemplatetierrevisioncleanup/nstemplatetier_revision_cleanup_controller.go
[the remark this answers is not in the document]
metlos: Usually, if you detect the resource is being deleted, you just stop the reconcile by returning early without an error. If there are finalizers, you first do the clean up logic, remove the finalizer and then return, again not performing the "normal" reconciliation logic.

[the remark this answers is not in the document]
fbm3307: @metlos I have updated it  now PTAL.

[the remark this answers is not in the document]
MatousJobanek: @metlos the point was that if there is no finalizer added, then as soon as the resource has the deletion  timestamp set, then it should be immediately deleted.

metlos (approved):

PR #1156, opened 2025-03-14
revert go-version-file usage in govulncheck gh action

rsoaresd:
# Description
There is a bug when using `go-version-file`: https://github.com/golang/go/issues/70036, which leads to pick go 1.24 as the go version. Reverting to the first approach of using `go-version-input`. Slack thread: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1741939607080469?thread_ts=1741889746.719249&cid=CHK0J6HT6

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -7,11 +7,16 @@ on:
 jobs:
   govulncheck:
     name: govulncheck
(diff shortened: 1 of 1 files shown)


xcoulon (approved):

MatousJobanek (approved):
/lgtm

PR #1158, opened 2025-03-18
KUBESAW-201: Replace go-bindata with go:embed

fbm3307:
We want to replace go-bindata with go:embed so that we don't rely on 3rd party libraries and use functionality provided as part of Go itself. 

This is just a small starting PR . I have just replaced gobindata with go embed for registration-service template.
rest will follow after this PR


alexeykazakov (approved):

mfrancisc (commented):
Looks good 👍 

just one minor suggestion.

mfrancisc (approved):
Great job 🚀 

Thanks for addressing my comments.

MatousJobanek (approved):

PR #1165, opened 2025-04-04
KUBESAW-273: Replace go-bindata with go:embed

fbm3307:
We want to replace go-bindata with go:embed so that we don't rely on 3rd party libraries and use functionality provided as part of Go itself.

This PR is 3rd PR in the series and succeeds PR - https://github.com/codeready-toolchain/host-operator/pull/1161 .
This is for converting nstemplatetiers to use goembed instead of gobindata


pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -2,24 +2,28 @@ package nstemplatetiers
...
 	"k8s.io/apimachinery/pkg/runtime"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const NSTemplateTierRootDir = "templates/nstemplatetiers"
+
 // CreateOrUpdateResources generates the NSTemplateTier resources from the cluster resource template and namespace templates,
 // then uses the manager's client to create or update the resources on the cluster.
-func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runtimeclient.Client, namespace string, assets assets.Assets) error {
+func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runtimeclient.Client, namespace string, nstemplatetierFS embed.FS, root string) error {
```
metlos, line 23: The `nstemplatetierFS` and `root` always use the same values apart from 2 tests. Could we do away with those parameters (ideally, without sacrificing the tests either)?

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -28,16 +32,33 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
...
+	}
+	if len(paths) == 0 {
+		return fmt.Errorf("could not find any user templates")
+	}
+
 	files := map[string][]byte{}
-	for _, name := range assets.Names() {
-		if name == "metadata.yaml" {
+	for _, name := range paths {
+		if name == "templates/nstemplatetiers/metadata.yaml" {
```
metlos, line 46: This feels tightly coupled to `deploy.NSTemplateTiersFS`. This would be ok if we didn't use the `nstemplatetierFS` as a parameter but IMHO is stylistically wrong when we're allowing passing in an arbitrary FS.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -28,16 +32,33 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
...
 			continue
 		}
-		content, err := assets.Asset(name)
+
+		parts := strings.Split(name, "/")
+		// skip any name that does not have 4 parts
```
metlos, line 51: Again, this is assuming the structure of the fs (that we don't restrict in any sense) and therefore feels tightly coupled to the expected `deploy.NSTemplateTiersFS`.

metlos (commented):
generally LGTM but I'd like us to explore the ways to improve the locality of the code and data.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: Okay so this is done here because this [function](https://github.com/codeready-toolchain/host-operator/pull/1165/files#diff-75fb59490e920da60f86dae8e59891d725272ef82d0515a32447036e0b576e40R65) in [toolchain common](https://github.com/codeready-toolchain/toolchain-common/blob/f85b77b1e378ee8ec2c367559e13b0d35470f007/pkg/template/nstemplatetiers/nstemplatetier_generator.go#L150) expects only the last two parts of the folder structure. 
and the test are written accordingly. 

moreover, it has to be in line with how our template folder structure is. similar is for user tiers, notification and others.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: i am not sure i understand it correctly, do you mean to remove these parameters?

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: updated it

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
metlos: Yes, I think ideally the `nstemplatetiersFS` and `root` params should be removed because in the runtime code they always have the same value.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
metlos: I think that illustrates my point :) There's a function in `toolchain-common` that assumes the paths are 2 levels deep. Here, you assume that the paths are 4 levels deep but neither that function nor this function is enforcing that in any sense through the type-system or function signature.

This IMHO makes the code more fragile than it could be. If you removed the two parameters to this function, the assumption about 4 levels would suddenly be local to this function and the callers of this function would no longer need to care about it. This would IMHO improve the code.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: Okay so  i removed the parameters from the main function and and created an another sub func added the parameters there so that i can use it for test and still it addresses the above concern.
PTAL @metlos .

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: i updated the function PTAL

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -51,6 +52,9 @@ func roles(_ string) []string {
 	return []string{"admin"}
 }
 
+func TestMo(t *testing.T) {
```
metlos, line 55: Some leftover?

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -52,3 +40,45 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
 		return applyCl.ApplyObject(ctx, toEnsure, commonclient.ForceUpdate(true))
 	}, namespace, metadata, files)
 }
+
+func LoadFiles(nsTemplateTiers embed.FS, root string) (metadata map[string]string, files map[string][]byte, err error) {
```
metlos, line 44: Maybe add some comment here about the assumptions on the directory structure in the supplied FS?

Also, if the tests were in the same package as this file, this function could be private which again would improve the code cohesion and lower the possibility of misuse.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -2,42 +2,30 @@ package nstemplatetiers
...
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const NSTemplateTierRootDir = "templates/nstemplatetiers"
```
metlos, line 20: This can be private.

metlos (commented):
Thanks for taking my suggestions into account 🙏🏼 

I only have a couple more nitpicks, no need to implement them if you feel that's not necessary now or if you made those choices deliberately with some future use in mind that I didn't see.

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
fbm3307: yes, removed it now

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
fbm3307: added the comments.
i did try to make the func private but the test package didnt pick it up. and hence kept it as is

metlos: The reason why the test didn't see the function when you made it private is because of the fact that, for some reason that I don't fully understand, the test lives in another package, `nstemplatetier_test`, than the code itself - `nstemplatetier`. If the test file declared the same package as the the code, you'd be able to test the private functions, too.

fbm3307: yes but  the reason, we cannot do that is, the import cycles aren't allowed in test, and we do need them in our test so we cannot have the same package

pkg/templates/nstemplatetiers/testtemplates/fakenstemplatetiers/test/cluster.yaml
```diff
@@ -0,0 +1,129 @@
+apiVersion: template.openshift.io/v1
```
mfrancisc, line 1: minor - do we need to have this nested folders or could it just be :
`testtemplates/testtier/cluster|ns_dev|spacerole_admin|tier.yaml` 

I mean removing the `fakenstemplatetier` folder.

mfrancisc (commented):
Great Job 👍 

make/go.mk
```diff
@@ -33,10 +33,7 @@ USERTEMPLATES_TEST_BASEDIR = test/templates/usertiers
 REGISTRATION_SERVICE_DIR=deploy/registration-service
 
 .PHONY: generate
-generate: install-go-bindata generate-metadata generate-assets 
-
-install-go-bindata:
-	@go install github.com/go-bindata/go-bindata/...
+generate: generate-metadata
```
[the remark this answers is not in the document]
fbm3307: it is not related to gobin data and its required to get the commit sha.

pkg/templates/nstemplatetiers/testtemplates/fakenstemplatetiers/test/cluster.yaml
fbm3307: so the logic is such that in it does require 4 step folder structure as it in deploy/ for actual prod . thats how we are getting the path. and hence this structure requires `faketesttemplatetier` . so for test cases we need to have same folder structure. 
https://github.com/codeready-toolchain/host-operator/pull/1165/files#diff-75fb59490e920da60f86dae8e59891d725272ef82d0515a32447036e0b576e40R44

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -52,3 +40,47 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
...
+		}
+
+		parts := strings.Split(name, "/")
+		// skip any name that does not have 4 parts
+		if len(parts) != 4 {
+			return nil, nil, fmt.Errorf("unable to load templates: invalid name format for file '%s'", name)
+		}
+
+		fileName := parts[2] + "/" + parts[3]
+		content, err := deploy.NSTemplateTiersFS.ReadFile(name)
```
mfrancisc, line 79: shouldn't this be"
```suggestion
		content, err := nsTemplateTiers.ReadFile(name)
```

pkg/templates/nstemplatetiers/testtemplates/fakenstemplatetiers/test/cluster.yaml
mfrancisc: yeah, it's a bit weird but not sure how we can make it more flexible.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -52,3 +40,47 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
...
+			continue
+		}
+
+		parts := strings.Split(name, "/")
+		// skip any name that does not have 4 parts
+		if len(parts) != 4 {
+			return nil, nil, fmt.Errorf("unable to load templates: invalid name format for file '%s'", name)
+		}
+
+		fileName := parts[2] + "/" + parts[3]
```
mfrancisc, line 78: minor - I was wondering if we could have this as part of the `GetAllFileNames` ( and maybe it rename it to `GetAllTemplateNames` ), but I guess the `nsTemplateTiers.ReadFile(name)` would than fail. 

Or if there's a way using `nsTemplateTiers.ReadDir("templates/nstemplatetiers")` to just pass the content of the directory, so that we don't have to worry about the 4 parts . As all the other logic requires 2 parts only, see : https://github.com/codeready-toolchain/toolchain-common/blob/c8834962f739648258091d29f9dfe916f9bdadcb/pkg/template/nstemplatetiers/nstemplatetier_generator.go#L150-L152

So ideally also this logic should require 2 parts and not 4. 
WDYT?

fbm3307: @mfrancisc as discussed i did give it a try.. the `.ReadDir()` returns the name object as 3 parts. so probably it will be almost the same logic as now. 
i do understand there might be some extra layer of logic as in toolchain-common with the current code. but i didnt want to change the logic in toolchain common, and hence did the change here.

metlos (approved):

.gitignore
```diff
@@ -9,7 +9,6 @@ tmp/
 
 # generated files
 deploy/**/metadata.yaml
-*_assets.go
```
MatousJobanek, line 12: do not remove this - it already causes unexpected changes included in the commit in member operator.

fbm3307: Going to revert this,
 but i didn't understand it completely, did you mean, it caused some changes since i removed it from member operator so i don't remove it from here..?

MatousJobanek: the `template_assets.go` files are already present in our projects locally. When you removed the .gitignore line, then git automatically picks these files and adds them to the commit. We have to manually delete the `template_assets.go` files to make sure that they are not included in the commit. This is not obvious and a bit tedious work.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
mfrancisc: Thanks for looking into it. Yeah I agree it's fine to keep it as is for now, and later refactor the toolchain common logic maybe.

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -52,3 +40,47 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
...
+	paths, err := templates.GetAllFileNames(&nsTemplateTiers, root)
+	if err != nil {
+		return nil, nil, err
+	}
+	if len(paths) == 0 {
+		return nil, nil, fmt.Errorf("could not find any ns templates")
+	}
+	files = make(map[string][]byte)
+	for _, name := range paths {
+		if name == root+"/metadata.yaml" {
```
MatousJobanek, line 68: I would say that using `filepath.Join` is the best practice for constructing the paths
```suggestion
		if name == filepath.Join(root, "metadata.yaml") {
```

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
MatousJobanek, line 78:
```suggestion
		filepath.Join(parts[2], parts[3])
```

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -52,3 +40,47 @@ func CreateOrUpdateResources(ctx context.Context, s *runtime.Scheme, client runt
 		return applyCl.ApplyObject(ctx, toEnsure, commonclient.ForceUpdate(true))
 	}, namespace, metadata, files)
 }
+
+// LoadFiles takes the file from deploy/nstemplatetiers/<tiername>/<yaml file name> . the folder structure should be 4 steps .
+// as the cologic here is written accordingly
+func LoadFiles(nsTemplateTiers embed.FS, root string) (metadata map[string]string, files map[string][]byte, err error) {
+	// load templates from assets
+	metadataContent, err := nsTemplateTiers.ReadFile(root + "/metadata.yaml")
```
MatousJobanek, line 48:
```suggestion
	metadataContent, err := nsTemplateTiers.ReadFile(filepath.Join(root, "metadata.yaml"))
```

pkg/templates/nstemplatetiers/testtemplates/fakenstemplatetiers/test/cluster.yaml
```diff
@@ -0,0 +1,129 @@
...
+      hard:
+        count/deployments.apps: "30"
+        count/deploymentconfigs.apps: "30"
+        count/pods: "50"
+        count/virtualmachines.kubevirt.io: "2"
+    selector:
+      annotations: null
+      labels:
+        matchLabels:
+          toolchain.dev.openshift.com/space: ${SPACE_NAME}
```
MatousJobanek, line 21: could you please point me to the place where these templates are currently used/defined? I cannot find anything corresponding to it.
My question is where (and why) did we get these files.

MatousJobanek (approved):

PR #1166, opened 2025-04-09
Update idler email

kyetter:
Let me know what you think about this!


deploy/templates/notificationtemplates/sandbox/idlertriggered/notification.html
```diff
@@ -41,11 +41,12 @@
 >
 
     <p>
-        You are receiving this email because one or more of your workloads in the Developer Sandbox has been running for an extended period and has been automatically idled.
+        Thank you for using the Developer Sandbox! We wanted to let you know that we have scaled down one or more of your workloads.
```
rajivnathan, line 44: I like this friendlier tone much more! I think the previous wording makes it seem like something bad happened and that was maybe part of the problem. Maybe we can make it even more clear that this is an expected outcome with something like 

```suggestion
        Thank you for using the Developer Sandbox! We wanted to let you know that we have scaled down one or more of your workloads. This is completely automatic and no action is required by you.
```

but what you have looks good to me as well

kyetter: I appreciate the suggestion! Let me tweak a little bit

alexeykazakov (approved):
I like it!

deploy/templates/notificationtemplates/sandbox/idlertriggered/notification.html
```diff
@@ -41,11 +41,12 @@
...
     </p>
 
     <p>
-        Our system automatically idles workloads that run for an extended time or enter a crashloop state. No further action is required by you. You can <a href="https://developers.redhat.com/learn/openshift/revive-inactive-openshift-pods-scaled-zero">restart your workload,</a> {{.AppType}} {{.AppName}}
+        Our system automatically idles workloads that run for an extended time or enter a crashloop state.
+        You can <a href="https://developers.redhat.com/learn/openshift/revive-inactive-openshift-pods-scaled-zero">restart your workload,</a> {{.AppType}} {{.AppName}}
```
alexeykazakov, line 49: Could you please also update the tests?
https://github.com/codeready-toolchain/host-operator/blob/d7cb28224462803ed805a465c3cbaeb92fb422ea/pkg/templates/notificationtemplates/notification_generator_test.go#L68-L69

alexeykazakov (approved):

PR #1171, opened 2025-04-30
Deploy the registration service using the new SSA client

metlos:
This was not as simple as initially thought because SSA client does not report the fact whether the objects were updated or not back to the caller.

This was relied upon when determining the "deploying" and "deployed" phases of the registration service as reported in the `ToolchainConfig.status`.

This has been discussed and we agreed to drop the distinction and merely consider the registration service not deployed or deployed.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -70,7 +70,8 @@ var (
 )
 
 const (
-	memberClientTimeout = 3 * time.Second
+	memberClientTimeout      = 3 * time.Second
+	hostOperatorFieldManager = "kubesaw-host-operator"
 )
 
 func init() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -294,6 +294,7 @@ func main() { // nolint:gocyclo
 		Client:         mgr.GetClient(),
 		GetMembersFunc: commoncluster.GetMemberClusters,
 		Scheme:         mgr.GetScheme(),
+		FieldManager:   hostOperatorFieldManager,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "ToolchainConfig")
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -293,6 +294,7 @@ func main() { // nolint:gocyclo
 		Client:         mgr.GetClient(),
 		GetMembersFunc: commoncluster.GetMemberClusters,
 		Scheme:         mgr.GetScheme(),
+		FieldManager:   hostOperatorFieldManager,
```
MatousJobanek, line 297: do we need to store it as a field of the Reconciler type? Couldn't we move it as an exported (global) constant and use it directly when creating the SSA client? The field is not set in the unit tests at all, so using the constant should work fine.

MatousJobanek (approved):
Cool :+1: Thanks for the simplification

controllers/toolchainconfig/toolchainconfig_controller_test.go
```diff
@@ -233,8 +233,8 @@ func TestReconcile(t *testing.T) {
 			config := commonconfig.NewToolchainConfigObjWithReset(t,
 				testconfig.AutomaticApproval().Enabled(true))
 			hostCl := test.NewFakeClient(t, config)
-			hostCl.MockCreate = func(ctx context.Context, obj runtimeclient.Object, opts ...runtimeclient.CreateOption) error {
-				return fmt.Errorf("create error")
+			hostCl.MockPatch = func(ctx context.Context, obj runtimeclient.Object, patch runtimeclient.Patch, opts ...runtimeclient.PatchOption) error {
```
fbm3307, line 236: Just for understanding, do we not require to test any create option here.?

fbm3307 (approved):

PR #1177, opened 2025-05-22
Remove the advanced ns tiertemplate

metlos:
This removes the `advanced` namespace template tier because it is no longer used in production.

Related PRs:
* toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1151

### DIFF
--- a/pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
+++ b/pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
@@ -24,7 +24,6 @@ import (
 )
 
 var expectedProdTiers = []string{
-	"advanced",
 	"base",
 	"base1ns",
 	"base1nsnoidling",
@@ -51,7 +50,6 @@ func roles(_ string) []string {
 }
 
 func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
-
 	s := scheme.Scheme
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -24,7 +24,6 @@ import (
 )
 
 var expectedProdTiers = []string{
-	"advanced",
```
rsoaresd, line 27: Should we also clean the test files? For example:

- https://github.com/codeready-toolchain/host-operator/blob/master/controllers/space/space_controller_test.go#L2256
- https://github.com/codeready-toolchain/host-operator/blob/master/controllers/spacecompletion/space_completion_controller_test.go#L58
- https://github.com/codeready-toolchain/host-operator/blob/master/controllers/toolchainconfig/configuration_test.go#L458
- (...)

rsoaresd (commented):
Great work 🚀

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
metlos: I originally wanted to keep this because 'advanced' would just become "another name", but you're right, let's get rid of that word in the nstemplatetier context everywhere so that people don't fallback to the old assumptions - that tier has been bundled for ages.

The latest commit should not have any 'advanced' tiers mentioned in the tests.

PR #1180, opened 2025-06-04
remove the bundled NSTemplateTiers that are no longer used

metlos:
This compiles the list of the `NSTemplateTier`s that are bundled in the binary and makes sure that any `NSTemplateTier` that is annotated as bundled and is NOT in that list, is automatically deleted if no longer used by any space.

This will provide a nice cleanup mechanism of the bundled `NSTemplateTier` that we no longer use.

It does not affect any user-defined `NSTemplateTier`s.

Related PRs:
- api: https://github.com/codeready-toolchain/api/pull/480
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1154

### DIFF
--- a/pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
+++ b/pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
@@ -9,10 +9,12 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/host-operator/deploy"
 	"github.com/codeready-toolchain/host-operator/pkg/apis"
+	"github.com/codeready-toolchain/host-operator/pkg/constants"
 	"github.com/codeready-toolchain/host-operator/pkg/templates/nstemplatetiers"
 	commontest "github.com/codeready-toolchain/toolchain-common/pkg/test"
 	"github.com/gofrs/uuid"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -50,8 +52,7 @@ func roles(_ string) []string {
 	return []string{"admin"}
 }
 
-func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
-
+func TestSyncResourcesWitProdAssets(t *testing.T) {
 	s := scheme.Scheme
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
@@ -60,7 +61,7 @@ func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
 	cl := commontest.NewFakeClient(t)
 
 	// when
-	err = nstemplatetiers.CreateOrUpdateResources(context.TODO(), s, cl, namespace)
+	err = nstemplatetiers.SyncResources(context.TODO(), s, cl, namespace)
 
 	// then
 	require.NoError(t, err)
@@ -109,10 +110,8 @@ func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
 	}
 
 	t.Run("failures", func(t *testing.T) {
-
 		namespace := "host-operator" + uuid.Must(uuid.NewV4()).String()[:7]
 		t.Run("nstemplatetiers", func(t *testing.T) {
-
 			t.Run("failed to create nstemplatetiers", func(t *testing.T) {
 				// given
 				clt := commontest.NewFakeClient(t)
@@ -124,7 +123,7 @@ func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
 					return clt.Client.Create(ctx, obj, opts...)
 				}
 				// when
-				err := nstemplatetiers.CreateOrUpdateResources(context.TODO(), s, clt, namespace)
+				err := nstemplatetiers.SyncResources(context.TODO(), s, clt, namespace)
 				// then
 				require.Error(t, err)
 				assert.Regexp(t, "unable to create NSTemplateTiers: unable to create or update the 'base' NSTemplateTier: unable to create resource of kind: NSTemplateTier, version: v1alpha1: an error", err.Error())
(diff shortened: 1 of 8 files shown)


controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -67,6 +99,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, nil
 	}
 
+	if tier.Annotations[toolchainv1alpha1.BundledAnnotationKey] == templates.BundledWithHostOperatorAnnotationValue {
+		if !slices.Contains(r.BundledNsTemplateTierKeys, request.NamespacedName) {
```
MatousJobanek, line 103: why checking the annotation is not enough?

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -67,6 +99,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			// this tier used to be bundled with the operator but it is not anymore.
+			// if it is not used by anything, we are safe to delete it.
+			if used, err := r.isTierUsed(ctx, tier); err != nil {
+				return reconcile.Result{}, err
```
MatousJobanek, line 107: I would rather make this as part of the finalizer logic - in other words, the "delete" operation can be called, but the actual removal would be blocked until there is at least one space that uses the tier

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -67,6 +99,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			} else if !used {
+				return reconcile.Result{}, r.Client.Delete(ctx, tier)
```
MatousJobanek, line 109: wouldn't make more sense to move deletion of the tiers into the `pkg/templates/nstemplatetiers/nstemplatetier_generator.go` file? This file is responsible for creating the tiers, so it would also make sense to call deletions there.

metlos: I think that the role of the reconciler is to sync the cluster state with the desired state, which is exactly what we're doing, so IMHO the reconciler is the right place. If we should move this logic into the generator it would no longer be a generator, but a "syncer" and basically become a poor man's one-time-reconciler itself.

controllers/nstemplatetier/nstemplatetier_controller.go
MatousJobanek: I'm idiot, of course :man_facepalming:

controllers/nstemplatetier/nstemplatetier_controller.go
metlos: If we do this just once at a startup, the consequence is that if a tier is not bundled anymore but is used by any space, it will not be cleared up until the next operator restart.

If we kept the logic in the controller (and added a watch on space deletion to the nstemplatetier controller) we could remove the tier without needing the restart.

metlos: This is enough of a corner case that I don't lean either way and am fine moving the logic to the tier generation at the operator startup. I'm just pointing out the consequence.

MatousJobanek: As mentioned in the other message, the check of the NSTemplateTier usage can be moved to a finalizer https://github.com/codeready-toolchain/host-operator/pull/1180/files#r2128994031 this means that the "generator" can call delete the NSTemplateTier as soon as it's not bundled anymore, but the finalizer will block the removal until it's not used by any Space.

MatousJobanek (commented):
Let's clarify the goals because the PR includes two things that are indepentent:

Finalizers:
* each NSTemplateTier should have finalizer
* the finalizer should block the removal of the NSTemplateTier until there is at least one Space that uses the tier (no matter if it's bundled or not)
* each NSTemplateTier is deletable, which means that anyone can call "delete" on the CR (no matter if it's bundled or not). The tier should be removed only when it's not used (see the previous point).

Deletion of obsolete tiers
* we want to delete tiers that are not bundle anymore, this means that the generator
    * adds annotation to each bundled tier  
    * calls delete on NSTemplateTiers that have the annotation but are not bundled

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -183,7 +180,108 @@ func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
...
+		} {
+			t.Run(fmt.Sprintf("tier: bundled=%v, annotatedBundled=%v, used=%v", setup.bundled, setup.annotatedBundled, setup.used), func(t *testing.T) {
+				// given
+				testTier := &toolchainv1alpha1.NSTemplateTier{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "testTier",
+						Namespace: namespace,
+						// in production, the finalizer is added by the controller.
+						// Here we add it manually to be able to test the annotation removal.
+						Finalizers: []string{"dummy"},
```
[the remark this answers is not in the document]
metlos: I rewrote the tests. There was no need for the table test anymore after the simplifications, so this whole section looks quite differently now.

test/nstemplatetier/nstemplatetier.go
```diff
@@ -246,12 +247,23 @@ func WithParameter(name, value string) TierOption {
...
+// MarkedBundled marks the tier as bundled by adding the appropriate annotation
+func MarkedBundled() TierOption {
+	return func(tier *toolchainv1alpha1.NSTemplateTier) {
+		if tier.Annotations == nil {
+			tier.Annotations = map[string]string{}
+		}
+		tier.Annotations[toolchainv1alpha1.BundledAnnotationKey] = constants.BundledWithHostOperatorAnnotationValue
+	}
+}
+
```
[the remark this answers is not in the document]
metlos: I reformulated the tests to actually use this.

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -183,7 +180,108 @@ func TestCreateOrUpdateResourcesWitProdAssets(t *testing.T) {
...
+				if setup.used {
+					space := &toolchainv1alpha1.Space{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "test-space",
+							Namespace: testTier.Namespace,
+							Labels:    map[string]string{hash.TemplateTierHashLabelKey(testTier.Name): "1234"},
+						},
+					}
+					objs = append(objs, space)
+				}
```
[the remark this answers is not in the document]
metlos: I reformulated the tests and actually dropped the table-driven test here because it's not needed anymore with the simplified logic in the generator that merely cruds.

PR #1182, opened 2025-06-12
Add a finalizer to the NSTemplateTier to prevent its deletion

metlos:
Add a finalizer to the NSTemplateTier to prevent its deletion as long as some spaces are still derived from it.

Related PR:
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/485


rsoaresd (approved):
🚀

rsoaresd:
I saw that the e2e tests are failing here:

```
=== NAME  TestFeatureToggles/provision_space_with_enabled_feature
    clean.go:188: 
        	Error Trace:	/tmp/toolchain-e2e/testsupport/cleanup/clean.go:188
        	            				/usr/local/go/src/sync/once.go:74
        	            				/usr/local/go/src/sync/once.go:65
        	            				/tmp/toolchain-e2e/testsupport/cleanup/clean.go:100
        	            				/tmp/toolchain-e2e/testsupport/cleanup/clean.go:77
        	            				/usr/local/go/src/runtime/asm_amd64.s:1695
        	Error:      	Received unexpected error:
        	            	context deadline exceeded
        	Test:       	TestFeatureToggles/provision_space_with_enabled_feature
        	Messages:   	The proper cleanup of the NSTemplateTier 'ftier' and related resources wasn't finished within the given timeout
        	            	the NSTemplateTier 'ftier' is still present in the cluster: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:ftier GenerateName: Namespace:
```

MatousJobanek:
but in most of the cases (if not all except for this one) we don't return an error if the CR is pending deletion because of some dependent CRs. The reason is that the other controllers watch the CRDs it depends on, this is an exception. So it's rather the logic of the finalizer you would need to implement anyway.

MatousJobanek:
I see that the e2e tests are still failing, According to the logs it's a timing issue because the returned error triggers postponed reconcile loops with exponential increase of the time gaps. In other words, returning an error is not enough. 
Multiple options from the top of my head:
* in the finalizer logic, ignore Spaces that are already being deleted. In other words, list all Spaces with the label, filter out those that have the deletion timestamp set and return the error only when there is a Space with the label that is not being deleted. This should help and it's completely fine to ignore Spaces that are being deleted.
* modify the cleanup logic in e2e tests so the deletion of NSTemplateTiers is triggered as the last one after all other cleanups are completed (this seems to be pretty complicated)
* modify the cleanup logic so it keeps updating for example an annotation in the NSTemplateTier CR to trigger the reconcile every time when the NSTemplateTier is pending deletion (this sounds much simpler)
* add a watch to the controller with a predicate & mapper so it listens to deletion of Spaces (in my opinion it's an overkill to workaround problem in e2e tests)

Anything else?

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -255,6 +257,56 @@ func (r *Reconciler) updateStatus(ctx context.Context, tmplTier *toolchainv1alph
 	return r.Client.Status().Update(ctx, tmplTier)
 }
 
+func (r *Reconciler) handleDeletion(ctx context.Context, tier *toolchainv1alpha1.NSTemplateTier) error {
+	list := &toolchainv1alpha1.SpaceList{}
```
fbm3307, line 261: Just curious, What if we check first if there is already a finalizer present or not , then can we directly return ?

fbm3307 (approved):
Great Work , 
just  a curiosity question .

controllers/nstemplatetier/nstemplatetier_controller.go
fbm3307: Okay , so in the current logic, i see it checks `util.IsBeingDeleted(&s)` ,  would checking `util.HasFinalizer` would be before listing spaces , am i understanding this correctly?

MatousJobanek: I mean that we filter out those that are being deleted

controllers/nstemplatetier/nstemplatetier_controller.go
```diff
@@ -255,6 +257,56 @@ func (r *Reconciler) updateStatus(ctx context.Context, tmplTier *toolchainv1alph
...
+		if util.IsBeingDeleted(&s) {
+			// As an optimization, we skip the spaces that are being deleted.
+			// There is no functional relationship between the spaces and the tier, we just don't want
+			// there to be spaces that refer to a non-existing tier. So if the space is being deleted,
+			// we can pretend like it doesn't already exist.
+			//
+			// This speeds up the deletion of the tier because it doesn't linger around for the spaces
+			// that are being deleted (which can take a good amount of time).
+			continue
+		}
```
MatousJobanek, line 280: Thanks for adding this :+1: This should help with the e2e tests :rocket:

MatousJobanek (approved):

controllers/nstemplatetier/nstemplatetier_controller.go
fbm3307: I was actually curious about checking a generic finalizer before listing spaces.

MatousJobanek: as for your question Feny (I misunderstood it before), we don't need to make yet another check because if the resource is being deleted and it doesn't have any finalizer, then it should be removed from the etcd, which means that we shouldn't get any other reconciles.

fbm3307: Thanks for Explaining !!

PR #1186, opened 2025-07-03
React on NSTemplateTier deletions

metlos:
We've added the finalizer on the NSTemplateTier but the reconciler would never receive the delete events, because it only reacted on the generation changed.

This is a followup of https://github.com/codeready-toolchain/host-operator/pull/1180.

Related PRs:
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1154

### DIFF
--- a/controllers/nstemplatetier/nstemplatetier_controller.go
+++ b/controllers/nstemplatetier/nstemplatetier_controller.go
@@ -21,7 +21,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):

MatousJobanek (changes requested):
The GenerationChangedPredicate doesn't have any impact on the deletion event. it defines predicate for the update event only https://github.com/kubernetes-sigs/controller-runtime/blob/252af6420feba970d0020ba4c029a0a1ce3f9688/pkg/predicate/predicate.go#L191-L224

metlos:
> The GenerationChangedPredicate doesn't have any impact on the deletion event. it defines predicate for the update event only https://github.com/kubernetes-sigs/controller-runtime/blob/252af6420feba970d0020ba4c029a0a1ce3f9688/pkg/predicate/predicate.go#L191-L224

That's why the `IsBeingDeletedPredicate` reacts on the update (and generic) events ;)

The delete of any kubernetes object creates two events:
1. an update event for setting the non-nil deletion timestamp
2. the delete event that is actually only trigger AFTER the object has disappeared from etcd.

The `GenerationChangedPredicate` filters out the above update event because it doesn't change the generation. Therefore we may never invoke the finalizer removal logic. This possibly doesn't happen frequently in practice because our controllers usually have other watches, too, but nevertheless, the possibility is IMHO real.

MatousJobanek:
After making a very quick test, I have apologize for making a mistake - the delete call (that sets the not-nil `deletionTimestamp`) does really trigger the update event and not the delete event. My bad, I misunderstood the purpose of the delete event. Sorry about that.

Anyway, as part of the quick test I also figured out that setting the `deletionTimestamp` also updates the `generation` number, which means that it does go through the `GenerationChangedPredicate`.
You can double check it by watching the resource that is being deleted.
This also explains why it works for all controllers that we have with the predicate now.

metlos:
Yep. My mistake in all this was testing this on a ConfigMap that for some reason doesn't set or increment the generation at all. I'm dropping this PR with a relief :)

PR #1203, opened 2025-09-23
[SANDBOX-1407] don't leave garbage behind when deleting no-longer bundled NSTemplateTiers

metlos:
When a `NSTemplateTier` that is no longer bundled with the host operator is deleted, we did not clean up the `TierTemplate`s that came along with it. These stay in the cluster unused indefinitely. 

Because the host operator is responsible for the lifetime of the bundled `NSTemplateTier`, it should also clean up the `TierTemplate`s created for the `NSTemplateTiers`.

## Summary by CodeRabbit

- Bug Fixes
  - Removing an outdated tier now also deletes its associated templates, preventing orphaned resources.
  - Improved error handling: missing referenced templates are skipped with logs, and multiple errors are aggregated for clearer reporting.

- Tests
  - Expanded tests to verify both the tier and its associated templates are removed during sync.
  - Added and refactored test helpers, including a new fluent constructor for TierTemplate test objects and option helpers to simplify test setup.

### DIFF
--- a/config/crd/bases/toolchain.dev.openshift.com_toolchainconfigs.yaml
+++ b/config/crd/bases/toolchain.dev.openshift.com_toolchainconfigs.yaml
@@ -407,6 +407,11 @@ spec:
                             type: array
                             x-kubernetes-list-type: atomic
                         type: object
+                      workatoWebHookURL:
+                        description: |-
+                          WorkatoWebHookURL is used by the UI to push events to Marketo for analytics purposes.
+                          The webhook URL is unique per environment.
+                        type: string
                     type: object
                   spaceConfig:
                     description: Keeps parameters necessary for configuring Space


rsoaresd:
/retest

failed due to missing configuration for sandbox ui e2e tests

pkg/templates/nstemplatetiers/nstemplatetier_generator_test.go
```diff
@@ -206,3 +215,12 @@ func verifyTierTemplate(t *testing.T, cl *commontest.FakeClient, namespace, tier
 	require.Fail(t, fmt.Sprintf("the TierTemplate for NSTemplateTier '%s' and of the type '%s' wasn't found", tierName, typeName))
 	return ""
 }
+
+func newTierTemplate(name, ns string) *toolchainv1alpha1.TierTemplate {
```
mfrancisc, line 219: minor - this function looks like a duplicate of what we already have in `nstemplatetier_controller_test.go` and `tiertemplate_guard_controller_test.go`. I'm wondering if we could extract it to common function ( if there is one already ).

mfrancisc (commented):
Looks good 👍 

I have one question about the use-case - I though that we want to delegate the management of the tiers outside . So for example if a tier is managed via ArgoCD then it shouldn't be deleted by host operator, since that might create some conflicts. 
But maybe I'm missing something .

mfrancisc:
Also - do we need an e2e test for this ?

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
```diff
@@ -63,20 +64,41 @@ func removeNoLongerBundledTiers(ctx context.Context, client runtimeclient.Client
...
 
 	return err
 }
 
+func deleteReferencedTierTemplates(ctx context.Context, client runtimeclient.Client, tier *toolchainv1alpha1.NSTemplateTier) error {
+	for _, ref := range GetNSTemplateTierRefs(tier) {
+		template := &toolchainv1alpha1.TierTemplate{}
+		if err := client.Get(ctx, runtimeclient.ObjectKey{Name: ref, Namespace: tier.Namespace}, template); err != nil {
+			if kerrors.IsNotFound(err) {
+				continue
```
fbm3307, line 91: should we add some logger/statement so we would know what happened.?

fbm3307 (commented):
looks good , just a minor thing

pkg/templates/nstemplatetiers/nstemplatetier_generator.go
mfrancisc: I was thinking the same, but maybe logging what was deleted , not what was skipped, not found.

metlos: done

fbm3307 (approved):

PR #1207, opened 2025-09-29
Fix: Check age of spacebinding before deletion

fbm3307:
this is to fix the flakiness of tests, examples and discussion of failures slack thread https://redhat-internal.slack.com/archives/CHK0J6HT6/p1758276534724369
here  the `space_binding_cleanup_controller`  checks the age of the spacebinding along with if related space is present or not before deleting so that its not deleted in case the client cache is  updated with spacebinding first and its yet to be updated with related space

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Enforces a short delay (2s) before removing SpaceBindings when the target Space/MUR is missing; requeues if binding is too new.
  * Logs now show binding age and remaining delay for clearer visibility.

* **Bug Fixes**
  * Prevents premature SpaceBinding deletion during transient missing-resource conditions.

* **Tests**
  * Expanded age-based coverage validating requeue timing and eventual deletion across multiple scenarios.

### DIFF
--- a/controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
+++ b/controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
@@ -24,7 +24,7 @@ import (
 
 const norequeue = 0 * time.Second
 const requeueDelay = 10 * time.Second
-const deletionDelay = 30 * time.Second
+const deletionDelay = 2 * time.Second
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
--- a/controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
+++ b/controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
@@ -24,6 +24,7 @@ import (
 
 const norequeue = 0 * time.Second
 const requeueDelay = 10 * time.Second
+const deletionDelay = 30 * time.Second
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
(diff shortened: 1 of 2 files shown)


controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -79,11 +80,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	space := &toolchainv1alpha1.Space{}
 	if err := r.Get(ctx, spaceName, space); err != nil {
 		if errors.IsNotFound(err) {
-			logger.Info("the Space was not found", "Space", spaceBinding.Spec.Space)
+			withinDelayPeriod, age, timeLeft := checkSpaceBindingDeletionDelay(spaceBinding)
+			if withinDelayPeriod {
+				logger.Info("Space not found, SpaceBinding too young - waiting", "space", spaceBinding.Spec.Space, "age", age.String(), "timeLeft", timeLeft.String())
+				return ctrl.Result{RequeueAfter: timeLeft}, nil
+			}
+			logger.Info("Space not found, SpaceBinding old enough - deleting", "space", spaceBinding.Spec.Space, "age", age.String())
```
MatousJobanek, line 88: shouldn't this go into `deleteSpaceBinding` so it's consistent behavior for both Space and MUR?

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
```diff
@@ -201,6 +208,19 @@ type SpaceBindingRequestAssociated struct {
...
+// Also returns the age and time left in delay period for logging.
+func checkSpaceBindingDeletionDelay(spaceBinding *toolchainv1alpha1.SpaceBinding) (withinDelayPeriod bool, age time.Duration, timeLeft time.Duration) {
+	spaceBindingAge := time.Since(spaceBinding.CreationTimestamp.Time)
+	if spaceBindingAge < deletionDelay {
+		// Calculate how much time is left in the deletion delay period
+		timeLeft := deletionDelay - spaceBindingAge
+		return true, spaceBindingAge, timeLeft
+	}
+	return false, spaceBindingAge, 0
+}
```
MatousJobanek, line 222: considering:
* the size of the function
* the number of returned values
* the fact that it's used only at one place
* and the additional if statement at the place where it is called

wouldn't it be better to better to incorporate it into the logic of `deleteSpaceBinding` function?

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
```diff
@@ -117,25 +117,65 @@ func TestDeleteSpaceBinding(t *testing.T) {
...
+				fakeClient := test.NewFakeClient(t, sbYoung, toolchainconfig)
 				reconciler := prepareReconciler(t, fakeClient)
 
 				// when
-				res, err := reconciler.Reconcile(context.TODO(), requestFor(sbPublicViewerRedhatView))
+				res, err := reconciler.Reconcile(context.TODO(), requestFor(sbYoung))
+
+				// then
+				require.NoError(t, err)
+				require.True(t, res.RequeueAfter > 24*time.Second && res.RequeueAfter <= 25*time.Second) // approximately 25 seconds left
```
MatousJobanek, line 132: it's better to use the built-in `Greater` and `Less` functions:
```suggestion
				assert.Greater(t, res.RequeueAfter, 24*time.Second)
				assert.Less(t, res.RequeueAfter, 25*time.Second)
```

controllers/spacebindingcleanup/spacebinding_cleanup_controller.go
fbm3307: yea i was thinking about that too.. update it now

PR #1257, opened 2026-04-22
SANDBOX-1282 | refactor: replace "wrapf" calls with stdlib wrap calls

MikelAlejoBR:
There is no need to keep the legacy "wrapf" calls to wrap errors anymore, so the idea is to remove them in favor of the wrapping calls of the stdlib.

## Related PRs

- https://github.com/codeready-toolchain/host-operator/pull/1257
- https://github.com/codeready-toolchain/member-operator/pull/744
- https://github.com/codeready-toolchain/toolchain-common/pull/526
- https://github.com/codeready-toolchain/toolchain-e2e/pull/1275

## Jira ticket
[[SANDBOX-1282]](https://redhat.atlassian.net/browse/SANDBOX-1282)

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling across the app to preserve underlying error context, leading to clearer and more informative error messages during user signup approval, template processing, notification generation, and startup operations.

### DIFF
--- a/pkg/templates/notificationtemplates/notification_generator.go
+++ b/pkg/templates/notificationtemplates/notification_generator.go
@@ -60,7 +60,7 @@ func templatesForAssets(notificationFS embed.FS, root string, setName string) (m
 		}
 		segments := strings.Split(path, "/")
 		if len(segments) != 5 {
-			return nil, fmt.Errorf("unable to load templates: %w", errors.New("path must contain directory and file"))
+			return nil, errors.New("unable to load templates: path must contain directory and file")
 		}
 		directoryName := segments[3]
 		filename := segments[4]
--- a/pkg/templates/notificationtemplates/notification_generator.go
+++ b/pkg/templates/notificationtemplates/notification_generator.go
@@ -60,7 +60,7 @@ func templatesForAssets(notificationFS embed.FS, root string, setName string) (m
 		}
 		segments := strings.Split(path, "/")
 		if len(segments) != 5 {
-			return nil, errors.Wrapf(errors.New("path must contain directory and file"), "unable to load templates")
+			return nil, fmt.Errorf("unable to load templates: %w", errors.New("path must contain directory and file"))
 		}
 		directoryName := segments[3]
 		filename := segments[4]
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):

mfrancisc (approved):

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -60,7 +60,7 @@ func templatesForAssets(notificationFS embed.FS, root string, setName string) (m
 		}
 		segments := strings.Split(path, "/")
 		if len(segments) != 5 {
-			return nil, errors.Wrapf(errors.New("path must contain directory and file"), "unable to load templates")
+			return nil, fmt.Errorf("unable to load templates: %w", errors.New("path must contain directory and file"))
```
metlos: Agreed with coderabbit here. `errors.New` doesn't make sense here. I'd just use `fmt.Errorf` with just the single string. Alternatively, you can use `errors.New` with that string or even a global private error var with this error.

pkg/templates/notificationtemplates/notification_generator.go
```diff
@@ -75,7 +75,7 @@ func templatesForAssets(notificationFS embed.FS, root string, setName string) (m
 			template.Subject = string(content)
 			notificationTemplates[directoryName] = template
 		default:
-			return nil, errors.Wrapf(errors.New("must contain notification.html and subject.txt"), "unable to load templates")
+			return nil, fmt.Errorf("unable to load templates: %w", errors.New("must contain notification.html and subject.txt"))
```
metlos, line 78: Same as above.

metlos (approved):
Yay for using the stdlib :) Just minor comments below that don't block the merge IMHO.

pkg/templates/notificationtemplates/notification_generator.go
MikelAlejoBR: Yeah, I went with a simple `errors.New` because we don't need to `errors.As` or `errors.Is` it

pkg/templates/notificationtemplates/notification_generator.go
MikelAlejoBR: Makes sense, done!

PR #1265, opened 2026-05-21
Add tier for -claw namespace

alexeykazakov:
Related to https://github.com/codeready-toolchain/sandbox-sre/pull/3177
Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/1282

## Summary by CodeRabbit

* **New Features**
  * Introduced new workspace tier with dedicated cluster resource management, namespace isolation through network policies, RBAC controls, and configurable idle timeout capabilities.

* **Configuration**
  * Added resource quota constraints across multiple workspace tier templates to limit resource consumption.

* **Tests**
  * Extended test suite to validate the new tier configuration and template references.

### DIFF
--- a/deploy/templates/nstemplatetiers/base1ns/ns_dev.yaml
+++ b/deploy/templates/nstemplatetiers/base1ns/ns_dev.yaml
@@ -120,6 +120,15 @@ objects:
         cpu: 10m
         memory: 64Mi
 
+- apiVersion: v1
+  kind: ResourceQuota
+  metadata:
+    name: compute-spacerequests
+    namespace: ${SPACE_NAME}-dev
+  spec:
+    hard:
+      count/spacerequests.toolchain.dev.openshift.com: "1"
+
 - apiVersion: networking.k8s.io/v1
   kind: NetworkPolicy
   metadata:
--- a/deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
@@ -50,6 +50,37 @@ objects:
   - kind: User
     name: ${USERNAME}
 
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: spacerequest-manage
+    namespace: ${NAMESPACE}
+  rules:
+  - apiGroups:
+    - toolchain.dev.openshift.com
+    resources:
+    - spacerequests
+    verbs:
+    - get
+    - list
+    - watch
+    - create
+    - update
+    - patch
+    - delete
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: ${USERNAME}-spacerequest-manage
+    namespace: ${NAMESPACE}
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: Role
+    name: spacerequest-manage
+  subjects:
+  - kind: User
+    name: ${USERNAME}
+
 parameters:
 - name: NAMESPACE
   required: true
--- a/deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
+++ b/deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
@@ -50,37 +50,6 @@ objects:
   - kind: User
     name: ${USERNAME}
 
-- apiVersion: rbac.authorization.k8s.io/v1
-  kind: Role
-  metadata:
-    name: spacerequest-manage
-    namespace: ${NAMESPACE}
-  rules:
-  - apiGroups:
-    - toolchain.dev.openshift.com
-    resources:
-    - spacerequests
-    verbs:
-    - get
-    - list
-    - watch
-    - create
-    - update
-    - patch
-    - delete
-- apiVersion: rbac.authorization.k8s.io/v1
-  kind: RoleBinding
-  metadata:
-    name: ${USERNAME}-spacerequest-manage
-    namespace: ${NAMESPACE}
-  roleRef:
-    apiGroup: rbac.authorization.k8s.io
-    kind: Role
-    name: spacerequest-manage
-  subjects:
-  - kind: User
-    name: ${USERNAME}
-
 parameters:
 - name: NAMESPACE
   required: true
--- a/deploy/templates/nstemplatetiers/claw/cluster.yaml
+++ b/deploy/templates/nstemplatetiers/claw/cluster.yaml
@@ -1,0 +1,35 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: claw-cluster-resources
+objects:
+- apiVersion: quota.openshift.io/v1
+  kind: ClusterResourceQuota
+  metadata:
+    name: for-${SPACE_NAME}-claw
+  spec:
+    quota:
+      hard:
+        count/deployments.apps: "5"
+        count/pods: "10"
+        count/routes.route.openshift.io: "3"
+        count/services: "5"
+        count/secrets: "50"
+        count/configmaps: "10"
+    selector:
+      annotations: null
+      labels:
+        matchLabels:
+          toolchain.dev.openshift.com/space: ${SPACE_NAME}
+- apiVersion: toolchain.dev.openshift.com/v1alpha1
+  kind: Idler
+  metadata:
+    name: ${SPACE_NAME}-claw
+  spec:
+    timeoutSeconds: ${{IDLER_TIMEOUT_SECONDS}}
+parameters:
+- name: SPACE_NAME
+  required: true
+- name: IDLER_TIMEOUT_SECONDS
+  # 12 hours
+  value: "43200"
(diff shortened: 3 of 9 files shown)


xcoulon (approved):

deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
```diff
@@ -50,6 +50,37 @@ objects:
...
+  metadata:
+    name: ${USERNAME}-spacerequest-manage
+    namespace: ${NAMESPACE}
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: Role
+    name: spacerequest-manage
+  subjects:
+  - kind: User
+    name: ${USERNAME}
```
MatousJobanek: users should be already able to create & manage SpaceRequests in their namespaces via aggregated roles managed by OLM. this part is not needed

deploy/templates/nstemplatetiers/claw/cluster.yaml
```diff
@@ -0,0 +1,35 @@
...
+  metadata:
+    name: for-${SPACE_NAME}-claw
+  spec:
+    quota:
+      hard:
+        count/deployments.apps: "5"
+        count/pods: "10"
+        count/routes.route.openshift.io: "3"
+        count/services: "5"
+        count/secrets: "50"
```
MatousJobanek, line 17: I guess that the number of secrets could be limited to 20 or something like that

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - list
+    - watch
+  - apiGroups:
+    - "route.openshift.io"
+    resources:
+    - routes
+    verbs:
+    - get
+    - list
+    - watch
```
MatousJobanek, line 54: couldn't we simply grant `view` cluster role to the user in this namespace? so they could see also deployment and other things? or is there anything we want to hide from them?

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
```diff
@@ -0,0 +1,82 @@
...
+  - apiGroups:
+    - ""
+    resources:
+    - pods/log
+    verbs:
+    - get
+    - list
```
MatousJobanek: do we want to let users to also ssh into the pod?

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - watch
+  - apiGroups:
+    - ""
+    resources:
+    - secrets
+    verbs:
+    - create
+    - update
+    - patch
+    - delete
```
MatousJobanek: isn't it missing read permissions?

alexeykazakov: No, it's by design. OpenClaw will have user's permissions in the -dev namespace. So we don't want it to indirectly be able to read the secrets from -claw namespace.

deploy/templates/nstemplatetiers/claw/cluster.yaml
alexeykazakov: Users will create secrets for credentials for external services. 20 should be fine in most cases but I wanted to be on the safe side.

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Yeah, I guess "view" would be fine.

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Ideally yes but I will need to think about possible security side-effects.

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
MatousJobanek: how would it be able to do that? this is Role granted to user, it's not visible anywhere in the `-dev` namespace.

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Now I remember why we can't do that. Users would be able to ssh to the proxy to see the credentials mounted to them (from secrets). And openclaw will have user's permissions. This is the same reason for not allowing users to read the secrets in -claw NS. We don't want OpenClaw be able to get access to credentials at all (users permissions == claw permissions).

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Yeah.. you are probably actually right. I'm still thinking about that space manager SA in the -dev namespace which we don't have. Let me think about it more carefully.

deploy/templates/nstemplatetiers/base1ns/spacerole_admin.yaml
alexeykazakov: Right. Addressed here: https://github.com/codeready-toolchain/host-operator/pull/1265/changes/e75981c9e6acd82637456f59f367483d1ca3d42d

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Addressed here: https://github.com/codeready-toolchain/host-operator/pull/1265/changes/e75981c9e6acd82637456f59f367483d1ca3d42d

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Addressed here: https://github.com/codeready-toolchain/host-operator/pull/1265/changes/e75981c9e6acd82637456f59f367483d1ca3d42d

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
alexeykazakov: Addressed here: https://github.com/codeready-toolchain/host-operator/pull/1265/changes/e75981c9e6acd82637456f59f367483d1ca3d42d

But we will also need to create a SA for the claw in the -dev namespace: `claw-workspace` SA. This SA will be used by OpenClaw to access -dev namespace. See https://github.com/codeready-toolchain/sandbox-claw-operator/blob/master/docs/proposals/namespace-isolation-design.md#claw-workspace-serviceaccount

I think we can stick with the option of creating this SA by Dashboard when provisioning the claw (vs. adding it to the namespace template). At least for now.

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
MatousJobanek: why and how would open-claw have the same permissions as users in this -claw namespace?

MatousJobanek: ok, seeing the commit (and your response in the other thread) I guess that you agreed that there is no security risk :-) All good then :-)

deploy/templates/nstemplatetiers/claw/spacerole_admin.yaml
MatousJobanek: yup, SA in the -dev namespace is a necessary piece of the puzzle, but it doesn't open any risks in terms of accessing secrets in -claw namespace, because SR shouldn't create the namespace-manager namespace in the -dev namespace when its name is not defined in the tier 
https://github.com/codeready-toolchain/host-operator/blob/f9148fc142c10b830ab9ebb9cde5e567294ec3c6/controllers/spacerequest/spacerequest_controller.go#L113-L121
and
https://github.com/codeready-toolchain/host-operator/blob/f9148fc142c10b830ab9ebb9cde5e567294ec3c6/controllers/spacerequest/spacerequest_controller.go#L438-L444
but it's worth double-checking it in stage cluster

deploy/templates/nstemplatetiers/base1ns/ns_dev.yaml
```diff
@@ -120,6 +120,15 @@ objects:
...
+  kind: ResourceQuota
+  metadata:
+    name: compute-spacerequests
+    namespace: ${SPACE_NAME}-dev
+  spec:
+    hard:
+      count/spacerequests.toolchain.dev.openshift.com: "1"
+
 - apiVersion: networking.k8s.io/v1
   kind: NetworkPolicy
```
MatousJobanek, line 133: I'm wondering if we would need also a NetworkPolicy for communication between `-claw` and `-dev` namespace, but probably not. WDYT?

MatousJobanek (approved):
Looks good 🚀

deploy/templates/nstemplatetiers/base1ns/ns_dev.yaml
alexeykazakov: It might be useful so the claw could access services running in -dev. But we would also need to whitelist internal cluster traffic in the proxy to make it work.

PR for the proxy side: https://github.com/codeready-toolchain/claw-operator/pull/139
I also added the NP creation by Dashboard to the design: https://github.com/codeready-toolchain/sandbox-claw-operator/blob/master/docs/proposals/namespace-isolation-checklist.md#8-dashboard-flow
I would probably avoid adding this NP to the -dev template so we don't introduce another NP for all users even if they don't deploy any Claws.

PR #1273, opened 2026-07-06
SANDBOX-1889: update CRD default phoneLookupMode to disabled

fbm3307:
Generated CRD update reflecting the default phoneLookupMode change from "log" to "disabled" in the API types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default phone verification behavior so phone risk checks are now skipped unless explicitly enabled.
  * Clarified the configuration description to match the new default setting.


fbm3307:
wrong branch

PR #1281, opened 2026-07-21
refactor: remove revision checks from ToolchainStatus

xcoulon:
now that the host and member operators and registration service expose
their current commit as a metric and that we have external CronJobs to
retrieve their associated upstream commits, we don't need the built-in
revision checks

see https://redhat.atlassian.net/browse/SANDBOX-1737

## Summary by CodeRabbit

- **Changes**
  - Toolchain status now reports host operator readiness based on deployment health.
  - Registration service readiness is determined by its health endpoint.
  - Status reporting no longer depends on GitHub revision comparisons or source commit information.
  - Revision status is presented as up to date for the host operator while deployment readiness is evaluated.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -42,10 +42,8 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/controllers/toolchaincluster"
 	"github.com/codeready-toolchain/toolchain-common/controllers/toolchainclustercache"
 	"github.com/codeready-toolchain/toolchain-common/controllers/toolchainclusterresources"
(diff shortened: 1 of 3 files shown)


xcoulon:
draft for now: we need to collect the latest deployed commit and the latest commit on GitHub for the registration service

rajivnathan (approved):

metlos (approved):
The code itself looks fine. Do we use the VersionCheckManager anywhere else or can this be removed, too, from `toolchain-common`? Are there any other runtime consequences of dropping this check than just the informational purposes?

alexeykazakov (approved):

PR #1291, opened 2026-08-07
test: remove flakiness in TestDeleteSpaceBinding

xcoulon:
the following test case is flaky, which causes the CI to fail in the
"Verify Dependencies" jobs in the `api` and `toolchain-common` projects.

```
--- FAIL: TestDeleteSpaceBinding (0.03s)
    --- FAIL: TestDeleteSpaceBinding/with_public-viewer_disabled (0.03s)
        --- FAIL: TestDeleteSpaceBinding/with_public-viewer_disabled/lara-redhat_SpaceBinding_requeues_when_redhat_space_is_missing_and_SpaceBinding_is_too_young (0.01s)
            spacebinding_cleanup_controller_test.go:165:
                	Error Trace:	/tmp/replace-verify.Y3t/host-operator/controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go:165
                	Error:      	"0s" is not greater than "0s"
                	Test:       	TestDeleteSpaceBinding/with_public-viewer_disabled/lara-redhat_SpaceBinding_requeues_when_redhat_space_is_missing_and_SpaceBinding_is_too_young
```

See [here](https://github.com/codeready-toolchain/toolchain-common/actions/runs/31172030720/job/92845733004) for example of the failure.

Workaround: increase the deletion delay to 3 seconds and compare the time left to 2 seconds instead of 1 second (to avoid rounding errors?)

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted SpaceBinding cleanup timing to provide a slightly longer delay before deletion.
  * Improved cleanup timing validation to ensure resources are handled consistently during the extended delay.


alexeykazakov (approved):

fbm3307 (approved):
Thanks for the fix .

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
fbm3307, line 123: instead of increasing the deletion threshold to 3 sec , may be we can do something of this sort 


```suggestion
sbYoung.CreationTimestamp = metav1.NewTime(time.Now().Add(-500 * time.Millisecond))
// ...
require.Greater(t, res.RequeueAfter, 0*time.Second)
require.LessOrEqual(t, res.RequeueAfter, deletionDelay)
// roughly "delay minus ~age"
require.Greater(t, res.RequeueAfter, deletionDelay-time.Second)
```

fbm3307 (commented):
This is just another way i was thinking to change the assertion and not update the spacebinding deletion threshold to 3 sec, 
but that's just a suggestion  , we can use this next time if the same flake returns

controllers/spacebindingcleanup/spacebinding_cleanup_controller_test.go
xcoulon: ok, let's keep this idea for later if the flakiness persists

